PR #2, opened 2019-07-24
Run micro-dev on port set by .env

callaghanc:
When `micro-dev` runs, it will try to use port 3000, and if that port is occupied it will use a random port. The public app uses port 3000, which is the default for Next.js apps, so if we run the listing service alongside the public app, `micro-dev` will see that port 3000 is busy and will use a random port. This makes it a little extra work to use the listing service from the public app, because you would have to update the listing service's port number in the API call every time you restarted the listing service. Being able to set `micro-dev`'s port via a local .env file will save that small extra step.

### DIFF
--- a/services/listings/README.md
+++ b/services/listings/README.md
@@ -1,3 +1,18 @@
 # Listings Service v0
 
 This is a minimal listings service which stores listings in individual JSON files and serves them via the [micro](https://github.com/zeit/micro) framework.
+
+## Running the service in development
+Set the port you would like the service to run on via a .env file.
+
+In this directory, create a .env file with the following contents:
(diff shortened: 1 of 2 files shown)


bk3c:
This didn't work right for me as-is, and when I tried to fix things in the idiomatic way, I ran straight into https://github.com/zeit/micro/issues/337 and related issues with hooking up dotenv. I think the simplest thing this early in the game will be to switch over to Koa, which will match our modern aesthetics but provide many more feature options in a relatively small package. See #5 as my suggested alternate plan.

bk3c:
Closing in favor of #5.

PR #5, opened 2019-07-29
Replace micro with Koa as the listings service framework.

bk3c:
Based on #2, I think micro might be a little too micro for our use case, especially since we're not particularly sensitive to size & dependency footprint. We could have used Express, but since any services running in node instead of rails would be greenfield, it seems better to be able to optimize for using async/await and go directly to the more modern alternative.

Koa [documentation](https://koajs.com/) and [repository](https://github.com/koajs/koa).

Some decent starters that I looked at for reference include:

- https://github.com/ddimaria/koa-typescript-starter
- https://github.com/javieraviles/node-typescript-koa-rest


callaghanc:
@bk3c i'm happy to switch to something that supports more of what we need to do! let's merge this and close #2

PR #6, opened 2019-07-31
Install Polyglot

callaghanc:
Install Polyglot and create a simple usage example.

The use of `_app.js` to instantiate Polyglot and provide it to the pages is just an initial easy way to do the setup and provide a working example of Polyglot to try out. We do not need to stick with this setup and we will likely evolve it as we do more work on the app and start to form more structure.


bk3c:
Merging this non-final implementation so we can see all the pieces working together.

PR #7, opened 2019-08-06
add the storybook viewport plugin and fix some (unrelated) story errors

bk3c:
a couple minor notes:

- the story errors were just typescript syntax that I'm not sure how we missed before, but were easy enough to fix. NB: declare things as string not String was most of them.

- I thought I added a prettier block to package.json but didn't so there's some noisy reformatting here. Will plan to address that asap.


bk3c:
Counting @jaredcwhite's thumbs-up as an approval and merging.

PR #8, opened 2019-08-07
Basic Table component + responsive mobile collapse

jaredcwhite:
Here's my take on how we could implement the tables that will appear on both the listings list page and the listing detail page. For mobile in particular, I noticed that sometimes the tables looks like tables and are horizontally scrollable, and in other cases it collapses down to header cells on left and value cells on right in a column (so each row becomes a block column of these cell pairs), so I accounted for both cases. I also like how easy it is to transform incoming JSON into formatted JSX that the table can display.

TIA for your thoughts and feedback.

### DIFF
--- a/apps/public/lib/unit_summaries.tsx
+++ b/apps/public/lib/unit_summaries.tsx
@@ -1,0 +1,14 @@
+import * as React from 'react'
+
+export const unitSummariesTable = (listing): Array<Object> => {
+  const unitSummaries = listing.unit_summaries.general.map(unitSummary => {
+    return {
+      unit_type: <strong>{unitSummary.unit_type}</strong>,
+      minimum_income: <><strong>${parseInt(unitSummary.min_income_range.min, 10)}</strong>/month</>,
+      rent: <><strong>${parseInt(unitSummary.rent_range.min, 10)}</strong>/month</>,
+      availability: <><strong>1</strong> unit</>
+    }
+  })
+
+  return unitSummaries
+}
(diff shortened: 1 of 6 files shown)


bk3c:
The basic class setup seems straight-forward enough. I know you were looking into some existing libraries like react-table -- did you just decide that was too much for what we needed right now?

My one code structure comment is that maybe the row and cell classes should be in the same file since they're so small, inextricably linked, and you already have multiple cell classes. I could certainly be convinced otherwise, but my gut is that might be easier to read and import.

Otherwise LGTM.

callaghanc:
responsiveness looks great! tried the table out on mobile in both responsive-collapse and non-responsive-collapse mode, collapsed mode looks good and horizontal scrolling works in non-responsive mode. am also inclined to think that we could put the row and cell components together, maybe even in the basic_table.tsx file along with the table itself?

jaredcwhite:
Sounds good re: putting it all into one file. I'll update that and then it sounds like this'll be ready to merge

callaghanc:
i just remembered, did we decide we wanted to go with the new table style as seen on: https://housing.sfgov.org/listings/for-sale?

jaredcwhite:
> i just remembered, did we decide we wanted to go with the new table style as seen on: https://housing.sfgov.org/listings/for-sale?

It looks like they have the old table style in some places and the newer in others, so I think if we end up going that route it's just a matter of using alternate classes as another option

callaghanc:
ok, then we can do that other style in another pr

bk3c:
Works locally and LGTM. I think a next step here would be to build a couple stories in SB and some tests for the component, but that can be another PR too.

PR #9, opened 2019-08-07
Start to build out site and listing page layout

callaghanc:
This PR contains:

- Some basic fleshing out of the site header and footer components. The styling is purely WIP, to allow us to start to see our work situated in something that looks more like a real site. All the exact colors, fonts, sizing, positioning, etc. will be further refined as we continue on.
- A high-level responsive layout for the listing page, so that it will have the sidebar content below the image hero on mobile and then the sidebar moves to the right on desktop. Lmk your thoughts on this layout. I've barely used Flexbox before, so open to all info about best practices.
- A refactor of the more-specific PropertyTitleImage component into the more general ImageCard component. 
- Removal of the About page, which was just for test/experimental purposes.


apps/public/components/listings_list.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { PropertyTitleImage } from '@dahlia/ui-components/src/cards/property_title_image';
```
bk3c: I don't have a specific problem with this change, but I thought we were going in the direction of more semantically aligned components? This seems like the other direction.

apps/public/pages/listing.tsx
```diff
@@ -2,7 +2,7 @@ import { Component } from 'react';
 import ReactDOMServer from 'react-dom/server';
 import Layout from '../layouts/application';
 import { Listing } from '../components/listings_list';
-import { PropertyTitleImage } from '@dahlia/ui-components/src/cards/property_title_image';
```
bk3c: Ditto my comment above

apps/public/pages/listing.tsx
```diff
@@ -46,66 +46,89 @@ export default class extends Component<ListingProps> {
...
+            subImageContent={
+              <>
+                <p className="t-alt-sans uppercase">{oneLineAddress}</p>
+                <p className="text-gray-700">{listing.developer}</p>
+                <p className="text-xs"><a href={googleMapsHref} target="_blank" aria-label="Opens in new window">View on Map</a></p>
+              </>
+            }
+          />
+
+          <aside className="w-full md:absolute md:right-0 md:w-1/4">
```
bk3c: I didn't know this was an HTML element before today.

bk3c:
I also want to bundle the UI components into a single module export at the root to keep the imports under control, especially since I don't feel like we need to worry much about supporting tree shaking or anything too advanced. For another PR once this one gets merged, but putting it out there so we all remember.

apps/public/components/listings_list.tsx
callaghanc: in group discussion, decided to go with components that are more generic - e.g. ImageCard instead of PropertyTitleImage - and to in the future keep assessing whether we may want to create more listing-specific components

PR #12, opened 2019-08-09
Fill in Listing Sidebar components

jaredcwhite:


bk3c:
Tested locally, and otherwise looks good.

PR #13, opened 2019-08-12
Prettier the existing codebase and add a git hook for prettier + eslint

bk3c:
- restore the table below the image that I messed up in the last merge
- run prettier against all of the TSX files so we have a good baseline
- add husky and set up git hooks to run prettier + eslint automatically

Note: If everyone's environment is set up correctly, lint-staged shouldn't actually do anything because both eslint and prettier should be integrated with the IDE / code editor.


jaredcwhite:
Hi all — I feel like some of the stylistic choices it's applying regarding whether a bit of JSX is contained in one line or not, etc., are pretty constraining/arbitrary. Also seeing a ton of semicolons creep back in…anyway, all that to say I'd love to discuss some of the options for what we are looking for in a linter before merging this in.

bk3c:
I have to admit that I'm in the pro-semicolon camp, but all of it is certainly open for discussion. Let's take a little time during our next meeting to talk it through.

jaredcwhite:
Yeah, I guess the jury is still out on semicolons in JS/TS…did a little searching around and everyone's got a different opinion. Next examples have them, Gatsby don't… I always used them up until about a year or so ago, then I got in the habit of dropping them and I've rarely run into any problems with it.

Anyway, here's some ammo on the against side I came across ;)
https://medium.com/@eugenkiss/dont-use-semicolons-in-typescript-474ccfe4bdb3

callaghanc:
I am still making up my mind on semicolons and therefore will accept either way for now. I hate having to type them, but in my IDE I use Prettier to autoformat so that's mostly taken care of. Visually I find it easier to look at code from a high level without semicolons because it's less cluttered, but when one has to get into it a bit I sometimes prefer semicolons because they can make it more clear where things end. Looking forward to discussing this and also the JSX formatting in our next meeting.

apps/public/lib/unit_summaries.tsx
```diff
@@ -1,14 +1,26 @@
...
+import * as React from "react"
 
 export const unitSummariesTable = (listing): Array<Object> => {
   const unitSummaries = listing.unit_summaries.general.map(unitSummary => {
     return {
       unit_type: <strong>{unitSummary.unit_type}</strong>,
-      minimum_income: <><strong>${parseInt(unitSummary.min_income_range.min, 10)}</strong>/month</>,
-      rent: <><strong>${parseInt(unitSummary.rent_range.min, 10)}</strong>/month</>,
-      availability: <><strong>1</strong> unit</>
+      minimum_income: (
```
bk3c: I agree this was more concise by hand

bk3c:
I switched to no semicolons and reran everything per our meeting decision. Everything still appears solid with a merge from master.

PR #14, opened 2019-08-12
Install Jest and Storyshots and create snapshots

callaghanc:
This PR:
- Sets up Jest to work with Typescript in shared/ui-components.
- Also sets up Storyshots with Jest to generate snapshot tests from the Storybook stories.
- Adds one example non-snapshot Jest test.

Still in progress, pending some updates and discussion.

Question for all: How would we like to organize the tests in shared/ui-components? There are two general ways I'm considering at the moment:

  1) All tests are in a top-level `__tests__` directory. The tests in the directory can be organized into subdirectories. Stories stay in the stories folder.
  2) All component tests are kept next to the component they test. Or, perhaps next to the stories they test in the case of Storyshot snapshots? Should a component + stories about it + snapshots + jest tests all live next to each other?

I don't feel very strongly either way yet. I'll have more of an opinion as I continue to work with the Jest/Storybook/Storyshoots combo and see what patterns emerge. What are your thoughts on this topic and are there other organizations you would like to consider? This is easy to re-organize as we go along so not a super critical decision.

Additionally, I'm looking into how the snapshots will work once there are many components. Right now the snapshots appear to be stored by Storyshots all in one file, so I'm looking into whether Storyshots continues that approach regardless of number of components or whether it breaks up the snapshots into multiple files.


callaghanc:
Added Storyshots configuration to generate one snapshot per story file, so we don't have one giant snapshot file for all stories. We can further adjust the Storyshots configuration as needed.

jaredcwhite:
Looks awesome, I'm excited to dig into this more. I don't have a strong preference either in terms of test location, although I'm slightly leaning in your (2) direction just because I noticed that convention in some other pattern library-style code bases and it seemed pretty nice when the tests are 1:1 with a single component.

bk3c:
I also lean toward tests in the same place as components for this kind of component library. I think it's more of a coin flip within an app or service (where the tests might lean a little more toward e2e or multi-component), but especially given a non-trivial directory structure it seems nice to not navigate twice.

callaghanc:
Cool, well I suppose I have a tiny leaning towards the same direction, not really for any good technical reason yet but rather for the novelty value because I haven't worked with that arrangement before and I know it's super widely used, so I'm interested to try it out for my own curiosity. If we all have a slight leaning in that direction let's go with that for now and see how it goes. I can see about adjusting the config to allow components, stories, and snapshots to cohabitate.

PR #17, opened 2019-08-21
Add in the individual units tables

jaredcwhite:
The tables should be in pretty good shape, and I even got a rudimentary version of the click-to-expand functionality we have on the current site—but I'm sure we'll want to replace that with something more modular/accessible. This is proof-of-concept stage…


shared/ui-components/src/types.ts
```diff
@@ -15,4 +15,5 @@ export interface Listing {
   developer: string
   credit_history: string
   rental_history: string
+  units: any
```
bk3c: I'd love to define this type somewhere/how and not have it be a complete "any"

shared/ui-components/src/page_components/unit_tables.tsx
```diff
@@ -0,0 +1,103 @@
...
+  if (lastDigit >= 1 && lastDigit <= 3 && !(lastTwoDigits >= 11 && lastTwoDigits <= 13)) {
+    suffix = oneToThreeSuffixes[lastDigit - 1]
+  } else {
+    suffix = standardSuffix
+  }
+
+  return `${num}${suffix}`
+}
+
+const unitFloorRange = (units: any) => {
```
bk3c: maybe getUnitFloorRange or unitGetFloorRange? It took me a minute of wondering why we needed an object to represent the floor and area ranges

services/listings/listings/triton.json
```diff
@@ -76,6 +76,56 @@
...
+      "num_bedrooms": null,
+      "number": "265",
+      "priority_type": null,
+      "reserved_type": null,
+      "sq_ft": 750,
+      "status": "available",
+      "unit_type": "one_bdrm",
+      "created_at": "2019-06-03T17:42:43.134Z",
+      "updated_at": "2019-07-10T16:43:16.766Z",
+      "listing_id": 3,
```
bk3c: probably for another PR, but I do think we should have some sort of externally-facing GUID for key objects; both for references like this, but also for the URL bar, etc. I've had good luck with https://github.com/ai/nanoid, but there are many options

bk3c: Now tracked in #18.

shared/ui-components/src/page_components/unit_tables.tsx
```diff
@@ -0,0 +1,103 @@
+import * as React from "react"
+import { BasicTable } from "@dahlia/ui-components/src/tables/basic_table"
+
+const unitTypeLabel = unitType => {
+  const unitTypeLabels = {
```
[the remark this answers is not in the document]
jaredcwhite: Yeah I'm not keen on this approach either, but I wasn't sure where we actually do want this logic. Thanks for pointing out this could very well be something we push to the backend.

shared/ui-components/src/types.ts
jaredcwhite: Yeah I'm getting lazy about the typing on JSON-backed objects since I'm not really sure what the units format is going to be.

bk3c:
I can't tell if this is somehow environmental for me, but it seems like the listing page is broken? It just returns a 404.

shared/ui-components/src/types.ts
```diff
@@ -32,3 +34,39 @@ export interface Address {
...
+}
+
+export type UnitGroup = [string, [Unit]]
+export type GroupedUnits = UnitGroup[]
+export interface UnitGroupSummary {
+  unit_type_label: string
+  area_range: string
+  floor_range: string
+}
+export type UnitGroupWithSummary = [string, Units, UnitGroupSummary]
```
bk3c: Is this supposed to be Unit instead of Units?

PR #23, opened 2019-08-23
Refactor Polyglot support into a global helper

jaredcwhite:
Here's an alternative approach to using Polyglot throughout the system. It does use a "global" variable which is usually a code smell, but in this case since a translation layer really is something you want to make available globally, I would argue it's an acceptable tradeoff.

I also abstracted out the Polyglot support so all you need to do is call `t("whatever")` and you're good to go…no need to know that Polyglot is used under the hood per se.

Thoughts? Suggestions?

P. S. Also experimenting with submitting this PR from my GH fork :)


bk3c:
Some reading on the subject to consider as I was looking through the approach:

- https://www.jonathan-petitcolas.com/2017/03/27/internationalizing-react-application-using-polyglot.html
- https://github.com/airbnb/polyglot.js/issues/25

As the example in the second link starts to show, I don't think the polyglot is going to work with SSR out of the box since it normally wants to grab the locale from the browser

bk3c:
It's young and poorly named, but maybe also worth looking at https://github.com/koala-interactive/frenchkiss.js, since it seems built for SSR by folks who use nextjs.

jaredcwhite:
@bk3c I was able to add a couple of new pages (`/es` homepage and `/es/listings`) and essentially reuse the original page components but with adding in a Spanish translation before they get rendered. Seems like it would be pretty straightforward to add in other languages in a similar fashion.

bk3c:
@jaredcwhite some more food for thought:
- https://github.com/zeit/next.js/issues/2665
- https://github.com/zeit/next.js/issues/5509

jaredcwhite:
@bk3c Thanks Ben — that second one with the exportPathMap example looks like it's very much what Cat was suggesting, so I'll give it a whirl!

jaredcwhite:
@bk3c OK, that approach worked! Now the language-specific routing is fully dynamic…just a couple of code changes are needed to add a new translation and that's it.

bk3c:
@jaredcwhite - Everything seems to be working as designed, but the language isn't reflected in the links -- e.g. if you are on the spanish home page and you click on a link, that will revert you to english. Do you want to merge this and handle that in a new PR, or is that small enough to knock out in this one?

jaredcwhite:
> @jaredcwhite - Everything seems to be working as designed, but the language isn't reflected in the links -- e.g. if you are on the spanish home page and you click on a link, that will revert you to english. Do you want to merge this and handle that in a new PR, or is that small enough to knock out in this one?

Good catch! That sounds like a small-ish fix, so if you're cool with it, LGTM  and I'll file an issue to work on for the links.

PR #33, opened 2019-09-04
Listing sidebar

software-project:
Adds sidebar to listings


jaredcwhite:
@software-project This looks cool! One item I'll mention is another PR up is up now which uses a dedicated `Button` react component, so we'd probably want to switch from the `.btn` markup/styling and utilize `Button` instead. Not sure about the timing on that…

PR #47, opened 2019-09-12
Use locale-specific links to keep using the same language

### LINKED ISSUE #32: Ensure locale-specific routes are used
If you're using a language other than English (aka `/es`) and go to the Listings page, it should stay with the same locale by going to `/es/listings`, not just `/listings` which would switch you back to English.

jaredcwhite:
Fixes #32


jaredcwhite:
@bk3c I really liked the LocalizedLink component idea, so I worked that into the mix and now things feel even cleaner. Abstraction FTW! 🤜 🤛

Issue #18, opened 2019-08-21
Use UUIDs to refer to listings and other key objects


bk3c:
As we start having interdependencies between services that don't share the same database, it's going to be important to have stable, universal, readable IDs for any object in the system that will be persistently referenced outside of 1 backend service.

Some implementation options include:
- https://github.com/ai/nanoid (my preference)
- https://github.com/dylang/shortid
- https://github.com/kelektiv/node-uuid

jaredcwhite:
Closed as part of #49

PR #49, opened 2019-09-17
Add an Occupancy table to the Listing page

jaredcwhite:
This is working pretty well now, but some of the code is basically a port from the old CoffeeScript so I want to refactor a little, improve the typing, and maybe add a Storybook definition as well.


jaredcwhite:
Fixes #49 — and also #18 (since I was in the guts of unit and table stuff anyway)

apps/public/lib/getOccupancyDescription.tsx
```diff
@@ -0,0 +1,15 @@
+import t from "@bloom/ui-components/src/helpers/translator"
+import { Listing } from "@bloom/ui-components/src/types"
+
+const getOccupancyDescription = (listing: Listing) => {
```
bk3c: My only question here is whether this should be a function with "get" in the name here, or whether it can just be a subcomponent that renders in?

services/listings/listings/gish.json
```diff
@@ -58,20 +58,20 @@
...
-      {
-        "min_income_range": { "min": "2135.0", "max": "2135.0" },
-        "occupancy_range": { "min": 1, "max": 2 },
-        "rent_as_percent_income_range": { "min": null, "max": null },
-        "rent_range": { "min": "854.0", "max": "854.0" },
-        "unit_type": "Studio"
-      }
-    ]
-  },
+  "unit_summaries": [
```
bk3c: Do we think this summary information could be generated from the base unit information? I'm fine if we want to file an issue and push doing so to another PR, but storing the info twice seems like asking for trouble. (Even in the relatively short term given the likelihood of replacement by a DB-backed system once we have editing in a partners portal)

bk3c:
TritonWaitlist

apps/public/lib/getOccupancyDescription.tsx
jaredcwhite: I'm not sure what you mean here. It's just returning a translated string…so not really rendering anything per se.

services/listings/listings/gish.json
jaredcwhite: Good call, I imagine we want the backend to just generate that, right? Sounds like something I could work on for the next PR

Issue #42, opened 2019-09-10
Create the mobile state for the open listing sidebar


callaghanc:
An open listing is a listing whose application due date is in the future. On mobile, the content of the sidebar on an open listing page should be displayed differently than on desktop.

On mobile:
 - The application due date section of the sidebar should appear below the page's listing image header and above the unit table.
- All sections of the sidebar **except** for the application due date, What to Expect, and Contact sections should appear between the unit table and the listing details accordion.
- A section titled Process should appear in the listing details accordion after the Eligibility section. That section should contain the What to Expect and Contact sections from the sidebar. See the attached screenshots for the image that should be used in the Process accordion section header - you can get the image from either a regional DAHLIA site or from the SF DAHLIA site.

<img width="594" alt="Screen Shot 2019-09-11 at 6 20 44 PM" src="https://user-images.githubusercontent.com/1117776/64809754-cf15f980-d54e-11e9-937a-2f72fa718923.png">
<img width="585" alt="Screen Shot 2019-09-11 at 6 21 54 PM" src="https://user-images.githubusercontent.com/1117776/64809756-cf15f980-d54e-11e9-922f-6256508568d4.png">
<img width="585" alt="Screen Shot 2019-09-11 at 6 22 10 PM" src="https://user-images.githubusercontent.com/1117776/64809758-cf15f980-d54e-11e9-9e4c-f82cbe761c83.png">
<img width="586" alt="Screen Shot 2019-09-11 at 6 22 29 PM" src="https://user-images.githubusercontent.com/1117776/64809759-cf15f980-d54e-11e9-9153-bf20d2c11a10.png">

callaghanc:
Notes for rewriting this issue and perhaps breaking out another issue:

- for both open and closed listing mobile states, have application date component in the top placement (see sf dahlia closed listing for example)
- on mobile, for open listing, two sidebar sections appear outside of accordion: the apply section and the need help section
- on mobile, for closed listing, the sidebar apply section and need help section do not appear
- for both open and closed listings, on mobile, the accordion section with sidebar components will be called Process and will appear as the second accordion item after Eligibility
- content of mobile accordion item representing sidebar is everything from the sidebar minus apply and need help sections

PR #51, opened 2019-09-30
Mobile sidebar

software-project:
Fixes mobile side bar from issue #42

### DIFF
--- a/apps/public/pages/listing.tsx
+++ b/apps/public/pages/listing.tsx
@@ -82,17 +82,21 @@ export default class extends Component<ListingProps> {
               </>
             }
           />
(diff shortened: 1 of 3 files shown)


bk3c:
I am going to merge this because I think it's a step in the right direction and I want to unblock the other PRs behind it, but I noticed some layout things that @software-project should take a look at (unless @slowbot has a different opinion or wants to handle them):

* padding in the sidebar boxes sometimes seems insufficient in mobile mode, and also doesn't seem like enough the main text body in "non-mobile" but smaller width windows
* there's no down-arrow to open the accordion, so you'd have to get lucky tapping. cf. the current screen on SF DAHLIA which looks a little better from my POV.

@slowbot what's the best workflow to track these kinds of things for you? Would it work if I just filed a follow-up Github issue, or would you prefer otherwise?

PR #55, opened 2019-10-02
Component consolidation WIP

slowbot:
Adding custom config settings
Creating component classes for various headings and lists elements

tailwind config settings are not compiling down to the css level


jaredcwhite:
@slowbot Just a quick overview here on GH…it definitely looks like everything should be "working right", so I'll check out this branch locally and mess around with the Tailwind install to ensure it can get built correctly.

PR #59, opened 2019-10-04
Fix unit summaries for listings directory

software-project:
Unit summaries got changed after I worked on then and path have been changed. This one updates the path.


bk3c:
Tested locally and LGTM.

PR #61, opened 2019-10-09
Responsive navbar

jaredcwhite:
Pull in a navbar component from Bulma and ensure Tailwind vars show up as available Sass vars for Bulma and other potential uses in the future


bk3c:
I'm fine with the Bulma approach, but I'd like to add a little more documentation so future developers don't get confused by the two CSS frameworks. See my commit below for a starting point that you (and I imagine @slowbot eventually) should feel free to extend as much as is reasonable, whenever suits your schedule.

bk3c:
Also, from local testing: the top "This is a preview..." text gets smashed and hidden at narrow window sizes. I'm fine if you want to punt that to another PR, just file an issue if so.

jaredcwhite:
@bk3c Doc looks good, I'll think about any additions I could make there. About the preview text, I noticed that's been wonky from the get-go. I'll work on it before we close out this PR.

Issue #60, opened 2019-10-07
Investigate merging the styles and ui-components packages


jaredcwhite:
Given the high degree of coupling between styles and ui-components and the desire to move individual stylesheets into component folders, this issue investigates how to merge the two and evaluating that change before a final merge commitment.

PR #71, opened 2019-10-18
Merge the styles module into ui-components

jaredcwhite:
Implementation of #60

So this is looking pretty good to me overall. I did notice one warning in the CSS build process which I think is simply around the fact that component-specific SCSS files importing a shared resource (in this case `mixins.scss`) can trigger a warning around source ordering in the final CSS bundle. But functionally everything seems to be working fine.


shared/ui-components/src/headers/SiteHeader/SiteHeader.tsx
```diff
@@ -53,7 +53,7 @@ class SiteHeader extends React.Component<SiteHeaderProps, SiteHeaderState> {
     )
   }
 
-  handleMenuToggle() {
+  handleMenuToggle = () => {
```
bk3c: @jaredcwhite This is a weird-to-me pattern that the linter wants so that there's not a reference problem when you call bind() that I don't completely understand. I don't think I broke anything. See also https://palantir.github.io/tslint/rules/no-unbound-method/, and maybe you can explain it to me.

jaredcwhite: @bk3c Looks weird, but it seems to work so let's roll with it. Merge away! (I also removed a couple extra file references from index.scss…holdover from the previous merge.)

PR #74, opened 2019-10-19
Add a basic CircleCI configuration and update testing-related dependencies

### LINKED ISSUE #69: Configure Circle CI
After some research and polling of the team, we've decided to move forward with Circle CI instead of semaphore, which (entirely) coincidentally is also the SF DS standard. I've enabled it for the repo but done no configuration, so the work here is to do the basic setup per their docs to have the root project test:all target run in CI.

My knowledge of Circle is relatively minimal, so I'm more than happy to entertain other configuration suggestions from anyone with both a well-formed opinion and a desire to do the work.

Extra credit for making sure that all the tests pass at the end of this ticket.

bk3c:
The red X by each PR made me sad, so this is a start. Fixes #69.


bk3c:
This is the most basic configuration I could come up with. Lots to do from here, but it's a start.

bk3c:
Merging now, so we can start to have working CI.

PR #79, opened 2019-10-22
Use env vars to configure data service URL and local development ports/URLs

jaredcwhite:
This is a pretty straightforward PR, but there are a couple of Bash shell voodoo tricks in order to use env vars within the package.json commands, so happy to discuss that approach. ~Also, I added some rule comments at the top of next.config.js so we don't get eslint errors every time we touch that file (yay!).~ Ah, looks like somebody else beat me to it in master. 🤓


jaredcwhite:
Looks like we got a bunch of branches based on this PR now, so unless I hear any "nays" to the contrary, I'll go ahead and merge this in later today.

akegan:
@jaredcwhite when I talked to @bk3c yesterday he mentioned that the only update he wanted before merge was to rename the env var from DATA_SERVICE to LISTING_SERVICE since we're anticipating adding more services in the near-ish future

jaredcwhite:
> @jaredcwhite when I talked to @bk3c yesterday he mentioned that the only update he wanted before merge was to rename the env var from DATA_SERVICE to LISTING_SERVICE since we're anticipating adding more services in the near-ish future

Sounds good, I'll change that now.

PR #81, opened 2019-10-25
Change json and variables to camel case

### LINKED ISSUE #67: remove snake_case from all json and variables
We're a good chunk of the way there based on the work in #63, so this should be a straight-forward chore for someone to knock out when they need a lower-brainpower task and still remember search regexes.

software-project:
Fixes #67 
Replaces jsons and variables with snake case. Includes also translations.


jaredcwhite:
Looks good to me! :+1:

Issue #66, opened 2019-10-16
Ability to list different submit application instructions for each listing


jacchau:
AC: 
1. A listing can either (1) Not accept postmark OR (2) Accept postmark 
2. If the listing (1) Does not accept postmark, then display 'Applications must be received by the deadline and postmarks will not be considered.' in the 2. Submit a Paper Application instructions
3. If the listing (2) Accepts Postmark, then display the following 'Applications must be received by the deadline. If sending by U.S. Mail, the application must be postmarked by [Application Due Date] and received by mail no later than [Postmark Due Date]. Applications received after [Postmark Due Date] via mail will not be accepted even if they are postmarked by [Application Due Date]. [Developer] is not responsible for lost or delayed mail.'

Notes:
Application Due Date, and Developer should be existing fields. Postmark Due Date and the Accepts Postmark? should be new fields that need to be created.

Example of Submit Application instructions on Archer Studios:
https://sj.housingbayarea.org/listings/5
Applications must be received by the deadline. If sending by U.S. Mail, the application must be postmarked by November 1, 2019 and received by mail no later than November 5, 2019. Applications received after November 5, 2019 via mail will not be accepted even if they are postmarked by November 1, 2019. Charities Housing is not responsible for lost or delayed mail.

Pivotal corresponding issue for SJ Bloom: https://www.pivotaltracker.com/story/show/169185822

jaredcwhite:
Completed with #82

PR #82, opened 2019-10-26
Ability to show postmarked application message

jaredcwhite:
Fix for #66. This replicates the work recently done for Regional Dahlia.

NOTE: This includes the work done for #79, so best to merge that first. Also includes a fix for a past merge problem on master that resulted in missing sidebar content.


jaredcwhite:
@bk3c I've rebased from master and updated the labels for the various fields/translations, so I'm hoping that's more clear now. Let me know!

bk3c:
@jaredcwhite Everything looks good to me code-wise. Can we make an example out of this PR and see if we can get some tests worked in here? I'm conscious of the amount of infrastructure we still need as @akegan is working on tests for #84, so maybe this is an opportunity for the two of you to team up / work in parallel?

I don't think we need extensive checking of every edge case, but it seems like we're just about at the point where having some tests in every new feature PR is something we should start to expect.

jaredcwhite:
@bk3c Yep, I'll confer with her about the direction we're going on the tests so I can continue along the side lines. Were you thinking E2E tests for this as well as Storybook, or you think Storybook is sufficient? (Given it's mainly just text differences, I'm inclined to think SB is fine, but YMMV)

jaredcwhite:
@bk3c One step forward: got Storyshot tests working for this, so that's cool.
One step back: spent a couple hours banging my head against the Typescript/Webpack wall trying to boot up Storybook. At this point it's down to issues relating to the JSON data and importing it in as a fixture for the stories. Hopefully I can get that nailed down pretty soon.

shared/ui-components/src/page_components/listing/listing_sidebar/Apply.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+import Apply from "./Apply"
+import Archer from "@bloom-housing/listings-service/listings/archer.json"
+
+const listing = Object.assign({}, Archer)
+
+storiesOf("Listing|Sidebar Apply", module).add("hard application deadline", () => {
+  listing.applicationDueDate = "2021-11-30T15:22:57.000-07:00"
+  listing.acceptsPostmarkedApplications = false
+
+  /* eslint-disable @typescript-eslint/ban-ts-ignore */
```
jaredcwhite: I had to disable TS checking here because otherwise _tons_ of errors were coming up related to the JSON data for listings/units. I initially started to try to fix those errors, but then ran into errors elsewhere in the listing service and website, so I think that should be filed as a separate issue.

jaredcwhite:
Big update on the branch now — I got Storybook and storyshots and Typescript to play well together (no small feat!), and based on a suggestion from Andrea I relocated the translation files to the ui-components module so that Storybook and the tests had full access to the translations (plus multiple downstream apps can utilize translations directly from the module).

Happy to discuss further on tomorrow's call.

PR #65, opened 2019-10-16
Shared interfaces

software-project:
Bases on PR #63 
Fixes #28

PR #84, opened 2019-10-29
Add housing counselors page

akegan:
This PR:
- Adds a housing counselor component
- Adds a housing counselor page and a link to it via "Get Assistance" in the nav bar. 
- Updates page header to accept a subtitle prop
- Fixes disclaimer page to not 404 and cause infinite reloads

Still TODO:
- [x] Finish transition to new types file structure (once packages are working again)
- [x] Pull housing counselor url into an env var once #79 is merged
- [x] Determine how we want to lock down props on page headers.


shared/ui-components/src/headers/page_header/page_header.tsx
```diff
@@ -1,6 +1,13 @@
 import * as React from "react"
 
-const PageHeader = (props: any) => {
+interface PageHeaderProps {
+  inverse?: boolean
+  subtitle?: string
+  // TODO: should this be locked down to string?
+  children: any
```
akegan: @jaredcwhite Ben mentioned that you've been working to lock down props and types. Any thoughts on what we should do for this page header component?

Seems like options are:
1. Specify title as a specific prop, lock it down to string
1. Specify the type of children as something other than any (not sure what) and continue passing in children

shared/ui-components/src/types.ts
```diff
@@ -1,103 +0,0 @@
-export interface Listing {
```
akegan: Not sure why this wasn't deleted in #65 @software-project any reason why it wasn't deleted in particular?

shared/ui-components/src/headers/page_header/page_header.tsx
jaredcwhite: @akegan Yeah, I think we've been using `React.ReactNode` most places rather than `any` in recent work and that seems to work pretty well (supports a string as well as JSX markup). I'd prefer having the title be inside the component tag rather than use a title prop but that's just my take.

akegan: ok, updated

akegan:
@bk3c or @jaredcwhite do you have any feedback on tests I could/should add?

bk3c:
I agree with adding it to the Navbar, but would also appreciate a @slowbot opinion. My default would be that we should expose 100% of functionality in the reference implementation so it's a true feature reference, and then each implementation will remove as-desired.

bk3c:
@akegan with regards to tests, the ideal thing to do would be do have a file that reliably mocks a couple of data cases, and then have cypress tests that evaluate correct rendering of that data. We don't really have the mock data pipeline figured out with a good example yet, so maybe @software-project or @jaredcwhite can jump in and help with that if it's more than you want to take on right now. The cypress part should be relatively straight-forward. 

One other thing that might be good is to have the app check for a null HOUSING_COUNSELOR_SERVICE_URL, and then not render the navbar if that's the case. Extra credit, and totally deferrable, but I think that's the right mechanic.

akegan:
OK, thanks for the review. I'll update to show the nav bar item if the URL is present, and look into tests. I might ask to pair with @software-project or @jaredcwhite on the tests if I don't get very far

akegan:
I've pushed up stories with storyshots, however they include the translation missing issue.

apps/public/cypress/integration/navigation.ts
```diff
@@ -40,7 +40,7 @@ describe("Navigating around the site", function() {
     cy.contains("Endorsement Disclaimers")
 
     // Click on the listings page link in the header nav
-    cy.get("header a")
+    cy.get(".navbar")
```
akegan: Looks like naming of this section changed and tests were broken on master, so I updated "header" to "navbar"

shared/ui-components/src/page_components/__snapshots__/HousingCounselor.stories.storyshot
```diff
@@ -0,0 +1,134 @@
...
+      href="www.counselor.org"
+      target="_blank"
+    >
+      Counselor Name
+    </a>
+  </h3>
+  <p
+    className="text-sm text-gray-700 pb-3"
+  >
+    {{ Missing Translation Phrases }}
```
akegan: example translation missing issue

jaredcwhite:
@akegan I forgot the final outcome of yesterday's conversation — did you want to me to merge this in as is or were you going to change something in the tests first? Let me know…thanks!

akegan:
As-is is fine with me for now until we resolve the translations question.

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

jaredcwhite:
All done! 👍

PR #88, opened 2019-11-02
Set up for Heroku deployment of the listings service

bk3c:
In the short term, this will power the front-end at bloom.exygy.dev, but it's also a (trivial) template for Heroku deployment of services more generally.


bk3c:
Holding on #89, since Heroku can't resolve the current package names.

bk3c:
This ended up being just adding the Procfile + a bunch of other cleanup that we should have done sooner.

bk3c:
Note: the listings service is now running at listings.bloom.exygy.dev.

PR #96, opened 2019-11-18
Configure cypress for circleci

software-project:


bk3c:
Ok, this was long-fought to get working, but I think it's now in a place where the tests are (unfortunately) legitimately failing and it's not an infrastructure problem. @jaredcwhite, I know you can't really test locally, but if you can do a quick code read and make sure the current CircleCI looks like it's back on track to being fixable, then I'll get this merged and we can move on to that.

PR #111, opened 2019-12-06
Additional Information section

### LINKED ISSUE #104: Add to and improve the Additional Information section
The "Required Documents" heading is missing and the "Important Program Rules" section is missing. See attached screenshot from previous codebase:

Screen Shot 2019-11-22 at 4 02 00 PM

jaredcwhite:
Fixes #104

### DIFF
--- a/apps/public/pages/listing.tsx
+++ b/apps/public/pages/listing.tsx
@@ -227,8 +227,15 @@ export default class extends Component<ListingProps> {
             >
               <div className="listing-detail-panel">
                 <div className="info-card">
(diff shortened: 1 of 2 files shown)


bk3c:
This LGTM to me and tests good locally, but I just want to check with @slowbot before we merge it, since the resulting interface seems confusing. My concern is how the various rules are listed in multiple place on the page, and it might be as easy for the user to get confused as it was for me doing the QA. Is this something we want to re-evaluate for Bloom?

bk3c:
Ok, met with Jesse about this and we're going to just update the text strings in a future PR but keep the layout the same for now.

PR #122, opened 2020-01-03
Add meta tags for page title and social graph

### LINKED ISSUE #105: Add Open Graph Tags Preview Image, Title, and Description for shared links
Add Preview Image, Title, and Description for shared links on social (i.e. Facebook, Twitter, Linkedin, iMessage)

AC:
The homepage, listings page, and individual listings have the following previews: 

**Homepage**, **Listings Page**, **Privacy Page**, and **Disclaimer Page**
Image: Hero Image
Title: [City/ County] Affordable Housing Portal
Description: Bloom Housing Portal: Search and apply for affordable housing in [City/ County's], built in partnership with Exygy.

**Individual Listing**
Image: Listing Image
Title: [Building Name]
Description: Bloom Housing Portal: Apply for affordable housing at [Building Name] in [City/ County], built in partnership with Exygy.

jaredcwhite:
Fixes #105


jaredcwhite:
> This looks generally good, but I'm wondering if we should put all of the meta tag syntax in a component and pass in params as needed? Maybe not since the (single) listing page is different enough but it seems like a block we're likely to edit and it'd be good not to do it in 3+ places. I'm fine with either way, but just want to make sure we consider it.

Good point…I think we could abstract this out a bit. I actually found a third-party component called Next SEO that is sort of along these lines, but it seemed maybe overkill to pull that in for our pretty straightforward purposes.

jaredcwhite:
@bk3c All right, I abstracted out a component for this so it's cleaner in the page code. Let me know what you think.

shared/ui-components/static/locales/general.json
```diff
@@ -44,7 +44,7 @@
   "listings": {
     "additionalInformation": "Additional Information",
     "allUnits": "All Units",
-    "allUnitsReservedFor": "All units reserved for {{type}}",
+    "allUnitsReservedFor": "All units reserved for %{type}",
```
bk3c: Did you mean to make all of these changes in string interpolation syntax? I don't think I have a strong opinion about what we use, but the %{str} format seems to be pretty confined to ruby.

jaredcwhite: @bk3c So that's actually the syntax for the Polyglot library…those interpolated sections never got updated from the old Angular stuff. Figured I'd do it while in there messing with page titles.

bk3c:
@jaredcwhite good to go here as well, based on our conversation. Feel free to resolve the merge conflict and merge at will.

PR #139, opened 2020-01-18
Upgrade to React 16.12 + Next 9.2 + Storybook 5.3

bk3c:
This PR upgrades basically all of the front-end dependencies to their latest versions. It started as Storybook-only, but that broke the build (I think due to excessive tree-shaking by the next build command) so I combined what was about to be the following PR into this one.

@jaredcwhite There are some changes / new features this enables, notably around CSS handling. Can you investigate a little bit and if it looks like it's worth making any further changes just add them on to this branch and I'll review, or make yourself an issue for later. #141 is in that category.

Ref: https://github.com/martpie/next-transpile-modules/releases/tag/3.0.0 and https://nextjs.org/blog/next-9-2


bk3c:
An implementation note and future TODO here on the Storybook side: I spent a decent amount of time trying to clean up our Storybook configuration to match the new default patterns they're moving to, and I think I've mostly gotten there. There's still some commented out things to remove, and I think we can actually remove even more config, but I want to take in slow in the interest of breaking as little as possible.

On that note, jest storyshots for UI components are still broken as I write this with a `Cannot find module '@storybook/react' from 'loader.js` error that I can't seem to get to the bottom of. I think I'm willing to merge this anyway so that we can stay current on everything else, and start to use the addon-docs, which seems a good match for our use cases. That said, @jaredcwhite I'd be excited if you figured out a fix without banging your head on it too much (and without having to revert back to the full babel transpile).

Also, I couldn't figure out which story actually had a translation in it so I could look at that. Maybe we should add a quick hand-testing checklist somewhere, for future-us?

jaredcwhite:
@bk3c Sure, taking a look now. Would it be OK to update at least one story to the new format in v5.3 (since that's a big reason to upgrade)? I'd love to make sure we're ready to move forward along those lines.

jaredcwhite:
@bk3c OK, so I churned for a while on the `Cannot find module '@storybook/react' from 'loader.js'` error because it's just so strange and seems like it shouldn't be happening. Tried a bunch of different tweaks to various configs, package.json, searching for all sorts of possible solutions in SO/GH/etc. … even tried debugging the exact JS file buried in node_modules that was throwing the error. Simply wasn't able to resolve it. I think maybe we should try to file an issue over at storybook and see if anyone has any ideas.

PR #151, opened 2020-02-06
Listings loader and filtering via jsonpath

jaredcwhite:
I looked into JSONQuery and JSONPath and the latter seemed pretty solid to me, so I've implemented that here. All listings come back from the endpoint by default, but if you supply a `jsonpath` query param, it'll filter down the array accordingly (an example URL is in a comment). Let me know what you think.


jaredcwhite:
https://5e3b7a2b930cbe1dcc444f9c--clever-edison-cd22c1.netlify.com/

jaredcwhite:
@bk3c Tests added and are passing now!

PR #152, opened 2020-02-06
rename app directories to include -reference

bk3c:
This PR just changes the directory names for the public and (future) partners apps to be public-reference and partners-reference instead. This is to support downstream forks using a directory structure like:

apps/
|--public-reference
|--public-jurisdiction1
|--public-othersite

where they can continue to keep the reference directory easily current with upstream, and then diff/merge from there to individual sites as appropriate.

Aside from the actual rename, I cleaned up as many references as I could find, and fixed a couple of minor linting flags in config files where we didn't have /*eslint-env node*/ set.

Note that I did *not* change the yarn targets, just the directories. It doesn't seem likely to cause confusion as-is, and less typing is better.

MERGE REMINDER: As soon as we merge this, we'll need to go update netlify to look in the right place for what to deploy.


jaredcwhite:
@bk3c Looks good to me!

bk3c:
@jaredcwhite I'm going to merge anyway, but next time if you can use the approve button (hidden on the "files changed" tab), then it will give the whole thing a nice green check.

jaredcwhite:
Will do!

PR #156, opened 2020-02-12
Support external online application urls

### LINKED ISSUE #154: Ability to Apply Online by directing to a third party site
**Summary:** Alameda County's pilot Eden Housing will work with a third party vendor to allow for application intake to occur online. Alameda Bloom needs to allow for the application button to redirect to this third party site

**User Story:**
As a property manager
I want to redirect users to a third party vendor for application intake within the listing
So that users can apply online using a third party service

**Acceptance Criteria**
- Admin can indicate that a listing is accepting online applications
- Admin can indicate the link to the third party service URL for online application intake
- Given the listing accepts online application, when the user views the listing, then the 'Download Application' button is changed to 'Apply Online' and when the user clicks on the 'Apply Online' button, it redirects the user to a third party service's 'URL'

**Notes**
Is the 'accepting_online_applications_boolean' being used? If not, would it make sense to repurpose that field?

jaredcwhite:
Fixes #154

### DIFF
--- a/shared/core/src/listings.ts
+++ b/shared/core/src/listings.ts
@@ -3,7 +3,8 @@ import { Address } from "./general"
 import { Preference } from "./preferences"
 
 export enum AttachmentType {
-  ApplicationDownload = 1
+  ApplicationDownload = 1,
+  ExternalApplication = 2
 }
 
 export interface Attachment {
@@ -15,6 +16,7 @@ export interface Attachment {
 export interface Listing {
   acceptingApplicationsAtLeasingAgent: boolean
   acceptingApplicationsByPoBox: boolean
+  acceptingOnlineApplications: boolean
   acceptsPostmarkedApplications: boolean
   accessibility: string
   amenities: string
--- a/shared/ui-components/src/atoms/Button.scss
+++ b/shared/ui-components/src/atoms/Button.scss
@@ -7,6 +7,7 @@
   @apply px-6;
   @apply py-4;
   @apply text-lg;
+  @apply text-center;
   @apply uppercase;
   @apply font-alt-sans;
   @apply inline-block;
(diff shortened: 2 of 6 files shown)


shared/core/src/listings.ts
```diff
@@ -3,7 +3,8 @@ import { Address } from "./general"
 import { Preference } from "./preferences"
 
 export enum AttachmentType {
```
bk3c: this should really be ApplicationType, no?

When we have applications in the system, nothing will be attached, per se.

bk3c: but both of the enum options look right to me.

shared/core/src/listings.ts
```diff
@@ -15,6 +16,7 @@ export interface Attachment {
 export interface Listing {
   acceptingApplicationsAtLeasingAgent: boolean
   acceptingApplicationsByPoBox: boolean
+  acceptingOnlineApplications: boolean
```
bk3c: do we need to maintain this as a separate boolean, or can we just derive it from the existence of an Application URL?

@slowbot - food for future thought: what would it mean for the UI in different combinations of application methods here? Could we have a listing that people could apply to in multiple ways (e.g. online or paper)? How should we render in that case? (note: we do not need to answer these questions now to merge, just want to surface them while the iron is hot)

shared/core/src/listings.ts
jaredcwhite: Well this is related to the earlier work we did to support file "attachments" and multiple versions at that (like PDFs for English, Spanish, etc.) and in lieu of using a third party service this is the current arrangement. It's not really an application type per se.

shared/core/src/listings.ts
jaredcwhite: re: the first question… I feel like some of these booleans could be set by a smarter backend rather than being data we're implicitly storing. Not sure how we want to handle it at present.

jaredcwhite:
> Other than that, it would be great to have a Cypress test here since I'd like us to get in the habit of testing conditionals, but I can live without it if that seems like too much work relative to the feature size.

If we have Storybook stories around each setting/value combo (and hopefully get those automated tests working again), do we need to write something separate with Cypress?

jaredcwhite:
Related improvements to be filed by @bk3c as new issues…

PR #160, opened 2020-02-19
Application marketing state

### LINKED ISSUE #153: New Pre-Application Listing State
**Description:** Create a 'Pre-Application' listing state for property managers/ developers to market their listings prior to when the application period is opened

**User Story**
As a property manager
I want to publish a 'Pre-Application' listing
So that the property manager can market the listing prior to the application period is opened

**Acceptance Criteria**
1. The admin can update the 'Application Open Date' on a listing.
2. When the 'Application Open Date' is less than system current date, then the listing is in 'Pre-Application' state.
3. When that the listing is in the 'Pre-Application' state, then the following fields are changed and hidden in the listing:
- change 'Application Deadline [Application Due Date] to 'Application Open Period from [Application Open Date] to [Application Due Date]' in the listing page
- Hide 'Download Application' button and replace with sub text 'Application will be available for download and pick up on [Application Open Date]
- Hide 'Contact Leasing Agent' section
- change 'Application Deadline: [Application Due Date]' to 'Coming Soon: [Application Open Date]'
- Hide 'Waitlist box on listing sidebar' component

jaredcwhite:
Fixes #153


shared/ui-components/src/page_components/listing/listing_sidebar/Apply.stories.tsx
```diff
@@ -27,7 +27,7 @@ storiesOf("Listing|Sidebar Apply", module).add("accepts postmarked applications"
   /* eslint-enable @typescript-eslint/ban-ts-ignore */
 })
 
-storiesOf("Listing|Sidebar Apply", module).add("shows multiple download URLs", () => {
+storiesOf("Listing Sidebar|Apply", module).add("shows multiple download URLs", () => {
```
bk3c: This story doesn't seem to render any differently for me than the baseline. Am I missing something subtle, or are we missing the data needed to power it correctly?

jaredcwhite: @bk3c I believe it will look the same unless you click the download button to see multiple download links. I'd be fine with making this a default story and removing one of the other ones if that simplifies things.

jaredcwhite:
> …If you can convert them quickly then I think it would be a win to do here, but we can just make a separate issue if you feel like that's too much.

I feel like we could file a separate issue that covers converting the older formats _plus_ adding new stories for some of the other components, and just generally rounding out what we're adding to Storybook.

shared/ui-components/src/page_components/listing/listing_sidebar/Apply.stories.tsx
bk3c: Ok, that was the subtle thing I was missing. No need to change the default, but maybe we can figure out a way to include instructions when we re-write the stories to CSF.

PR #163, opened 2020-02-21
Upgrade to jest 25 and fix Storyshots

bk3c:
Still working out the details here, but I think I finally figured out where we went wrong with the Storyshots jest configuration (even if how to fix it precisely is still eluding me). It seems like the jest custom loader needs to always be run from the root of a monorepo with pointers into the subprojects, not run isolated in each how we had it.

For reference, see:

- https://github.com/facebook/jest/issues/5420
- https://github.com/facebook/jest/issues/3112#issuecomment-398581705
- The [ts-jest documentation](https://kulshekhar.github.io/ts-jest/user/config/)
- The [jest config documentation](https://jestjs.io/docs/en/configuration)
- https://medium.com/@takamura.developer/yet-another-monorepo-setup-tutorial-34744315318b
- http://orlandobayo.com/blog/monorepo-testing-using-jest/ (I'm trying to just use ts-jest and remove the babel dependency and associated config, so this isn't a perfect analog)


jaredcwhite:
Well I came, I yarn installed, but I did not conquer. 🙁 

Tried to re-enable the storyshot tests but got the same error as before:
`Cannot find module '@storybook/react' from 'loader.js'`

jaredcwhite:
@bk3c Good news! I give this another go, and on a whim tried adding `@storybook/react` to the `devDependencies` of the root `package.json`, and it finally started working! I don't recall if I tried that in the past or not, but at any rate, it's now running through the stories. Cleaning up some test failures and then hopefully can commit a working PR.

bk3c:
I'm not sure this is actually the completely right solution, in part based on all of the warnings I'm getting for @sb/react's peerDeps not being there at the top level. If I had to guess, declaring it as a top-level dependency just causes the package to get hoisted to the root node_modules so it's working right in the short term, even if the full jest monorepo config isn't getting evaluated the "proper" way.

That said, I think it's more important to have it working at all in the short term, and we can clean things up more later (hopefully as the tooling evolves to make it a little easier -- it looks like SB 6.0 might be a step in the right direction).

PR #177, opened 2020-03-05
Add a basic toml file for the public app

bk3c:
I'd like to see if we can keep the non-environment-specific configuration in code so it's more visible and easier to update with everything else, so this is a first cut at a TOML file to bake in the invariant parts of the netlify configuration.


bk3c:
Merging now that everything is working on Netlify

Issue #174, opened 2020-03-04
Add Google Tag Manager support


bk3c:
Everyone is keen to get good analytics data as soon as we launch, so let's make sure we have GA/GTM enabled from the get-go:

- We'll want to have different tags on a per-site basis, both to separate jurisdictions and also to separate staging and production. I'm thinking the easiest way to do this is by setting the tags as a Netlify environment variable, but I'm open to suggestion.
- We should also make sure to set up a property for the reference implementation, so we can test appropriately. I'm not sure what the standard way to test GA correctness is, but if we can do something in CI, that would be great.
- Capturing external link clicks is a key requirement, and any other useful events that we can get is a nice-to-have bonus -- the more data the better.
- There seem to be a significant number of existing GA modules, so I think that's a fine route to go if it offers helpful functionality over just embedding the vanilla JS ourselves.

jaredcwhite:
@bk3c I got Google Analytics working in a branch, but after some review of the past Dahlia codebase, I realized we weren't using Analytics at all, but Google Tag Manager solely. Along those lines, Dahlia didn't have any click tracking or anything unique to Google going on within the components from what I can gather. I think that was all managed through GTM features.

Let me know how you want to handle this in Bloom.

bk3c:
After doing some research, it looks like using GTM [seems to offer a superset of functionality](https://www.analyticsmania.com/post/google-tag-manager-vs-google-analytics/) available by calling Analytics directly, so I'd lean toward that direction so that it's easier to use other GTM features in the future (e.g. adding in tracking code from other platforms as well).

I'd also definitely lean towards an existing module to help us deal with the react (and to a lesser extent next.js) complexity, so long as we can find something that's not too complicated and reasonably maintained. See https://ryanelainska.com/blog/google-tag-manager-in-react for an example, though it looks like the fork used there might be dead?

No matter where we end up implementation-wise, it's definitely important to be able to capture non page load events like outbound link clicks, since those are basically our conversion metric until we have applications built out in the system (and even then, will be the conversion metric for listings that have their own application flow)

jaredcwhite:
@bk3c That seems like a good approach (the blog post example). Regarding link clicks and so forth, I think that's all able to be handled via GTM. I'm not too familiar with it but here's an article to that effect: https://www.analyticsmania.com/post/track-outbound-links-with-google-tag-manager/

jaredcwhite:
[...]

PR #179, opened 2020-03-07
Add React GA component for analytics support

jaredcwhite:
This is working fine, but it might not be the correct approach. Further comments in the related issue #174


jaredcwhite:
Closing, due to a change to using Google Tag Manager instead.

Issue #169, opened 2020-03-02
Export UI Components at the root of the package


bk3c:
The imports for ui components seem a little awkward / heavyweight, but I think we can clean that up pretty easily by re-exporting the components at the package root so they can be imported without knowing the full internal paths.

bk3c:
Note: I started this for @bloom-housing/core already, and I'll put up a PR once we merge #168 so we don't have a zillion confilcts.

PR #181, opened 2020-03-08
Add an index.ts to shared/core and update imports

bk3c:
This is the first half of #169. After this, we can do the same for ui-components, and then further consolidate the imports to 1 statement per package.


bk3c:
Note this will probably break the build until it's published, since Netlify will still be downloading the old public packages from NPM which don't have the index yet. @jaredcwhite, if you can test locally that would be great.

Issue #150, opened 2020-02-05
Add documentation of the translations process and string locations


bk3c:
I expect this will be one of the most common sources of customization and changes, so I think we should invest in having good developer documentation for how to do it, since the actual work should be super straight-forward once you find all the files in the right places.

My initial vision in just a Markdown doc in the top-level docs directory, similar to my poorly-stubbed out Styling.MD, but I'm wide open to other options.

For extra credit, add useful content to Styling.MD too :)

bk3c:
The other thing I think we need to do is try to rationalize the hierarchy of string keys, which seems to be a mess. Are there best practices for that kind of thing?

@jaredcwhite do you think it makes sense to file a separate issue for that, or do you want to roll it in to this work? It's probably a never-ending gardening task over the long term, but I'm hoping that we can do a big push and get to a better starting point.

PR #185, opened 2020-03-16
New and updated locale strings

jaredcwhite:
Incremental improvements on #150 (there's probably more work to do…)

Also branched off of #183 so this should be merged after that.


bk3c:
@jaredcwhite Can you see if you can rebase this against master? Since we're doing squash merges it doesn't track those changes easily.

PR #187, opened 2020-03-18
Error page and Hero style updates

### LINKED ISSUE #170: Add user-friendly custom error pages
Since we don't have the additional info pages defined yet they're currently showing a 404, but it's a very generic / unhelpful 404 error page. Clearly we should fix the actual link in this case too, but I think it would be prudent to also define branded and user friendly error pages as well, in case someone has a URL typo or anything else that causes a page not to be served.

@jacchau can you work on defining the text and desired look for this page?
@jaredcwhite while we're waiting for that, can you go ahead and get the pages roughed out?

jaredcwhite:
Closes #170

### DIFF
--- a/apps/public-reference/pages/_error.jsx
+++ b/apps/public-reference/pages/_error.jsx
@@ -6,7 +6,7 @@ import MarkdownSection from "@bloom-housing/ui-components/src/sections/markdown_
 import PageContent from "../page_content/homepage.mdx"
 
 export default () => {
-  const pageTitle = <>Page Not Found</>
+  const pageTitle = <>{t("error.notFound.title")}</>
 
   return (
     <Layout>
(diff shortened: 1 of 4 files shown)


apps/public-reference/pages/_error.jsx
```diff
@@ -0,0 +1,27 @@
...
+export default () => {
+  const pageTitle = <>Page Not Found</>
+
+  return (
+    <Layout>
+      <Head>
+        <title>{pageTitle}</title>
+      </Head>
+      <Hero title={pageTitle} buttonTitle={t("welcome.seeRentalListings")} buttonLink="/listings">
+        Uh oh, we can’t seem to find the page you’re looking for. Try going back to the previous
```
bk3c: Are we hard-coding the text here because we're worried about not being able to resolve translations, or can we make this a localized string too?

jaredcwhite: @bk3c Good catch, I totally spaced. Moved those messages to the locale file now.

PR #194, opened 2020-03-26
Update dev and build dependencies

bk3c:
Updates to a variety of build tools and dependencies, including Typescript to 3.8. We also should upgrade Prettier to 2.0, but I'm saving that for another PR since a lot of files are going to change due to new rules, especially the trailing comma.


bk3c:
@jaredcwhite The new ESLint Typescript rules are catching a problem in services/listings/src/lib/listings_loader.ts having to do with returning a promise while it's expecting a void. Can you take a look and see if you can resolve that before we merge this?

I did a quick once over and looked at the current node docs for `fs`, and it looks like the most robust solution would be to use the node calls that now return native promises natively. Alternatively, I think it'd actually be reasonable to just convert the whole thing to synchronous calls, and make sure they only get called at startup rather than every request, which is probably better for performance anyway. 

Of course if there's a minor syntax fix to just keep the linter happy, I can also live with that since we'll be moving to a DB-backed solution soon enough.

jaredcwhite:
I fixed the TS errors with a bit simpler refactor, so we should be all good now.

PR #195, opened 2020-03-27
Insert GTM tags into page

### LINKED ISSUE #174: Add Google Tag Manager support

jaredcwhite:
Closes #174 

Out of the box nothing will happen on any new builds currently, but if the GTM_KEY env var is set, then it inserts the head script and body noscript tag. It's also flexible enough to accommodate any other sort of scripts which might be needed in future app variants.

### DIFF
--- a/apps/public-reference/src/customScripts.ts
+++ b/apps/public-reference/src/customScripts.ts
@@ -1,0 +1,22 @@
+export const headScript = () => {
+  const gtmKey = "TEST"
+  return `
+    (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    '//www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    console.info("TESTING the GTM, very kewl!")
+    })(window,document,'script','dataLayer','${gtmKey}')
+  `
+}
+
+export const bodyTopTag = () => {
+  const gtmKey = "TEST"
+  return `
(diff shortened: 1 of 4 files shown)


apps/public-reference/src/customScripts.ts
```diff
@@ -0,0 +1,34 @@
+export const headScript = () => {
+  const gtmKey = process.env.gtmKey
```
bk3c, line 2: Do we want to have the process env key be all caps to be consistent with the other ones in .env?

jaredcwhite: The env keys within the app are different than the actual server-side env keys. There's a mapping from node to SRR/client-side that happens in next.config.js — so you'll want your actual env var set via `GTM_KEY`. Let me know if that works.

jaredcwhite:
@bk3c I rebased on master and verified it's working. Here's the review app — you can look at the head and top of body tags and see the relevant GTM code snippets. (just using TEST123 as a test key I defined in the Netlify environment settings).

https://5e8d54f6821f5038b9543374--clever-edison-cd22c1.netlify.com

As far as click handlers and all that, my understanding is it's all defined using the GTM software…not sure there's anything we need to add to the codebase on our end. (And I didn't see anything like that in Dahlia.)

PR #197, opened 2020-03-28
Disable the units list accordion

### LINKED ISSUE #190: Lock the unit feature accordions so that they're not expandable
Given feedback from Eden Housing, the rent matrix is too complex to show for each of the 29 units. They'd prefer to show only the summary of the unit features and have the unit summary accordion locked. We've received this similar requests from Charities in SJ before as well. I think it'd be good to have a flag in the listing json to lock the unit feature accordion

As an admin,
I can lock the unit feature accordion on the listing
So that I can hide the individual unit feature details for the listing

AC:

The admin has the ability to lock the unit feature accordion so that individual unit details are not displayed

jaredcwhite:
Closes #190

### DIFF
--- a/apps/public-reference/pages/listing.tsx
+++ b/apps/public-reference/pages/listing.tsx
@@ -256,6 +256,7 @@ export default class extends Component<ListingProps> {
                       <UnitTables
                         units={listing.units}
                         unitSummaries={listing.unitsSummarized.byUnitType}
+                        disableAccordion={listing.disableUnitsAccordion}
                       />
                     }
                   />
--- a/services/listings/listings/gish.json
+++ b/services/listings/listings/gish.json
@@ -109,6 +109,7 @@
     }
   ],
   "buildingTotalUnits": 5,
+  "disableUnitsAccordion": true,
   "units": [
     {
       "id": "MFaowAxySJW3utZBPNxrg",
--- a/shared/core/src/listings.ts
+++ b/shared/core/src/listings.ts
@@ -35,6 +35,7 @@ export interface Listing {
   depositMin: string
   depositMax?: string
   developer: string
+  disableUnitsAccordion?: boolean
   id: string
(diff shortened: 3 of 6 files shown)


jaredcwhite:
@jacchau You can test this out here:
https://5e7eb917f0aa9e1785384adc--clever-edison-cd22c1.netlify.com/listing/sz0urxnbdaxhmdfgcgha7

Gish has a disabled accordion, the other two listings expand as before.

jacchau:
Looks good @jaredcwhite, let's ship it!

apps/public-reference/pages/listing.tsx
```diff
@@ -256,6 +256,7 @@ export default class extends Component<ListingProps> {
                       <UnitTables
                         units={listing.units}
                         unitSummaries={listing.unitsSummarized.byUnitType}
+                        disableAccordion={listing.disableUnitsAccordion}
```
bk3c: nitpick: should this be Accordion = disabled or something similar? I try to avoid having things phrased negatively in general, and it seems like there might eventually be more than two states?

jaredcwhite: Well if we go with HTML conventions, everything's enabled by default unless there's a specific `disabled` attribute. I think this boolean makes sense, but the term "accordion" could be renamed perhaps. Maybe `disableExpansion` ?

jacchau:
what is the finalized feature tag attribute and syntax? Eden Housing has requested to disable the accordion. I will update the listing json once this has been finalized and merged @jaredcwhite @bk3c

jaredcwhite:
@jacchau You'll want to add `"disableUnitsAccordion": true,` to the listings JSON

jaredcwhite:
@bk3c OK to merge this in and we can circle back around later on if we come up with a better approach?

PR #205, opened 2020-04-02
Update dependencies for each app

bk3c:
Catch up the three jurisdiction apps to match the latest versions in the public-reference package.json


bk3c:
My turn for the wrong repo today. :(

PR #209, opened 2020-04-03
Use newest ui-components and core in reference app

bk3c:
and add homepage fields to shared package.json files


bk3c:
Verified that it fixes the Netlify build issues.

PR #214, opened 2020-04-07
Upgrade prettier

### LINKED ISSUE #200: Upgrade to Prettier 2.x
There's no immediate urgency on this, but also no blockers so we can get to it as soon as someone has some time, and ideally there are no other major PRs outstanding since applying the new rules (especially trailing comma) will affect a huge number of files.

Also, please make sure to do this in its own PR that doesn't have anything else in it, so the review isn't confusing. Maybe just one commit for the dependency change, and then one or more for the application of the new rules across the repo. I noticed that the current ESLint rules might not be applied to Markdown files, so maybe we could add that as well, if everyone thinks it's a good idea?

Ideally, whoever takes this on should then also be in charge of applying the new rules to the jurisdiction-specific sites in housingbayarea as soon as this upstream work is merged.

software-project:
Fixes #200


software-project:
@bk3c I've tried adding MDX linting as well, but there are some issues with it. Could you have a quick look at it with your professional eye? Maybe I'm missing something.

PR #218, opened 2020-04-09
Import mapbox css and add css loader

### LINKED ISSUE #204: Mapbox console warnings on listing page
Seen on the Alameda site, but I'm pretty sure it's a core problem:

1. `This page appears to be missing CSS declarations for Mapbox GL JS, which may cause the map to display incorrectly. `
2. `A cookie associated with a cross-site resource at http://mapbox.com/ was set without the 'SameSite' attribute. It has been blocked...`

Neither appears to be a major user-facing problem at the moment, but I think better to fix stuff like this sooner rather than later.

software-project:
Fixes #204


jaredcwhite:
LGTM 👍

PR #226, opened 2020-04-21
Add lerna

### LINKED ISSUE #210: Set up lerna (or similar) to keep version numbers and publishing in sync
We're now doing updates to shared packages frequently enough that we're starting to hit the kinds of problems that lerna is meant to solve, so it would be great to add it, or another similar tool to our build process. The main goal is to make sure we have consistent versioning across packages, and that we publish updates to shared packages to NPM when appropriate.

Eventually this can mean auto-publishing to a NPM tag like `next`, but I'd err on the side of taking ne step at a time.

There shouldn't be any need to do this for downstream repos, since it only really makes a difference for shared packages published to NPM.

software-project:
Fixes #210


lerna.json
```diff
@@ -0,0 +1,10 @@
+{
+  "packages": [
+    "apps/public-reference",
+    "services/listings",
+    "shared/*"
+  ],
+  "version": "0.0.9",
+  "npmClient": "yarn",
+  "useWrokspaces": true
```
jaredcwhite, line 9: I think the `useWrokspaces` line should be removed…?

software-project: I think you are right.

bk3c: My reading of the docs is that we probably do want to useWorkspaces (note: spelled correctly), especially since we're already using them -- unless there's something I'm missing?

software-project: We can either use useWorkspaces flag or define packages. Both ways should work the same.

bk3c:
After some local testing, I'm turning useWorkspaces back on. @software-project is completely right that it shouldn't make a difference in most cases, but if you try to do `lerna bootstrap` without it enabled, it seems to get confused between the yarn hoisting and the lerna config. The other option would be to switch to straight lerna and turn off yarn workspaces entirely, but they seem to be reasonably working for us, so I opted for what seemed like the smaller change.

PR #232, opened 2020-04-27
Version "0.1" of Submit an Application form

jaredcwhite:
All right, at only a "tech demo" stage it's already a huge set of changes! I'll try to break it down:

----

* General comment about all the styling: it's currently a mix of inline styles, tailwind, and usual sass stuff so obviously this will all need to get cleaned up soon.
* Similar comment regarding the form fields: plenty of room for future DRY componentization.

----

* Added a new "child" layout for the specific design of the form pages.
* Using the React Contexts API to store JSON state from multiple form steps/pages.
* The `ApplicationConductor` is responsible for managing the overall lifecycle of the multi-step form. I could have called this a "wizard" but that's such a 90s UI term. 😉
* Using `localStorage` to autosave form submissions and load them up again, so if you refresh the browser or close it and open it again, all the data should come back—until you complete the form and then click "Return to Start" which will clear the storage.
* The form submit event in each step in the multi-step form is handled by its own object in `lib/app_submission_steps`…this seems like a lot of overhead right now, but later on when we need to add a bunch of additional logic, validation, etc., I feel like keeping that out of the form page itself is a good separation of concerns (i.e., let's keep model/controller stuff out of the view).
* So far I'm really liking React Hook Form and feel it's the right balance of simplicity and flexibility for our needs.
* Weird glitches with the statically-generated vs. client-side React state with regard to the `ProcessStepIndicator` component took a lot of time to resolve. I finally found a tip from the Next.js community on how to hook into React's component lifecycle to contain certain code in a "client-only" block via the `useEffect` hook and tracking state. I rolled that up in a simple `onClientSide()` function, so any time we need to show certain things or run certain logic only on the client side, that's an easy approach.
* All right, probably enough info to throw at you right now. Let's chat about feedback and immediate improvements and then what makes sense to tackle for the next version.


jaredcwhite:
Review URL: https://5ea74edebe844e2a43507ae5--clever-edison-cd22c1.netlify.app/applications/new

software-project:
Looks good, nice work!!

jaredcwhite:
Added a new Field component to wrap around the form inputs, bake in some design patterns. Will more to do…tests to write…

james-wills-sf:
Hey @jaredcwhite! This is James from the DS housing team. Are there any documents around the technical decisions made so far for this project? For example I see in the PR comment that styling is currently a mix of things, do we have an idea of which direction we're going to go when these are consolidated (tailwind, sass, or inline styles)?

 I'm still very new to Dahlia so I don't have feedback on integration specifics, I'll leave those to Andrea.

jaredcwhite:
Hi @james-wills-sf! I'm not sure I can speak to the project as a whole (I believe Ben Kutler @bk3c is working on a broader architectural diagram at present), but in terms of this PR (submitting a listing application):

* On the JS side, we're using React Hook Form to handle the all the input and client-side validation, along with the React Context API to manage state. I've written some small objects to function basically as a go-between, so there's more structure around the context<->form communications (plus handling autosave of form data). There's also the step by step through multiple form pages.
* On the CSS side, Jesse James @slowbot is working on consolidating remaining styles into component-level SCSS files using Tailwind theming. Look at `src/prototypes/AppCard.scss` in the `shared/ui-components` module as an example of that approach.

Let me know if you have any questions!

jaredcwhite:
I made some improvements to the file naming conventions for the form step objects…seems a bit cleaner now.

jaredcwhite:
LGTM!

PR #233, opened 2020-04-28
Add pretty URL slugs to listing pages

### LINKED ISSUE #215: Display meaningful urls and deeplinks for listing
Currently the url for a listing is determined by the listing id such as https://ac.housingbayarea.org/listing/sz0urxnbdaxhmdfgcgha8. 

The listing id is currently manually created in the listing json, in the future these should be autogenerated when a user creates a listing in the Partners Portal. 

The goal here is to create a mechanism where the listing url displays more meaningful information such as the name of the property or the address. This would provide more information in the link itself, increase SEO for the listing, and enable deeplinks for the future. 

Cosniderations:
- The text/ information displayed on the listing url should be unique. If we use the property name, we have to consider rather there would be multiple listing for the same property in the future. What is the best technical design to display unique information for the listing about the property?

bencpeters:
# Add pretty URL slugs to listing pages

This PR adds descriptive url "slugs" to the end of listing page URLs:

`/listing/jsQvgvy6eJy56TNCFfQp8` => `/listing/jsQvgvy6eJy56TNCFfQp8/the_triton_55_triton_park_lane_foster_city_ca`

The bare "id-only" URLs will still work; they issue an immediate 301 redirect to the pretty version. This PR adds a general purpose `Redirect` component to accomplish this.

The url slug is available as an attribute `urlSlug` on the `Listing` type.

Minor version bump of `eslint-typescript` to [allow function definitions](https://github.com/typescript-eslint/typescript-eslint/pull/1736). This was necessary to avoid an eslint error on [this line](https://github.com/bloom-housing/bloom/blob/add-url-slugs/apps/public-reference/pages/redirect.tsx#L15).

Closes #215

### DIFF
--- a/apps/public-reference/cypress/integration/navigation.ts
+++ b/apps/public-reference/cypress/integration/navigation.ts
@@ -13,8 +13,23 @@ describe("Navigating around the site", function() {
     cy.contains("Rent affordable housing")
   })
 
-  it("Loads a listing page directly", function() {
+  it("Loads a listing page directly by id", function() {
     cy.visit("http://localhost:3000/listing/Uvbk5qurpB2WI9V6WnNdH")
+
+    // Check that the listing page sidebar apply section text is present on the page
+    cy.contains("Get a Paper Application")
+
+    // Check that the URL got re-written with a URL slug
+    cy.location("pathname").should(
+      "eq",
+      "/listing/Uvbk5qurpB2WI9V6WnNdH/archer_studios_98_archer_street_san_jose_ca"
+    )
+  })
+
+  it("Loads a listing page directly with a full url", function() {
+    cy.visit(
+      "http://localhost:3000/listing/Uvbk5qurpB2WI9V6WnNdH/archer_studios_98_archer_street_san_jose_ca"
+    )
 
     // Check that the listing page sidebar apply section text is present on the page
     cy.contains("Get a Paper Application")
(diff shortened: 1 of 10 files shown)


jaredcwhite:
@bencpeters Nice solution! I'm wondering if we even need the guid in the the URL though? The number of available listings isn't that high so certainly no real risk of slug collision.

bencpeters:
> @bencpeters Nice solution! I'm wondering if we even need the guid in the the URL though? The number of available listings isn't that high so certainly no real risk of slug collision.

Personally, I like having the guid in there - it allows us to keep a "stable" URL in cases where the listing `name` changes (or maybe the address is incomplete/incorrect to start with and gets updated?). I could also see a higher chance of slug collisions in a multi-unit building - maybe the address is the same for all of them if the listing mistakenly doesn't include a unit number?

I don't feel super strongly on it though.

jaredcwhite:
> it allows us to keep a "stable" URL in cases where the listing `name` changes

That's a good point. Probably wouldn't happen after a listing goes live, but during development I could see that. Maybe we can knock the id length down to 8 characters or something of that nature?

bencpeters:
> That's a good point. Probably wouldn't happen after a listing goes live, but during development I could see that. Maybe we can knock the id length down to 8 characters or something of that nature?

Ya, that seems reasonable. Where are those IDs generated right now? I just saw them listed in the JSON data files?

jaredcwhite:
Now that you mention it, they are totally arbitrary (hard-coded into the JSON). So perhaps we can file that as a separate issue related to the future Listings service.

bk3c:
FWIW, I'd definitely be in favor of shorter nanoids. Perhaps 8 characters, which should give us a [1% chance of collision every ~239 years](https://zelark.github.io/nano-id-cc/) if we generate 1/hr? I'm open to other lengths, though.

PR #237, opened 2020-05-04
WIP: Typeorm test

software-project:


bk3c:
The one thing I find a little disappointing about this is that it seems to mess with the ability to have a shared TS interface between the front and back ends, or at the very least would push us toward having ORM entities defined in a duplicative way. There could be a potential workaround by using something like https://github.com/typeorm/typeorm/blob/master/extra/typeorm-model-shim.js on the front end to ignore the annotations, but that seems awkward at best.

This isn't a deal-killer for me, but am I missing something about how this is all supposed to work?

software-project:
@bk3c I think we could use EntitySchemas: https://typeorm.io/#/separating-entity-definition but it would also require a level of duplication. Will have a look on Prisma, this does feel a bit disappointing, especially taking my struggles here and a bit nasty interface as well. So fingers crossed ;)

jaredcwhite:
Yeah, I would argue that if we're keeping with full-stack Node/Typescript/etc. in order to reap the benefits, we would need…to be able to reap those benefits. 😄

Seems like Prisma's typing solution for the client is pretty solid:

https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/advanced-usage-of-generated-types#overview

bk3c:
@software-project should we close this in favor of #282, or are there things here that we should figure out a merge path for?

software-project:
@bk3c, yes that makes sense, will do it now

Issue #227, opened 2020-04-22
Choose a ORM for node-based services


bk3c:
Assuming we continue down the path of node-based reference implementations for the core services, the next major decision we'll need to make is an ORM for database operations. Postgres is the presumptive database, perhaps with an HA variant in the future, but not on the immediate horizon.

In addition to good postgres support, we also want enough Typescript support that we won't need to be casting every result set or otherwise writing too much boilerplate.

Options to consider include:

- [TypeORM](https://typeorm.io/)
- [Prisma2](https://www.prisma.io/)
- [Objection](https://github.com/Vincit/objection.js) / [Knex](http://knexjs.org/)
- [Sequelize](https://github.com/sequelize/sequelize)

I think we're open to other options, as well, so long as they meet our core needs and have a fighting chance at long-term sustainability.

There's also still the option of choosing something like [NestJS](https://docs.nestjs.com/techniques/database) as a more full featured backend framework, which would then likely imply (though not conclusively) using their TypeORM integration. It's still not clear we need all of those features, though.

bk3c:
[Relevant NPMTrends view.](https://www.npmtrends.com/knex-vs-sequelize-vs-objection-vs-prisma-vs-typeorm)

software-project:
I had a brief look on all of above solutions. Sequelize seems the most mature solution, but also the most 'old school'. Knex is more or less the same. Prisma2 is in beta and from what I found not super stable yet. So my contender is TypeORM and I'm going to prototype it tomorrow. Apollo popped up couple times as well. Did we shut GraphQL doors or is it worth testing it as well?

bk3c:
@software-project I definitely think the door is still open for GraphQL, but I think I'd like to keep that as a separate decision from the ORM unless there's a compelling reason we need to make the the decision together. 

If you want to prototype the two together since that seems like better leverage for your time, I'm totally fine with that -- let's just try to keep the decisions and timelines separate on the other end (and comment on the GraphQL learnings in #77)

software-project:
I think I will keep it separate for now, cheers.

jaredcwhite:
Venturing forth a very uninformed opinion that Prisma 2 might deserve a closer look if for no other reason than it's the ORM of choice for Redwood.js, and the lead dev there (Tom Preston-Werner) is no slouch. So it might be improving at a pretty rapid clip going forward…

software-project:
[...]

PR #254, opened 2020-05-08
WIP: Prisma test

software-project:


bk3c:
See details in #227, but we've decided to move forward with TypeORM.

PR #264, opened 2020-05-11
Minor dependency updates to build and test tools

bk3c:
There shouldn't be any meaningful changes here, just trying to keep up to date on the JS package treadmill.


bk3c:
@jaredcwhite approves via slack

PR #188, opened 2020-03-18
Jc museum test listings

jacchau:
Add three new test case scenarios based on https://docs.google.com/spreadsheets/d/1_RfaHaQdR61MUomxtU5nnLPgSQMxzCGF9TrsS8xUfyc/edit#gid=0

PR #282, opened 2020-05-13
Nestjs with typeorm

software-project:
NestJs backend with typeORM.


software-project:
@bk3c @bencpeters @pbn4 @jaredcwhite As we decided to move forward with NestJS, do we still want to evaluate option not to use Services or we want to keep with convension and continue with services?

pbn4:
I prefer services approach as they promote code reuse and allow for different presentation layers. Here we have HTTP, but introducing websockets seems likely and then we could make those two use the same service. Also that will keeps things DRY.

jaredcwhite:
> I prefer services approach as they promote code reuse and allow for different presentation layers. Here we have HTTP, but introducing websockets seems likely and then we could make those two use the same service. Also that will keeps things DRY.

I think maybe we're talking past each other as does often happen when discussing service objects.

What I'm not saying is we put model-level logic directly in controllers.

What I am suggesting is that a controller is responsible for pulling data out of the model layer that has been requested by that HTTP request. If that data is composed in a way that conforms specifically to a particular request, then it makes sense for that composition to occur in a controller.

However, if data is often expected to be in a composed/transformed state, regardless of the particulars of a specific controller/request, then that composition/transformation should happen in the model layer.

For example, we group the Unit data into together in different ways to provide Unit Summaries. Ideally we could add this to the model layer. Some pseudo-code using domain modeling:

```
// Listings Controller
listings = Listings.findAll({includeUnitSummaries: true})

// Listings model
static findAll(options) {
  listings = ListingsRepository.findAll(other_options) // load raw DB objects
  if (options.includeUnitSummaries) {
    listings = listings.each(listing => {
      listing.addUnitSummaries()
    })
  }
  return listings
}
addUnitSummaries() {
  this.unitSummaries = UnitSummaries.transformUnits(this)
}

// UnitSummaries model
static transform(listing) {
  // return the unit summaries
}
```

Again, no need for a separate service object layer if we use domain modeling best-practices to compose data transformations out of model logic as requested by the controller.

There could be a place for service objects at some point down the road when a really advanced operation needs to happen which spans multiple domains, but meanwhile what I would like to avoid is this pattern:

`HTTP Request -> Controller -> Service Object -> Model`

for every GET, POST, PUT, etc. in a REST API. That seems to me like an unnecessary level of abstraction.

software-project:
@jaredcwhite I'm a fan of Single Responsibility Principle. Ruby/ Rails wise I wasn't super hard core on them, as Active Record is not exactly SRP and ppl tended to separate validation object with Virtus model and then put the logic to a service. That is a bit too much. Regarding the case you put out there, that piece of code should belong to a service rather then a model, as there is a logic that transforms other object. 
Another thing is that we are calculating unitSummaries with each request, which is not really optimized, as we are talking about doing things the proper way, in this particular case I would store it in the db and calculate it once, on save, but in a service object. And then show/ index action can we without a service object and just return a findAll collection.

jaredcwhite:
@software-project I agree it would be great to transform that data once and cache it in some fashion until there's a later change to units data.

But I don't agree with the assertion that it violates SRP for a model to provide transformed data from a related model. My basic rule of thumb is:

Can I drop into a REPL and ask `Listing` to provide me all expected data of Listing along with associations to related data models like `Unit` and `Attachment`?

Can I ask a `Listing` to update some data of a listing?

Can I ask an associated model like `Unit` to update some data of a unit, and expect the parent `Listing` to reflect that updated unit data?

This is all straightforward domain modeling and fully congruent with the ActiveRecord OOP pattern. Nothing in those questions above is begging to be answered "Nope, sorry, can't be done. You need a separate service object to handle that."

(For the record, I equally view most third-party additions to "the Rails way" over the years to be entirely over-engineered and usually unnecessary, Virtus included.)

services/listings/.gitignore
```diff
@@ -0,0 +1,34 @@
+# compiled output
```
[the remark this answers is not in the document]
software-project: I missed this, I used nestjs generator to create the app and it added this automatically. Will remove.

services/listings/Procfile
```diff
@@ -1,2 +0,0 @@
-release: yarn herokusetup
```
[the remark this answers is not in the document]
software-project: It's needed, will move it over

software-project:
@bk3c I've updated the code according to your change request. DB seed should be working now as well. I have two things left to do `jsonpath` and tests.

bk3c:
@software-project it looks like there's some of #316 mixed in here. Ideally let's try to avoid that in the future, but I expect to be able to merge that one as soon as the small fixes for Any props are in so it's fine this time around.

software-project:
@bk3c yes, before nohoist thing I was trying to align dependencies, so we can avoid nohoist. It didnt work, so I can remove the upgrades.

backend/core/src/migration/1588694149283-InitialDB.ts
```diff
@@ -0,0 +1,61 @@
...
+      undefined
+    )
+    await queryRunner.query(`CREATE TYPE "attachment_type_enum" AS ENUM('1', '2')`, undefined)
+    await queryRunner.query(
+      `CREATE TABLE "attachment" ("id" uuid NOT NULL DEFAULT uuid_generate_v4(), "label" character varying, "fileUrl" character varying, "type" "attachment_type_enum", "listingId" uuid, CONSTRAINT "PK_d2a80c3a8d467f08a750ac4b420" PRIMARY KEY ("id"))`,
+      undefined
+    )
+    await queryRunner.query(
+      `ALTER TABLE "unit" ADD CONSTRAINT "FK_72719e0518e726a1de304c6738e" FOREIGN KEY ("listingId") REFERENCES "listing"("id") ON DELETE NO ACTION ON UPDATE NO ACTION`,
```
bencpeters, line 29: This might not be necessary to think about yet, but do we want all these 1:many listings to live on after their parent listings are removed? I was surprised to see the `on delete no action` here - I'd expect a `cascade` normally?

backend/core/src/seed.ts
```diff
@@ -0,0 +1,59 @@
...
+    const listings = (await listingsLoader("listings")) as OldListing[]
+
+    for await (const listing of listings) {
+      l = new Listing()
+      const entities = {}
+      for await (const key of Object.keys(listing)) {
+        if (!skipped.includes(key)) {
+          l[key] = listing[key]
+        } else {
+          if (key != "id") {
```
bencpeters, line 27: /pedantic `else if ()` & `!==` rather than `!=` (just JS best practices, doesn't change anything in this script)

backend/core/src/seed.ts
```diff
@@ -0,0 +1,59 @@
...
+import { Listing } from "./entity/Listing"
+import { Listing as OldListing } from "@bloom-housing/core"
+import listingsLoader from "./lib/listings_loader"
+import { Unit } from "./entity/Unit"
+import { Attachment } from "./entity/Attachment"
+import { Preference } from "./entity/Preference"
+
+const skipped = ["id", "units", "attachments", "preferences"]
+const types = { units: Unit, attachments: Attachment, preferences: Preference }
+let l
```
bencpeters, line 13: /pedantic, but is there any reason to declare `l` in this outer scope? it seems like it could be a `const` scoped to the `listing` loop instead?

backend/core/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "removeComments": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "target": "es2017",
+    "sourceMap": true,
+    "outDir": "./dist",
+    "baseUrl": "./",
+    "incremental": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true,
```
bencpeters, line 14: Not sure if this matters, but it's missing a couple keys that are in the root `tsconfig.json` for the mono repo:

`isolatedModules`, `forceConsistentCasingInFileNames`, `skipLibCheck`, `strict: false`.

There might be good reasons to have module-specific tsconfigs, but I think we probably want to keep them as consistent as possible, no?

ormconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
bk3c, line 1: Is there a reason we want to have an ormconfig at the top level in addition to `backend/core`? I think it's probably better to keep everything there unless the tooling doesn't support it.

backend/core/listings/gish.json
```diff
@@ -0,0 +1,237 @@
+{
```
bk3c, line 1: We don't need/want this whole directory now that things are in the database, correct? Ideally I'm thinking that we should convert the test data over to seeds, and then also have tests against the API that validate we're serving correctly. (I'm fine if we do the latter in a separate PR.)

Rather than use the 3 listings here as test data, it would be even better to replace them with the work that Jacky started in #188. That way there's less possibility of collision or confusion with the various real properties. I'm not sure the data in that PR matches our latest thinking in terms of fields and format, but I can't think of a better time to sort it out and then we can close out that PR.

software-project: I've added seed file that reads that files and inserts them to db. I can change those 3 files to seed.json, would that work?

ormconfig.json
software-project: ormconfig didn't play nicely with workspace config that we have. I think that might me a subject for a separate issue.

backend/core/tsconfig.json
software-project: Will try adding them and see how it plays out.

backend/core/src/seed.ts
software-project: 👍

backend/core/src/migration/1588694149283-InitialDB.ts
software-project: good catch

backend/core/src/app.module.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Unit } from './entity/Unit';
+import { Preference } from './entity/Preference';
+
+@Module({
+  imports: [TypeOrmModule.forRoot({
+    type: "postgres",
+    host: "localhost",
+    port: 5432,
+    database: "bloom",
+    synchronize: true,
```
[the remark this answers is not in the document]
software-project: Ideally we would use ormconfig file, but it doesnt play nicely with workspaces or I just couldnt set it up corectly

backend/core/package.json
```diff
@@ -0,0 +1,103 @@
...
+  "version": "0.1.0",
+  "description": "Listings service reference implementation for the Bloom affordable housing system",
+  "author": "Marcin Jedras <marcin.jedras@exygy.com>",
+  "private": true,
+  "license": "GPL-3.0",
+  "scripts": {
+    "clean": "rimraf dist",
+    "build": "rimraf dist && nest build",
+    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
+    "start": "yarn run build && node dist/main",
```
bk3c, line 12: this didn't actually work for me out of the box -- it looks like TS is compiling such that `main.js` is in `dist/src` rather than just `dist`. Are other people having the same experience, or is this a problem specific to my local environment?

ormconfig.json
bk3c: Ok, I think I got this solved by deleting both ormconfig files and putting everything into .env and .env.template, which has the added benefit of less confusion when we deploy to heroku. @software-project are you ok if I commit those changes directly here?

backend/core/src/app.module.ts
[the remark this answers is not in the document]
bk3c: See my comment below, but I think I got this figured out based on [TypeORM documentation here](https://typeorm.io/#/using-ormconfig/using-environment-variables).

backend/core/src/migration/1588694149283-InitialDB.ts
bk3c: I think `NO ACTION` here actually isn't that bad, since we should more or less never be doing hard deletes from the database, and if it happens by mistake I think we want to limit the damage / preserve the evidence. I realize now that I've said this that we'll inevitably have to write some sort of cleanup script that will be more difficult without cascading deletes, but am I crazy to think that it's worth a little more complexity in that case to be safer otherwise?

bk3c:
@bencpeters re: column names, I totally agree with you that having to quote them in SQL is highly annoying (and I always end up doing manual SQL at some point). My only concern is that the plugin you referenced seems like it's not all that well maintained, so I wonder if we're better off doing `@Entity("table_name")` right in the definition file, which seems to be the more documented / more supported if definitely more manual way to do it. Do you think that's too much work?

bk3c:
Much work still to do, but consensus is to merge this so we can work on pieces of it in parallel since it shouldn't affect the existing (soon to be legacy) listings service in production.

Issue #248, opened 2020-05-07
Browser Timeout


slowbot:
**Goal**
As a housing applicant I don’t want to inadvertently expose sensitive personal information

**Acceptance Criteria**

- If an applicant is idle on a page for more than “5?” min they will be give a browser warning that reads:
   _To protect your identity, your session will expire in one minute due to inactivity. You will lose any unsaved information and be logged out if you choose not to respond._ 
- The user can select Ok and reset the timer and close the browser alert.
- After one additional minute if they have not acknowledged the browser alert, they are shown another browser alert that reads
   _Your session has expired._
- The user is redirected to the Sign In screen with a notification that reads:
- We care about your security. We logged you out due to inactivity. Please sign in to continue.

**Documentation**
![image](https://user-images.githubusercontent.com/13915/81323237-ebb8f180-9049-11ea-9387-47da3ab77828.png)

**Engineering Notes**

bk3c:
@bencpeters UI pieces of this aside, how do you see idle session timeout working within an OIDC context? Some quick googling seems to show that having an overall max session is easily supported, but it's not clear to me how we'd do a ~5min idle timeout.

- The most secure option seems to be that each JWT would be issued with a 5 minute expiration, and then the auth would be continually renewed by the apps based on user action. This seems like it would be very chatty on the network side, but maybe we could do some optimization by having the app code keep an internal timer and then only auto-renew every few minutes.

- The most efficient option would be to set a longer (maybe 30m) session timeout in the token itself, and then have the frontend app responsible for sending an explicit logout request after its own internal timer expired. This might be enough security given the risk vectors involved, but I'm open to feedback if this seems too dangerous.

bencpeters:
I think the best way to do this is probably to have a short lived access token based on the timeout that is periodically refreshed. I'm not super worried about having a call to the auth server every 5 min or so, for a token refresh, I don't think that's prohibitive from a network traffic point of view.

[...]

PR #284, opened 2020-05-13
Add OIDC-based Auth Service

bencpeters:
# Add OIDC-based Auth Service

This PR adds an authorization service and client implementation based on Open ID Connect standards.

The auth service is a Node server using the [oidc-provider](https://github.com/panva/node-oidc-provider) library with a hard-coded client configuration. It has an example username/password login view based on the examples from the library, and a placeholder `authenticate` method that I was holding off on writing until the DB/ORM decision was made. We can update this method to handle usernames/passwords in Postgres if desired.

The client side is a React Context that wraps the [oidc-client](https://github.com/IdentityModel/oidc-client-js) library. Right now it consists of a Login button that redirects to the auth server for authentication as per the Oauth flow, a Redirect component which handles the post authentication redirect. As per best security practices, the access tokens are only stored in memory, so the context also automatically makes a background request silently to try to get an access token on initial load (whether or not this will succeed depends on whether there's a valid session stored in the browser for the auth server - we can configure the longevity/persistence of this to accommodate use cases like #248).

There are still a few TODOs on this PR:
- [ ] Implement a "real" authorize method (right now it accepts any email/password combination since there's no database attached)
- [ ] Add a persistent session store (right now it uses an in-memory store, so sessions are not persisted when the auth server restarts). This could either be a key-value optimized NoSQL store like Redis, or we could just use a Postgres database.
- [ ] Add a keystore for configuring JWT signing keys (right now it uses a Dev-only signing key)
- [ ] Refactor the Auth Client to be a stand-alone package in `shared` so that other apps can use it (do we need to do this now, or should this become a future task?)

I wanted to get this POC up and running so that we can confirm this is the right direction to go in though. I'll add a README to the package before final PR, but you can start the server in the usual way:

```
cd services/authentication
yarn install
yarn run dev
```

By default it will start up on `localhost:3002`.

The main questions to focus on for feedback:

- Is this login flow acceptable? (clicking on a "login" button, or some other prompt in an app, which redirects to the auth server page for authentication - right now via email/pw, but could be SSO or any other method too). After initial login as long as the auth server session is persisted, this happens in the background with no user intervention.

- How do we want to deploy this **initially**? Email/pw login? Google SSO? Something else?

- Opinions on adding another technology (Redis) vs. Postgres for persistent sessions? Is this going to be deployed on Heroku? Or Docker containers?

- Do we want to handle registration in this PR, or in another one down the line?

- Do we want to use a service like Firebase (other options would probably be Auth0, Okta, or AWS Cognito) to handle user accounts? In theory, some of these services (especially Auth0 & Okta) could even handle the OIDC server side of this as well, although we'd have a bit less control over it. But then we have vendor lock-in/not as OSS friendly, etc. 

Anything else?


services/authentication/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "yarn": "^1.22"
+  },
+  "scripts": {
+    "clean": "rimraf dist",
+    "build": "yarn run clean && tsc -p tsconfig.prod.json",
+    "start": "yarn install && yarn run build && node dist/src/index.js",
+    "dev": "NODE_ENV=development ts-node-dev src/index.ts",
+    "test": "jest"
+  },
+  "babel": {
```
bk3c, line 38: Just to mostly contradict myself on the record about babel versioning above, I think we're mainly using it for testing, and I would be thrilled to take it out of the process entirely in favor of ts-jest running TS more natively. I made a couple of attempts with limited luck and I don't want to make it a requirement to move forward with this, but maybe starting a new service is the right time to give it another shot?

services/authentication/package.json
```diff
@@ -0,0 +1,52 @@
+{
```
bk3c, line 1: I think we also need to add this as a workspace in the root package.json so dependency hoisting works correctly.

services/authentication/src/views/login.ejs
```diff
@@ -0,0 +1,150 @@
+<!DOCTYPE html>
```
bk3c, line 1: It looks like @jaredcwhite may be already moving down this road in parallel. but I think we want the UI to live with the apps not with the service, so long as that doesn't make life exceedingly difficult. I expect both the branding and potentially the UX more generally to be pretty different between public and partners, so I'm not sure that we save much by combining any forms.

bk3c:
Some thoughts on the questions posed above:

> How do we want to deploy this initially? Email/pw login? Google SSO? Something else?

I think let's start with Google SSO for Exygy accounts, so that we can land this first version without having to build out too much backend, and we can get some experience with what SSO might look like.

> Opinions on adding another technology (Redis) vs. Postgres for persistent sessions? Is this going to be deployed on Heroku? Or Docker containers?

The plan is to deploy on Heroku unless there's a reason we shouldn't, and I'd lean toward postgres for persistent sessions so that we only have one database technology in play, unless there's reason to think it won't be performant or is missing things we need.

> Do we want to handle registration in this PR, or in another one down the line?

I'd vote for another one down the line, especially if we're starting with SSO.

> Do we want to use a service like Firebase (other options would probably be Auth0, Okta, or AWS Cognito) to handle user accounts? In theory, some of these services (especially Auth0 & Okta) could even handle the OIDC server side of this as well, although we'd have a bit less control over it. But then we have vendor lock-in/not as OSS friendly, etc.

Totally reasonable question, and I would lean toward Auth0 or Okta if we wanted to go the 3rd party route, but I'm still leaning toward doing it ourselves for both the control and vendor vs OSS reasons you've mentioned. If we find that it's difficult to implement what we want and it seems like using a 3rd party might be a better tradeoff, I'm open to re-evaluating this in a couple weeks

bk3c:
@bencpeters I added a minor commit just to fix the workspaces thing and add the auth service to the dev:all script so I didn't need to run it in two windows, but feel free to adjust as needed if you'd prefer a different implementation.

services/authentication/package.json
bk3c: Fixed

.eslintignore
```diff
@@ -0,0 +1 @@
+apps/public-reference/public/auth/oidc-client.rsa256.slim.min.js
```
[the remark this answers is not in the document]
bencpeters: That's a good question - I can look more into whether that's possible.

Some background for why this exists:

The silent sign in method works by opening up a background iFrame and attempting an auth request for a new access token with it (this is a standard methodology to avoid insecurely storing refresh tokens in a public app with OIDC). The redirect page that opens up in that invisible iFrame needs to communicate back to the main frame with the response. I was originally doing this using the same `Redirect` component that I'm using for the regular login flow, which works nicely within the React app. This approach worked totally fine in Firefox and Safari. Unfortunately, in Chrome, for some reason as soon as the iFrame communicates back to the parent, the CPU use of the tab spikes to 100% and everything freezes (no DOM interactions possible at all, the debugger doesn't have the DOM loaded - I need to use the task manager to kill the tab). After much trial and error, I was able to get the silent sign in working in chrome by creating a redirect page that doesn't involve React, hence this `public/auth` directory. The problem is that since it needs to run outside of the React context, it doesn't have access to the build pipeline of dependencies.

I don't *really* know why loading the app in the iFrame was messing things up in Chrome, but my gut instinct is that it's probably something to do with how next.js pre-rendering is resolving things </hand-waving>. I did find some reference to a similar problem in [this issue](https://github.com/IdentityModel/oidc-client-js/issues/462#issuecomment-355963315) describing an Angular implementation, and one of the core maintainers mentioned that a simple HTML file for the redirect was the "preferred solution" anyway.

One idea would be to add a CDN dependency for the OIDC client library... but that also runs into potential versioning mis-match issues.

It's possible that this callback file could access the main build file exported by next.js, although I'm not really sure how to do that... and it would also require `Oidc` to be added as a global to `window`, which isn't happening with ES6-style imports in the React app right now.

**Basically, I don't love this solution, but it seemed like the easiest of the sub-optimal options to me. Open to suggestions from the team!**

apps/public-reference/package.json
```diff
@@ -28,6 +28,7 @@
     "next-plugin-custom-babel-config": "^1.0.2",
     "node-polyglot": "^2.4.0",
     "node-sass": "^4.12.0",
+    "oidc-client": "^1.10.1",
```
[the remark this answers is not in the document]
bencpeters: This is the client portion of the app. `oidc-provider` is for the auth service (since it is an OIDC server). The `oidc-client` is the client side code that implements the login flow within the SPA.

Since OIDC is a standard, any client that implements the standard correctly would work. I actually initially started writing a simple custom implementation on this, but then decided to go with a standard certified library (mostly for the "background signin" functionality).

[This page](https://openid.net/developers/certified/) lists OIDC-certified libraries. `oidc-client` is the certified JS library for browser-based apps (e.g. SPAs). `openid-client` is a library for node.js apps that need to act as open-id clients of some auth server (e.g. for making federated data requests).

services/authentication/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "name": "@bloom-housing/authentication-service",
+  "version": "0.0.1",
+  "description": "Authentication service for Bloom affordable housing system",
+  "main": "dist/index.js",
+  "author": "Ben Peters <bencpeters@gmail.com>",
+  "license": "GPL-3.0",
+  "private": true,
+  "dependencies": {
+    "@types/node": "^13.13.4",
```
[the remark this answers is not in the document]
bencpeters: Good call, I totally agree, just missed that this was a node 13 dependency for this library.

services/authentication/package.json
bencpeters: I can take a look!

services/authentication/package.json
bencpeters: 🙏 Thanks!

services/authentication/src/views/login.ejs
bencpeters: I agree in theory. However, for security reasons, OIDC interactions for authenticating need to happen from the auth server.

That being said, this is just a sample implementation of a login form from the `oidc-provider` library. I'd definitely rather use our components and (obviously) branding, so I'm totally happy to swap it out. However, to use those components I would have had to set up a React app in this service, so I wanted to make sure that we actually wanted to start out with a email/pw form (vs. e.g. Google SSO) before I put in the work to do that.

bencpeters:
> I think let's start with Google SSO for Exygy accounts, so that we can land this first version without having to build out too much backend, and we can get some experience with what SSO might look like.

@bk3c - Sounds good. I'll add this then in lieu of username/pw form. I believe the experience will then be:

- First time, background login fails, need to click on login. Login prompt is a Google screen that either prompts you to sign in if not signed in (or pick an account to use if you're signed into multiple), and then give Oauth consent after you sign in.

- After that, background login should succeed as long as you're signed in to your google account & don't revoke consent. (it's possible there might be some complications if logged into multiple google accounts, I'll investigate).

> The plan is to deploy on Heroku unless there's a reason we shouldn't, and I'd lean toward postgres for persistent sessions so that we only have one database technology in play, unless there's reason to think it won't be performant or is missing things we need.

Postgres should be fine.

bencpeters:
Closing since we're going in a different direction.

PR #295, opened 2020-05-15
Add translations for strings

### LINKED ISSUE #293: Do another sweep for untranslated strings in public-reference
When I did a diff of public-reference against public-smc, I noticed:

```
pages/privacy.tsx
  const pageTitle = <>Privacy Policy</>
```

was fixed in the smc app (thanks, @software-project), but it didn't make it upstream. If we can do another thorough search through the reference implementation, that would be great. I'll add any others that I notice here.

software-project:
Fixes #293

### DIFF
--- a/apps/public-reference/layouts/application.tsx
+++ b/apps/public-reference/layouts/application.tsx
@@ -7,11 +7,11 @@ import {
   SiteFooter,
   FooterNav,
   FooterSection,
-  ExygyFooter
+  ExygyFooter,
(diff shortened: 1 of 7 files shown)


software-project:
@bk3c I've added missing translations. One thing left are mdx files translations. On that note I would prefer to have all translations in one translation file, so it's easier for someone who actually knows the language to translate it, rather than go through MDX files.

software-project:
@bk3c both fixed

PR #300, opened 2020-05-18
Add css purging

### LINKED ISSUE #78: Add purgecss
Appears to be a best practice. See https://github.com/zeit/next.js/pull/9145 for more details. That issue only modifies the example, though.

See also the caution in that PR about dynamically rendered CSS, if that applies to us?

pbn4:
Should fix #78

### DIFF
--- a/apps/public-reference/next.config.js
+++ b/apps/public-reference/next.config.js
@@ -4,6 +4,7 @@
 const withTM = require("next-transpile-modules")(["@bloom-housing"])
 const withSass = require("@zeit/next-sass")
 const withCSS = require("@zeit/next-css")
+const withPurgeCss = require('next-purgecss')
 const withMDX = require("@next/mdx")()
 const axios = require("axios")
 if (process.env.NODE_ENV !== "production") {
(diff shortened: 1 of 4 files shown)


bk3c:
Ok, I agree with @pbn4 that this implementation as it currently exists isn't doing anything. I spent more time than I should have trying to see if there was an easy fix, but I'm a little lost in the whole import ui-components / tailwind / sass / postcss flow. The best suggestion I could find was to manually [trigger purgecss at the end as described here](https://nextjs.org/learn/basics/assets-metadata-css/styling-tips), but I couldn't get that to `next build` cleanly either.

@pbn4 - If you want to take another try and see if you can get it to work that would be awesome, but if you just run into the same wall that I did, I'm fine with just deferring this indefinitely. It's not a problem that's worth too much time in the short term.

jaredcwhite:
I took a stab at this since I was working on adding PurgeCSS to a couple of other projects. I was able to get it working, but there are some visual glitches due to missing styles we'll want to fix with configuration (whitelist) options:

https://5ed2a19087e8c7ce8c3a8364--clever-edison-cd22c1.netlify.app/listing/jsqvgvy6ejy56tncffqp8/the_triton_55_triton_park_lane_foster_city_ca

vs.

https://bloom.exygy.dev/listing/jsqvgvy6ejy56tncffqp8/the_triton_55_triton_park_lane_foster_city_ca

jaredcwhite:
Documentation here: https://purgecss.com/guides/next.html#next-js-plugin

PR #303, opened 2020-05-20
Fix/nestjs with typeorm

pbn4:


package.json
```diff
@@ -2,13 +2,18 @@
...
+      "shared/core",
+      "shared/ui-components",
+      "apps/public-reference",
+      "services/listings"
+    ],
+    "nohoist": [
+      "**/@types/mocha", "**/@types/jest", "**/@types/gl-matrix"
+    ]
+  },
+ "repository": "https://github.com/Exygy/bloom.git",
```
software-project, line 16: Shouldn't that be https://github.com/bloom-housing/bloom ?

services/listings/ormconfig.json
```diff
@@ -4,23 +4,14 @@
...
    "subscribers": [
-      "dist/subscriber/**/*.ts",
       "src/subscriber/**/*.ts"
-   ],
-   "cli": {
-      "entitiesDir": "src/entity",
-      "migrationsDir": "src/migration",
-      "subscribersDir": "src/subscriber"
-   }
-}
```
software-project, line 26: I thought cli config is required?

package.json
pbn4: I did not change it. It's marked as change because there is a slight indent misalignment

services/listings/ormconfig.json
pbn4: Actually I was not able to find 'cli' key anywhere in typeorm docs. Do you have any reference to manual with it?

software-project: It's here: https://typeorm.io/#/using-cli

PR #311, opened 2020-05-22
Zoom in listing map and update style

bk3c:
Minor tweaks to #308, zooming in 1 more level to show basic POIs more clearly, and switching to the brighter / more conventional "streets" style.


jaredcwhite:
@bk3c This looks great to me. @slowbot Any objections to merging this in?

PR #316, opened 2020-05-27
Upgrade Next.js

### LINKED ISSUE #182: Upgrade to Next.js 9.4
Looks like Next.js 9.3 is out, with [some new features that are meaningful for us](https://nextjs.org/blog/next-9-3#next-gen-static-site-generation-ssg-support).

The minimal version of this is just to update the dependencies, since nothing major seems to be deprecated, but I think it's probably also worth investigating:

- [ ] If we can use the native SCSS handling now and simplify the config, and if there's an opportunity / need to otherwise refactor the CSS to take advantage of their new support
- [ ] Whether it's worth switching to getServerSideProps for listings data, in anticipation of a database-backed service that changes much more frequently than the code does (obviously it shouldn't make any positive difference at the moment, since any content updates will also be repo updates)

 This is probably a good time to update other related dependencies as well, in the general effort to stay current.

software-project:
Fixes #182 

Notes:
We had a call with @jaredcwhite yesterday regarding CSS. We decided not to remove css/sass plugins for now. Although Next.js recommends doing so it would also require us to move all styles to one file from components. Next.js displays a warning now because of that: `Warning: Built-in CSS support is being disabled due to custom CSS configuration being detected.`

Regarding `getServerSideProps` it's not compatible with `_app`. Right now it's only possible to use it on the Page. But it will be possible in future.


bk3c:
@jaredcwhite It looks like a props object with implicit Any type previously snuck through in a couple of form components. Can you define the prop types for:

- shared/ui-components/src/forms/ErrorMessage.tsx(3,23)
- shared/ui-components/src/forms/Field.tsx(4,16)

before we merge this, so Storybook works again?

bk3c:
I also upgraded `@storybook/preset-typescript`, which it looks like we should have done before but it only really broke with the other dependency upgrades here. It's not clear to me whether we're getting everything that we can / should be able to get from the component introspection, so there's probably more work to do there, but on the plus side the "Docs" tab in the Storybook UI seems more functional now than it was before this PR.

Issue #305, opened 2020-05-20
Consolidate SMC CSS overrides into Core


bk3c:
When doing SMC QA over at housingbayarea, we noticed that the styles in `/pages/AdditionalResources.scss` were being applied globally, rather than just to the intended page. While there's probably a future path for having that kind of component-specific styling (with the proper .module.scss as @software-project pointed out), it sounds like it will be better in the short term to just roll the style variants back up into the core CSS and adjust the class usage on the relevant page appropriately.

slowbot:
@bk3c The first pass of this is done to fix the global css overrides. 
https://github.com/housingbayarea/bloom/pull/57

Remaning to do include:
- Replace info-card reference with info-card component from bloom core
- Move shadow-left style to bloom core 
- Create component for info-card-grid with header and card list in bloom-core
- Resolve other jurisdiction Resources pages

slowbot:
@software-project Do you think you have time to take these to dos on early next week? Happy to jump on a call and talk through it.

software-project:
@slowbot will do

slowbot:
@slowbot I created a prototype of info-card-grid markup and styles here: https://github.com/bloom-housing/bloom/pull/307

You can use this as the starting point for the new component

bk3c:
I think this is done. @slowbot can you confirm?

bk3c:
Done per @software-project!

PR #307, opened 2020-05-21
Translation nav

slowbot:
Adding placeholder component for translation navigation

PR #319, opened 2020-05-27
Add info card grid

software-project:
Adds InfoCardGrid for #305


bk3c:
@software-project Now that we're using Lerna, you don't need to increment version numbers in any of your PRs, we do it as part of the publishing process and doing it sooner just breaks CI since the new versions aren't published to NPM yet. Otherwise LGTM.

software-project:
@bk3c uhhh, I was trigger happy on using `yarn new-release` then, will delete version update.

software-project:
@slowbot next step would be merging it to `translation-nav` or merging translation-nav first and then this to master. After that I can push the changes upstream and update the components there.

bk3c:
@software-project I actually wan't aware of `yarn new-release`, so thanks for teaching me something new today (and reverting the commit in addition). 

I think #307 is waiting on @jaredcwhite's review, so I'm good with any merge sequence that makes sense to the two of you, including merging this there or merging that to master when it's ready and then this afterward.

So LGTM and ready to merge as desired.

PR #326, opened 2020-06-02
Add Equal Housing Opportunity icon

### LINKED ISSUE #320: Add the Equal Housing Opportunity logo to the public-reference footer
There might be some accompanying text, which @slowbot is working on defining both for the current jurisdictions in housingbayarea, and for a generic version here.

The "original" versions of the logo [are here for reference](https://www.hud.gov/library/bookshelf11/hudgraphics).

slowbot:
Fixes #320. 

Adding Equal Housing Opportunity and pattern for placement in footer.

To do:
- Needs to be added to specific jurisdictions


bk3c:
@jaredcwhite Is it overkill to wrap the image in a trivial react component (maybe taking light/dark or white/black as props), or is there a better way to expose the SVG up to the app for inclusion in `layouts/application.tsx`?

@slowbot Once we get that decided, we can add it to public reference since I'm assuming that most jurisdictions will want this, and then we can copy it over in housingbayarea as appropriate with very little effort.

shared/ui-components/src/footers/SiteFooter/SiteFooter.stories.tsx
```diff
@@ -49,9 +52,6 @@ export const withLogoAndCredits = () => (
   <SiteFooter>
     <FooterSection>{logo}</FooterSection>
     <FooterSection>{credits}</FooterSection>
-    <FooterSection>
-      <img className="h-16 w-16" src="/images/eho-logo-white.svg" alt="Equal Housing Opportunity Logo" />
-     </FooterSection>
     <FooterNav copyright="Alameda County © 2020 • All Rights Reserved">
```
bk3c: I somehow never noticed that we had Alameda hard coded into the story here (and into the reference implementation FooterNav). Can we take this opportunity to clean it up so the reference implementation and ui-components are generic?

jaredcwhite:
@bk3c The footer shared components don't have any hardcoded content and all images/text/etc. comes from the particular app implementation. Updating the reference app and stories here is fine, but that doesn't have to be tied to any particular release.

It looks like the SJ and SMC footers in housingbayarea pull text from the JSON translations, so we'll just want to update ALA to do that. Adding the EHO icon is similarly something we'll do for each of those footers.

bk3c:
@jaredcwhite So does that mean you want to move this out of `ui-components`? I don't think I have a strong opinion either way, though I will note that we'll probably see this logo on close to every Bloom site. Fine if we want to push it to the implementations (incl. reference), though.

jaredcwhite:
Yeah we've not been including any file-based image assets (aka JPGs) from ui-components to date, so if we were to start doing that I'd want to make sure we understand the rationale and test how that works with Next.js' export pipeline.

Issue #325, opened 2020-06-01
Scaffold out all of the steps in the Application Form


bk3c:
From our design/eng sync meeting this morning, the consensus is that it will be helpful to build out all of the form steps first, and then we can fill in the fields later with a much lower chance of merge conflicts, etc.

Latest Figma link from @slowbot: https://www.figma.com/file/ubE2C6P3XRpR0UAu4RJov0/Core-App-Flow-2?node-id=0%3A1

Write up on conditionals
https://docs.google.com/document/d/1KTfmzP8SgqjIhYXHfg_E4p6gTHE1P4ulRhb1tMHDy1I/edit?usp=sharing

User Flow:
https://app.flowmapp.com/share/3367021645a029eb2c66ff35de1e21a6/userflow/69488/

bk3c:
@jaredcwhite I'm marking this Ready for Dev because I think we need to get started writing code to find out any open questions in the details, but feel free to post those questions back here if/when they turn up.

jaredcwhite:
@slowbot I'm afraid I don't have permissions to access the Figma link. Do I need to sign up for an account, and if so, how?

jaredcwhite:
@slowbot OK I created a jared.white@exygy.com account. I guess I just need to be added to the invite list.

jaredcwhite:
@bk3c @software-project I realized I hadn't added any of the new form steps to the static routes in `next.config.js`…I just put a PR up which includes the first step — aka

```js
              "/applications/start/choose-language": {
                page: "/applications/start/choose-language",
              },
```

and stepping through the forms works fine after the initial step, but if you directly go to one of the step URLs you get a 404.

I'm wondering if there's any more elegant solution we can come up at this point than hardcoding all the URLs in `exportPathMap`…maybe we can write a little utility to glob the file paths and construct a list automatically? Thoughts?

bk3c:
Calling this one generally done, with follow-on issues to come for the last few details.

PR #330, opened 2020-06-04
WIP: Application Forms scaffolding

jaredcwhite:
Addresses #325 

I moved the old demo form to /applications/archived/new. The new flow begins at /applications/start/choose-language

Only a few steps in the full flow done so far. BTW, I'm finding it a challenge to ascertain if conditional steps should be handled by state management within a single page, or if we should define each one as a separate page and use the Next router to go between them. I'm leaning towards managing state within a single page as the initial approach, but we can certainly revisit that as time goes on.

Not too concerned with visual fidelity right now. I expect that to be done in a future pass.


jaredcwhite:
All right, I think this PR is ready for wider attention…I got most of the steps in there (basically just missing some of the intermediate conditional lists). We'll also need to search-and-replace the ` <h5 className="font-alt-sans text-center mb-5">LISTING</h5>` heading on all the pages once we figure out how we're tracking which listing the application is for. The step logic objects in `src/forms/applications` are only present for the first few steps, but it should be pretty straightforward how to add new ones.

apps/public-reference/lib/ApplicationConductor.ts
```diff
@@ -38,9 +38,11 @@ export default class ApplicationConductor {
...
 
-  reset() {
+  reset(shouldSync = true) {
     this.application = blankApplication()
-    this.context.syncApplication(this.application)
+    if (shouldSync) {
+      this.context.syncApplication(this.application)
+    }
     if (typeof window != "undefined") {
       window.localStorage.removeItem("bloom-app-autosave")
```
bk3c, line 47: Can we replace all localStorage calls with sessionStorage, so that form information doesn't persist longer than anyone expects?

jaredcwhite:
> 1. …wondering if we should think about migrating back toward JSON fields now that we've landed MD support there?

Agreed, I'll take out the MDX sidecar files—that was sort of a holdover approach from the past and I'd prefer to stick with JSON strings and/or future CMS-ish things we might implement.

> 2. The other thing I definitely anticipate loading from the backend before too long is the configuration that defines which application questions will be asked for any particular listing…This seems like a good match for the ApplicationConductor as a place to read that kind of config in the future and drive the logic

Yep, that makes sense to me. I think it'd be pretty straightforward to conditionally show or hide fields or even change the routing flow based on that particular config.

apps/public-reference/lib/ApplicationConductor.ts
jaredcwhite: Will do.

PR #331, opened 2020-06-04
Add Authentication to Nest.js Backend

### LINKED ISSUE #133: Build out User/Authentication Service v0
**Infrastructure:**

- [ ] Database + ORM
- [ ] Schema create migrations
    - users (history only in v0)
    - permissions
- [ ] Object versioning

**User Endpoints:**

- [ ] /authenticate => returns JWT

**Admin Endpoints:**

- [ ] /user/{ID}/permissions => shows permissions for userID
- [ ] /user/{ID}/setPermission => sets a permission for userID

bencpeters:
# Add Authentication to Nest.js Backend
Adds authentication & user models to Nest.js backend.

This adds two new Nest.js modules: `user` and `auth`. For the moment users are simple - email, bcrypt-based password hash, timestamps, and uuid.

Initial login happens via a POST to `/auth/login` with `username` and `password`.

Todo:
- [x] Add JWT strategy
- [x] Function/strategy documentation
- [x] Coordinate on Frontend UI with @jaredcwhite 
- [x] Tests

BK: I'm going to call this fixes #133, with the idea that we have an implemented plan, and we can add more specific issues going forward.


backend/core/package.json
```diff
@@ -25,11 +25,15 @@
     "@nestjs/cli": "^7.0.0",
     "@nestjs/common": "^7.1.0",
     "@nestjs/core": "^7.1.0",
+    "@nestjs/passport": "^7.0.0",
     "@nestjs/platform-express": "^7.1.0",
     "@nestjs/typeorm": "^7.1.0",
+    "bcrypt": "^4.0.1",
```
bk3c, line 31: Can you take a quick look at using `scrypt` instead of `bcrypt`? My understanding is that it's generally preferred from a security perspective these days ([e.g. here](https://latacora.singles/2018/04/03/cryptographic-right-answers.html)), and it's built into Node 10+ so it would save us an external package to keep current in a security sensitive (if admittedly pretty stable) place.

I'm ok with punting it to a subsequent PR if it's somehow a more difficult switch than I think. bcrypt isn't all that bad, but I'd just like to save us all the migration logic later.

backend/core/src/entity/User.ts
```diff
@@ -0,0 +1,17 @@
+import { Entity, PrimaryGeneratedColumn, Column, CreateDateColumn, UpdateDateColumn } from "typeorm"
+
+@Entity()
+class User {
+  @PrimaryGeneratedColumn("uuid")
```
bk3c, line 5: I'm going to file a separate issue about this for discussion, but I was sort of hoping that we could use nanoids as the native primary keys across the board. I don't think it's a merge blocker while we sort it out, whichh hopefully will be quickly.

backend/core/package.json
bencpeters: I think it's worth making this choice _before_ we have data rather than migrating later ;-). Migrating hashes is pretty problematic for the obvious reason that we don't actually have the original PW. 

I can use `scrypt`, it's _slightly_ more effort that `bcrypt` because you have to save your `salt` as well to verify, but I can just concat that onto the `hash` string itself just like `bcrypt` does internally.

backend/core/src/entity/User.ts
bencpeters: Ya, I'm pretty neutral on this, but definitely be good to be consistent.

backend/core/src/entity/User.ts
```diff
@@ -0,0 +1,33 @@
...
+  PrimaryGeneratedColumn,
+  Column,
+  CreateDateColumn,
+  UpdateDateColumn,
+  Index,
+} from "typeorm"
+
+@Entity()
+@Index("user_email_lower", { synchronize: false })
+class User {
```
bk3c, line 12: I'm ashamed to admit I just spend time chasing this, but `user` is essentially a reserved word in postgres because it's a built in function. Can we call the table something else without making life too complicated? I'd be fine with pluralizing everything for the DB rails-style, or just picking something else in this case -- whatever the consensus practice we prefer is.

apps/public-reference/pages/sign-in.tsx
```diff
@@ -1,13 +1,38 @@
...
+
+  const onSubmit = async (data: { email: string; password: string }) => {
+    const { email, password } = data
+
+    try {
+      const res = await axios.post(`${apiBase}/auth/login`, { username: email, password })
+      const { accessToken } = res.data
+      console.log(`Got access token '${accessToken}'`)
+    } catch (err) {
+      const { status } = err.response
```
bk3c, line 29: This throws an uncaught error when the backend is down, I think because that case leaves `status` undefined. Can we handle that case a little more cleanly?

bencpeters:
@bk3c This is ready for merging now I think.

backend/core/src/entity/User.ts
bk3c: Captured in #348.

PR #332, opened 2020-06-04
Make listings service and tests use active record

pbn4:
Master branch uses Data Mapper and the PR changes that to use Active Record

##### Data Mapper vs Active Record:
- No need for class-transformer in Data Mapper (otherwise typescript complains about missing database access methods),
- Explicit dependencies in Data Mapper (one can see exactly what service depends on because it's injected in the constructor),
- Active Record requires entity classes to extend `BaseEntity` and based on what I see in [typeorm-model-shim](https://github.com/typeorm/typeorm/blob/master/extra/typeorm-model-shim.js) this is not going to work with models exporting since model shim is only providing empty definitions for decorators

### DIFF
--- a/backend/core/src/entity/Listing.ts
+++ b/backend/core/src/entity/Listing.ts
@@ -1,11 +1,11 @@
-import { Entity, PrimaryGeneratedColumn, Column, OneToMany } from "typeorm"
+import { Entity, PrimaryGeneratedColumn, Column, OneToMany, BaseEntity } from "typeorm"
 import { Unit } from "./Unit"
 import { Preference } from "./Preference"
 import { Attachment } from "./Attachment"
 import { Address, UnitsSummarized } from "@bloom-housing/core"
 
 @Entity()
-class Listing {
+class Listing extends BaseEntity {
   @OneToMany((type) => Preference, (preference) => preference.listing)
   preferences: Preference[]
   @OneToMany((type) => Unit, (unit) => unit.listing)
--- a/backend/core/src/listings/listings.service.ts
+++ b/backend/core/src/listings/listings.service.ts
@@ -1,6 +1,4 @@
 import { Injectable } from "@nestjs/common"
-import { InjectRepository } from "@nestjs/typeorm"
-import { Repository } from "typeorm"
 import { Listing } from "../entity/Listing"
 import { amiCharts } from "../lib/ami_charts"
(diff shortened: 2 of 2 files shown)


jaredcwhite:
@pbn4 Thanks for digging into this. I like what I see so far! Your concern about `class-transformer` seems related to the spec test, yes? That seems like a reasonable trade-off to me.

If the TS shim for the frontend doesn't work though with Active Record, that might be a deal breaker. I'll give this a try and see if I can get that working.

jaredcwhite:
@pbn4 OK, awesome—the TypeORM shim does work! (I had to add `BaseEntity` to it, not sure why it wasn't already there.) Both Next.js is compiling/running successfully and VS Code is loading in all the type definitions through `backend/core` in the monorepo.

I also confirmed that no actual code from TypeORM leaked into the frontend. In fact, at first I didn't get the Webpack alias config right, and when I inspected the compiled Next.js chunks, there was a ton of TypeORM stuff in there. Not good. But once I got Webpack aliasing and the shim working, nothing's there now. So no worries about a TypeORM dependency for the frontend. :+1:

shared/core/index.ts
```diff
@@ -1,6 +1,6 @@
 export * from "./src/HousingCounselors"
 export * from "./src/ami_charts"
 export * from "./src/general"
```
bk3c, line 3: I realize that being able to do this is more-or-less the whole point of the exercise here, but exporting across package boundaries like this doesn't feel right to me. If everyone is good with the path in this PR generally, I almost wonder if we shouldn't just retire `shared/core` as a separate directory, and then publish the interfaces out of `backend/core` directly (taking care not to publish the whole backend for size reasons). I'm not sure what the existing best practices / prior art is here, so open to suggestions...

jaredcwhite: @bk3c I think whether the interfaces are in shared/core or backend/core, the entire package gets installed locally to node_modules to the purpose of being accessible by the Next app's Webpack—however, virtually nothing gets imported into the app bundle due to the TypeORM shim we use. So I don't necessarily feel we need to change this approach (we might, and still do, have interfaces that aren't DB-related that we can export to anywhere—thus shared/core still seems relevant).

bk3c: Yeah, ok. We can revisit this in a few months and see if the pattern still makes sense once we've built out more.

Issue #140, opened 2020-01-20
Pick a solution for API documentation and build v1 for the listings service


bk3c:
In order to make the new contributor path as viable as possible, we should have web-based API documentation for services that's complete and helpful. This is both for FE-developers knowing what they need to consume, as well as service developers knowing what they need to provide in a non-reference implementation.

The ideal solution will yield great output while adding minimal developer overhead to keep documentation current. Output doesn't need to be fancy (e.g. english-only is fine for the foreseeable future) so long as it's reasonable to navigate.

Some potential solutions include:

- [Swagger UI](https://swagger.io/tools/swagger-ui/)
- [Stoplight](https://stoplight.io/)

Or we could use a more general documentation platform as part of an effort to have broader documentation, e.g.:

- [Docusaurus](https://v2.docusaurus.io/)

bk3c:
Now that we made the choice for NestJS on the backend, it seems like their `swagger-ui-express` recipe might be an easy first step in the right direction. @pbn4 I recall you suggesting this earlier on -- want to take a first stab at turning it on?

PR #336, opened 2020-06-05
Add swagger docs generation

pbn4:
Related issue: https://github.com/bloom-housing/bloom/issues/140

Swagger docs are available under: `/docs` path. 

It's not complete because `Listing`, 'Attachment', 'Preference', 'Unit' entities have not been anotated. I did not annotate them because there is a mismatch between TypeORM annotations and the entity class themselves and it's mostly related to nullability. @software-project Could you help me with determining what fields should be nullable (a nullable property should be of type 'T | null')?  

IIRC Typescript is missing some reflection features that help avoid such duplication in decorators as e.g. 'isArray'. This is going to get worse once we decorate entities with @ApiProperty.


pbn4:
Models are now automatically anotated with nestjs/swagger plugin. To make use of this plugin it's best if we stick with NestJS file naming conventions. By default plugin generates annotations for *.entity.ts and *.dto.ts files.

pbn4:
Closing this in favor of https://github.com/bloom-housing/bloom/pull/357

PR #337, opened 2020-06-06
Add js factories with fishery

software-project:
A test PR to see if typescript factories are for us. Factories are make with Fishery: https://thoughtbot.com/blog/announcing-fishery-a-javascript-and-typescript-factory-library
Library is based on most popular factory solution for Rails. 
As I was working on that I took a look at out field types and made comments in places we should address them, eg. string fields that should be numeric or fields that could be nullable (@pbn4 a kickstart for our Monday task)

### DIFF
--- a/backend/core/test/factories/listing.ts
+++ b/backend/core/test/factories/listing.ts
@@ -1,0 +1,79 @@
+import { Factory } from "fishery"
+import { nanoid } from "nanoid"
+import { Listing } from "../../src/entity/Listing"
+
+export default Factory.define<Listing>(({ sequence, factories }) => ({
+  id: nanoid(),
+  preferences: [],
+  units: [factories.unit.build()],
+  attachments: [],
+  acceptingApplicationsAtLeasingAgent: true,
+  acceptingApplicationsByPoBox: true,
+  acceptingOnlineApplications: false,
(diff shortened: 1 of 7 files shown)


backend/core/test/factories/listing.ts
```diff
@@ -0,0 +1,79 @@
...
+  applicationDueDate: "2019-12-31T15:22:57.000-07:00",
+  applicationOpenDate: null,
+  applicationFee: "30.0",
+  applicationOrganization: `Triton-${sequence}`,
+  applicationAddress: {
+    city: `Foster City-${sequence}`,
+    street: "55 Triton Park Lane",
+    zipCode: "94404",
+    state: "CA",
+    county: "San Jose",
```
bencpeters, line 25: Do we want to use [faker](https://github.com/marak/Faker.js/) for making more varied data here? It has address utilities...

If determinism is an issue, faker can de made deterministically random using `faker.seed(deterministicSeed)` (so the random data it generates will be the same with the same seed value)

bk3c: I agree faker could be useful here, perhaps in combination with a set of static test data that we know will be consistent for human QA. I think it can wait for a later PR, though.

PR #341, opened 2020-06-12
Add first come first serve option for listings

software-project:
First come first serve option for latest San Mateo and Alameda listings. @slowbot @bk3c


jaredcwhite:
I might be missing a bit of context. Is this intended for listings where both the open date and the due date are null at once? Also yes, I agree with @bk3c that we should keep all the tests and storyshot data from before and add new tests for this particular data state.

software-project:
@bk3c @jaredcwhite that's a good point. I've reverted Archer's date change and added a story for the new case.

software-project:
@bk3c @jaredcwhite can we merge this one?

PR #342, opened 2020-06-13
Allow Markdown in Listing pages' expandable text

### LINKED ISSUE #314: Support Markdown text in some listing fields
There are a number of listing fields that would benefit from some basic formatting ultimately entered by the Partners users without any engineering effort desired. Current consensus is that Markdown should be sufficient for the foreseeable requirements (e.g. carriage returns, occasional bullets) and that allowing raw HTML is undesirable from a security and usability perspective.

The goal with this issue is to allow at least some Listing fields in Markdown, and to have the appropriate HTML rendered for display. I think having the field definitions fixed in Typescript is fine -- they don't need to be changeable on the fly or anything. Obviously we need to make sure that generic plain text in a Markdown field also renders as it does now with no extra wackiness.

I'd suggest prototyping this with the current JSON-based storage, knowing that it's soon to be replaced by something DB/TypeORM-backed, but presumably just stored as a string in the DB.

jaredcwhite:
Fixes #314


jaredcwhite:
Example (scroll down to Credit History): https://5ee41dcad71752f52cc5b6b6--clever-edison-cd22c1.netlify.app/listing/jsqvgvy6ejy56tncffqp8/the_triton_55_triton_park_lane_foster_city_ca

So the surface area of this PR is pretty small…it's just an enhancement to the ExpandableText component.

But it raises a larger question about how we're using Markdown. So initially we picked the MDX library because of its popularity (used by Gatsby, for instance) and it seemed to get the job done. However, I was surprised when I started working on this issue and discovered that MDX out of the box _only_ works with a static build process. To process arbitrary Markdown string data and also parse at runtime, you have to install the separate [MDX Runtime library](https://mdxjs.com/advanced/runtime). I initially went with that approach, but I felt a bit uneasy about it because it comes with a scary warning message that it adds lots of extra size to the JS bundle and you have to be cautious with user input (and once we have more of a CMS environment, we'll need to consider that).

So I shopped around a bit and found another library called [Markdown-to-JSX](https://github.com/probablyup/markdown-to-jsx). It seems fairly well maintained judging from the commit history, and the nice thing about it is it boasts a small bundle size, works in static or runtime environments, and—bonus—you can configure it so (unlike standard Markdown) it strips out all HTML and only processes Markdown syntax. That seems like a big win for all of our content whether coming from JSON in the repo or from APIs.

My thought is we merge this in for now, but consider migrating from MDX to this broadly speaking. The only blocker there could be if Markdown-to-JSX' ability to embed React components in Markdown doesn't work as well as MDX. It claims to support it but I didn't get that far.

Let me know what you think :)

jaredcwhite:
@bk3c All right, I got the tests working. Not quite sure what you mean re: MDX and editable fields, but sure that's something we can work out down the road.

bk3c:
I guess I'm ok doing something custom if we have to, but is there a reason that we didn't just add the @types package from NPM?

jaredcwhite:
I tried that first, but oddly enough it was missing that one option I was using.

bk3c:
🤦 It looks like they landed `disableParsingRawHTML` on the same day they released, so not 100% surprising it's not in the types. 

I'll merge this as-is in the meantime, but would you be willing to do a 1-liner (+comment) PR upstream and submit to DefinitelyTyped, and then we can switch over whenever that's accepted? If you feel like it's going to take you too off track from application form work, maybe @bencpeters or @pbn4 could be enlisted to help -- I'm thinking the testing + filling out the PR template is 98% of the work here.

jaredcwhite:
Sure I can do that, but it looks like there's a lot of ceremony around submitting a PR (testing, listing, etc.) and even more so for first-time contributors, so let's file an issue for that and I'll keep it on my radar.

PR #346, opened 2020-06-16
Allow external application link

software-project:
Moved from Eastlake listing PR on housingbay. Allows online application with an external link.


shared/ui-components/src/page_components/listing/listing_sidebar/Apply.tsx
```diff
@@ -3,7 +3,6 @@ import { Listing, Attachment, AttachmentType } from "@bloom-housing/core"
 import moment from "moment"
 import t from "../../../helpers/translator"
 import Button from "../../../atoms/Button"
-import LinkButton from "../../../atoms/LinkButton"
```
jaredcwhite, line 6: I'd like to maintain consistency and continue using the `LinkButton` component. Was there a feature it needs to do in this case that it wasn't providing?

software-project:
@jaredcwhite @bk3c this one should also be ready now

PR #353, opened 2020-06-18
Fix all the latest form build errors

jaredcwhite:
Sorry about all the errors folks! Just running the local dev server doesn't always catch these things, I need to run `yarn build` to verify.


jaredcwhite:
FYI, I just took out the old demo forms…probably don't need those any more even for reference.

PR #355, opened 2020-06-19
Create UserContext to track login state

bencpeters:
# Create UserContext to track login state

This creates a `UserContext` component in `shared/ui-components` to provide auth logic to any app implementations that need it.

On successful login/registration, it stores the JWT token in either `local` or `session` storage (configurable via a prop on the `UserProvider`; defaults to `session`). On reload/mounting, this component checks for a stored token, if the token is still valid, it loads the user's profile into memory (no user info other than the access token itself is stored in either session or local storage).

This component also schedules/performs a background token refresh 60 seconds before the token expires to keep the user logged in during a session.

The context object makes the following available to consumers:

```typescript
const {
  login: (email: string, password: string) => Promise<void>
  register: (user: CreateUserDto) => Promise<User>
  // Flag for whether a login or register operation is pending
  loading: boolean
  profile?: User
  // Authenticated client for making requests to the API with auth. Configured with auth headers
  // and baseURL, to allow usage: `axiosClient.get('/users/profile')`
  axiosClient?: AxiosInstance
} = useContext(UserContext);
```


apps/public-reference/pages/create-account.tsx
```diff
@@ -1,21 +1,29 @@
...
+  FormCard,
+  Icon,
+  LinkButton,
+  UserContext,
+} from "@bloom-housing/ui-components"
 import FormsLayout from "../layouts/forms"
 import { emailRegex } from "../lib/emailRegex"
 
 export default () => {
+  const { register: registerUser } = useContext(UserContext)
```
jaredcwhite, line 15: I know we have the register vs. register conflict here, but I'm wondering if we can rename the original function name so it doesn't have to get rewritten?

backend/core/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.2.3",
   "description": "Listings service reference implementation for the Bloom affordable housing system",
   "author": "Marcin Jedras <marcin.jedras@exygy.com>",
-  "private": true,
+  "private": false,
```
bk3c, line 6: Is the intent here to start publishing this to NPM? I'm not morally opposed to superseding the current `shared/core` package if we think packaging from here makes more sense (though I do have size concerns we'd need to address with a tight `.npmignore`), but I got the impression for a comment on another issue that we didn't want to do that?

bencpeters: Ooops, no! This was actually a mistake, I didn't mean to commit it. I was having trouble getting `ui-components` package to import the `backend-core` package, and I briefly thought this might fix the workspace setup (turns out I just had the `index.js` file in the wrong directory at first). I think this should probably stay `private: true` for now at least.

bk3c: #whywehavecodereview

bencpeters:
Ok, refactored into hopefully more modular logic.

pbn4:
@bencpeters In application API PR I added this https://github.com/bloom-housing/bloom/pull/357/files#diff-1e1ad015a963d36214620212ec471e07R11  Maybe it's worth changing this field to 'usernameField' here too?

pbn4:
@bencpeters If access token is a refresh token too then user will be logged in forever if a browser is left open in the background. Also stealing an access token allows such access. 

I am used to the approach where access tokens and refresh tokens are separated and only refresh tokens are revokable. In such setup access token is short lived, same as here, but the refresh token is e.g. valid for 7 days. Section [using refresh tokens](https://auth0.com/learn/refresh-tokens/)

Issue #130, opened 2020-01-10
Build out Application Service v0


jacchau:
**Infrastructure:**
- [x] Database + ORM
- [ ] Schema create migrations
    - listing
    - unit
    - preference_rule
    - income_eligibility_chart (fka AMI chart)
- [ ] Object versioning

**Build out endpoints:**
- [ ] /Applications/submit => returns ID
- [ ] /Applications/{userID}/list => returns list of applications received for userID
- [ ] /Applications/{ID} => returns Application JSON

bk3c:
@pbn4 I realize that we're still working on a number of things on the Infrastructure list above, but do the endpoints below make sense to stub out? I'm not wedded to the exact naming or semantics listed above (e.g. the `list` endpoint taking the userID as a query param might be better), but it seems like maybe we can build little hard-coded stubs for each of them to get the API details ironed out and the front-end components in-progress?

pbn4:
@bk3c Here's my proposition:

```
POST /users/<userId>/applications        => returns appId
GET /users/<userId>/applications         => returns applications array for a given user
GET /users/<userId>/applications/{appId}  => returns single Application JSON
PUT /users/<userId>/applications/{appId}  => updates existing application
```

Then it **should** play nicely with permissions e.g. user can access only those that match his uuid and a super user can access any. 

For super users I'd create a separate endpoint for querying 

```
GET /applications?userId=id&listingId=id => returns all applications (queryable by user and listing id)
```

What do you think?

bk3c:
@pbn4 That looks great to me as a starting point, so let's go with it and we can adjust later if need be.

As a way to get started while we figure out the eventual object model, what do you think about just having an `applications` table with a big jsonb column for the content of the application, along with the normal meta fields for ID, created_at, etc? I'm mostly sure it's not where we want to end up, but it seems like the quickest way to get basic functionality of saving and loading.

pbn4:
I recognize that we had a discussion about having permission guard at service level. According to [this](https://docs.nestjs.com/faq/request-lifecycle) services can not be decorated with guards decorators. Maybe it's a good moment to think about https://github.com/bloom-housing/bloom/issues/315

Issue #315, opened 2020-05-27
Permissions library? / object model v1


bk3c:
Now that we're settling on a NestJS / Passport [vision of authentication](https://docs.nestjs.com/techniques/authentication) (authn), at least in the short-medium term, we can turn our attention to authorization (authz), where there are more complex decisions for us to make.

In general, I think we should expect to use a version of role-based access control (RBAC) for the foreseeable future. If there's something off the shelf that would let us implement attribute-based access control instead, and we thought it would really be more robust for a similar amount of work, that could also be something worth discussing.

At runtime, we're looking for an authz solution that can easily plug into the [Guards](https://docs.nestjs.com/guards) infrastructure provided by NestJS, and where the authn part can remain with Passport so we get the benefit of all the strategies included there. We also want the underlying permission information to be stored in the database, presumably via TypeORM. In an ideal world, we'd have a clear path to splitting permissions storage by module (e.g. storing permissions on Listings separately than permissions on Applications, in case we ever split those two services), but so long as we have a pattern we can duplicate when the time comes there doesn't seem to be a big downside to just storing all the permissions information in the same place for now.

In addition to roles, the other key concept that we need to be able to support is permissions at various levels of a scope hierarchy. For example, the permission to edit a listing might be implemented only for that property (e.g. for a property manager), but it might also be implemented at a city or county level for the appropriate jurisdiction staffs. I'll cover the object model for that hierarchy in a separate issue, but whatever code is doing the evaluation will need to be able to traverse appropriately.

bk3c:
Some libraries that we might consider using, stealing from, or otherwise being inspired by:

- [CASL](https://github.com/stalniy/casl)
- [Casbin](https://github.com/casbin/node-casbin)
- [AccessControl](https://github.com/onury/accesscontrol) (less maintained, but leans toward ABAC in a maybe-good way)

And see also for reading:

- https://www.ekransystem.com/en/blog/rbac-vs-abac

hsluoyz:
Hi guys, Casbin provides a Nest.js middleware, which supports TypeORM exactly: https://github.com/dreamdevil00/nest-authz.

It also has an example: https://github.com/dreamdevil00/nest-authz-example

software-project:
@pbn4 for reference here is Pundit that I used to use on Ruby side: https://github.com/varvet/pundit

bk3c:
Some thoughts based on a design conversation that I had with @bencpeters:

[...]

PR #357, opened 2020-06-19
Add Application api stubs 

pbn4:
Closes https://github.com/bloom-housing/bloom/issues/130 and https://github.com/bloom-housing/bloom/issues/140

@bk3c I decided to move swagger + listings service stuff here since there were conflicts and the swagger part overlaps. It was easier for me to simply copy the listings changes here. 

@bk3c @bencpeters @software-project @jaredcwhite I renamed files containing entities to `className.entity.ts` so that swagger plugin can pick them up by suffix. Otherwise we would have to define DTOs fo every entity class manually with all the properties basically copied from the original entity. 

[\*] Make Listings service use swagger plugin
[\*] Implement `/applications` route 
[\*] Implement `/user/{userId}/applications`
[\*] Write e2e tests for `/user/{userId}`
[] Write e2e tests for `/applications` (not doing that since roles e.g. 'admin' are missing, route is protected by an auth guard requiring JWT token to contain roles prop)

At the moment API accepts arbitrary json as application (see application prop in POST/PUT body dto)


backend/core/src/application/application.controller.ts
```diff
@@ -0,0 +1,18 @@
+import { ClassSerializerInterceptor, Controller, Get, Query, UseInterceptors } from "@nestjs/common"
+import { ApplicationListQueryParams } from "./application.dto"
+import { ApplicationService } from "./application.service"
+import { plainToClass } from "class-transformer"
+import { Application } from "../entity/application.entity"
+
+@Controller("application")
+export class ApplicationController {
```
jaredcwhite, line 8: For consistency (and also because Application can have many meanings), can we rename this to `ApplicationsController` (plural) like we have `ListingsController` (and same for the service)?

backend/core/src/application/application.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from "@nestjs/common"
+import { InjectRepository } from "@nestjs/typeorm"
+import { Application } from "../entity/application.entity"
+import { Repository } from "typeorm"
+
+@Injectable()
+export class ApplicationService {
+  constructor(
+    @InjectRepository(Application) private readonly applicationRepo: Repository<Application>
```
jaredcwhite, line 9: Can we use the same ActiveRecord-style pattern here as we do for the Listings?

backend/core/src/application/application.controller.ts
pbn4: Done

backend/core/src/application/application.service.ts
pbn4: Done. @bencpeters I guess the same applies to user and auth modules?

pbn4:
@bencpeters Should I create a module for Applications the same way you did for User and Auth?

pbn4:
Another question. What's the problem here?

```
// Use require because of the CommonJS/AMD style export.
// See https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require
import dbOptions = require("../../ormconfig.test")
```

 @bencpeters Are you the author? Is this related to our monorepo setup? I have a side project using NestJS and I did not hit that anywhere.

pbn4:
@software-project I have put first test for applications in test/applications, is that correct? 

@jaredcwhite @bencpeters @bk3c @software-project  Regarding our discussion about having a test DB I made an initial step in this direction [here](https://github.com/bloom-housing/bloom/pull/357/files#diff-b6549d14994f69cfcba779f356a42529). I created `ormconfig.test.ts` a separate config for the test DB and added `clearDb` a utility dropping every table so the test DB is clear. In my test I call it in `beforeAll` since there is a single test at the moment. Normally I believe it should be called in `beforeEach` so that our tests are independent from each other.

pbn4:
I recognize that we had a discussion about having permission guard at service level. According to [this](https://docs.nestjs.com/faq/request-lifecycle) services can not be decorated with guards decorators. 

So far I added auth guards a at controller level: 

```
...
@Controller("user")
@UseGuards(JwtAuthGuard)
@UseGuards(OwnerGuard)
@UseInterceptors(ClassSerializerInterceptor)
export class UserApplicationsController {
...
```

where OwnerGuard is defined as follows: 

```
import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common"

@Injectable()
export class OwnerGuard implements CanActivate {
  canActivate(context: ExecutionContext): boolean {
    const request = context.switchToHttp().getRequest()
    return request.user.id === request.params["userId"]
  }
}
```

Maybe it's a good moment to think about https://github.com/bloom-housing/bloom/issues/315  ?

pbn4:
Ready for review

pbn4:
@software-project I took your listings seeding code and wrapped into a service so that seeder now allows using dependency injection and services (it was useful when creating a user with hashed password)

backend/core/src/listings/listings.service.ts
```diff
@@ -1,23 +1,39 @@
...
+}
 
 @Injectable()
 export class ListingsService {
-  public async findAll() {
-    const listings = await Listing.find({ relations: ["units", "attachments", "preferences"] })
+  constructor(@InjectRepository(Listing) private readonly repo: Repository<Listing>) {}
+
+  public async list(options?: ListingsListQueryParams): Promise<ListingsListResponse> {
+    let listings = await this.repo.find({ relations: ["units", "attachments", "preferences"] })
```
jaredcwhite, line 20: Whoops, was this a regression? (changing from `Listing.find`?)

pbn4: Fixed

pbn4:
@jaredcwhite their purpose is described in here https://github.com/bloom-housing/bloom/issues/130 . 

>  One seems to be for public access of data,

What do you mean by public? One must have an admin role to access `/applications`. This is created specifically for superusers querying over applications table. 

> Can we simplify the API endpoint down to one and add logic there to allow or deny requests based on if auth is required?

Answer to this question depends on the answer to the previous one.

backend/core/src/listings/listings.module.ts
```diff
@@ -0,0 +1,13 @@
+import { Module } from "@nestjs/common"
+import { TypeOrmModule } from "@nestjs/typeorm"
+import { ListingsService } from "./listings.service"
+import { ListingsController } from "./listings.controller"
+import { Listing } from "../entity/listing.entity"
+
+@Module({
+  imports: [TypeOrmModule.forFeature([Listing])],
```
bencpeters, line 8: Maybe list the other related entities here (`unit`, `attachment`?) rather than at the root level?

backend/core/src/seeder/listings-seeder/listings-seeder.service.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+  beforeEach(async () => {
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [ListingsSeederService],
+    }).compile();
+
+    service = module.get<ListingsSeederService>(ListingsSeederService);
+  });
+
+  it('should be defined', () => {
+    expect(service).toBeDefined();
```
bencpeters, line 16: Do we want this placeholder test here? I saw you deleted the other ones...

backend/core/src/seeder/listings-seeder/listings-seeder.service.ts
```diff
@@ -0,0 +1,50 @@
...
+    const skipped = ["id", "units", "attachments", "preferences"]
+    const types = { units: Unit, attachments: Attachment, preferences: Preference }
+
+    const connection = getConnection()
+    const listings = listingsSeeds as any[]
+
+    for await (const listing of listings) {
+      const l = new Listing()
+      const entities = {}
+      for await (const key of Object.keys(listing)) {
```
bencpeters, line 21: You can use `for (const <property> in <object>)` to iterate through objects.

backend/core/src/user-applications/user-applications.controller.ts
```diff
@@ -0,0 +1,63 @@
...
+import { ApplicationCreateDto } from "./application.create.dto"
+import { ApplicationUpdateDto } from "./application.update.dto"
+import { ApplicationDto } from "../applications/applications.dto"
+import { JwtAuthGuard } from "../auth/jwt.guard"
+import { OwnerGuard } from "../auth/owner.guard"
+import { ApiBearerAuth } from "@nestjs/swagger"
+
+@Controller("user")
+@UseGuards(JwtAuthGuard)
+@ApiBearerAuth()
```
bencpeters, line 22: Isn't this duplicating the JWT logic from the Auth guard? It checks for a `Bearer token` too...

backend/core/src/application/application.service.ts
bencpeters: Yes

bencpeters:
> @bencpeters Are you the author? Is this related to our monorepo setup? I have a side project using NestJS and I did not hit that anywhere.

@pbn4 So, the issue is that we need access to `ormconfig.ts` in two different places. One is imported using TS-style imports into the application loader, the other is imported by the `typeorm` cli utility for configuration. The type ORM CLI appears to require Common JS style exports with the export being an object with keys and values:

```javascript
export = {
  type,
  url,
  ...
}
```

But for the application, we want to be able to import all of the exports as an object that we can spread:

```javascript
    TypeOrmModule.forRoot({
      ...dbOptions,
      autoLoadEntities: true,
    }),
```

If we didn't have to support common JS too, we would just do this:

```typescript
// Export definition (in ormconfig.ts)
const options = { ... };
export default options;

// Import
import dbOptions from '../ormConfig'
```

But the CLI doesn't appear to be able to pick that up correctly.

jaredcwhite:
> @jaredcwhite their purpose is described in #130 .
> 
> > One seems to be for public access of data,
> 
> What do you mean by public? One must have an admin role to access `/applications`. This is created specifically for superusers querying over applications table.
> 
> > Can we simplify the API endpoint down to one and add logic there to allow or deny requests based on if auth is required?
> 
> Answer to this question depends on the answer to the previous one.

I guess what I'm driving at is I would expect a single API endpoint (`/applications`) to deal with the data model "applications". What that endpoint is allowed do (aka when querying or saving data) depends on the request auth. Let me know if we should hop on a chat to discuss further.

backend/core/ormconfig.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  // Pass
+}
+
+export = {
+  type: "postgres",
+  host: "localhost",
+  port: 5432,
+  database: "bloom_test",
+  synchronize: true,
+  logging: false,
```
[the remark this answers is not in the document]
pbn4: Done

backend/core/package.json
```diff
@@ -17,7 +17,9 @@
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
     "test:e2e": "jest --config ./test/jest-e2e.json",
     "typeorm": "ts-node ./node_modules/.bin/typeorm",
+    "typeorm-test": "ts-node ./node_modules/.bin/typeorm --config ./ormconfig.test.ts",
     "herokusetup": "node heroku.setup.js",
+    "test:seed": "ts-node src/seed.ts --test",
```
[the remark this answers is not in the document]
pbn4: Yup it actually makes sense once I looked at other scripts in package.json, those starting with `test:`  are actually testing something. :)

backend/core/src/app.module.ts
```diff
@@ -1,28 +1,31 @@
...
+import { ApplicationsModule } from "./applications/applications.module"
 
 @Module({
   imports: [
     TypeOrmModule.forRoot({
       ...dbOptions,
       autoLoadEntities: true,
     }),
-    TypeOrmModule.forFeature([Listing, Attachment, Preference, Unit]),
+    TypeOrmModule.forFeature([Listing, Attachment, Preference, Unit, Application]),
```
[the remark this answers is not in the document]
pbn4: I agree. The same applies also for listing related entities registered here, since I created a module for them too.

backend/core/src/auth/jwt.strategy.ts
```diff
@@ -17,7 +17,7 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
     super({
       jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
       passReqToCallback: true,
-      ignoreExpiration: false,
+      ignoreExpiration: true,
```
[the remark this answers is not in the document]
pbn4: That was my dev setup that accidentially got into the PR

backend/core/src/applications/applications.service.ts
```diff
@@ -0,0 +1,10 @@
+import { Injectable } from "@nestjs/common"
+import { Application } from "../entity/application.entity"
+import { pickBy, identity } from "lodash"
+
+@Injectable()
+export class ApplicationsService {
+  async find(userId?: string, listingId?: string): Promise<Application[]> {
+    return await Application.find(pickBy({ userId, listingId }, identity))
```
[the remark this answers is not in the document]
pbn4: Done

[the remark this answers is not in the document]
pbn4: Done

backend/core/src/listings/listings.module.ts
pbn4: Done

backend/core/src/seeder/listings-seeder/listings-seeder.service.spec.ts
pbn4: Done

backend/core/src/seeder/listings-seeder/listings-seeder.service.ts
pbn4: That code I was just copying and wrapping into a service, but OK I'll change that.

pbn4: Done

backend/core/src/user-applications/user-applications.controller.ts
pbn4: It is not. It only adds metadata for swagger generation. See 
https://github.com/nestjs/swagger/blob/master/lib/decorators/api-bearer.decorator.ts and https://github.com/nestjs/swagger/blob/master/lib/decorators/api-security.decorator.ts

backend/core/src/auth/login.dto.ts
```diff
@@ -0,0 +1,14 @@
...
+export class LoginDto {
+  @IsEmail()
+  email: string
+
+  @IsString()
+  password: string
+}
+
+export class LoginResponseDto {
+  @IsEmail()
```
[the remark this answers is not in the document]
pbn4: The fact that it worked means it was not validated after all. I fixed that but the quesiton is whether we want to validate our responses.

backend/core/src/entity/application.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  @UpdateDateColumn()
+  updatedAt: Date
+
+  @ManyToOne((type) => User, (user) => user.applications)
+  user: User
+
+  @Column()
+  @IsString()
+  userId: string
```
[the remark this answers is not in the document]
pbn4: It allows flat response object structure (one could achieve this with querybuilder and without new column too, but much more complicated), otherwise a response would contain a user object:

```
{
  ...
  user: {
    id: id
  }
},
```
 instead of 

```
{
  ...
  userId: id
}
```

I'm eager to hear comments about this approach.

backend/core/src/listings/listings.module.ts
pbn4: Done

backend/core/src/seeder/listings-seeder/listings-seeder.service.spec.ts
pbn4: Removed.

backend/core/src/auth/owner.guard.ts
```diff
@@ -0,0 +1,9 @@
+import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common"
+
+@Injectable()
+export class OwnerGuard implements CanActivate {
+  canActivate(context: ExecutionContext): boolean {
+    const request = context.switchToHttp().getRequest()
+    return request.user.id === request.params["userId"]
```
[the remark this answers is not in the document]
pbn4: Good point, I tried. Here is my stackoverflow question https://stackoverflow.com/questions/62712757/how-to-create-explicit-dependencies-between-auth-guards-in-nestjs

backend/core/src/auth/roles.guard.ts
```diff
@@ -0,0 +1,25 @@
+import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common"
+import { Reflector } from "@nestjs/core"
+
+@Injectable()
+export class RolesGuard implements CanActivate {
```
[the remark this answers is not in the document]
pbn4: Answered that in question about OwnerGuard

backend/core/src/applications/applications.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+import { ApplicationsListQueryParams } from "./applications.dto"
+import { ApplicationsService } from "./applications.service"
+import { Application } from "../entity/application.entity"
+import { JwtAuthGuard } from "../auth/jwt.guard"
+import { ApiBearerAuth } from "@nestjs/swagger"
+import { RolesGuard } from "../auth/roles.guard"
+import { Roles } from "../auth/roles.decorator"
+
+@Controller("applications")
+@UseGuards(JwtAuthGuard)
```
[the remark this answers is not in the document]
pbn4: > > @jaredcwhite their purpose is described in #130 .
> > > One seems to be for public access of data,
> > 
> > 
> > What do you mean by public? One must have an admin role to access `/applications`. This is created specifically for superusers querying over applications table.
> > > Can we simplify the API endpoint down to one and add logic there to allow or deny requests based on if auth is required?
> > 
> > 
> > Answer to this question depends on the answer to the previous one.
> 
> I guess what I'm driving at is I would expect a single API endpoint (`/applications`) to deal with the data model "applications". What that endpoint is allowed do (aka when querying or saving data) depends on the request auth. Let me know if we should hop on a chat to discuss further.
> 
> **Update:** sorry, I just noticed your link to #130. Maybe we can discuss with Ben then…I'm not convinced admin routes vs user-specific nested routes makes sense, it seems to make the codebase much more complex.

@jaredcwhite I find API easier to reason about when the endpoint semantics (instead of a query param) tell you that these applications are in a relation to a specific user. I don't have any strong arguments for that and an `/applications` with query params and request auth handling would probably do the work too. I just doubt that it would be less complex than what I proposed. Maybe we can engage more people into the discussion since we are probably going to create few similar endpoints in the future @software-project @bk3c @bencpeters 

BTW: `/applications` is optional and I created it as a convenience. Everything related to applications that we need at the moment is doable using `/users/id/applications`, correct? Superuser interface is not really specified at the moment and I created /applications just as a proposition of how such access to arbitrary application could work.

pbn4:
> > @jaredcwhite their purpose is described in #130 .
> > > One seems to be for public access of data,
> > 
> > 
> > What do you mean by public? One must have an admin role to access `/applications`. This is created specifically for superusers querying over applications table.
> > > Can we simplify the API endpoint down to one and add logic there to allow or deny requests based on if auth is required?
> > 
> > 
> > Answer to this question depends on the answer to the previous one.
> 
> I guess what I'm driving at is I would expect a single API endpoint (`/applications`) to deal with the data model "applications". What that endpoint is allowed do (aka when querying or saving data) depends on the request auth. Let me know if we should hop on a chat to discuss further.
> 
> **Update:** sorry, I just noticed your link to #130. Maybe we can discuss with Ben then…I'm not convinced admin routes vs user-specific nested routes makes sense, it seems to make the codebase much more complex.

@jaredcwhite I find API easier to reason about when the endpoint semantics (instead of a query param) tell you that these applications are in a relation to a specific user. I don't have any strong arguments for that and an `/applications` with query params and request auth handling would probably do the work too. I just doubt that it would be less complex than what I proposed. Maybe we can engage more people into the discussion since we are probably going to create few similar endpoints in the future @software-project @bk3c @bencpeters 

BTW: `/applications` is optional and I created it as a convenience. Everything related to applications that we need at the moment is doable using `/users/id/applications`, correct? Superuser interface is not really specified at the moment and I created /applications just as a proposition of how such access to arbitrary application could work.

pbn4:
@bencpeters @jaredcwhite I addressed your comments, waiting for replies now. :) 

@bencpeters Weirdly some of the comments were visible here in conversation tab, but not all of them. Some were in 'files changes' tab, so if I missed something please let me know.

pbn4:
@bencpeters Could you elaborate on how we can make explicit dependencies between decorators?

backend/core/src/entity/application.entity.ts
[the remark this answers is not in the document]
pbn4: @bencpeters up

backend/core/src/applications/applications.controller.ts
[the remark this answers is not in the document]
pbn4: @bencpeters As discussed elsewhere this must be JWT because roles are required on user object. Do you agree?

pbn4:
@jaredcwhite I decided to go with what you originally proposed: a single controller and service, not a single service for two controllers. Mainly because my approach had to check path parameter userId against userId that was specified in request body for e.g. POST creating an application (so that I was sure a user if creating an application for himself) and since I was checking the auth data in the service anyway I realised this is not much different than what you proposed and the entire  `/user/:userID` prefix is unnecessary. 

Also, I removed the superuser route for the moment. There is an open issue for how we are going to handle permissions  https://github.com/bloom-housing/bloom/issues/315 so I'm going to bring that back once we have this resolved and actually need a route like that.

Issue #127, opened 2020-01-10
Set up basic nextjs shell and routes for Partners Reference app


jacchau:
- [x] Set up the basics of a nextjs app (presumably v9.4) in `apps/partners-reference`, matching the decisions that we made for the public reference app wherever appropriate, and documenting where things are different.

- [x] Have a home page for already-logged in users that just says "Welcome to the Partners Portal" or something similar, pending a more real design from @slowbot.

- [x] Users who haven't authenticated yet should get a login screen, using the form in #285 with modifications as needed.

- [ ] Show the currently logged in user in a top nav position, with an option to log out as a menu item below that (pending a more fancy design)

- [x] Basic Cypress testing for all of the above.

bk3c:
I'll mark this Ready for Dev when once we merge #285 and #316. 

cc: @jaredcwhite -- am I missing any other dependencies?

bk3c:
I'm calling this basically done from #500. We can improve the sign-in/out once #484 (or a successor) lands, and there's plenty to build on from there.

PR #358, opened 2020-06-22
Add basic Nextjs config for Partners

software-project:
@bencpeters here is an initial partners app for #127 
Should be ready for you to take over on authentication side.


bencpeters:
Alright, I just added a sign in page - for now it just signs in and redirects to "/" (we don't have any notifications/flashes set up at the moment to show success, correct?).

I did not add a registration page, do we want one on this app?

PR #364, opened 2020-06-26
More minor dep updates

bk3c:
Another couple weeks, another raft of dependencies to update. There **should** be no noticeable changes here.


jaredcwhite:
Looks good to me too. Quick visual review of the review app seems fine.

PR #365, opened 2020-06-29
Make WhatToExpect pick fields conditionally

pbn4:


shared/ui-components/src/page_components/listing/listing_sidebar/WhatToExpect.tsx
```diff
@@ -8,25 +8,33 @@ interface WhatToExpectProps {
...
+  allInfoWillBeVerified =
+    allInfoWillBeVerified || allInfoWillBeVerified == ""
+      ? allInfoWillBeVerified
+      : t("whatToExpect.allInfoWillBeVerified")
+
+  bePreparedIfChosen =
+    bePreparedIfChosen || bePreparedIfChosen == ""
+      ? bePreparedIfChosen
+      : t("whatToExpect.bePreparedIfChosen")
+
```
jaredcwhite, line 29: Can we simplify the conditionals here a bit? Maybe like (pseudocode):

```
if (!(value)) {
  value = t("whatToExpect.xyz")
}
```

pbn4: Hey the problem with this conditional is that I wanted empty strings "" defined in listing's s JSON to be the desired value in the template. Since "" is falsy then then 

```
if (!(value)) {
  value = t("whatToExpect.xyz")
}
```
evaluate to true and the default value will be used. One might think that leaving the prop undefined or null will work, but that requires adding a property in every other JSON definition. 

@jaredcwhite  I'm merging as is now, because of the deadline we have for launch, but if you are not happy with that I will fix that in a separate PR.

pbn4: BTW I just realised I can edit your comments. o_o

PR #379, opened 2020-06-30
Make backend/core public on NPM

bk3c:
Now that `ui-components` has an explicit dependency on `backend/core` to get the DTO objects, we need to make it public on the NPM repository in order for things to build without an implicit import via the monorepo. I guess the implicit build isn't the end of the world, but the package is small enough that I think we can make it public in the short term as we assess our strategy.


backend/core/.npmignore
```diff
@@ -0,0 +1,103 @@
...
+# Compiled binary addons (http://nodejs.org/api/addons.html)
+build/Release
+
+# Dependency directories
+node_modules/**
+services/listings/node_modules/**
+apps/public-reference/node_modules/**
+shared/core/node_modules/**
+shared/ui-components/node_modules/**
+jspm_packages/
```
bencpeters, line 67: This stuff probably doesn't need to be in this nested `.npmignore`?

backend/core/.npmignore
```diff
@@ -0,0 +1,103 @@
...
+*.storyshot
+
+# Cypress test output videos
+apps/**/cypress/videos
+apps/**/cypress/screenshots
+
+# Complied Typescript
+dist
+services/listings/dist/**
+apps/public-reference/.next
```
bencpeters, line 25: Not relevant?

backend/core/.npmignore
```diff
@@ -0,0 +1,103 @@
...
+# Storybook build output
+storybook-static
+
+# Storyshots and other tests
+__tests__/**
+*.storyshot
+
+# Cypress test output videos
+apps/**/cypress/videos
+apps/**/cypress/screenshots
```
bencpeters, line 20: Doesn't apply here?

PR #381, opened 2020-07-01
Fix upstream pieces missing from HBA PR#76

bk3c:
...in order to get Paper pickup information to render properly.


shared/core/src/listings.ts
```diff
@@ -24,6 +24,8 @@ export interface Listing {
   acceptingApplicationsByPoBox: boolean
   acceptingOnlineApplications: boolean
   acceptsPostmarkedApplications: boolean
+  applicationPickUpAddress: Address
+  applicationPickUpAddressOfficeHours: string
```
jaredcwhite, line 28: Should these new fields be optional?

bk3c: Good catch.

PR #384, opened 2020-07-01
TypeORM shim to use entities in Next.js frontend

jaredcwhite:
(Coming soon with DB backend, not yet activated)

This was part of a previous PR that got overwritten, but since we haven't yet fully switched to the DB backend, we wouldn't have wanted to roll it out without keeping it under wraps for the time being. So this PR adds the functionality for the frontend to potentially use the backend TS interfaces but doesn't yet activate it.

### DIFF
--- a/apps/public-reference/src/typeorm-model-shim.js
+++ b/apps/public-reference/src/typeorm-model-shim.js
@@ -1,0 +1,183 @@
+/* eslint-disable */
+
+// columns
+
+/* export */ function Column(typeOrOptions, options) {
+  return function (object, propertyName) {}
+}
+exports.Column = Column
+
+/* export */ function CreateDateColumn(options) {
+  return function (object, propertyName) {}
+}
+exports.CreateDateColumn = CreateDateColumn
+
+/* export */ function ObjectIdColumn(columnOptions) {
+  return function (object, propertyName) {}
+}
+exports.ObjectIdColumn = ObjectIdColumn
+
+/* export */ function PrimaryColumn(typeOrOptions, options) {
+  return function (object, propertyName) {}
+}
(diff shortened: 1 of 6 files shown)


apps/public-reference/src/typeorm-model-shim.js
```diff
@@ -0,0 +1,183 @@
+/* eslint-disable */
+
+// columns
```
bencpeters, line 3: I assume this file is essentially cut & pasted from [the shim provided by TypeORM](https://github.com/typeorm/typeorm/blob/master/extra/typeorm-model-shim.js). Do we want to host our own version of this, or should we literally use the shim provided by TypeORM? We could adjust the webpack config to look in `node_modules`:

```javascript
resolve: { // see: https://webpack.js.org/configuration/resolve/
  alias: {
    typeorm: path.resolve(__dirname, "../node_modules/typeorm/typeorm-model-shim")
  }
}
```

jaredcwhite: I had to make one modification, but it was something they actually should have included in that file, so theoretically if it got upstreamed we could pull it from the module dir. But for now we need to keep it in-repo.

bencpeters: @jaredcwhite Does [this PR](https://github.com/typeorm/typeorm/pull/6354) resolve the issue?

jaredcwhite: @bencpeters Unfortunately not — the shim is missing `BaseEntity` for some reason, I had to add it in.

bk3c:
Closing this in favor of swagger-axios-codegen and/or its successor.

PR #395, opened 2020-07-06
Fix the JS error on the error page

### LINKED ISSUE #345: JS Error on 404 pages
I'm not sure when this crept in…possibly with the Next.js upgrade…but 404 Not Found pages are producing a JS error. It's OK in production as you only notice it if you look at the developer console, but it's fairly annoying in development.

Error message:
`Unhandled Promise Rejection: TypeError: c.join is not a function. (In 'c.join("")', 'c.join' is undefined)`

bk3c:
Fixes #345, which was annoying me this morning.

I'm not sure if I'm missing some intended behavior with pageTitle being a JSX.Element here, but it was the unhappiness causing the error, and it seems to work reasonably as a string.

### DIFF
--- a/apps/partners-reference/pages/_error.jsx
+++ b/apps/partners-reference/pages/_error.jsx
@@ -3,7 +3,7 @@ import Head from "next/head"
 import { Hero, MarkdownSection, t } from "@bloom-housing/ui-components"
 
 export default () => {
-  const pageTitle = <>{t("error.notFound.title")}</>
+  const pageTitle = t("error.notFound.title")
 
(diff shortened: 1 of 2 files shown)


bk3c:
Also note we're still getting a warning from https://github.com/vercel/next.js/blob/master/errors/custom-error-no-custom-404.md for not having a 404 page separately defined and ready for static optimization. Doesn't seem like the highest priority, but is there an easy way to make that without being a duplicate of what we've got, or is there otherwise a better fix for that?

PR #398, opened 2020-07-07
Add utility hook + component to require login to access a page

### LINKED ISSUE #387: [Partners] Catch unauthenticated users and redirect to login page
There's currently a login page for partners, but the routing logic doesn't enforce that you have to go there if you're not logged in. Other than the Privacy Policy and maybe some other tangential pages, everything should require the user to be authenticated before they see anything, and get redirected otherwise.

If it's more expedient to redirect literally 100% of pages to the login screen rather than make any exceptions, that's probably fine for the short-medium term.

bencpeters:
# Add utility hook + component to require login to access a page

This PR adds a utility component `<RequireLogin>` and hook (`useRequireLoggedInUser`) to help manage protected front end routes.

`<RequireLogin>` only renders its children if the user is signed in, or the path doesn't require login. Login is based on props using either a blacklist (`requireForRoutes`) or whitelist (`skipForRoutes`), both of which can be RegExp. If a user isn't logged in and the requested path matches the appropriate list, the user will be redirected to the `signInPath` prop. If no lists are provided, login is required for all paths other than the sign in path.

Closes #387

### DIFF
--- a/apps/partners-reference/pages/_app.jsx
+++ b/apps/partners-reference/pages/_app.jsx
@@ -1,7 +1,12 @@
 import React from "react"
 import App from "next/app"
 import "@bloom-housing/ui-components/styles/index.scss"
-import { addTranslation, ConfigProvider, UserProvider } from "@bloom-housing/ui-components"
+import {
+  addTranslation,
+  ConfigProvider,
+  UserProvider,
+  RequireLogin,
+} from "@bloom-housing/ui-components"
 
 class MyApp extends App {
   constructor(props) {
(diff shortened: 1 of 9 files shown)


jaredcwhite:
@bencpeters Looks good! I'm not sure what the hook would be used for…maybe you could add a brief example. Otherwise, LGTM

bencpeters:
@bk3c [added documentation](https://github.com/bloom-housing/bloom/blob/add-login-required-redirect/docs/Authentication.md#front-end-authenticationuser-management). Are we good to merge?

PR #402, opened 2020-07-08
Fix Waitlist and ApplicationStatus

### LINKED ISSUE #380: Implement all permutations of the availability/waitlist component on the sidebar
**Application Status:**
- Has Due Date: Application Due Date: Weekday, Month Day, Year @ 5PM
- Has Open Date: Applications Open: Weekday, Month Day, Year @ 5PM
-  FCFS: First Come First Serve

**Waitlist:**
- Has Available Units no Waitlist:
  - **Do not display**
- If first come first serve
  -  **Do not display**
- Has Available Units and Waitlist: 
   - Available Units and Waitlist (header)
   - Once ranked applicants fill all available units, the remaining ranked applicants will be placed on a waitlist for those same units. 
   - “#” Available Units (Semi bold)
   - “#” Open Waitlist Slots (Semi bold)
- Has Only Waitlist:
   - Waitlist is Open (header)
   - Submit an application for an open slot on the waitlist.
   - “#” Current Waitlist Size
   - “#” Open Waitlist Slots (Semi bold)
   - “#” Final Waitlist Size

software-project:
Fixes #380 
SMC listings changes for Waitlist and ApplicationStatus


software-project:
@slowbot @bk3c this one should be ready for a review. One outstanding question that comes to my mind is if we should change ApplicationStatus when there are not available units and waitlist spots are finished?

shared/ui-components/src/page_components/listing/listing_sidebar/ApplicationSection.tsx
```diff
@@ -3,12 +3,19 @@ import moment from "moment"
...
 import Waitlist from "@bloom-housing/ui-components/src/page_components/listing/listing_sidebar/Waitlist"
-import { openDateState } from "../../../helpers/state"
 
 export interface ApplicationSectionProps {
   listing: Listing
 }
 
+// Hide waitlist for FCFS
+const hideWaitlist = (listing: Listing) => {
+  const applicationsClosed =
```
bk3c, line 13: Ok, I think this should actually be applicationsOpen and then you can take out the ! when you evaluate it, since that will be more forgiving of null values for the waitlist data elements on the listing. The current behavior is that if waitlistMaxSize and waitlistCurrentSize are both null, it shows the box open, but my understanding (unless corrected by @slowbot) is that we shouldn't show the waitlist component in that case since there isn't one.

I guess we could also modify this to explicitly check for nulls, which is fine too but I think might be tougher to read and understand later.

bk3c:
Current review app: https://5f0cf0c3e45bb4b57d420ffb--clever-edison-cd22c1.netlify.app/listings

software-project:
> Current review app: https://5f0cf0c3e45bb4b57d420ffb--clever-edison-cd22c1.netlify.app/listings

thanks @bk3c I'm working on your RR right now

PR #409, opened 2020-07-10
Add listings status

### LINKED ISSUE #279: Preview Listing Status
As we start to have more listings in the pipeline, there's going to be a lot more listing data QA to do with housing developers and jurisdictions. Rather than send folks to a separate staging site, it would be useful to designate listings in "preview" status (or maybe "unapproved'? I'm open to the best naming) at which point they're only visible to logged in users with appropriate permissions, but completely hidden from the general public.

We'll definitely need to have this working before we allow listings to be created and edited from the Partners app, since we'll presumably have those being stored in the same production database while the editing and approvals are in process. (I suppose we could also have Partners talk to a separate staging DB, but that doesn't sound better to me.) That said, I think there's a benefit to implementing the basics here sooner, since this would be useful functionality today, at least as soon as the dependancies (e.g. login) were ready.

Key pieces I foresee:

- [ ] deciding how we want to store listing status (enum?) and adding the field to all of our current listings
- [ ] filtering by status within the listings service -- we definitely don't want to take the security risk of doing it on the frontend
- [ ] implementing a security model of who can see a listing in preview mode. We could probably start as simply as a single global bit per site, and then get more sophisticated as we build out the permissions model for Partners

The first two parts have (I think) no dependencies, so we can knock them out whenever time becomes available.

pbn4:
Closes #279

I'm closing https://github.com/bloom-housing/bloom/pull/399 in favor of this since I got some merge conflicts with upstream there and it was easier to move it to another branch.


pbn4:
@bk3c I also made Listings controller work and have written appropriate tests for it. Including filtering for only those with status equal to "active".

pbn4:
Rebased to latest upstream.

backend/core/src/entity/listing.entity.ts
```diff
@@ -108,6 +113,11 @@ class Listing extends BaseEntity {
   urlSlug?: string
   @OneToMany((type) => Application, (application) => application.listing)
   applications: Application[]
+  @Column({
+    type: "enum",
+    enum: ListingStatus,
```
bencpeters, line 118: Should we add a default here at the database level to make creation easier?

backend/core/src/listings/listings.controller.ts
```diff
@@ -1,15 +1,13 @@
...
 export class ListingsController {
   constructor(private readonly listingsService: ListingsService) {}
 
   @Get()
-  public async getAll(
-    @Query() queryParams?: ListingsListQueryParams
-  ): Promise<ListingsListResponse> {
-    return await this.listingsService.list(queryParams)
+  public async getAll(@Query("jsonpath") jsonpath?: string): Promise<ListingsListResponse> {
+    return await this.listingsService.list(jsonpath)
```
bencpeters, line 11: `return await` is redundant - you only need `await` if you're using the result:

```typescript
const res = await ...
// synchronous code
return res;
```

backend/core/test/applications/applications.e2e-spec.ts
```diff
@@ -10,6 +10,7 @@ import { UserModule } from "../../src/user/user.module"
 import { ListingsModule } from "../../src/listings/listings.module"
 import { EntityNotFoundExceptionFilter } from "../../src/filters/entity-not-found-exception.filter"
```
bencpeters, line 11: I think the linter isn't running on this file - these imports aren't used now, they should be removed (and the linter should catch that...)

shared/ui-components/package.json
```diff
@@ -26,6 +26,7 @@
     "@types/enzyme-adapter-react-16": "^1.0.6",
     "@types/jest": "^25.2.2",
     "@types/jwt-decode": "^2.2.1",
+    "@types/react-input-mask": "^2.0.5",
```
bencpeters, line 29: Is this related to this PR? I didn't see any other changes to `ui-components`...

bk3c: No, but I think we merged it in from a different PR in the last couple days. It would be good to remove it here (and from the yarn.lock) when cleaning up the merge conflicts.

backend/core/src/entity/listing.entity.ts
pbn4: Sure.

pbn4:
> Looks good overall! A couple small suggestions, but no blockers.
> 
> 2 other questions/thoughts I had:
> Do we feel that we're still in a stage of development where it's easier to just update the initial migration than it is to add additional migrations? I feel like at some point it's easier to change models via additional migrations, rather than requiring everyone to reset their database when the initial migration changes...

Yeah I think we do not use DB backed service publicly anywhere so technically nothing changed. I'm favoring this solution because having many migration files slows down application startup in development mode. https://github.com/typeorm/typeorm/issues/4136 

> For some reason, I don't love "unapproved" as the status - maybe "pending" is a better word? If this is too pedantic, feel free to ignore, I don't have super strong feelings about it, it just doesn't sound right to me.

I changed it to "pending", sounds good to me.

shared/ui-components/package.json
pbn4: I removed it, although it is still being displayed in Github's PR diff view..

Issue #410, opened 2020-07-10
Commas in addresses aren't in the correct location


jaredcwhite:
Haha, I just noticed this. We're displaying a comma after the state abbreviation and before the zip code. Typically you'd see it before the state.

**Current:**
55 Triton Park Lane
Foster City CA, 94404

**Expected:**
55 Triton Park Lane
Foster City, CA 94404

(Interestingly, it's also incorrect on housing.sfgov.org's listing heroes)

slowbot:
@jaredcwhite Feel free to correct. Thanks for catching this

PR #411, opened 2020-07-10
5.2 Review Summary step and routing updates

### LINKED ISSUE #273: Application: Review
**Goal**: As an applicant I would like to review the details of my application before submission with the opportunity in order to make changes if needed

**Acceptance Criteria**:
- Applicant should be presented with a summary of their application details presented in editable sections
   - You
   - Household Members
   - Household Details
   - Income
   - Preferences
- Helper text below the summary should read: This is your last chance to edit before submitting.
- Options at the bottom of the page should be:
   - Confirm
   - Save and Come Back Later

image

jaredcwhite:
**7/27:** Latest RA: https://5f1f62f8451881028f60666d--clever-edison-cd22c1.netlify.app/applications/start/choose-language

---

Closes #273 and #410

## FYI: Updates to general form flow

I added a few methods to `ApplicationConductor` to handle routing in a way that allows a form submit to go to the next form URL _or_ the Review Summary.

There's also a `completeStep` method to help advance from through the steps shown in the progress component.

There's a new `conductor.canJumpForwardToReview()` conditional block at the bottom of some of the form pages to show the link to go back to Review if that's appropriate.


jaredcwhite:
I need to go through again and move labels into the language JSON as needed, but otherwise this should be pretty solid as far as PR review goes.

apps/public-reference/lib/ApplicationConductor.ts
```diff
@@ -41,8 +43,12 @@ export default class ApplicationConductor {
...
   }
 
-  advanceToNextStep() {
-    this.application.completedStep += 1
+  completeStep(step) {
+    this.application.completedStep = Math.max(step, this.application.completedStep)
+  }
+
+  canJumpForwardToReview() {
+    return this.application.completedStep == 4
```
pbn4, line 51: Can that become `return this.application.completedStep == totalNumberOfSteps() -1`?

apps/public-reference/pages/applications/contact/alternate-contact-type.tsx
```diff
@@ -20,11 +20,12 @@ export default () => {
...
   const onSubmit = (data) => {
     application.alternateContact.type = data.type
+    conductor.completeStep(1)
     conductor.sync()
     if (data.type == "noContact") {
-      Router.push("/applications/household/live-alone").then(() => window.scrollTo(0, 0))
+      conductor.routeTo("/applications/household/live-alone")
     } else {
-      Router.push("/applications/contact/alternate-contact-name").then(() => window.scrollTo(0, 0))
+      conductor.routeTo("/applications/contact/alternate-contact-name")
```
pbn4, line 28: Why not routeToNextOrReturnUrl here?

jaredcwhite: I think it's because there's not a return link at the bottom of this page…I started to add one but realized if somebody changes the value here, they'd have to go to the next page to complete the related information.

jaredcwhite:
> Maybe it's me being nitpicky but I'd create a data structure defining all possible steps in a single place so we could get rid of hardcoded step numbers and e.g. define possibility of going forward to review based on how many steps have been completed in relation to the length of this data structure.

That sounds good to me—let the conductor strictly handle actions but base the logic on a separate object defining the form flow. Do you want to write up something or pseudocode it, or should I give it try?

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -4,21 +4,78 @@ Display a summary of application fields with edit links per section
...
 import { useForm } from "react-hook-form"
 import { AppSubmissionContext } from "../../../lib/AppSubmissionContext"
 import ApplicationConductor from "../../../lib/ApplicationConductor"
 import { useContext } from "react"
 
+const EditLink = (props: { href: string }) => (
+  <div className="float-right">
+    <Link href={props.href}>
+      <a className="button is-unstyled is-borderless uppercase text-base my-2 text-primary-dark">
+        Edit
```
bencpeters, line 19: This text should probably be localized, no?

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -4,21 +4,78 @@ Display a summary of application fields with edit links per section
...
+      </a>
+    </Link>
+  </div>
+)
+
+const ReviewItem = (props: { label: string; sublabel?: string; children?: any }) => (
```
bencpeters, line 25: A better way to type functional components taking children:

```tsx
import type { FunctionComponent } from "react"
const ReviewItem: FunctionComponent<{ label: string; sublabel?: string }> = (props) => (
```

(This will automatically give you `children` with an appropriate type, rather than just using `any`, which will defeat any upstream typechecking. If you want to explicitly type: `children: JSX.Element` is appropriate I think...

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -4,21 +4,78 @@ Display a summary of application fields with edit links per section
...
+    {props.sublabel && (
+      <>
+        <br />
+        <span className="-mt-3 block text-base">{props.sublabel}</span>
+      </>
+    )}
+  </p>
+)
+
+const reformatAddress = (address: any) => {
```
bencpeters, line 45: Why isn't this typed? At the very least, we need it to be `object` or `Record<string, any>`, since we're pulling keys from it, but it seems like it'd be very easy to specify what this object looks like. If we just throw `any` all over the place, Typescript doesn't actually give us anything other than extra boilerplate...

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -4,21 +4,78 @@ Display a summary of application fields with edit links per section
...
+  const { street, street2, city, state, zipCode } = address
+  const newAddress = {
+    placeName: street,
+    street: street2,
+    city,
+    state,
+    zipCode,
+  } as Address
+  if (newAddress.street == null || newAddress.street == "") {
```
bencpeters, line 54: use `===` for comparisons.

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -39,29 +94,168 @@ export default () => {
...
-              filled={true}
-              onClick={() => {
-                //
-              }}
-            >
-              Next
-            </Button>
+        <div className="form-card__lead">
+          <h2 className="form-card__title is-borderless">
+            Take a moment to review your information before submitting your application.
```
bencpeters, line 105: This should be localized?

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -39,29 +94,168 @@ export default () => {
...
+
+        <h3 className="px-8 py-4 bg-gray-200">
+          You
+          <EditLink href="/applications/contact/name" />
+        </h3>
+
+        <div className="form-card__group mx-0">
+          <ReviewItem label={"Name"}>
+            {application.applicant.firstName} {application.applicant.middleName}{" "}
+            {application.applicant.lastName}
```
bencpeters, line 117: Are we guaranteed to have all 3 values here? (e.g. is `middleName` optional) If not, we'll have extra spaces with this, I think I'd probably do:

```tsx
{["firstName", "middleName", "lastName"].map(key => application.applicant[key]).filter(value => value && value.length > 0).join(" ")}
```

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -39,29 +94,168 @@ export default () => {
...
+          )}
+
+          {application.applicant.workInRegion == "yes" && (
+            <ReviewItem label={"Work Address"}>
+              <MultiLineAddress address={reformatAddress(application.applicant.workAddress)} />
+            </ReviewItem>
+          )}
+        </div>
+
+        {application.alternateContact.type != "" &&
```
bencpeters, line 158: Use `!==` (or better yet, `.length !== 0` for checking empty string)

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -39,29 +94,168 @@ export default () => {
...
+          application.alternateContact.type != "noContact" && (
+            <>
+              <h3 className="px-8 py-4 bg-gray-200">
+                Alternate Contact
+                <EditLink href="/applications/contact/alternate-contact-type" />
+              </h3>
+              <div className="form-card__group mx-0">
+                <ReviewItem
+                  label={t(
+                    "application.alternateContact.type.options." + application.alternateContact.type
```
bencpeters, line 168: stylistic, but I'd generally suggest using string interpolation:

```tsx
`application.alternateContact.type.options.${application.alternateContact.type}`
```

shared/ui-components/src/locales/general.json
```diff
@@ -2,14 +2,12 @@
...
-        "pleaseEnterZipCode": "Please enter your Zipcode",
         "selectOption": "Please select one of the options above.",
         "householdTooBig": "Your household size is too big.",
         "householdTooSmall": "Your household size is too small."
       },
-      "radio": {
-        "no": "No",
-        "yes": "Yes"
+      "general": {
+        "saveAndReturn": "Save & return to review"
```
bencpeters, line 10: Maybe "Save & Return to Review"?

jaredcwhite:
@bencpeters Definitely agree we can break out some of the form repetitions into components, as we've been discussing…I think that's be a separate issue/PR and warrant some consensus on the best way to do that. Also needs localization (on my todo list).

Regarding the stylistic conversation, I'll admit I have a pretty strong bent towards OOP methodologies wherever possible￼ (an ongoing frustration of mine with regards to recent trends in the JS world). I'm all for making code simpler and easier to reason about, but I feel like having a single object responsible for everything regarding navigation through all the forms is desirable.￼￼ Even more to the point, I prefer keeping logic out of React hooks as much as possible￼ when it comes to something not immediately pertaining to the UI visuals.

I think your concern about the tick￼/return state makes sense, but given that the state is only set within the same operation as form submission, it seems to do the trick.

bencpeters:
> Regarding the stylistic conversation, I'll admit I have a pretty strong bent towards OOP methodologies wherever possible￼ (an ongoing frustration of mine with regards to recent trends in the JS world). I'm all for making code simpler and easier to reason about, but I feel like having a single object responsible for everything regarding navigation through all the forms is desirable.￼￼ Even more to the point, I prefer keeping logic out of React hooks as much as possible￼ when it comes to something not immediately pertaining to the UI visuals.

Haha we'll have to agree to disagree then - I'm pretty strongly in the functional camp myself. But the "react-way" to do a single object responsible for all navigation through the forms would be to add a context. The issue with the conductor as its written right now is that it doesn't use any of React's built-in conventions for state - it's literally being recreated on every render. The solution right now works because `onSubmit` is synchronous - as soon as you add an async call in there pre-navigation (to persist to the backend, for example), it could break (it might still work because of javascript closure rules - in that case confusingly you'd technically actually have a redefinition of the onSubmit handler with the re-render (along with a new instance of the conductor class), but since it's not a singleton, the "old" instance is bound to the currently executing version of the onSubmit... but it feels very messy to me)

If you really want to keep this "master class" stateful paradigm, you should move the state to the `ApplicationContext`, which at least persists between re-renders. I think the "React Way" to write this would be to refactor the conductor class into an `ApplicationProvider` which provides `ApplicationContext` storing both the application data itself and any helper methods used for navigation. That's totally fine, but to me it feels like the choice of where to navigate next "belongs" to the currently displayed component - the component is the one giving the user a UI choice, so I'd personally advocate keeping the state local to the component.

jaredcwhite:
@bencpeters Thanks for the followup suggestions! I'll do some experimenting to see if I can account for the re-render/async-scenario or will otherwise have to migrate to context/state mechanics. I'll look more into the provider pattern as well.

shared/ui-components/src/locales/general.json
jaredcwhite: Fine by me, but I'll let @slowbot weigh in

apps/public-reference/pages/applications/review/summary.tsx
jaredcwhite: Hmm, that's pretty verbose and not a pattern we're using anywhere else, but I'll see if I can type `children` more accurately. (needs to handle either string or element(s))

apps/public-reference/pages/applications/review/summary.tsx
jaredcwhite: Ah, I forgot to mention that we have a slight mismatch between our addresses in applications and the addresses used for listing display from past work, so we'll need to align the two data structures. Probably should be done as part of the TS data modeling for applications.

apps/public-reference/pages/applications/review/summary.tsx
jaredcwhite: Good thinking…I believe HTML will swallow any extra whitespace so might not need exact accuracy in this context.

jaredcwhite:
Update: after reading up more on rendering lifecycles and hooks, I think `useMemo` is the way to go here…it ensures the `conductor` object is only initialized once and subsequent re-renders won't re-initialize it. I hadn't fully grasped it wasn't persisting in the first place…

shared/ui-components/src/locales/general.json
slowbot: No need for caps here since it link, plus the fact that copy should be written for more inclusive literacy levels: "Save and return to review" is preferred  @jaredcwhite

apps/public-reference/pages/applications/review/summary.tsx
bencpeters: Ya, that's the advantage of using the generic type `FunctionComponent` - it takes care of typing `children for you. It's not really much more verbose though:

```
const ReviewItem = (props: { label: string; sublabel?: string; children?: any }) => (
const ReviewItem: FunctionComponent<{ label: string; sublabel?: string }> = (props) => (
```

I _think_ you could use `children: JSX.Element` (or maybe `React.ReactNode`?), but probably worth testing to be sure.

apps/public-reference/pages/applications/review/summary.tsx
bencpeters: Ya, I guess that's probably true. Still feels wrong though ;-)

jaredcwhite:
All right…this is quickly becoming a beast of a PR, so with the latest changes to fix build errors, sync up with master, and localize the text labels, I'm hoping we can merge in soon and create issues as needed for any follow-up improvements.

jaredcwhite:
Latest RA: https://5f1f5c78525fc21f4158905e--clever-edison-cd22c1.netlify.app/applications/start/choose-language

jaredcwhite:
Done! Review RA: https://5f1f62f8451881028f60666d--clever-edison-cd22c1.netlify.app/applications/start/choose-language

bk3c:
FYI, the tests breaking are because of #457, which I think doesn't need to be a blocker for merging this since it's unrelated, but we should probably fix sooner rather than later.

bk3c:
Ok, I went ahead and reverted yarn.lock to the version that was on master, which is not the same as the output of running `yarn install` against what was already there. Some newer stuff had crept in somehow, and it probably wouldn't have hurt anything, but I don't want to get in the habit of changing dep versions unless it's on purpose. Running `yarn install` after the revert correctly shows `Already up-to-date`.

apps/public-reference/pages/applications/contact/address.tsx
```diff
@@ -19,14 +19,18 @@ import FormsLayout from "../../../layouts/forms"
...
 import { PhoneField } from "@bloom-housing/ui-components/src/forms/PhoneField"
 
 export default () => {
   const context = useContext(AppSubmissionContext)
   const { application, listing } = context
-  const conductor = new ApplicationConductor(application, listing, context)
+  const conductor = useMemo(() => new ApplicationConductor(application, listing, context), [
+    application,
+    listing,
+    context,
```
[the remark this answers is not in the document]
jaredcwhite: Thanks @bencpeters, I just wrote up issue #459 so we should look at addressing this in detail as we review the way we're very much not DRY-ing across all the form pages.

jaredcwhite:
@bencpeters Yeah, I think we want the validation to happen on the Save and Return…it's a little weird since it looks like a link not a button. Will go over that UX in more detail with @slowbot. Looks like @bk3c got the yarn.lock thing fixed. So, good to merge?

Issue #274, opened 2020-05-11
Application: Terms


slowbot:
**Goal**: As a developer or city I want to ensure that the applicant has agreed to terms of the housing opportunity 

**Acceptance Criteria**: 
- Applicant should be presented with a full explanation of the terms of the opportunity before submitting their application
- Applicant should be prompted to agree with the terms before moving forward
   - Label should read: I agree and understand that I cannot change anything after I submit.
- Options at the bottom of the page should read: 
   - Submit
   - Save and Come Back Later
   - If no option selected a presence validation reads: You must agree to the terms in order to continue

![image](https://user-images.githubusercontent.com/13915/81966543-88d9d400-95ce-11ea-8d89-e07b05bc151a.png)

bk3c:
@slowbot Can you copy the text of the terms (as best we understand it for now) into this issue as text instead of a graphic, so it's easy to just copy into the form?

@pbn4 No need to start on this until you have the backend endpoint on the other end to receive the information, but I figured that once we merged and deployed that, you'd be the best person to write the matching front-end code. Feel free to check-in with @jaredcwhite for his implementation thoughts, but I think it should be straight-forward reading from the sessionStorage he's set up and then just doing an axios call (I think through the UserContext from @bencpeters) to your endpoint with the JSON.

PR #413, opened 2020-07-13
Application terms

pbn4:
Closes https://github.com/bloom-housing/bloom/issues/274


pbn4:
@bencpeters  I am not yet sure how to connect it to user login, since the API requires access token which can be undefined. @jaredcwhite @bk3c  Is profile data and access token guaranteed to be present when submitting in normal conditions? In other words: is it required for user to be logged in to submit an application?

pbn4:
Ready for review. I added possibility to submit an application for unauthenticated users and some tests for that.

pbn4:
@bk3c I'm unable to make multiline strings in terms with carriage returns so I left parsing raw HTML enabled for now. Could you point me to how it's done?

pbn4:
@slowbot Thanks for taking time. 

There is one outstanding problem with rendering data from listing that I'm trying to solve with Jared now. Application due date is fixed at the moment.

shared/ui-components/src/applications/Client.d.ts
```diff
@@ -0,0 +1,457 @@
+import {
+  OpenAPIClient,
+  Parameters,
+  UnknownParamsObject,
+  OperationResponse,
+  AxiosRequestConfig,
+} from "openapi-client-axios"
+
+declare namespace Components {
```
bencpeters, line 9: Is this file auto-generated? If not, it seems like a bunch of duplicate definitions, there must be a better way to keep this in sync with `backend/core`?

pbn4:
> It's not totally clear to me how much of backend/core/client.ts is autogenerated - I made some style & clean-up comments but they obviously only apply to portions that are not auto generated. Let's get a little bit of documentation for developers around exactly when generating the client files (e.g. when yarn generate:client should be run), and what, if anything, needs to be done before running that command.

Command `yarn generate:client` fetches input swagger json from a running service. So you need to start it with `yarn dev` from `backend/core` and then run `yarn generate:client`. 

> It also seems like we're moving to having a bunch of duplicate type definitions here - there are the types in backend/core/client.ts, and the shared/ui-components/Client.d.ts definitions, in addition to the entities themselves and (presumably) the DTOs used to validate incoming data with the backend. If these additional definitions are totally auto-generated by a single command, then that's fine, but if there are steps required to keep them in sync, that seems less than ideal to me.

As discussed outside github:  `shared/ui-components/Client.d.ts` was a client generated by another tool that we are not going to use and should not appear in this PR, sorry for that. I removed it. :)  The only valid client is `backend/core/client.ts` and it's entirely autogenerated, so the duplication with DTOs is not a problem (actually they are a source of truth for that client because they are what NestJS/swagger takes to generate `/api-json` file).

PR #423, opened 2020-07-15
Update occupancy description to match most housingbayarea sites

bk3c:
This was customized for both Alameda and SMC, so I think we're better off just having the more standard text here in core. My understanding is that the "under 6" language is an SF thing, and it'll be easy enough for them to override the string in their own implementation.

(cc: @akegan, as perhaps the smallest possible customization to try next week)

### DIFF
--- a/shared/ui-components/src/locales/general.json
+++ b/shared/ui-components/src/locales/general.json
@@ -507,8 +507,8 @@
   },
   "whatToExpect": {
     "label": "What to Expect",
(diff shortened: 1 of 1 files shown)


bk3c:
@slowbot done.

bk3c:
Ok, I added the basic fixes from #424 to make the language neutral. Still leaving that issue open so we can get the specific cases handled in the next PR.

PR #429, opened 2020-07-17
Misc Dependency Updates

bk3c:
Another batch of what-should-be minor dependency updates.

Note that this includes a full refresh (delete + regenerate) of yarn.lock, since there was a minor security issue for lodash and that was the easiest way to clean out all of the 2nd and 3rd level deps that otherwise wouldn't get upgraded.


jaredcwhite:
Looks good to me. @bk3c mind fixing the lockfile so we can merge?

PR #430, opened 2020-07-17
Add Markdown option to InfoCard

### LINKED ISSUE #401: Update InfoCards to support Markdown
There have already been requests for InfoCards like credit history and rental history to support Markdown features like carriage returns and bulleted lists. Seems like it should be pretty straight-forward to get converted over?

Ideally maybe we should also have a Storybook story / tests that support.

jaredcwhite:
Fixes #401 

There's a second InfoCardGrid story variant showing the Markdown formatting.

### DIFF
--- a/shared/ui-components/package.json
+++ b/shared/ui-components/package.json
@@ -61,6 +61,7 @@
     "@types/node": "^12.12.47",
     "@types/node-polyglot": "^0.4.34",
     "@types/react-dom": "^16.9.5",
+    "@types/react-input-mask": "^2.0.5",
     "axios": "^0.19.2",
     "bulma": "^0.7.5",
     "jwt-decode": "^2.2.0",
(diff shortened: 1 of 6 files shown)


bk3c:
Is there ever a case where we wouldn't want markdown enabled? I'm thinking the normal text case should work exactly the same way if it's parsed as markdown, and we'll have to enable it at for all of the components anyway. Could we just have InfoCard either always be markdown-enabled, or at least default to having it enabled with no prop?

jaredcwhite:
So based on our convo this morning, I'll opt this into Markdown out-of-the-box and we can follow up on the housingbayarea repo to tweak the content there.

jaredcwhite:
@bk3c I ended up making this conditional on the prop data type (string = process as Markdown, otherwise just render as a standard `children` prop) because I realized we sometimes have other components embedded within an InfoCard (for example ExpandableText, which itself does Markdown processing).

I also fixed an issue with missing types for React Input Mask while I was looking at storyshots…but I also discovered that we're getting Storybook errors consistently now regarding an inclusion of the backend's User entity, so I'll file a separate issue for that.

PR #439, opened 2020-07-21
Fixes for partners login and add specs

software-project:
@bk3c as requested initial cypress tests for partners. I had some issues with login, an attribute was off and there was a race condition in reducers. @bencpeters helped me out with that one.


apps/partners-reference/cypress/integration/navigation.ts
```diff
@@ -0,0 +1,23 @@
...
+    cy.contains("Sign In")
+  })
+
+  it("Signs in", () => {
+    cy.visit("/")
+    cy.fixture("user").then((user) => {
+      cy.get("input#email").type(user.email)
+      cy.get("input#password").type(user.password)
+      cy.get(".button.is-filled").contains("Sign In").click()
+      cy.contains("This will be the home page")
```
bencpeters, line 14: Do we maybe want to check some other text on the home page that isn't likely to change as we build out the app? This basically guarantees that we'll have to update this test as soon as we modify the home page. Maybe check that the title is "Housing Portal", or something like that?

shared/ui-components/src/authentication/UserContext.tsx
```diff
@@ -140,11 +140,13 @@ export const UserProvider: FunctionComponent = ({ children }) => {
     accessToken: state.accessToken,
     initialStateLoaded: state.initialStateLoaded,
     login: async (email, password) => {
-      dispatch(signOut())
```
bencpeters, line 143: Unless it's causing problems, this should still be here I think. The goal is a "belt and suspenders" check to clear out any existing user information before we login the user.

apps/partners-reference/cypress/integration/navigation.ts
```diff
@@ -0,0 +1,23 @@
+describe("Navigating around the site", () => {
+  it("Loads the homepage directly", () => {
```
bencpeters, line 2: What this test is really checking is that our redirect works. Let's call it out explicitly: `it("Redirects to sign-in without a logged-in user"`

apps/partners-reference/cypress/integration/navigation.ts
```diff
@@ -0,0 +1,23 @@
...
+    cy.fixture("user").then((user) => {
+      cy.get("input#email").type(user.email)
+      cy.get("input#password").type(user.password)
+      cy.get(".button.is-filled").contains("Sign In").click()
+      cy.contains("This will be the home page")
+    })
+  })
+
+  it("Applications page", () => {
+    cy.visit("/")
```
bencpeters, line 19: This is going to fail if the user hasn't already logged in - it's dependent on state from the previous test (logging in) to succeed. More broadly speaking, since we require login for basically all pages on the partners app, a lot of the tests are going to be dependent on login. I think we should extract that into a helper function that can be run `beforeEach` in order to allow us to decouple tests/run them in isolation.

We can probably do this in 2 ways:

1) Extract the login logic in the previous test into a helper function.
2) Don't rely on the UI - make a request to the API programmatically with Cypress to get an access token, then write it to localStorage before the rest of the test runs. This is probably better fitting with Cypress best practices. I can post a commit of what I'm thinking...

bencpeters:
I've updated this branch with implementations of all of my suggested changes, but it'd be good to get one other set of eyes on it pre-merge.

shared/ui-components/src/authentication/api_requests.ts
```diff
@@ -10,7 +10,7 @@ export const renewToken = async (client: AxiosInstance) => {
 
 export const login = async (apiBase: string, email: string, password: string) => {
   const res = await axios.post<{ accessToken: string }>(`${apiBase}/auth/login`, {
-    username: email,
+    email: email,
```
bk3c, line 13: My only concern here is that I don't think we're ready to say that username == email address across the board, since a small but important part of our user population seems not to have email addresses. I realize that isn't what this PR is about, so we can handle it elsewhere if you think that's more appropriate, but I just wanted to make sure we weren't taking a step in the wrong direction.

(I do think it's 100% fine that the test username is an email address, since I think that's the normal case and we want to encourage it where appropriate. I just want to make sure we're not setting up a model where engineers assume that's the only option.)

bencpeters:
This PR isn't really taking a stand on it one way or another - it's just
matching the client code with what has been implemented in the controller
from other PRs.

Agreed that this issue probably needs more discussion, generally, but I
think this PR probably isn't the place for it.

shared/ui-components/src/authentication/api_requests.ts
bk3c: Ok, split this off to #461

PR #444, opened 2020-07-23
Fix minor errors with the storybook build

bk3c:
This removes a couple of unused imports that were causing build errors for Storybook, but then I just went ahead and disabled that error entirely in the tsconfig.json used for Storybook since it turns out there's one that's now autogenerated into `backend/core/client.ts` and that seemed like it might be too complicated to fix. We should watch and make sure we don't silently ignore typescript errors that we care about in the real apps.

This also has a yarn.lock update that seems unrelated since there are no package.json updates anywhere. I'm guessing there was a missing yarn.lock update in one of the commits we merged yesterday.

### DIFF
--- a/backend/core/src/entity/user.entity.ts
+++ b/backend/core/src/entity/user.entity.ts
@@ -8,8 +8,6 @@ import {
   OneToMany,
 } from "typeorm"
 import { Application } from "./application.entity"
-import { Exclude } from "class-transformer"
-import { ApiHideProperty } from "@nestjs/swagger"
 
 @Entity({ name: "user_accounts" })
 @Index("user_accounts_email_lower", { synchronize: false })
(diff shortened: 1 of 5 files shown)


bk3c:
@bencpeters I combed through the last few pull requests, and the package.json changes are in #442, but no matching yarn.lock update is in that PR so that's why yarn is catching up here. It also broke the `backend/core` build and deploy since there wasn't, but we didn't see it because we're not testing the new backend in CI yet. That should probably change asap.

I would also love it if we could get the warnings reduced enough to catch them at the linting stage (both in hooks and in CI). Just a matter of chipping away at the big pile we have now and then staying diligent about it. It looks like a huge chunk of them are because of explicit Any usage, so while it's more work in the short term we'll almost certainly end up with better code as we get all the types more explicit.

PR #464, opened 2020-07-30
Add idle timeout & automatic logout

### LINKED ISSUE #248: Browser Timeout

bencpeters:
# Add idle timeout & automatic logout

This detects an idle browser (based on keyboard & mouse events) and shows a prompt after 5 minutes. This prompt waits 1 further minute, then automatically logs the user out and redirects to home.

Adds a few utility components, including `Modal` (self-explanatory), and `UrlAlert` (which conditionally displays an `AlertBox` if a message is provided as a url param).

Question:
Right now, this only shows the idle timeout prompt if a user is logged in. Do we want it _also_ to apply to non logged in users? (e.g. an anonymous form application) In that case, we obviously couldn't log the user out, but we could clear `sessionStorage` and redirect back to home.

Screenshots:

image

image

Closes #248


bk3c:
I would definitely be in favor of this applying to non-logged in users too, ideally only once they've started an application so we know we have data to protect. If it's just someone browsing to a listing page then I don't think expiring the session makes any sense, and it would become annoying over time.

Also adding @slowbot to review the design and tweak the CSS if needed.

bencpeters:
> I would definitely be in favor of this applying to non-logged in users too, ideally only once they've started an application so we know we have data to protect. If it's just someone browsing to a listing page then I don't think expiring the session makes any sense, and it would become annoying over time.

Sounds good. I'll add that next week.

bencpeters:
Ok, updated with an idle timer for non logged in user on application forms only (redirects to homepage rather than sign-in, slightly different messages since there's no "sign out")

Changed the log out path to `/sign-in` (vs. `/`).

@slowbot to test:
Logged out user:
- Start an application
- Wait 5 minutes with no activity. A prompt should display
- Wait 1 minute, it should redirect to the homepage with a message

Logged in user:
- Log in with your account, then go anywhere on the site
- Wait 5 minutes with no activity, a prompt should display
- Wait 1 minute, then it should redirect to the sign in page with a message

Issue #453, opened 2020-07-27
Allow form pages to be easily configured off and on


bk3c:
The end goal is to have this configurable from the back-end, but after some discussion it seems like doing this on the front-end only first and then working back from there. Things that need to be configured include:

- [ ] which screens of the form are supposed to be used in which order 
- [ ] validation rules for various forms (e.g. the listing-specific thresholds for income / preference validation should be passed in from the config, and then the rules for evaluating them can be coded into each screen)
- [ ] conditional use of screens, if they're only supposed to be part of the flow based on other answers
- [ ] text configuration based on listing type (e.g. FCFS vs. Lottery, see also #424 and #278)

bk3c:
Just to document what's currently in my head, I think an ideal solution would have the configuration in a JSONable object that can easily be served from the backend in the future, and that had a format that was easy to expand as we add steps and options over time. For example, something like a `FormSequence` object with a structure like:

```js
{ steps: [
  {component: 'HouseholdMemberForm', props: {dataKey: 'HouseholdMember', title: 'Household Member'}},
  {component: 'IncomeForm', props: {dataKey: 'Income', title: 'Income', otherConfig: value}}
  ],
  metaField1: value,
  metaField2: value
}
```

The basic idea here is that the configuration just references an existing form component in the code base, and can pass in any props that are exposed but we want that to be pretty minimal and not try to represent the form itself in the config. As particular form steps are added or overridden, they can just be added to the array or the component name changed. Having the data key there is a way to be able to re-use the same component for multiple data elements, and to decouple the component from knowing too much about where the resulting data is stored (though I imagine it will still map the individual fields below the specified key). 

All of the details here are only meant to spark discussion -- I'm not sure that any of them are correct, and I welcome all ideas including completely different approaches.

bk3c:
cc: @james-willis-sf, since I know you had experience in this area and thoughts for how it might go. I expect we'll be starting on code for this soon, so would love to get any thoughts or feedback on your mind.

jaredcwhite:
@bk3c I think that's a good start. I'd probably suggest we add the relative URLs for each step so that we can anticipate where to go when clicking "Next" on a form, unless we want to derive the URLs from some other logic.

bk3c:
[...]

PR #467, opened 2020-07-30
Move initialization of the conductor object to React context

jaredcwhite:
Worked on this after some additional suggestions from @bencpeters — he gave this a thumbs-up pre-PR but I'm putting it out there for general feedback. It means that now conductor isn't re-initialized from form to form at all, which is fine as it's not supposed to be coupled to any individual form.


apps/public-reference/lib/ApplicationConductor.ts
```diff
@@ -62,13 +60,10 @@ export default class ApplicationConductor {
     }, 800)
   }
 
-  reset(shouldSync = true) {
+  reset() {
     this.application = blankApplication()
     this.listing = {} as Listing
-    if (shouldSync) {
-      this.context.syncApplication(this.application)
-      this.context.syncListing(this.listing)
```
bencpeters, line 70: Are these two functions used after this refactor? If not, we should remove them from the context.

apps/public-reference/pages/_app.jsx
```diff
@@ -5,20 +5,28 @@ import "@bloom-housing/ui-components/styles/index.scss"
...
     // Load autosaved listing application, if any
-    const autosavedApplication = loadApplicationFromAutosave()
+    const application = loadApplicationFromAutosave() || blankApplication()
     const savedListing = loadSavedListing()
-    this.state = { application: autosavedApplication || blankApplication(), listing: savedListing }
+    this.state = {
+      conductor: new ApplicationConductor(application, savedListing),
+      application: application,
+      listing: savedListing,
+    }
```
bencpeters, line 29: Do we ever envision a use case where a user might have multiple applications in progress at the same time? This logic here sets up `application`, `conductor`, and `listing` as global singletons without scoping by ID or anything like that.

Since our routing doesn't support unique application IDs yet, this might just be something to punt on until later, but it seems to me like keeping this state in an object that's more easily scoped (vs. the root app class) might make it easier to support that use case down the line?

apps/public-reference/pages/_app.jsx
```diff
@@ -98,6 +106,7 @@ class MyApp extends App {
     return (
       <AppSubmissionContext.Provider
         value={{
+          conductor: this.state.conductor,
```
bencpeters, line 109: I was just looking at the logic here for `syncApplication` and `syncListing` - it seems a bit problematic to me to have the "split state" between exposing those objects directly via the context, but then also having the conductor keep its own internal references to them. Right now if `syncApplication` or `syncListing` are used to change the global `application`/`listing` values to different objects, the conductor doesn't get updated - it will just keep the old values.

To me, it seems like the cleanest way to do this is to _only_ expose the conductor, and just let it be the single interface for dealing with these variables, but if you wanted to, you could also modify the `syncApplication` and `syncListing` functions to update the conductor as well as the state variables.

Worth noting that If you _do_ end up consolidating everything into the conductor, you'll basically have created a ContextProvider though - it'd be a quick & easy refactor at that point to update it so that rather than creating a context and giving it a reference to the conductor, you have the conductor be a React component that provides the context to the tree. I think this would result in something that's a bit a easier to scope, but it's optional ;-).

apps/public-reference/pages/applications/contact/address.tsx
```diff
@@ -24,12 +24,7 @@ import { PhoneField } from "@bloom-housing/ui-components/src/forms/PhoneField"
 
 export default () => {
   const context = useContext(AppSubmissionContext)
-  const { application, listing } = context
-  const conductor = useMemo(() => new ApplicationConductor(application, listing, context), [
-    application,
-    listing,
-    context,
-  ])
+  const { conductor, application, listing } = context
```
bencpeters, line 27: Much better!!!

I'd probably write it on a single line, rather than 2, but that's _purely_ stylistic:

```typescript
```

apps/public-reference/pages/_app.jsx
bk3c: I think we'd want to restrict users to having one application in progress for at least the short-medium term, more because I think the experience would be confusing to have multiples open at once, and we want to discourage half-completed applications. That said, I'm all in favor of scoping the variables underneath to give us more technical flexibility if we ever need it, so long as the overhead isn't too annoying.

apps/public-reference/lib/ApplicationConductor.ts
jaredcwhite: @bencpeters They should probably be renamed—they're only relevant when actually replacing the entire JSON object with a new object. I don't think `syncApplication` is used at all right now, and `syncListing` is used only once when loading a listing from the API.

apps/public-reference/pages/_app.jsx
jaredcwhite: @bencpeters Yeah, I thought about that too, but see my comment above…they should be used as close to never as possible since we're modifying the application date in-place as we go, and the listing is essentially read-only.

apps/public-reference/pages/_app.jsx
jaredcwhite: I was wondering about that as well when working on this update, but ending up with multiple applications getting autosaved and having to deal with all the mushrooming complexity of the data we're tracking across pages…oof, feels like a major headache.

bk3c: If it's a major headache, I'm fine with saying we're only going to allow one application at a time in-progress, so long as @slowbot is on board with that too.

apps/public-reference/pages/applications/contact/address.tsx
jaredcwhite: @bencpeters All right I checked in simplified syntax in all these cases. Build is failing though, but it looks like that's an issue on master so will investigate separately.

apps/public-reference/pages/_app.jsx
bencpeters: Cool, I’m totally fine punting on this then - we can tackle when/if it becomes relevant.

apps/public-reference/pages/_app.jsx
bencpeters: That makes sense, but to me the fact that they’re not even really used at present means we should either fix or remove them - right now it’s potentially dangerous code that isn’t really being tested, but could lead to inconsistent state if used down the line.

In this case “fixing” could be as easy as modifying the `sync` functions to also update `application`/`listing` on the conductor too...

jaredcwhite: Truth — for now I'd say let's put a pin in it…if we end up "merging" application and listing data models together in some fashion ala https://github.com/bloom-housing/bloom/issues/453, we might not even need to track both separately.

bk3c:
@bencpeters Do you feel like all of your feedback is sufficiently addressed for merge?

Issue #457, opened 2020-07-28
Don't show a Listing Map without a valid location


bk3c:
This is mostly just a code quality issue at this point since I can't think of a realistic situation where we'd have no location information at all, but since they tightened up the types for react-map-gl, we're now getting a type failure on jest tests because we're not checking for the undefined case. Our Address (correctly) allows those fields to be null, and the map component rightly complains if you try to put a marker without valid coordinates.

For bonus points, we can go ahead and implement #448 as we do this, but that's also fine to do in the more distant future so long as we gracefully handle the null case in the short term, which I assume means just not showing the map.

jaredcwhite:
Well that's weird…on the latest `review-form-step` branch, I can't get the "No overload matches this call" error with Marker to happen on my machine. Tests run fine. I wonder if there's some difference in env vars or something…

bk3c:
Try it now on master now that everything is merged? I can replicate it locally no problem.

PR #468, opened 2020-07-31
Feature/add nest command

dominikx96:
- Added useful commands to package.json to run frontend app & nest (a new backend)
- Added command to run migrations


package.json
```diff
@@ -28,6 +28,7 @@
     "dev:listings": "cd services/listings && yarn dev",
     "dev:backend": "cd backend/core && yarn dev",
     "dev:all": "concurrently \"yarn dev:listings\" \"yarn dev:app:public\"",
+    "dev:all:nest": "concurrently \"yarn dev:backend\" \"yarn dev:app:public\"",
```
bencpeters, line 31: I think we should probably update `dev:all` rather than adding a new command - my understanding is that we're phasing out the "old" JSON based backend anyway?

bk3c: Addressed through a more significant refactor.

bk3c:
Ok, I increased the scope of this PR by quite a bit, but hopefully in a useful direction. Thanks to @dominikx96 for starting off some overdue fixes. Now let's just get a few more eyes on it, and then we can merge.

bk3c:
Jest failure here is just due to #457, which we can fix separately.

README.md
```diff
@@ -46,11 +70,6 @@ yarn test:all
 ### Versioning
 
 We are using [lerna](https://lerna.js.org/) as a package versioning tool. It helps with keeping multiple package versions in sync.
-To upgrade version, first commit your changes to feature branch and push it to github. Then run
-```
-yarn new-version
```
jaredcwhite, line 51: We're removing this…why?

bk3c: Because I've been just incrementing the version with `lerna publish` when I do the publish, and I think that's easier to manage than having individual engineers try to bump versions and then keep it all in sync. I'm planning to do some maintainer-oriented documentation in another PR, but in the meantime I just wanted to remove a suggestion I didn't want anyone to follow in the short-term.

jaredcwhite: OK, got it.

PR #469, opened 2020-07-31
Updating dropdown styles

slowbot:
Updating styles for dropdown.

I'm assuming this is a Bulma component so I'm not sure the best way to customize. @jaredcwhite Let me know if I'm missing anything or if we should use the bulma config for some if these styles.

### DIFF
--- a/shared/ui-components/src/headers/SiteHeader/SiteHeader.stories.tsx
+++ b/shared/ui-components/src/headers/SiteHeader/SiteHeader.stories.tsx
@@ -36,7 +36,6 @@ export const withDropdownAndButton = () => (
       <a href="#" className="navbar-item">
         Get Assistance
       </a>
-      <hr className="navbar-divider" />
       <a href="#" className="navbar-item">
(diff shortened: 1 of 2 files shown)


jaredcwhite:
Eh, this is probably fine. Bulma's individual dropdown link styles are not as easily customizable via variables as I would like anyway.

bk3c:
@jaredcwhite It looks like Bulma is up to 0.9 now, and it's in the set of packages that I don't update by default because I'm not sure I'd catch any subtle breaking changes. Do you think this would be a good time to upgrade it while we're paying attention on the QA front?

bk3c:
After some quick conversation, we'll go ahead and merge this then do the bulma upgrade later.

jaredcwhite:
@bk3c Sorry I didn't grasp your question earlier when we talked…yes we should upgrade Bulma at some point just to stay current but I don't think there's anything recently added that affects the particular component we're using. So fine to punt.

Issue #276, opened 2020-05-11
Application: View Submitted Application


slowbot:
**Goal**: As an applicant I want to see a summary of the application that I submitted along with a confirmation number so that I have a backup copy for my records.

**Acceptance Criteria**: 
- An applicant should be presented with a summary of the application details they submitted for a specific opportunity similar to the application summary they are shown before submitting the application. Sections include:
   - You
   - Household Members
   - Household Details
   - Income
   - Preferences
- Applicant should be given their confirmation number as well that is associated with this application                     

![image](https://user-images.githubusercontent.com/13915/81966835-eec65b80-95ce-11ea-9f77-031ca27ca8ac.png)

Issue #275, opened 2020-05-11
Application: Confirmation


slowbot:
**Goal**: As an applicant I want a confirmation that my application has been submitted as well as a confirmation number so I can refer to my application in the future.

**Acceptance Criteria**: 
- Applicant should be presented with a screen confirming the submission of their application
- Applicant should be given a confirmation number unique to their application
- Applicant should be given instructions on what do with their confirmation number: 
   - _Please write down your lottery number and keep it in a safe place. We have also emailed this number to you if you provided an email address._
- Applicant should be emailed a copy of their confirmation number
- Applicant should be given a summary of what happens next in the application process
- Applicant should be given the contact information for the Leasing Agent for this listing
- Applicant should be prompted to Create a User Account
- Options at the bottom of the page should read:
   - Create an Account (Primary)
   - No thanks, I'm done
   - Browse more listings
   - View submitted application and print a copy
   
![image](https://user-images.githubusercontent.com/13915/81966686-b292fb00-95ce-11ea-9c63-5677bcf2a00f.png)

PR #472, opened 2020-08-03
Feature/confirmation

dominikx96:
I'm not sure what should be used as a "key" in FormSummaryDetails.tsx in maps:
- application.householdMembers
- accessibilityLabels(application.accessibility)

Maybe should we add unique IDs on the step when the user adding this information?

I also moved summary fragment to -> FormSummaryDetails.tsx, let me know if it's a good place for this kind of this component.


jaredcwhite:
@dominikx96 Looks good! and nice work on extracting the summary information out into a reusable component.

FYI, I'd like to get #467 merged in before rebasing and merging this… also on the question regarding item keys, maybe @bencpeters can speak to that as he's been working on adding keys in other places.

jaredcwhite:
@dominikx96 FYI, #467 got merged in

bencpeters:
> I'm not sure what should be used as a "key" in FormSummaryDetails.tsx in maps:
> 
>     * application.householdMembers
> 
>     * accessibilityLabels(application.accessibility)
> 
> 
> Maybe should we add unique IDs on the step when the user adding this information?

For purposes of React `map`, keys are basically just used to help React build out its virtual DOM and keep track of each element in an array - it's not super important what they are, they just need to be unique.

For `householdMembers`, I'd probably just use `key=`${member.firstName} ${member.lastName}` - that should be unique unless the user makes a repeat entry.

For `accessibilityLabels`, the key should just be the `item` itself - the text should be unique. You'll have to use `<Fragment key={item}>`, since the `<>` shorthand doesn't support `key` props...

dominikx96:
@slowbot so we should keep this component, but for now, #275 will be used in this step as default, right?

bk3c:
@dominikx96 Correct - #275 is the screen that shows in response to a successful submission at the end of the flow. #276 is the screen with the details of the application that a user can log back in to see at a later time. 

Not part of this pair of issues, but the full flow is that once we have the functionality to send an email confirmation, that will include a link back to the site where the user can log in and then see the #276 screen.

dominikx96:
@bk3c I updated this branch to be up to date with master. I also moved confirmation.tsx content to the new file: view-application.tsx.

A new confirmation.tsx is available in PR: #512 , so I think will be good to merge this first and next #512 .

dominikx96:
@jaredcwhite I merged master into this branch and resolved conflicts. I think the flow is correct ...terms -> summary -> confirmation -> (optional view-application after clicking on "View submitted application and print a copy").

apps/public-reference/pages/applications/review/summary.tsx
```diff
@@ -101,144 +48,7 @@ export default () => {
...
-            {application.incomeVouchers ? t("t.yes") : t("t.no")}
-          </ReviewItem>
-
-          <ReviewItem label={t("t.income")}>
-            ${application.income} {t(`application.financial.income.${application.incomePeriod}`)}
-          </ReviewItem>
-        </div>
+        <FormSummaryDetails application={application} editMode />
 
         <h3 className="form--card__sub-header">
```
jaredcwhite, line 53: The new Preferences section should probably get moved into the FormSummaryDetails component as well.

apps/public-reference/pages/applications/review/view-application.tsx
```diff
@@ -0,0 +1,81 @@
+/*
```
jaredcwhite, line 1: Let's move this to something like /applications/view — I think we'll want to use it for referring to applications after time has passed, so it's not really part of the review flow.

apps/public-reference/pages/applications/review/summary.tsx
dominikx96: Absolutely yes! I think I missed this element after the merge.

apps/public-reference/pages/applications/review/view-application.tsx
dominikx96: Sounds great, I changed the URL to /applications/view.

jaredcwhite:
I think this is ready to merge now…can we get the QA sign off from @slowbot?

PR #480, opened 2020-08-05
Feature/demographics

dominikx96:
I created generic component CheckboxGroup to do not repeat the same code or hardcode e.g labels (I moved all keys to FormOptions, translations to general json).


shared/ui-components/src/forms/CheckboxGroup.tsx
```diff
@@ -0,0 +1,40 @@
...
+                {item.label}
+              </label>
+            </div>
+          ))}
+      </div>
+    </div>
+  )
+}
+
+interface CheckboxSingle {
```
jaredcwhite, line 29: Just an FYI: our convention typically has been to put interface declarations above components in a file

shared/ui-components/src/forms/Select.tsx
```diff
@@ -0,0 +1,62 @@
...
+          )}
+          <FormOptions options={options} keyPrefix={keyPrefix} />
+        </select>
+      </div>
+      {error && errorMessage && <ErrorMessage error={error}>{errorMessage}</ErrorMessage>}
+    </div>
+  )
+}
+
+interface Props {
```
jaredcwhite, line 47: For this and the checkbox group, it'd be great to add a component name prefix to the interface (aka `SelectProps`)

shared/ui-components/styles/forms.scss
```diff
@@ -8,7 +8,7 @@
   label,
   .label {
     @apply text-gray-750;
-    font-size: 0.95rem;
+    font-size: .875rem;
```
jaredcwhite, line 11: I hesitate to sign off on a font size change to all labels at this point. Can you check with @slowbot to confirm this is what we want?

shared/ui-components/src/helpers/formOptions.tsx
```diff
@@ -5,6 +5,12 @@ interface FormOptionsProps {
...
 }
+
+interface CheckboxGroupItem {
+  name: string
+  checked?: boolean
+}
+
 export const contactPreferencesKeys = ["email", "phone", "letter", "text"]
 
 export const relotionshipKeys = [
```
jaredcwhite, line 16: I know this isn't your change, but while you're in here can you fix this typo? (relotionship)

apps/public-reference/pages/applications/review/demographics.tsx
```diff
@@ -42,29 +76,97 @@ export default () => {
       </FormCard>
 
       <FormCard>
-        <p className="text-bold">
+        <p className="form-card__back">
           <strong>
-            <Link href="/applications/preferences/general">
+            <Link href="/applications/contact/name">
```
jaredcwhite, line 81: I think this is the wrong URL for linking back…

shared/ui-components/styles/forms.scss
slowbot: I think this looks good and is more accurate to the original product.

apps/public-reference/pages/applications/review/demographics.tsx
dominikx96: oh, you're right, I'll fix it! I'm also thinking about these hardcoded links. Maybe will be better to create something like "routes" constant & helper function to detect prev/next component (not necessary, but helpful). What do you think?

dominikx96:
@jaredcwhite all suggestions implemented :) Please look if the back URL is correct now.

apps/public-reference/pages/applications/review/demographics.tsx
jaredcwhite: @dominikx96 Yeah, we have a project underway to consolidate all the form routes into a single config file, see #453 (and feedback is most welcome)

dominikx96:
@jaredcwhite FYI: I made small cleanup & added optional validation (if is required) for the checkbox group.

bk3c:
@dominikx96 Can you resolve this merge conflict? I think we're good to merge as soon as you do.

dominikx96:
@bk3c conflicts resolved! :)

Issue #367, opened 2020-06-29
Decide method for linking applications back to the relevant listing


bk3c:
We should be able to start very small here, since there will be a lot of need and room to expand over time as the application logic gets more customized to each listing (e.g. with preferences).

At a minimum, we should make sure we capture the listing ID and name when an user starts an application, so that we can display the active listing name to the user as they fill the form out, and we can submit the resulting ID to the backend with the application itself.

bk3c:
As @bencpeters pointed out, we need to not just link to the listing but we also need to potentially look at unit information to do things like validating that applicants match income requirements (which we're storing on a per-unit basis)

bk3c:
See also #518. Once we have that sorted out, I think the only other work we need to do here is to make Application a real typed object on the client side and assign the relevant listing object to it, then make sure that still works through submittal.

bk3c:
I think #460 covers the second part of the work here, so closing this issue in favor of that and #518.

PR #482, opened 2020-08-06
Link Common Application to a listing and fix submittal

bk3c:
The slightly more complicated part of this PR is changing the `swagger-axios-codegen` configuration to not generate static methods, so we avoid the build error that @jaredcwhite identified in `review/terms.tsx`. Static is probably the better way to go long term for a backend service proxy, but I didn't want to add another import linking the front-end to the back-end, so I didn't go that way for now. It doesn't seem like instantiating the service object will happen often as currently done so I'm not expecting memory or performance problems, but speak up if there are Bad Implications that I'm missing.

Now that I've dug into it a little more, I think the right thing to do for the medium-long term is to move the `swagger-axios-codegen` stuff over to the front-end (either ui-components, or maybe this is time to refactor some of the more logic-centric pieces to a different shared frontend package). It's already set up to to the generation from a live URL, and this can become part of the client build in a pretty straight-forward way. It should also remove the need for any imports from front-end of back-end pieces, and it will make the swagger.json definitions on the backend the definitive source for the API, which seems good from a documentation perspective. If people agree with this approach, I'll file a separate issue for it, and then we can decide if can wait until @pbn4 is back or we need to try and do it faster. Also if this all seems to work, I think we can finally abandon #384, which I'm assuming @jaredcwhite will not be sad about.

The uncomplicated part of this PR changes the test versions of the Triton listing in both the old and new services to link to the common application. This should allow folks to demo and qa end-to-end in the various dev environments without having to manually paste the URL into the browser's location bar. Needless to say the actual implementation here will undoubtedly change once #453 and #367 are sorted out, but this should be an improvement in the interim.


bk3c:
PS: to test, you should be able to run the whole way through the application, and then see the resulting JSON in your local database from psql or equivalent. Needless to say that having that work seems like a pretty big win, and necessary to unblock  all the other issues about viewing/exporting the resulting applications.

bk3c:
Note: this PR now includes a huge number of Storyshots changes from when I ran `test -u` for the small updates to the test data originally intended. As far as I can tell, it's just deleting the ones with the old `|` naming convention that got replaced with slashes in the SB6 upgrade. I think it's a good thing and fine to go ahead with, but would appreciate a second pair of eyes on that part too.

bk3c:
Ok, I'm going to go ahead an merge this so we don't hit a lot of conflicts, but would still love feedback on the longer term client codegen plan.

Issue #344, opened 2020-06-15
Accounts: My Applications


slowbot:
**Goal**: As a user I want to be able to save my application and finish them later as well as update my account informations.

**Acceptance Criteria**:
- A signed in user should be bale to navigate to their account via the main nav
   - When logged out this link reads "Sign In"
   - When logged in this link reads "My Account"
- "My Account" link should be a drop down that reveals a dropdown menu:
   - My Dashboard
   - My Applications
   - Account Settings
   - Sign Out
- "My Dashboard" view has two cards that link to
   - My Applications
   - Account Settings
   - Utilize component prototype Dash Card
- "My Applications" view should provide a list of all the applications this applicant has saved or finished
   - Each application should be represented by an application card:
      - Name of Listing
      - Application Deadline
      - Building Address
      - Application Status: In Progress
      - Edited Date
      - Continue Application (link)
         - Picks from the step that the applicant left off on
      - See Listing (link)
         - Takes applicant to listing 
      - Delete (link)
         - Deleting application will trigger a confirmation modal
   - If application is submitted:
      - Application Status: Submitted
      - View Application (link instead of Continue Application)
      - No delete action
   - If no applications then show a blank state with link to "Browse Rentals"
- "Account Settings" should provide the ability to update account information. See https://github.com/bloom-housing/bloom/issues/245

**My Account Dropdown**
![image](https://user-images.githubusercontent.com/13915/84707628-e639ab80-af13-11ea-8339-777fe9c267be.png)

**My Dashboard**
![image](https://user-images.githubusercontent.com/13915/84706106-5135b300-af11-11ea-8b3e-df6e17e4db00.png)

**My Application Blank State**
![image](https://user-images.githubusercontent.com/13915/84706686-43346200-af12-11ea-94bc-1a996b40dc09.png)

**Application Card: In Progress**
![image](https://user-images.githubusercontent.com/13915/84706809-7b3ba500-af12-11ea-9b04-3be047ad5bdb.png)

**Delete Confirmation**
![image](https://user-images.githubusercontent.com/13915/84706997-c950a880-af12-11ea-9141-37e095e95aa5.png)

**Application Card: Submitted**
![image](https://user-images.githubusercontent.com/13915/84707329-5c89de00-af13-11ea-998d-4bac125128a6.png)

Issue #475, opened 2020-08-04
Improve frontend protection from null or malformed JWTs


bk3c:
While doing QA on #464, I got into a situation where a valid JWT wasn't returned from the backend because I was running the old service instead of the new one, which caused a NextJS error and required a manual cookie wipe to resolve. It seems like the browser being unable to connect to the backend or otherwise getting an unexpected error is a case that we should handle a little better.

kathyccheng:
@jaredcwhite can you see if this is still an issue?

jaredcwhite:
@kathyccheng Sure, I'll take a look.

jaredcwhite:
Looks like we need to catch any `InvalidTokenError` thrown and deal with it accordingly. I'll work on a fix.

kathyccheng:
Ongoing tech debt

Issue #450, opened 2020-07-24
Cleanup: swapping JSX.Element out for ReactNode


jaredcwhite:
We have a lot of interfaces with stuff like `JSX.Element` all the way to `string | JSX.Element | JSX.Element[]`

As I recently learned, it'd be great to standardize everything around `ReactNode` (or `React.ReactNode` depending on how you import it), since that will cover all those use cases (and is more readable!).

bk3c:
It looks like there are 14 files with jsx.element, so I'm hoping it's not too huge of a lift. It's also nice to feel like we're getting a little more mature in our type usage.

Happy to have this happen whenever you need a break from more brain-intensive work.

bk3c:
One other note here: there's a school of thought that the better way to include `props.children` is by using a type of `React.FC` for the components, so that's probably an option for some of these replacements too. See [this blog post](https://medium.com/better-programming/the-right-way-to-create-function-components-in-react-with-typescript-59dcf4076b92) for the argument.

@jaredcwhite Do you think it would be worth writing up a quick list of Bloom-preferred coding patterns and practices for the rest of the team to provide feedback on?

jaredcwhite:
@bk3c I'll admit I'm not a huge fan of the added verbosity. Some docs I've seen recommend including the children props explicitly even when using the FC typing, so not sure it's a win.

On a general note though, I definitely think an overall coding style guide would be helpful. I'm also curious if we could draw on other open source guides of note and simply tweak as needed.

PR #484, opened 2020-08-06
Basic user account

software-project:
Fixes: #344 
@slowbot I think this one is ready for some love from your side.
@jaredcwhite @bk3c I figured out the shim thing. Michał moved some interfaces into a clint so shim is not needed after all. But methods from `ApiClientContext` are throwing a static declaration error. I think Jared mentioned it before. So for now this PR is not using API.


bk3c:
@software-project See #482 for a fix to the static declaration problem, and feel free to test/review that so we can get it merged and you can use the results. Also would love your input on the longer-term plan I'm proposing there.

bk3c:
@software-project Can you rebase this now that #482 is merged in? It should remove the static declaration problem and also take away all the Storyshots changes that were included there so it's easier to review.

bk3c:
@slowbot I'm pretty sure your error is caused by not having the new backend/core service running. Once this is rebased then `dev:all` should do the right thing. See also #475, which is meant to protect the user with a nicer error message in this case.

bk3c:
I appreciate that master got merged in here, but is there any way to either rebase or otherwise cleanup the diff? It's currently showing 79 files changed including a bunch of other things already merged, which just makes it a bear to review.

software-project:
@bk3c this one should be finally ready. I skipped API communication side, but updated Modal with @bencpeters work and also addressed his PR review notes.

apps/public-reference/pages/account/applications.tsx
```diff
@@ -0,0 +1,97 @@
...
+  DashBlock,
+  DashBlocks,
+  HeaderBadge,
+  LinkButton,
+  MetaTags,
+  Modal,
+  t,
+} from "@bloom-housing/ui-components"
+import Layout from "../../layouts/application"
+import Archer from "@bloom-housing/listings-service/listings/archer.json"
```
jaredcwhite, line 15: This is probably a holdover from past testing that should be removed now?

apps/public-reference/pages/account/applications.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export default () => {
+  const { applicationsService } = useContext(ApiClientContext)
+  const [applications, setApplications] = useState([])
+  const [deletingApplication, setDeletingApplication] = useState(null)
+
+  useEffect(() => {
+    // applicationsService.list().then((apps) => {
+    //   setApplications(apps)
+    // })
```
jaredcwhite, line 27: Same question as above…my understanding is that we can pull live data from the DB now instead of test data?

shared/ui-components/src/forms/ProgressNav.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react"
-import { onClientSide } from "../helpers/nextjs"
+import { OnClientSide } from "../helpers/nextjs"
```
jaredcwhite, line 2: Caps looks odd to me and doesn't match other hook/function names…was there a reason to change it?

shared/ui-components/src/page_components/AppStatusItem.tsx
```diff
@@ -0,0 +1,108 @@
...
+          </div>
+
+          <div className="status-item__action">
+            <p className="status-item__status">
+              <span className={"status-item__label " + statusClassNames}>
+                {t("application.status")}: {statusText}
+              </span>
+            </p>
+            {status == "submitted" && (
+              <a href="#" className="button small">
```
jaredcwhite, line 64: These links don't go anywhere at present?

shared/ui-components/src/page_components/DashBlock.tsx
```diff
@@ -0,0 +1,46 @@
...
+  subtitle?: string
+  icon: JSX.Element
+  children?: JSX.Element | JSX.Element[]
+}
+const DashBlock = (props: DashBlockProps) => {
+  const { href, title, subtitle, icon, children } = props
+  let content, wrapper
+  if (children) {
+    content = (
+      <div className="dash-item__content" style={{ background: "#f6f6f6" }}>
```
jaredcwhite, line 16: It'd be good to use the Tailwind theme classes here.

apps/public-reference/pages/account/applications.tsx
software-project: As we talked on the standup, @bk3c wanted to merge this one sooner than later, so I prepared a separate issue for that.

shared/ui-components/src/forms/ProgressNav.tsx
software-project: There is a lining error for that function, thats why I renamed it to capital case.

shared/ui-components/src/page_components/AppStatusItem.tsx
software-project: Believe we don't have a application page yet?

jaredcwhite: Ah, yeah I think #472 would need to get merged in first

apps/public-reference/pages/account/applications.tsx
jaredcwhite: Found it: #532. Sounds good.

shared/ui-components/src/forms/ProgressNav.tsx
bk3c: Yeah, this is what spurred me to file #509, which I believe will be the real fix and hopefully supersede this function, but in the meantime I think it's better that we just get linting to pass. I'm also fine if you think readability would be better to keep it as-is and then just turn off the eslint rule in this case until we fix it?

shared/ui-components/src/page_components/AppStatusItem.tsx
bk3c: merged!

shared/ui-components/src/page_components/HeaderBadge.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react"
+import HeaderBadge from "./HeaderBadge"
+import { withA11y } from "@storybook/addon-a11y"
```
bk3c, line 3: This is built in as of Storybook 6, so the import and decorator should be dropped.

shared/ui-components/src/page_components/AppStatusItem.tsx
software-project: hmmm, I think application view page should work a bit different if we want to use it for showing applications that are already submitted. I think applications should be loaded from API, not via AppSubmissionContext context. What do you think @bk3c @jaredcwhite ? Should I maybe add it to #532 ?

bk3c: I think we should load it from a context, so that we can take advantage of the authentication already being built in with that pattern, but I agree that we should either rename AppSubmissionContext, use ApiClientContext (probably my leading contender), or create a new one.

Adding it to #532 makes sense to me so we can get this merged.

shared/ui-components/src/page_components/UserNav.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from "react"
+import "./UserNav.scss"
```
jaredcwhite, line 2: I don't see a connection between what's in the CSS getting imported and this component?

shared/ui-components/src/page_components/DashBlock.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react"
+import "./DashBlocks.scss"
+
+interface DashBlockProps {
+  href?: string
+  title: string
+  subtitle?: string
+  icon: JSX.Element
+  children?: JSX.Element | JSX.Element[]
```
jaredcwhite, line 9: Let's avoid the use of `JSX.Element` per https://github.com/bloom-housing/bloom/issues/450

shared/ui-components/src/page_components/DashBlocks.tsx
```diff
@@ -0,0 +1,7 @@
+import React from "react"
+import "./DashBlocks.scss"
+
+const DashBlocks = (props: { children: JSX.Element | JSX.Element[] }) => (
```
jaredcwhite, line 4: Let's avoid the use of `JSX.Element` per https://github.com/bloom-housing/bloom/issues/450

shared/ui-components/src/page_components/UserNav.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { LocalizedLink } from "../atoms/LocalizedLink"
+import { t } from "../helpers/translator"
+import { NavbarDropdown } from "../headers/SiteHeader/SiteHeader"
+
+export interface UserNavProps {
+  signedIn: boolean
+  children: JSX.Element | JSX.Element[]
```
jaredcwhite, line 9: Let's avoid the use of `JSX.Element` per https://github.com/bloom-housing/bloom/issues/450

shared/ui-components/src/page_components/UserNav.tsx
software-project: Good catch 👍

bk3c:
Ok, merging per Slack conversation and we'll handle outstanding items in other issues.

PR #488, opened 2020-08-07
Fix minor issues with Cypress for partners-reference

### LINKED ISSUE #137: Yarn peerDep warnings related to ""workspace-aggregator..."
This has been happening for a while and we've been ignoring it without a bigger problem, but I just wanted to document the relevant Yarn issues that are causing the warnings for tracking, especially since it looks like they might be fixed soon. Check out:

https://github.com/yarnpkg/yarn/issues/5810
https://github.com/yarnpkg/yarn/pull/7812

bk3c:
- Fixes the port numbers in the partners-reference Cypress API tests to match the new numbering scheme in #468. 
- Makes explicit the peerDependencies for the Cypress webpack processor, which is the last thing for fixes #137. 
- Bumped the babel and cypress versions globally while I was at it.

### DIFF
--- a/apps/partners-reference/cypress/support/index.ts
+++ b/apps/partners-reference/cypress/support/index.ts
@@ -54,7 +54,7 @@ declare global {
 
 Cypress.Commands.add(
   "login",
-  (email: string, password: string, { apiBase = "http://localhost:3001" } = {}) => {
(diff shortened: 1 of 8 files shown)


bk3c:
I'm going to go ahead and merge this now that the port numbers have been officially changed in #468.

Issue #271, opened 2020-05-11
Application: General Pool


slowbot:
**Goal**: As an applicant I want to clarity about my rank if I have not claimed any preferences

**Acceptance Criteria**:
- If an applicant does not claim any preferences on an application for a housing opportunity then they should be notified that they are in the General Pool

![image](https://user-images.githubusercontent.com/13915/81748148-2eba0100-945e-11ea-9183-c18df128df3c.png)

PR #489, opened 2020-08-07
Add the Preferences form step

### LINKED ISSUE #270: Application: Preference
**Goal**: An an applicant I want to be able to claim a preference in order to receive a higher ranking

**Acceptance Criteria**:
- Applicant should be prompted to claim a preference
   - Preference should have a checkbox that allows you to opt in to the preference
   - More information about the preference should be available via show/hide and offsite link
   - Helper text should read: If you have one of these preferences, select it below
- Applicant should be able to opt out of a preference
   - Label should read: I don't want this lottery preference
   - Helper text should read: You'll still have the opportunity to claim other preferences.
- If no option is selected 
   - Inline validation should read: Please select one of the options above.
   - Page level validation should read: 
      - **Please select and complete one of the options below in order to continue.** We strongly encourage you to claim the preference before submitting to have a better chance in the lottery.

**MVP Option 1**
image

**MVP Option 2**
image

**Post MVP**
image

jaredcwhite:
Closes #270 and #271 

I have a question for @slowbot about the way we determine if either live or work preferences can be applied at all. Also I wasn't sure how to style/present the "read more" info for each preference.

Review app: https://5f2cb6f052ea8127b083e910--clever-edison-cd22c1.netlify.app/


apps/public-reference/pages/applications/household/add-members.tsx
```diff
@@ -73,25 +73,27 @@ export default () => {
...
-                householdSize={application.householdSize}
-                validate={true}
-                register={register}
-                error={errors.householdSize}
-                clearErrors={clearErrors}
-                assistanceUrl={t("application.household.assistanceUrl")}
-              />
-            </div>
+        <form onSubmit={handleSubmit(onSubmit)}>
+          <div className={errors.householdSize ? "" : ""}>
```
bencpeters, line 77: Is this intentional? As written, we're not adding classes either way, should just be `<div>`...

apps/public-reference/pages/applications/preferences/general.tsx
```diff
@@ -9,28 +9,30 @@ import FormsLayout from "../../../layouts/forms"
...
 export default () => {
   const { conductor, application, listing } = useContext(AppSubmissionContext)
   const currentPageStep = 4
 
+  useEffect(() => {
+    if (!application.preferences.none) {
+      // Only describe the General Pool if no preferences were selected
+      Router.push("/applications/review/demographics")
+    }
+  }, [application])
```
bencpeters, line 23: This dependency should _probably_ be `application.preferences` or `application.preferences.none` - we don't want this effect to run only when `application` changes?

apps/public-reference/pages/applications/household/add-members.tsx
```diff
@@ -73,25 +73,27 @@ export default () => {
...
+            <HouseholdSizeField
+              listing={listing}
+              householdSize={application.householdSize}
+              validate={true}
```
bencpeters, line 81: Totally pedantic, but `true` is assumed - you can just write `<HouseholdSizeField validate ...` (don't need to explicitly specify `validate={true}`)

apps/public-reference/pages/applications/preferences/select.tsx
```diff
@@ -0,0 +1,189 @@
...
+*/
+import Link from "next/link"
+import Router from "next/router"
+import { Button, ErrorMessage, FormCard, ProgressNav, t } from "@bloom-housing/ui-components"
+import FormsLayout from "../../../layouts/forms"
+import { useForm } from "react-hook-form"
+import { AppSubmissionContext } from "../../../lib/AppSubmissionContext"
+import ApplicationConductor from "../../../lib/ApplicationConductor"
+import FormStep from "../../../src/forms/applications/FormStep"
+import { useContext, useMemo, useState } from "react"
```
bencpeters, line 13: More pedantic changes, but "best practice" (and what we've been doing in other files) is to organize imports with third party imports first, then doing "local" imports to our packages.

apps/public-reference/pages/applications/preferences/select.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+        <form className="" onSubmit={handleSubmit(onSubmit)}>
+          <div className="form-card__group px-0 pb-3">
+            <p className="field-note">{t("application.preferences.selectBelow")}</p>
+          </div>
+
+          {preferenceOptions.map((option) => {
+            const [showMore, setShowMore] = useState(false)
+            return (
+              <div className="form-card__group px-0 border-b">
```
bencpeters, line 67: This needs a `key` property - anytime you use `map` in JSX, you _need_ to provide a `key` (you almost certainly have console warnings about this if you look).

apps/public-reference/pages/applications/preferences/select.tsx
```diff
@@ -0,0 +1,189 @@
...
+          <p className="field-note mt-5">{t("application.preferences.preamble")}</p>
+        </div>
+
+        <form className="" onSubmit={handleSubmit(onSubmit)}>
+          <div className="form-card__group px-0 pb-3">
+            <p className="field-note">{t("application.preferences.selectBelow")}</p>
+          </div>
+
+          {preferenceOptions.map((option) => {
+            const [showMore, setShowMore] = useState(false)
```
bencpeters, line 65: Using a hook inside the render is generally considered to be a mistake in React. One of the rules of hooks is that the number of hooks that you render needs to be invariant every time the component renders (e.g. can't change based on conditional logic), so you want to avoid burying hooks deep in the render logic where someone else might come along later and add conditional rendering - it's much more clear to have them at the top of your component function. This is doubly true for using them within an iteration - in this case `preferenceOptions` is a fixed length array, which is why it works, but if it was ever made variable (e.g. we loaded options asynchronously) this logic could cause problems. 

To fix this, you should either move the state up to the top of the function:

```typescript
const [showMore, setShowMore] = useState({})
...
// Read from the state for a specific option
showMore[option]

const toggleShowMoreForOption = (option) => setShowMore({ ...showMore, [option]: !showMore[option] })
```

Or refactor the iteration render into a subcomponent that can keep track of its own `showMore` state internally and take `option` as a prop.

apps/public-reference/pages/applications/preferences/select.tsx
```diff
@@ -0,0 +1,189 @@
...
+                  <button
+                    type="button"
+                    className="button is-unstyled m-0 no-underline has-toggle"
+                    aria-expanded={showMore ? "true" : "false"}
+                    onClick={() => {
+                      setShowMore(!showMore)
+                    }}
+                  >
+                    {!showMore && t("label.readMore")}
+                    {showMore && t("label.readLess")}
```
bencpeters, line 97:
```suggestion
                    {t(showMore ? "label.readLess" : "label.readMore")}
```

apps/public-reference/pages/applications/preferences/select.tsx
```diff
@@ -0,0 +1,189 @@
...
+                defaultChecked={application.preferences.none}
+                ref={register({
+                  validate: {
+                    somethingIsChecked: (value) => {
+                      let returnValue = value
+                      preferenceOptions.forEach((option) => {
+                        if (getValues(option)) returnValue = true
+                      })
+                      return returnValue
+                    },
```
bencpeters, line 133:
```suggestion
                    somethingIsChecked: (value) => preferenceOptions.some((options) => getValues(option)),
```

apps/public-reference/pages/applications/household/add-members.tsx
jaredcwhite: @bencpeters again, this is a mystery to me as well!

apps/public-reference/pages/applications/household/add-members.tsx
jaredcwhite: @bencpeters Huh, I don't know how changes to this page ended up in this PR. Maybe it's something Jesse worked on. I will double-check.

jaredcwhite:
Regarding your question about a schema for preferences, I'll let @bk3c weigh in on that as well… I don't think how we're doing it here is going to last much longer than the MVP…

apps/public-reference/pages/applications/preferences/select.tsx
jaredcwhite: @bencpeters I'll move it out to a subcomponent, as it does need to maintain separate state for each item.

apps/public-reference/pages/applications/preferences/general.tsx
jaredcwhite: I'm not sure TBH…ideally this URL never comes up in the flow if it's configured right assuming a preference was selected, so this is simply an escape hatch.

jaredcwhite:
@slowbot @bencpeters All right, I've made some final tweaks to the PR addressing most of the feedback. For adding the Preferences stuff to the Review page, let's tackle that in its own issue.

Ben P, I moved the checkboxes `map` code to its own subcomponent, which is fine…except passing all the different form hooks/functions feels rather messy to me. Not sure if there's a better approach but let me know if you have any suggestions.

RA: https://5f32b7a7b137dbc89f2ae28f--clever-edison-cd22c1.netlify.app

bencpeters:
> Ben P, I moved the checkboxes `map` code to its own subcomponent, which is fine…except passing all the different form hooks/functions feels rather messy to me. Not sure if there's a better approach but let me know if you have any suggestions.

Ya, I can definitely see the argument against passing all those props around, no disagreement that it's messy. If you look at my original comment I posted some suggested code for how to do it without creating a subcomponent (and that solution still has separate state for each individual checkbox)

jaredcwhite:
@bencpeters Ah…the alternative you'd posted didn't click with me the first time. Making the top-level state an object to hold the multiple toggles is clever, I'll try that approach!

jaredcwhite:
@bencpeters Yep, that works well.

Issue #474, opened 2020-08-04
Fix jest config for backend/core


bk3c:
I think we're getting close enough to using backend/core in production that we should make sure we can add it to CI asap, but I'm currently seeing a lot of failures when I run `yarn test:backend:core`.

I think the immediate problems at least mainly due to jest configuration errors, so hopefully that should be easy to clean up? [This example repo](https://github.com/cypress-io/cypress-and-jest-typescript-example) seems to have some configs that might be useful as a template.

At the very least, we probably want to remove the jest configuration from package.json since there are already a jest and jest-e2e JSON file, or maybe the other way around if that's the Nest preferred method?

bk3c:
Fixed in #493

PR #493, opened 2020-08-07
fix backend/core test failures

bk3c:
This is a downpayment on #474. It looks like it's not just the jest config, but the actual tests need help (or I'm using them so incorrectly that they need much better documentation).

Based on the error messages that I was getting, I first updated all the relevant tests to make sure we were using jest's expect and not the one from cypress/chai, which seems to be a common monorepo problem.

Once I got past that, things got more interesting with the entities not reflecting metadata correctly and the app not starting up in many cases. I tried to update the metadata to include the actual database column types, which seemed to help but then created a new set of problems. @bencpeters or @pbn4, do you know how this was working right before? Is it something that broke with the upgrade to jest26 or a newer nest version?

### DIFF
--- a/backend/core/jest.json
+++ b/backend/core/jest.json
@@ -1,8 +1,8 @@
 {
   "moduleFileExtensions": ["js", "json", "ts"],
-  "rootDir": "..",
+  "rootDir": ".",
   "testEnvironment": "node",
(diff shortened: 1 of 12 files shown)


bk3c:
@bencpeters 🏅

PR #495, opened 2020-08-07
Initial email notifications setup

### LINKED ISSUE #132: Build v0 of a Notifications Service to support sending Email Notifications
Initial functionality to allow:

- [ ] Sending an email notification. The calling service (e.g. the Application Service to call when a new application is submitted) should be able to queue up an email, and then it should be sent as soon as reasonable.
- [ ] ~~We need to catch the case where the email API isn't available for any reason, and make sure we can try again later.~~ Split off to #498
- [ ] The content of the message should also be passed in at send time, though maybe in some future version we could consider templates?

software-project:
Fixes #132


software-project:
@bencpeters @bk3c could you help resolve the issue with backend/core. Seems like I'm doing sth wrong with imports. When I'm starting backend there is a missing module error for `@nestjs/common`. I tried using different sendit package but got the same error, so looks like I did sth wrong.

bk3c:
For better or worse, I'm getting the exact same problem, and it's not obvious what the underlying cause is. My best guess at the moment is that it's because nest-sendgrid is being hoisted and @nestjs/common isn't, but looking at the code in node_modules it looks right (no hard-coded relative path or anything), so I don't understand why that would actually cause a problem :/

bk3c:
@software-project I don't love it as a solution, but putting nest-sendgrid on the nohoist list seems to fix the immediate problem.

software-project:
@bk3c working finally. Handlebar templates are working. Added nicer loading of templates. Translations are also working for templates. And finally added a test.

software-project:
@bk3c I've resolved first issue and it turned out views were not loading for dist folder. I was testing with IDE and it was loading differently. Please check if you have two email related env vars set correctly. For me test are working fine and emails are sending.

bk3c:
Views issue looks all resolved to me, but I'm still seeing the from: address as undefined when running `yarn test` only. If I do `yarn dev:all` having made no other changes, emails send great.

PR #504, opened 2020-08-11
Update to storybook 6 official + other minor dep updates

bk3c:
This moves us off the Storybook 6 release candidate now that they're official, plus a bunch of other minor dependency updates while I was at it.


jaredcwhite:
@bk3c Storybook and Jest tests work for me. LGTM :+1:

PR #506, opened 2020-08-11
Add a bunch of additional documentation

bk3c:
The included documentation basically fits into two categories:

1. DevOps-y stuff, which is mostly just catching up with what I think we're already doing, but needed some doc help

2. Object model, which is very much a work in progress but I wanted to get started with some form of documentation rather than waiting until we had the whole things figured out.

I'm inclined to merge this sooner rather than later, even though part 2 isn't all that correct, but happy for any feedback that folks may have on both content and format.


jaredcwhite:
@bk3c I like it. Merge early and often is my motto when it comes to docs. :)

PR #519, opened 2020-08-12
Add netlify.toml for partners and global bump node

bk3c:


bk3c:
Global node bump looks solid on both the front and back end. Merging to make it easier to debug any TOML issues specific to partners.

Issue #76, opened 2019-10-20
Finalize mobile header / navbar visual design


bk3c:
Follow up from #61, there are still some visual differences from the current production sites, so I just want to make sure they're all on purpose, or otherwise use this issue to get where we need to be. @slowbot, feel free to just add the "ready for development" tag whenever you're ready to hand back off to engineering.

bk3c:
Current:

![image](https://user-images.githubusercontent.com/975317/67162684-3a57a600-f31b-11e9-99ee-9c44761098c6.png)

bk3c:
Previous:

![image](https://user-images.githubusercontent.com/975317/67162699-66732700-f31b-11e9-948d-9deb8b40a982.png)

bk3c:
@slowbot going through some old issues, and I wonder if you can weigh in on what the right answer is here? As far as I can tell, the current behavior is still as described in the screen shots above -- if that's what we want, I'm happy to just close this, or we can just write up whatever changes need to be made...

slowbot:
@bk3c 
1. Prioritize removing the feedback bar at the top of the screen
2. Prioritize adding "menu" to the mobile menu icon
3. Lower priority, styling the logo on mobile

PR #521, opened 2020-08-13
Styled mobile navbar

dominikx96:
Styled mobile navbar to look like on the old design.

### DIFF
--- a/shared/ui-components/styles/navbar.scss
+++ b/shared/ui-components/styles/navbar.scss
@@ -1,5 +1,4 @@
 @import "utilities/setup_bulma";
-@import "~bulma/sass/utilities/mixins.sass";
 @import "~bulma/sass/components/navbar.sass";
 
 .navbar-notice {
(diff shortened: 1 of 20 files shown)


shared/ui-components/styles/navbar.scss
```diff
@@ -3,10 +3,15 @@
 @import "~bulma/sass/components/navbar.sass";
 
 .navbar-notice {
+  @apply hidden;
```
jaredcwhite, line 6: @slowbot Just checking…did we decide that the feedback notice should be hidden on mobile now?

dominikx96: @jaredcwhite I think yes, @slowbot mentioned about this in #76 . @slowbot could you confirm?

bk3c: I re-read #76, and I agree with this assessment.

PR #522, opened 2020-08-13
Added basic version of redirect

dominikx96:
I created a basic version of the redirect. I'm not sure about a couple of things so please leave your feedback :)

To achieve 2 things:
- redirect the user to the latest path
- show success modal

I had to save a path before the change. We cannot use history.go(-1), because it is impossible to attach extra params to the URL.

So, I created property in the UserContext to store the last path, and additional method to change localization with query param including user name (to show it in the modal).

Also, the login success modal seems to work only on the homepage.

### DIFF
--- a/apps/public-reference/pages/sign-in.tsx
+++ b/apps/public-reference/pages/sign-in.tsx
@@ -12,11 +12,11 @@ import {
   UrlAlert,
 } from "@bloom-housing/ui-components"
 import FormsLayout from "../layouts/forms"
-import { useRouter } from "next/router"
+import { useRedirectToPrevPage } from "../lib/hooks"
 
 const SignIn = () => {
-  const { login, redirectToPrevPage } = useContext(UserContext)
-  const router = useRouter()
+  const { login } = useContext(UserContext)
+  const redirectToPrev = useRedirectToPrevPage()
   /* Form Handler */
   const { register, handleSubmit, errors } = useForm()
   const [requestError, setRequestError] = useState<string>()
@@ -26,15 +26,14 @@ const SignIn = () => {
 
     try {
       const user = await login(email, password)
-      redirectToPrevPage()
-      // router.back()
-      // router.push(
-      //   `?success=${encodeURIComponent(
-      //     t(`authentication.signIn.success`, { name: user.firstName })
-      //   )}`
-      // )
+
+      redirectToPrev({
+        success: `${encodeURIComponent(
+          t(`authentication.signIn.success`, { name: user.firstName })
+        )}`,
+      })
     } catch (err) {
-      const { status } = err.response
+      const { status } = err.response || {}
       if (status === 401) {
         setRequestError(`${t("authentication.signIn.error")}: ${err.message}`)
       } else {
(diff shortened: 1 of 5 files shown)


apps/public-reference/pages/sign-in.tsx
```diff
@@ -26,11 +26,13 @@ const SignIn = () => {
...
-          t(`authentication.signIn.success`, { name: user.firstName })
-        )}`
-      )
+      redirectToPrevPage()
+      // router.back()
+      // router.push(
+      //   `?success=${encodeURIComponent(
+      //     t(`authentication.signIn.success`, { name: user.firstName })
+      //   )}`
+      // )
```
bencpeters, line 35: Let's make sure we don't leave these 👣 around!

dominikx96: Thanks for your feedback!

The hook solution makes sense, but I'm not sure about using referrer. I think it could not work in a few cases. e.g: the user will move around the page, (nextjs will serve pre-rendered page for the first request, next it will be a normal js SPA) so referrer will point for a bad path because of wont change during browse in a SPA mode.

If my thinking are correct - we should prepare a separated place to store prev path.

What do you think?

bencpeters: Ya, that's true. I had initially though it would be updated by Next, but I just tested it and it's not being updated.

I actually think a better way to handle this might be to include a `redirectUrl` as a url param on the sign in page. I can imagine some times when it might be useful to go to a new page (rather than just the last page) on sign-up, and that gives us a bit more fine-grained control over it. Then the redirect hook can just look in the current url params (which you can look up on `router.query`) and redirect based on that.

dominikx96: If we're ok with additional param - it's best solution 👍 
Updated hook & added to create account page.

@bk3c I did not update anchors to sign-in and create an account page yet, because I remember all hardcoded paths will be moved to separated file. Is it available now on master or work is in progress?

bk3c: @dominikx96 I think both of those pages will keep the same paths, which will continue to be (more or less) hard coded. The place where we're talking about configurable paths is within the application form itself. 

There is a slightly weird scenario where a user wants to sign in while they're halfway through filling out the form, but I think it should still work. We should know their current state and URL (even if dynamic in some way) at that time, and that can get passed to the redirect so they return to wherever they were.

bencpeters: > There is a slightly weird scenario where a user wants to sign in while they're halfway through filling out the form, but I think it should still work. We should know their current state and URL (even if dynamic in some way) at that time, and that can get passed to the redirect so they return to wherever they were.

I don't think this should be a problem with a query param based redirect. The actual form state lives in session storage/a root level context, so we won't lose that unless the user closes the window. The form stages all have a unique URL anyway, so getting redirected back to the appropriate point is easy.

dominikx96:
@bk3c added redirect in /applications/start/choose-language, I think it's ready for the merge.

bencpeters:
LGTM

PR #525, opened 2020-08-13
Various style updates including sticky footer

slowbot:
- Sticky footer
- Form card lead spacing
- Removing un-needed back buttons
- Date year field
- Error states for radios and checkboxes
- Adding label class to field component (I know this goes against best practice and open to suggestions)

### DIFF
--- a/shared/ui-components/src/forms/Field.tsx
+++ b/shared/ui-components/src/forms/Field.tsx
@@ -5,6 +5,7 @@ export interface FieldProps {
   error?: boolean
   errorMessage?: string
   controlClassName?: string
+  labelClassName?: string
   type?: string
   id?: string
   name: string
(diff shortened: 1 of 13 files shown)


jaredcwhite:
About the label classes…I'm not morally opposed to the idea, but maybe there's a way to create a descriptive prop instead and set the relevant class under the hood for the variation.

shared/ui-components/src/forms/Field.tsx
```diff
@@ -5,7 +5,8 @@ export interface FieldProps {
   error?: boolean
   errorMessage?: string
   controlClassName?: string
-  labelClassName?: string
+  caps?: boolean
+  labelClasses?: string
```
jaredcwhite, line 9: We can take this `labelClasses` prop out I think.

slowbot: Done.

PR #533, opened 2020-08-17
Do not show map if lat & lng are null

dominikx96:
I created a check for null (buildingAddress or buildingAddress.lat || buildingAddress.lng) - I'm not sure if the situation exists when buildingAddress has few records, but not have lat & lng.

### DIFF
--- a/apps/public-reference/pages/listing.tsx
+++ b/apps/public-reference/pages/listing.tsx
@@ -292,17 +292,19 @@ export default class extends Component<ListingProps> {
               </div>
             </ListingDetailItem>
 
-            <ListingDetailItem
-              imageAlt="neighborhood-buildings"
-              imageSrc="/images/listing-neighborhood.svg"
-              title={t("listings.sections.neighborhoodTitle")}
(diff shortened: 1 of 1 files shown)


apps/public-reference/pages/listing.tsx
```diff
@@ -292,17 +292,19 @@ export default class extends Component<ListingProps> {
...
-              subtitle={t("listings.sections.neighborhoodSubtitle")}
-              desktopClass="bg-primary-lighter"
-            >
-              <div className="listing-detail-panel">
-                <ListingMap address={listing.buildingAddress} listing={listing} />
-              </div>
-            </ListingDetailItem>
+            {listing?.buildingAddress.latitude && listing?.buildingAddress.longitude && (
+              <ListingDetailItem
+                imageAlt="neighborhood-buildings"
```
bk3c, line 297: Unrelated to this PR directly, but shouldn't alt tags be translatable strings, @slowbot?

bk3c: Ok, I split this off into #547, since it's really a separate thing are there are other instances as well.

PR #545, opened 2020-08-19
Feature/accessibility modal component

dominikx96:
I did a small refactor to create a react portal and render modals outside page content. I think it's a fine solution because page content should have "aria-hidden" attribute, so splitting those elements is the easiest way to achieve it.

I also added preview-body.html to support react portal in storybook preview.

Also, added modal close on outside click (it's an extra feature, @slowbot doesn't mentioned about it in the issue).

@bencpeters I'm not sure if this refactoring does not break your timeout modal - so, after positive review, I can inspect and change your component to work with that solution.


bk3c:
For the tabindex linter warnings, see https://webaim.org/techniques/keyboard/tabindex for more information.

@slowbot - Are we setting tabindex to 1 here since we want the modal to explicitly take priority, or can/should we be setting it to zero?

shared/ui-components/src/modals/Modal.tsx
```diff
@@ -26,46 +32,97 @@ export const Modal: FunctionComponent<ModalProps> = ({
   fullScreen = false,
   onClose,
   children,
+  ariaDescription,
 }) => {
+  // append modal to the root of app
+  const modalRoot = typeof document !== "undefined" ? document.querySelector("#__next") : null
+  const el = typeof document !== "undefined" ? document.createElement("div") : null
```
bencpeters, line 39: This should probably be wrapped in `useState` or similar, no? Right now every time this component re-renders it's going to create a new "div" element...

shared/ui-components/src/modals/Modal.tsx
```diff
@@ -26,46 +32,97 @@ export const Modal: FunctionComponent<ModalProps> = ({
...
+      enableBodyScroll(el)
+      modalRoot.removeChild(el)
+    }
+  }, [el, modalRoot, siteContainer, open])
+
+  useEffect(() => {
+    if (isEscClicked && onClose) {
+      onClose()
+    }
+  }, [isEscClicked, onClose])
```
bencpeters, line 64: I think it makes a little bit more sense to me to have `useKeyPress` take a `onKeyPress` handler instead of requiring us to define another effect here....

shared/ui-components/src/helpers/useOutsideClick.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useCallback, RefObject } from "react"
+
+type OutsideClickProps = {
+  ref: RefObject<Element>
+  callback: () => void
```
bencpeters, line 5: Nitpicky, but I'd use `() => unknown` for the type here (since it's not that you require the handling function not to return anything, you just don't care about the return value)

shared/ui-components/src/helpers/useKeyPress.ts
```diff
@@ -0,0 +1,37 @@
...
+  )
+
+  const upHandler = useCallback(
+    ({ key }: KeyboardEvent) => {
+      if (key === targetKey) {
+        setKeyPressed(false)
+      }
+    },
+    [targetKey]
+  )
```
bencpeters, line 22: Maybe simplify the repeated handler logic?:

```typescript
  const keyHandler = useCallback(
    (type: "up" | "down") => ({ key }: KeyboardEvent) => {
      if (key === targetKey) {
        setKeyPressed(type === "down")
      }
    },
    [targetKey]
  )

... 
// inside useEffect
const downHandler = keyHandler("down")
const upHandler = keyHandler("up")
```

bk3c:
I'll defer to @slowbot for the overall intent, but I'll say that if we call it "modal" then it should probably behave like that and "hold focus" until resolved. If we want it to just be a pop-up, then we should call it something different?

dominikx96:
> * If the modal has an input on it (e.g. a text field), how do the key listeners respond? I think it could be useful to allow the user to type on a modal at some point...

Hmm, not sure if I understand correctly. The modal has children prop and allows to put content e.g inputs etc. All focusable elements work as expected :)

shared/ui-components/src/modals/Modal.tsx
dominikx96: I refactored this hook to do not store state, I think it will be simpler & enough for most purposes.

shared/ui-components/src/modals/Modal.tsx
dominikx96: Correct! I moved these elements to the useState. Good catch, thanks 👍

shared/ui-components/src/helpers/useOutsideClick.ts
dominikx96: Correct, changed to unknown.

shared/ui-components/src/helpers/useKeyPress.ts
dominikx96: Refactored.

shared/ui-components/src/modals/Modal.tsx
```diff
@@ -35,14 +35,25 @@ export const Modal: FunctionComponent<ModalProps> = ({
...
+  )
+
+  const [siteContainer] = useState<Element | null>(
+    typeof document !== "undefined" ? document.querySelector(".site-container") : null
+  )
+  // append modal to the root of app
+  // const modalRoot = typeof document !== "undefined" ? document.querySelector("#__next") : null
+  // const el = typeof document !== "undefined" ? document.createElement("div") : null
+  // const siteContainer =
+  //   typeof document !== "undefined" ? document.querySelector(".site-container") : null
```
bencpeters, line 53: Remove 👣

dominikx96:
> @dominikx96 Is there a way within storybook to trigger the modal? Or is it just display.
> 
> 1. I think we should have the option to show or hide the background overlay
> 2. I do agree with be though that if have a UX scenario where we want another interaction pattern we can create another component for that
> 3. We do want the modal to be in focus but the tab index should not be greater than zero

Now it's just a display.
1. It's possible, FullScreen modal has additional prop "fullScreen" - when is set to true, background overlay is visible.
3. Fixed on the close button - now it has just 0 tabIndex.

@bk3c @bencpeters I also tested with timeout.tsx and works without change, so I think it's ready for merge.

bencpeters:
LGTM, I'd say it's ready to merge!

PR #550, opened 2020-08-19
Backend Record Localization

### LINKED ISSUE #360: Implement a method for storing and showing localized listing fields
The goal here is to have the data for each individual listing (what's currently in the listing.json file, and soon to be in the database) available in multiple languages to match the selected site language. In the event that a particular string isn't available in the desired language, we should show english instead as a fallback. Note that the first implementation is expected to cover (in addition to English):

- Spanish
- Chinese (Traditional)
- Vietnamese

I think the easiest way to accomplish this is to add a query parameter in the call from the front-end to the backend that specifies the preferred language, and then have the backend return in the best language it can. Ideally, we'd also include some metadata both confirming the language that was asked for and also the language of the response (or maybe the # fields delivered in each language?) for debugging purposes.

I suspect that implementing this might tip the balance in favor of having the browser connect to the backend in real-time rather than having Netlify connect at deployment time, which is fine with me if it happens, but certainly not required if we want to continue to avoid it. Open topic of discussion (it sounded like @jaredcwhite had an opinion, at least).

For the implementation in the new backend, we at least need:

- [ ] evaluation of a preferred language in the query string
- [ ] a way to store multiple language versions of all of the existing fields, lining up with the current JSON keys
- [ ] the logic to choose the correct language, ideally on a per-field basis so we only show an english fallback for any specific fields where the requested language isn't available
- [ ] a scheme to send back to the front-end whatever metadata we want

There's some demand for this potentially before we fully switch over from JSON to a database on the backend, so ideally we'd come up with a system that works with both versions of the backend. If it seems like it's going to be a ton of additional work to get the legacy backend Listings Service on board, let's talk about that because maybe we can/should put the effort into just moving to the new backend instead.

bencpeters:
This PR adds the ability to localize fields on individual records in the backend. Right now, translations are supported on `Listings` and `Preferences`, and are stored in the database in tables `listing_translations` and `preference_translations`. Each translation table contains a foreign key reference to the parent table, a language code (e.g. `"es"`) column, and all of the columns that make sense to translate on the parent record (basically all the `text` columns, although there are a few cases where we're using `text` columns essentially as enums, and we should probably use a smarter approach for those, rather than requiring identical translations throughout the database).

The desired language is specified on a request using the standard [`Accept-Language` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language). The backend looks through the list of languages, and uses the highest priority matching translation it can find. If no matching translations are found, it returns the original record unchanged. If a translation record _is_ found, but only has some fields filled out, the API merges existing translated fields with the underlying record as best it can. If languages are specified with locale strings (e.g. `en-US`), the backend tries first to match the language + locale combo, then tries to match just the language (`en` in this example), and then moves on to the next lower priority language.

This Listings controller now returns a `Content-Language` header based on translations performed on listings.

Additional changes:
- Added unit tests to the `ListingsService`
- Refactored the listings seed service to expose some utility functions which will be useful for database-less backend unit test mocks.

Closes #360

### DIFF
--- a/backend/core/src/entity/listing-translation.entity.ts
+++ b/backend/core/src/entity/listing-translation.entity.ts
@@ -1,0 +1,60 @@
+import { Entity, BaseEntity, Column, ManyToOne, PrimaryColumn } from "typeorm"
+import { Listing } from "./listing.entity"
+import { WhatToExpect } from "@bloom-housing/core"
+import { JoinColumn } from "typeorm/index"
+
+@Entity({ name: "listings_translations" })
+class ListingTranslation extends BaseEntity {
+  // Manually define this column ahead of languageCode so that it can appear first in the composite primary key that
+  // gets generated for better indexing performance.
+  // See https://github.com/typeorm/typeorm/issues/3069
+  @PrimaryColumn({ type: "uuid" })
+  listingId: string
+
+  @PrimaryColumn({ type: "citext" })
+  languageCode: string
+
+  @ManyToOne(() => Listing, (listing) => listing.translations, {
+    onDelete: "CASCADE",
+    nullable: false,
+  })
+  @JoinColumn({ name: "listing_id" })
+  listing: Listing
+
+  @Column({ type: "text", nullable: true })
+  accessibility: string
+  @Column({ type: "text", nullable: true })
+  amenities: string
+  @Column({ type: "text", nullable: true })
+  buildingSelectionCriteria: string
+  @Column({ type: "text", nullable: true })
+  costsNotIncluded: string
+  @Column({ type: "text", nullable: true })
+  creditHistory: string
+  @Column({ type: "text", nullable: true })
+  criminalBackground: string
+  @Column({ type: "text", nullable: true })
+  leasingAgentOfficeHours: string
+  @Column({ type: "text", nullable: true })
+  leasingAgentTitle: string
+  @Column({ type: "text", nullable: true })
+  name: string
+  @Column({ type: "text", nullable: true })
+  neighborhood: string
+  @Column({ type: "text", nullable: true })
+  petPolicy: string
+  @Column({ type: "text", nullable: true })
+  programRules?: string
+  @Column({ type: "text", nullable: true })
+  rentalHistory: string
+  @Column({ type: "text", nullable: true })
+  requiredDocuments: string
+  @Column({ type: "text", nullable: true })
+  smokingPolicy: string
+  @Column({ type: "text", nullable: true })
+  unitAmenities: string
+  @Column({ type: "jsonb", nullable: true })
+  whatToExpect?: WhatToExpect
+}
+
+export { ListingTranslation as default, ListingTranslation }
--- a/backend/core/src/entity/listing.entity.ts
+++ b/backend/core/src/entity/listing.entity.ts
@@ -5,6 +5,7 @@ import { Attachment } from "./attachment.entity"
 import { Address, UnitsSummarized, WhatToExpect } from "@bloom-housing/core"
 import { Application } from "./application.entity"
 import { ListingTranslation } from "./listing-translation.entity"
+import { Exclude } from "class-transformer"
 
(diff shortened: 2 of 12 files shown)


bk3c:
Code review to come shortly, but in terms of approach I'm not sure we can rely solely on the HTTP header, since we('re about to via #307) offer the ability to select language from the nav bar, which should change the language of the site but won't change the headers. Maybe we could take it both explicitly in a query param and also from the header, with the query param over-riding the header if it's there?

bk3c:
On the schema side, I'm wondering about the merit of having a more generic "translations" table, with columns for object type, property name, language code, and value. That way we won't end up with a XXX_translations table every time we add a new object with user-facing strings on it. For example, we're already planning to split off Properties (the real estate kind, not the object kind) into their own table, which will be 1:M with Listings. I can see there being at least a dozen more tables on the backend within the next few months...

I realize that the generic translations table has its own downsides in terms of size and management, so I'm not convinced it's better, but I think I'm more worried about the proliferation of tables than I am about one biggish one.

bencpeters:
> On the schema side, I'm wondering about the merit of having a more generic "translations" table, with columns for object type, property name, language code, and value. That way we won't end up with a XXX_translations table every time we add a new object with user-facing strings on it. For example, we're already planning to split off Properties (the real estate kind, not the object kind) into their own table, which will be 1:M with Listings. I can see there being at least a dozen more tables on the backend within the next few months...

Ya, I definitely thought about a few different schema options a bit, but there are a few major downsides to ^ that approach:

- You can't have multiple foreign keys on the same column. This means that you basically either forgo all the automatic data integrity benefits of foreign keys, or need to have a nullable column per translated entity defined.

- Modifying and viewing translations becomes less user-friendly since there are now many rows per entity-language pair.

- The schema is far less friendly for actually keeping track of what fields "should" be translated for a given entity type. With the translations tables, this is more or less self-documenting.

- It's not a natural fit for an object-oriented ORM like TypeORM. This isn't a show stopper, since you can obviously construct arbitrary queries with the QueryBuilder and `getRawMany`, but it's more manual coding work.

At the end of the day, I don't think there's a strong reason to be worried about a few(or even 10+) extra tables lying around - it's quite clear what their purpose is. We could even create a new schema `translations` to namespace them if desired - `translations.listings`, for example.

software-project:
In past for heavily i18n page we used json fields for translations instead od strings. Each json would contain en, fr, ch values etc. It was really easy to work with those fields as we had text fields with flag tabs. So we were just updating one record. Pulling data is also simple. Downside, bigger listings jsons, but I suppose we can return only current translation from backend.

jaredcwhite:
Chiming in here way after the fact…I think I agree with @software-project that it would be easier to manage translated strings in a JSON blob…I suspect there aren't a whole lot of fields that actually need to get translated since a decent portion of the data is dates, numbers, booleans, etc. And if we have the main field as the fallback, it wouldn't break anything (other than showing the wrong language) if there's a data key mis-match in the blob over time

software-project:
@kathyccheng @jaredcwhite @slowbot @pbn4 I had another look on this PR. So as discussed lately we don't want to translate listings. This PR adds translations for listings and preferences. That decision aside, I don't like the way it adds that translations. It adds another table for both listings and preference and introduce duplication on translated fields as 'base'/ english and other translations are in the separate table. Besides that column duplication it will require doing more queries to the DB to get those translations. The solution that I proposed before would eliminate both of those issues. From what Michał said the drawback of JSON column translation is that we can't query json fields in our ORM so we are not going to be able to eg. search text in those fields, at least till TypeOrm doesn't solve that issue.
My call would be not to merge that PR for 2 reasons: 1. We want to google translate listings; 2. Solution is not optimal.
Side note: I think we will be adding translation in the future as google translate is not perfect and one developer in future will be lets say less forgiving ;)

kathyccheng:
@bencpeters I think we're going to go with another approach here, discussion following up back in #360

Issue #518, opened 2020-08-12
Refactor the Apply UI component and attachment/applicationmethod schema


bk3c:
Split off from #516, where the first implementation had the IDs hard-coded in the apply links. As it currently stands, we consider external links "attachments", and I had hacked on top of that to point back to our internal application link. What I'm proposing we change to is:

- add a new TS interface called ApplicationMethod that has a 1:M relationship with listings (e.g. each listing can have one or more methods to apply).
- add multiple types to implement ApplicationMethod for internal online applications, links to external applications, download of PDF, pickup of paper application from leasing agent, etc.
- refactor all of the appropriate fields on Listing into listing.applicationMethod.whatever
- get rid of boolean fields for whether or not a listing has specific methods, and just use the presence of the relevant ApplicationMethods on a particular listing to define whether a method applies
- add logic to the Apply UI component to do the right thing passing through the listing ID for internal applications
- fix all of the other sidebar UI components to work with the new schema appropriately
- continue to have an Attachment object, but focus it holding metadata for 3rd party assets that we're going to manage in the system (e.g. images and PDFs).

cc: @akegan since this is a schema change that will move us further from the historical DAHLIA schema. I think it's the right direction for future flexibility and should be pretty easy to facade for compatibility in the SF case, but would be happy for your input here.

jaredcwhite:
@bk3c New ApplicationMethod model sounds good. I'm not sure we'd even need Attachment anymore…probably should be changed to more of a generic Asset model, and then have a polymorphic relationship with anything that needs to contain assets (Listings as well as other stuff down the road).

jaredcwhite:
So, are all of these flags going to be redundant given what we'll build into ApplicationMethod?

* `acceptingApplicationsAtLeasingAgent`
* `acceptingApplicationsByPoBox`
* `acceptingOnlineApplications`
* `acceptsPostmarkedApplications`
* `blankPaperApplicationCanBePickedUp`

bk3c:
Yes, I think those all disappear from the main listing, though maybe `acceptsPostmarkedApplications` needs to exist as a key on the mail-in method.

akegan:
@bk3c this seems reasonable and seems like something we'd be able to translate our listings into. The idea is that the listing will have a property called something like ApplicationMethods with type ApplicationMethod[], right?

jaredcwhite:
I think this is ready to close now. Any further thoughts @bk3c?

bk3c:
[...]

PR #551, opened 2020-08-20
Redo ApplicationAttachments to ApplicationMethods within ui-components

jaredcwhite:
**Don't merge!**

Regarding #518 

This only tackles the ui-components (aka Storybook) side of the equation. It needs a backend change as well, and then the Next reference app will need updating.


jaredcwhite:
@bk3c Started out using a number-based enum for this, but now I'm thinking maybe we want string-based so it's easier to inspect the underlying data. Thoughts? https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

bk3c:
Agreed. I can't see any benefit to using integers over strings, since it's not like we're going to be space-constrained at that level. I think this is consistent with w3hat @pbn4 already did for the Listing entity on the backend (see `backend/core/src/entity/listing.entity.ts`)

shared/ui-components/src/atoms/__snapshots__/ApplicationStatus.stories.storyshot
```diff
@@ -21,11 +17,30 @@ exports[`Storyshots Listing Sidebar/Application Status Due Soon And Vivid 1`] =
...
         "applicationPhone": "(408) 217-8562",
+        "assets": Array [
+          Object {
+            "fileId": "https://regional-dahlia-staging.s3-us-west-1.amazonaws.com/listings/archer/archer-studios.jpg",
+            "label": "building",
+            "referenceId": "Uvbk5qurpB2WI9V6WnNdH",
+            "referenceType": "Listing",
+          },
+        ],
         "attachments": Array [],
```
pbn4, line 43: Since attachments are no longer in use I guess those arrays can be removed, correct?

bk3c: Yes, please.

services/listings/listings/archer.json
```diff
@@ -16,9 +12,28 @@
...
+  ],
   "applicationOrganization": "98 Archer Street",
   "applicationPhone": "(408) 217-8562",
-  "blankPaperApplicationCanBePickedUp": true,
+  "assets": [
+    {
+      "referenceType": "Listing",
+      "referenceId": "Uvbk5qurpB2WI9V6WnNdH",
+      "label": "building",
+      "fileId": "https://regional-dahlia-staging.s3-us-west-1.amazonaws.com/listings/archer/archer-studios.jpg"
```
bencpeters, line 34: pedantic, but perhaps call this `fileUrl` to make it more clear what it is?

jaredcwhite: Well, eventually when we switch to a dedicated external file provider (Cloudinary maybe?), we'd only want to store the object ID of the file and the URL would be generated as a derivative of that.

jaredcwhite:
@bencpeters I agree with you regarding the JSON going over the wire. Keep it simple. From the backend perspective, I was imagining assets to be a polymorphic association so an asset could be related to anything, not just a listing. So we'd need to store both the ID and the object type of the association. I'm not familiar with the best way to implement that from a Nest/TS standpoint, I'm just drawing on my experience with polymorphic tables in Rails.

bencpeters:
> From the backend perspective, I was imagining assets to be a polymorphic association so an asset could be related to anything, not just a listing. So we'd need to store both the ID and the object type of the association. I'm not familiar with the best way to implement that from a Nest/TS standpoint, I'm just drawing on my experience with polymorphic tables in Rails.

Ah, ok, that makes sense in that case. From my perspective the best path forward then is for this type to not take a position on the reference type/id at all (leave it out of the type in `shared/core`) and leave that completely up to the backend - the purpose of this type is basically to define what the data coming over the wire will look like to front-end consumers, and I think we can do that without including those "implementation details" in the JSON response.

Issue #363, opened 2020-06-26
Translations: Add top nav language selection


bk3c:
With translations incoming soon for Alameda, we want to make sure we have the top navigation ready to allow users to select their preferred language. There should probably be a good way to define the list of language options on a per-site basis, so I'm expecting a default implementation in public-reference that's easy and documented or obvious for how to customize.

We need to account for some choices in the language nav menu that will truly change the site language, while others will just point to a one-off welcome page for that language, but not change anything else about the site.

Lastly, we need to decide if we should automagically change the language for new users based on their browser language settings. I'm in favor of this by default if we see an incoming user who has their browser set to prefer a language that we support and they haven't chosen anything otherwise, but I could be convinced otherwise if we think the best practice is to wait for an explicit choice regardless. On a related note, should we store the language preference in a cookie that lasts longer than the session? (cc: @antonelitum, since doing so would have cookie policy implications, but I think it's the right thing for the user)

I'll admit to confusion on my part with apologies if we already did this, but I think maybe we haven't yet?

jaredcwhite:
So the way it works at the moment is it's entirely route-based. So for example, these pages in English:

* https://bloom.exygy.dev
* https://bloom.exygy.dev/listings

become for Spanish:

* https://bloom.exygy.dev/es
* https://bloom.exygy.dev/es/listings

I'm thinking, regrading the top language selection, is if it's simply pointing to a "temporary" welcome page in another language, all it has to do is point to a specific page URL for that, and that's the end of it.

But if we want to it "change" the language of the site as a whole, we'll want it to redirect to the homepage URL of that language (/es or whatever) and then make sure all routes "hanging off" that homepage are linking to the correct language-specific routes.

I'm not sure we need to store any session/cookie info for the language, unless we end up in a scenario where a particular action has to redirect elsewhere and that action needs something more than a route to determine which language to use.

bk3c:
Agreed that the "welcome page" case should be pretty simple.

My suggestion for the "full language switch" case, if it's not too complicated or cumbersome, is that we figure out a way to have the links and routes generally resolve to maintain the same language without needing to explicitly put each language in the markup for each link, since that would explode quickly into an unmaintainable number of permutations.

dominikx96:
@slowbot Does any design for that component exist?

bk3c:
I think we're doing another bar across the top similar to https://housing.sfgov.org/, but I'll defer to @slowbot if he has more details.

bk3c:
[...]

PR #552, opened 2020-08-21
Feature/lang nav

dominikx96:
Issue: #363


shared/ui-components/src/navigations/LanguageNav/LanguageNav.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+    </li>
+    <li>
+      <a href="">Spanish</a>
+    </li>
+    <li>
+      <a href="">Chinese</a>
+    </li>
+    <li>
+      <a href="">Vietnamese</a>
+    </li>
```
bencpeters, line 23: This story should show how to use the actual UI element that it renders.

For example, the code here should be:

```jsx
const items = [...];
<LanguageNav item={items} />
```

That way a developer looking through story book can see both how the element looks and how to use it. In your example here, all the `<li>` elements are unused (since `LanguageNav` never renders `children`), and we don't see items actually getting rendered.

jaredcwhite:
@dominikx96 Let me know when I can do a final review, thanks!

dominikx96:
@jaredcwhite quick summary of what has been implemented: 

- created hook to switch language (it can be used for /start/choose-language as well)
- created language list with translatable labels
- now SiteHeader & Language are reusable (user needs to pass languaged in the props)
- defined each application form path in the next.config

I also think about:

- Maybe is there a better place to store the language list, not in the application.tsx? (in this place translator should work)
- Maybe instead of defining each path directly to the next.config, we can make it automagically? (it's the nodejs, so technically we can, but I'm not sure if it suits our requirements)

I'll be grateful for your feedback :)

jaredcwhite (changes requested):
@dominikx96 I like your approach here!

A couple of small suggestions:

* As you said, we should probably have one location where we configure the available language codes (`es`, `zh`, whatever). I'm wondering if that could be supplied through an environment variable, so that it's accessible by both the Next config and the frontend?
* When checking the current pathname in `useLanguageChange`, it should probably do a regex so it's looking at the front of the path string, as `/zh/essay` would currently match for "es" as well as "zh". Does that make sense?

dominikx96:
@jaredcwhite thanks for your feedback! 

- I defined regex to test different combinations: "/lang", "/lang/", "/lang/..." so I think now it should work in each case
- Moved language codes to the .env file, next I transform it to get label translation

Could you re-check it? :)

shared/ui-components/src/helpers/useLanguageChange.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  function change(prefix: string): void {
+    if (typeof window == "undefined") return
+
+    const pathname = window.location.pathname
+
+    let newPath = pathname
+
+    const pathIncludesLang = languages.filter((item) => {
+      const pattern = new RegExp(`(/${item.prefix}/$)|(/${item.prefix}$)|(/${item.prefix}/)`, "gm")
```
jaredcwhite, line 15: I think all you'd need to use here is something like this:

`^/(${item.prefix})(/|$)`

You either want to match `/es` (end of string) or `/es/` (plus whatever), and you'll have the prefix as a capture group in case you ever need it. You wouldn't want to match "/some/path/including/es/and/more"

shared/ui-components/src/helpers/useLanguageChange.ts
```diff
@@ -0,0 +1,35 @@
...
+
+      return item.prefix !== "" && pattern.test(pathname)
+    })
+
+    if (pathIncludesLang.length && prefix === "") {
+      newPath = pathname.replace(
+        `${pathIncludesLang[0].prefix}${pathname.length > 3 ? `/` : ``}`,
```
jaredcwhite, line 22: If I'm reading this rightly, a path like "/path/es/whatever" could become "/path/whatever", which isn't what we'd expect. It should match only when the language prefix is at the beginning of the URL.

jaredcwhite (changes requested):
@dominikx96 The new env configuration for languages looks good! I just saw a couple of outstanding issues in the `useLanguageChange` hook, so hopefully we can resolve that and then LGTM. :+1:

dominikx96:
@bk3c last improvement has been implemented, the branch is up to date with master.

jaredcwhite:
@dominikx96 Can you update the storshots? Then hopefully we'll get tests passing and can merge in.

dominikx96:
@jaredcwhite snapshots for header & lang nav updated, but CI still says that is an issue with:
Snapshot name: `Storyshots Listing/Map Show BW Map With Pin 1`

Locally, I have not any problems, all tests passed, do you have an idea what happens?

jaredcwhite:
LGTM!!

PR #556, opened 2020-08-21
Qa accessibility

slowbot:
- Adding Skip Nav
- Adding focus state to links and buttons


jaredcwhite:
Looks good to me. LGTM

PR #558, opened 2020-08-24
Qa accounts

slowbot:
Made some styles updates

- repositioned sign in / account link in nav
- created layouts for accounts pages
- various style tweaks to my applications
- hide dropdown parent link (probably a better way to do this @jaredcwhite


apps/public-reference/pages/account/applications.tsx
```diff
@@ -76,23 +76,25 @@ export default () => {
...
-                      application={application}
-                      setDeletingApplication={setDeletingApplication}
-                    ></AppStatusItem>
-                  ))}
-                  {applications.length == 0 && noApplicationsSection}
-                </Fragment>
-              </DashBlock>
-            </DashBlocks>
-          </div>
+          <section className="" style={{ background: "#f6f6f6" }}>
```
jaredcwhite, line 79: That's strange, I don't know how the hardcoded background snuck in here. @slowbot Do you mind relocating that to the stylesheet and switch to the Tailwind colors?

PR #559, opened 2020-08-24
Minor dependency updates: late August edition

bk3c:
There should be nothing of note here, but it would be great to get a few other pairs of eyes to make sure all is running well locally and there are no unintended consequences.


bk3c:
a couple of notes:

- I backed axios down to 0.19.2 since some dependencies we have use that pinned version, and we were running into conflicts based on two different versions since one was hoisted and the other wasn't. I don't think there's any huge rush to upgrade it, and we can get there once things catch up

- There was a bunch of cruft in the yarn.lock, so I regenerated a fresh one which took the file size down by a few hundred lines.

- There was an unrelated build problem introduced in #557 where we forgot to remove the reference to homepage.mdx from the public-reference error page. I went ahead and fixed that here so tests would pass (and shame on me for merging the other issue without noticing the failure)

- Cypress, Partners, and Backend tests are still failing for unrelated reasons.

jaredcwhite:
Ooo, Tailwind 1.7! There's a bunch of improvements to the `@apply` feature…it's behind a `applyComplexClasses` config flag though.

@bk3c Should we try to switch that on as part of this PR, or investigate it separately?￼

bk3c:
Turns out that Prettier 2.1 breaks Storybook, so I'm going to hold off merging this until they fix it. There's already a big issue thread, so I expect it'll get sorted out in the next ~48 hours.

bk3c:
@jaredcwhite Up to you. I was thinking we could wait and do the new tailwind usage in a follow-on PR, but if you think it'll be relatively quick and easy, feel free to make adjustments here while we're waiting on the SB issue to get resolved.

jaredcwhite:
@bk3c I'll give it a quick spin, but happy to keep it as-is if there are odd quirks.

jaredcwhite:
Well…that is weird. Just switching `applyComplexClasses` to true causes the app startup to freeze, Node pegged at 100% and my fans revved up. Switching it back to false and everything seems fine. Tried back and forth a few times. I'll poke around the Tailwind repo issues to see if anyone else is seeing a similar freeze, but otherwise I guess we'll need to leave it off for the time being.

jaredcwhite:
Ah, there IS an issue. I have to admit I'm surprised a problem this obvious made it into the 1.7 release. Oh well… https://github.com/tailwindlabs/tailwindcss/issues/2192

bk3c:
Ok, Storybook fix is in, so I'm going to go ahead and merge this. 

@jaredcwhite - feel free to make yourself an issue for the Tailwind `@apply` updates if you like, or we can just get back to it whenever it's working again.

PR #560, opened 2020-08-25
Fix/form return btn

dominikx96:
@slowbot I tried to investigate it, but I can't reproduce "submit" action after the back button click. Could you point on which step it occurs?

I also:

- made small cleanup & refactor e.g replace select with the component in checkbox group, when a user comes back, previously selected items didn't checked - now they are for few fields
- I improved error state - e.g on radio buttons - they weren't red
- Fixed some accessibility issues (missing label, incorrect input id)

MAYBE these fixes repaired the second point: "This either produces errors for required fields".


apps/public-reference/pages/applications/review/demographics.tsx
```diff
@@ -49,7 +49,13 @@ const Demographics = () => {
...
       register,
     }))
-  }, [])
+  }, [register])
+
+  const backPath = useMemo(() => {
+    return application.preferences.none
+      ? "/applications/preferences/general"
+      : "/applications/preferences/select"
+  }, [application.preferences.none])
```
bencpeters, line 58: Why are you using `useMemo` here? Since the result is a string, it shouldn't matter for purposes of a dependency list in `useEffect` (string literal equality in JS is fixed if the string value is fixed), and this isn't an expensive operation - I'd just do this for simplicity:

```
const backPath = `/applications/preferences/${application.preferences.none ? "general" : "select"}`
```

dominikx96: Agree, nice catch! 🥇

dominikx96:
@bencpeters @slowbot I created Form component to control all forms from one place. I prevented submission on "enter" click for forms elements except <button>.

slowbot (approved):
@dominikx96 LGTM

Issue #340, opened 2020-06-11
Add a check for pending migrations when starting the backend


bk3c:
Definitely one of those nice safety features that I miss from rails. It would be great to get a warning on server startup that there are pending migrations, in case one forgets to run them in their dev environment. We'll have a similar problem on Heroku for CI and other test apps, so I think we want an automated way to get to a good database state there (though I'm conflicted in whether we want to automate or just warn for local development). 

I'm similarly conflicted by whether we would want to automate for production DB migrations, but would lean toward a warning there as well to reduce the chance of having to back out a mistake. Ideally any production warning would be exposed in a way that it would be super-obvious and couldn't be missed in QA.

PR #562, opened 2020-08-25
Add warning when pending migrations are detected

pbn4:
Closes https://github.com/bloom-housing/bloom/issues/340

### DIFF
--- a/backend/core/ormconfig.ts
+++ b/backend/core/ormconfig.ts
@@ -28,18 +28,18 @@ const connectionInfo = process.env.DATABASE_URL
 export = {
   type: "postgres",
   ...connectionInfo,
-  // We don't want to auto-synchronize production data - we should deliberately run migrations.
-  synchronize: env === "development",
+  synchronize: false,
+  migrationsRun: false,
   logging: false,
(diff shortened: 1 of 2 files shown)


software-project (approved):
LGTM 👍

bencpeters:
> Looks good generally, but I'm wondering if we want to be more assertive here and exit out on startup with an error if things aren't in sync, instead of just a warning that someone could miss? Is that too much? I could be convinced either way.

That seems easier to check to me - a single test visit post deployment would reveal that the API was down. Since the migrations are presumably necessary for the API to run correctly anyway, it's really a choice between "very noticeable error on deploy" or "variable/unpredictable errors later". The only "advantage" to the non enforcement approach seems to me to be that the API _might_ degrade _somewhat_ gracefully (e.g. parts of the functionality that have nothing to do with the missing migration might still work), but I don't think that's a great thing to rely on, personally.

bk3c:
Ok, I made the change to not start up if there are pending migrations. @pbn4 does that look ok to you?

pbn4:
LGTM Thanks @bk3c

PR #564, opened 2020-08-25
Creating Print Styles

slowbot:
- Created custom media query breakpoint for print
- Form cleanup
- Created print style sheet with base style reset
- Created print show hide classes

TO DO: 
- We'll need to add a style to each public reference which prints the base URL so folks who print have a record of relevant URLS specific to their jurisdiction site
- Trying to update default border color value


jaredcwhite:
@slowbot Is this how it's looking for you when printing? Seems off to me… (using Safari 13 / macOS)
[Housing Portal.pdf](https://github.com/bloom-housing/bloom/files/5126550/Housing.Portal.pdf)

jaredcwhite:
OK, well let me know if you think we're good to merge.

bk3c:
@slowbot @jaredcwhite FYI: When I was resolving the merge conflicts I got a warning:

`The screen 'print' does not exist in your theme config` from navbar.scss

I'm going to go ahead and merge this anyway since that doesn't seem like it would break anything existing, but can you take a look and make sure that's not something we need to address?

PR #567, opened 2020-08-25
Add application confirmation + email service refactor

### LINKED ISSUE #416: Application: Email Confirmation
**If Lottery:**

Hello ______,

Thanks for applying. We have received your application for [Listing Name: link to listing].

Here is your confirmation number: [Confirmation Number]

What to expect next:
The lottery will be held on  [Lottery Date]. Applicants will be contacted by the agent in lottery rank order until vacancies are filled.

Should your application be chosen, be prepared to fill out a more detailed application and provide required supporting documents.

If you need to update information on your application, do not apply again. Contact the agent. See below for contact information for the Agent for this listing.

[Leasing Agent Name]
[Leasing Agent Title]
[Leasing Agent Phone]
[Leasing Agent Email]

Office Hours:
[Leasing Agent Office Hours]

How are we doing? We'd like to get your [feedback: link to feedback form].

Thank you,
Alameda County - Housing and Community Development (HCD) Department

**If no Lottery and Waitlist**

Hello ______,

Thanks for applying. We have received your application for [Listing Name: link to listing].

Here is your confirmation number: [Confirmation Number]

What to expect next:
Applicants will be contacted by the agent in waitlist order until vacancies are filled.

Should your application be chosen, be prepared to fill out a more detailed application and provide required supporting documents.

If you need to update information on your application, do not apply again. Contact the agent. See below for contact information for the Agent for this listing.

[Leasing Agent Name]
[Leasing Agent Title]
[Leasing Agent Phone]
[Leasing Agent Email]

Office Hours:
[Leasing Agent Office Hours]

How are we doing? We'd like to get your [feedback: link to feedback form].

Thank you,
Alameda County - Housing and Community Development (HCD) Department

**If First Come First Serve**

Hello ______,

Thanks for applying. We have received your application for [Listing Name: link to listing].

Here is your confirmation number: [Confirmation Number]

What to expect next:
Applicants will be contacted by the property agent on a first come first serve basis until vacancies are filled.

Should your application be chosen, be prepared to fill out a more detailed application and provide required supporting documents.

If you need to update information on your application, do not apply again. Contact the agent. See below for contact information for the Agent for this listing.

[Leasing Agent Name]
[Leasing Agent Title]
[Leasing Agent Phone]
[Leasing Agent Email]

Office Hours:
[Leasing Agent Office Hours]

How are we doing? We'd like to get your [feedback: link to feedback form].

Thank you,
Alameda County - Housing and Community Development (HCD) Department

software-project:
Fixes #416 

As part of this task I've also modified email service a little. I've added translations that are loaded from file and added partials.
What this PR is missing is a trigger for sending email confirmation and I'm not 100% if conditions for lottery/ no lottery listings are correct?


backend/core/src/shared/email.service.ts
```diff
@@ -14,13 +16,14 @@ export class EmailService {
   constructor(private readonly sendGrid: SendGridService) {
     const polyglot = new Polyglot({
       // TODO Add translations loaded from a file
```
bencpeters, line 18: Should we get rid of this TODO?

backend/core/src/shared/email.service.ts
```diff
@@ -34,6 +37,43 @@ export class EmailService {
     await this.send(user.email, "Welcome to Bloom", this.template("register-email")({ user: user }))
   }
 
+  public async confirmation(user: User, listing: Listing, application: Application) {
+    let whatToExpectText
+    const listingUrl = `${process.env.APP_URL}/listing/${listing.id}`
```
bencpeters, line 42: We might not want to address this now, but I know that in the past we've talked about having a single backend instance that serves multiple front ends. This would prevent us from using a single env variable to hold the front end `app_url` like this. We could perhaps have the client calling this API path provide a URL instead? (or inspect the `Request` object to pull it out?)

backend/core/src/shared/email.service.ts
```diff
@@ -34,6 +37,43 @@ export class EmailService {
...
+    const compiledTemplate = this.template("confirmation")
+
+    if (process.env.NODE_ENV == "production") {
+      Logger.log(
```
bencpeters, line 46: This seems like it might be more of a `debug` level logging? Or maybe that's a larger conversation that needs to be had about how much request-level logging is happening...

backend/core/src/shared/email.service.ts
```diff
@@ -46,6 +86,33 @@ export class EmailService {
     )
   }
 
+  private partial(view: string) {
+    return fs.readFileSync(
+      path.join(path.resolve(__dirname, "..", "..", "views").replace("/dist", ""), `/${view}`),
```
bencpeters, line 91: Is this meant for production builds when we're inside of a `dist` folder? Might it be cleaner to make sure that the build process copies over `views` instead?

backend/core/src/shared/email.service.ts
```diff
@@ -46,6 +86,33 @@ export class EmailService {
...
+  }
+
+  private translations() {
+    return JSON.parse(
+      fs.readFileSync(
+        path.resolve(__dirname, "..", "..", "locals/general.json").replace("/dist", ""),
+        "utf8"
+      )
+    )
+  }
```
bencpeters, line 114: It's probably a better practice to use the async versions of all these functions rather than blocking the whole app while they get read. This would then make this service an [async provider](https://docs.nestjs.com/fundamentals/async-providers), which _slightly_ complicates the declaration, but it's probably the "right" way to do it.

As long as this service is a [default scoped](https://docs.nestjs.com/fundamentals/injection-scopes#provider-scope) provider (basically a singleton loaded at initialization), it probably doesn't matter too much because the blocking code should only run once at initialization time. But if we ever wanted to re-scope this, it could definitely cause blocking issues.

backend/core/views/confirmation.hbs
```diff
@@ -0,0 +1,7 @@
+<h2>{{t "t.hello"}} {{> user-name}},</h2>
```
bencpeters, line 1: It seems like these views should probably be nested a level deeper - perhaps `views/confirmation_email/*`?

yarn.lock
```diff
@@ -3681,6 +3681,11 @@
     "@types/connect" "*"
     "@types/node" "*"
 
+"@types/body-scroll-lock@^2.6.1":
```
bencpeters, line 3684: Why did `yarn.lock` change without any dependency updates?

backend/core/src/shared/email.service.ts
bk3c: Yeah, this will actually be the case as soon as we expand the email functionality from Alameda to SMC, since both sites share the same backend. I'm fine if we want to do a separate PR for that, so @software-project feel free to either address now or file an issue for later as you prefer.

In terms of implementation, I think I prefer having the front-end set it explicitly, since I've seen referrer URLs get messed up a bunch of times.

For the longer term, I think what we actually want to do is add a `site` field to the user and/or listing, and then use that, so we're set up for future notification emails that might not be triggered by an explicit user action. There seems like there could be a bunch of subtleties to get that right in the farther future where we could have listings that could appear on multiple sites (e.g. a city site, county site, and regional site, all from the same backend) and we (maybe?) *also* want user accounts to transcend those sites? I'd be up for any solution that solves this most of the way in the short term, and then we can build more when we need it.

backend/core/src/shared/email.service.ts
bk3c: Maybe Logger.info? I'd be inclined to leave it on by default since it doesn't seem too noisy, but I agree we should get in the habit of being explicit about log levels.

backend/core/views/confirmation.hbs
bk3c: or `email_views/confirmation/...`, as long as that doesn't break the expected nestjs pattern

backend/core/views/register-email.hbs
```diff
@@ -1,6 +1,5 @@
 <h1>{{t "register.welcome"}} {{user.firstName}} {{#if (user.middleName)}}{{user.middleName}} {{/if}}{{user.lastName}}</h1>
 <p>
-  Thank you for setting up your account on the Bloom Housing Portal. It will now be easier for you to start, save, and submit online applications for listings that appear on the site.
+  {{"register.welcomeMessage"}}
```
bk3c, line 3: is this missing a call to the `t` function?

backend/core/views/partials/leasing-agent.hbs
```diff
@@ -0,0 +1,12 @@
+<p>
+  If you need to update information on your application, do not apply again. Contact the agent. See below for contact information for the Agent for this listing.
```
bk3c, line 2: #TranslationStringHere (and also in the `h4` below)

bk3c (changes requested):
I made a bunch of smaller comments, but I also think we need to actually add the call to EmailService.confirmation() from the controller where we're submitting the application?

backend/core/src/shared/email.service.ts
software-project: @bk3c I don't see `Logger.info` options are `'log' | 'error' | 'warn' | 'debug' | 'verbose'` I think thats sth you have added, Ben before to `welcome` action, so I just repeated the pattern. Not sure how error logging services work for NestJs. But I would say we don't want to have it here at all (if airbrake or sth can catch it) or have that bug tracking library here to catch it.

backend/core/src/shared/email.service.ts
software-project: that would be better, true. I'm not sure how to do that? Can you send me an example @bencpeters ?

backend/core/src/shared/email.service.ts
software-project: @bencpeters I tried using async version. It was working in development, but in test env sth was off with variable visibility and scoping and no partials were loaded. Is it ok if I add TODO for this one for future and leave it for now as we are under time preasure?

software-project:
@bk3c @bencpeters I've made some updates here. I've moved all the files and added them as assets, fixed missing translations. Only thing left is the URL thing, which I'm not sure how to fix quicly, so maybe we can address it in a separate PR next week?

backend/core/src/shared/email.service.ts
bencpeters: Ya, I did this in my authorization PR right [here](https://github.com/bloom-housing/bloom/blob/backend-authorization-and-roles/backend/core/nest-cli.json#L13). Basically you tell the Nest build system to copy the files over. The only subtlety is that the way our build system is set up right now, the TS files get compiled to `dist/src/path/to/file.js`, whereas when you specify an entry in `assets` it goes from `src/path/to/file` to `dist/path/to/file`, and what we want with `__dirname`-based relative paths is `dist/src/path/to/file`, hence the "custom" definition with `outDir`.

backend/core/src/shared/email.service.ts
bencpeters: Yes

bencpeters:
> @bk3c @bencpeters I've made some updates here. I've moved all the files and added them as assets, fixed missing translations. Only thing left is the URL thing, which I'm not sure how to fix quicly, so maybe we can address it in a separate PR next week?

I think the URL can be a pretty quick fix, in theory - all you have to do is add another parameter on the POST request for `url` or `app_url`, and then pass that into the `EmailService`. On the front end, just grab the URL base from the Next.js router (or `window` if the router doesn't have it, I can't remember).

I also think it's worth moving the partials into a subfolder pre-merge, but that should be super fast too.

software-project:
> > @bk3c @bencpeters I've made some updates here. I've moved all the files and added them as assets, fixed missing translations. Only thing left is the URL thing, which I'm not sure how to fix quicly, so maybe we can address it in a separate PR next week?
> 
> I think the URL can be a pretty quick fix, in theory - all you have to do is add another parameter on the POST request for `url` or `app_url`, and then pass that into the `EmailService`. On the front end, just grab the URL base from the Next.js router (or `window` if the router doesn't have it, I can't remember).
> 
> I also think it's worth moving the partials into a subfolder pre-merge, but that should be super fast too.

Partial already are in a subfolder. I will add `app_url`

software-project:
@bencpeters @bk3c I think all of the issues are addressed now. Tested locally and besides url being replaced by the email service it all works.

backend/core/client.ts
```diff
@@ -610,6 +613,7 @@ export enum EnumAttachmentType {
   'ApplicationDownload' = 'ApplicationDownload',
   'ExternalApplication' = 'ExternalApplication'
 }
+export type CombinedUserTypes = (User & any) | null;
```
bk3c, line 616: This feels a little hacky to me, but I'm fine if we clean it up later.

bk3c (approved):
Overall looks good and works for me locally too. We can come back to the APP_URL thing in a separate PR.

backend/core/client.ts
software-project: That's not me. It'ss autogenerated.

Issue #517, opened 2020-08-12
Update backend/core auth tests to work without a database


bk3c:
`backend/core` auth tests are failing in CI for lack of a database. I guess we can provision one if need be, but I'm hoping there's a reasonable way to do it with a fixture so we can keep the CI configuration simpler for the time being.

Split off from #511

bencpeters:
Ya, adding a fixture should be pretty trivial for the unit tests. It might a little bit trickier for the cypress/e2e tests, but probably still possible, I just haven't looked at it yet...

bencpeters:
An example of this are the tests in PR #550.

Do we want to focus around this & write tests for all the modules? Or just use that as a template going forward?

bk3c:
@bencpeters I think just using that as a template going forward is fine for now, though I'll never turn down more test coverage if you're feeling enthusiastic about adding more there. The main thing I want to do at this point is make sure all of the existing tests pass. I thought it was failing on auth, but it looks like I was wrong and it's failing on listings.spec.ts, so re-assigning to @pbn4.

pbn4:
@bk3c We have both end-to-end and unit tests and listings.spec.ts are e2e, but they are started together with unit tests for reason unknown. Normally those are started by a separate command `yarn test:e2e`. Should I move listings tests there and make the `yarn test` contain only unit tests? BTW End to end test contain also a suite for /applications route which is not run by the CI at the moment so we are not using all the tests we have anyway. @bk3c We can either mock the DB access for e2e or run them locally. What do you think?

PR #571, opened 2020-08-26
Fix modal ariaDescription on My Applications page + fix tests

bk3c:
Quick PR to add the now-required ariaDescription to the Modal component on the My Applications page. I just reused the same string that we're using for the modal title, but there's probably a better best practice per: https://www.w3.org/TR/wai-aria-practices-1.1/#alertdialog

This also fixes a number of test failures that snuck in from the last few PRs merged -- one each for public build, public cypress, and partners build (which still has an open issue I'll fix in a different PR).

### DIFF
--- a/apps/partners-reference/layouts/application.tsx
+++ b/apps/partners-reference/layouts/application.tsx
@@ -16,7 +16,12 @@ const Layout = (props) => (
     <Head>
       <title>{t("nav.siteTitle")}</title>
     </Head>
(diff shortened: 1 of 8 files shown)


bk3c:
Ok, Cypress works locally but is still broken on CI for lack of a database, similar to #517 on the jest side. I'm thinking we probably just need to take the plunge and set up a DB for the CI environment, but I think that's worth its own PR too.

PR #572, opened 2020-08-27
add postgres to CircleCI config + associated cleanups

### LINKED ISSUE #517: Update backend/core auth tests to work without a database

bk3c:
The goal of this PR will be to add a postgres database to our CI setup so we can run tests that need that. Will also try to keep cleaning and optimizing the configs as we go.

It looks like everything is working ok with in the core node image to test backend/core now, but adding a DB to Cypress tests will be a little more difficult since we'll have to [eject from the orb](https://github.com/cypress-io/circleci-orb#ejecting) in order to have multiple docker images. I'll leave that for a later PR, or maybe we're better off mocking the backend when we do front-end testing?

Fixes #517, if not how I had originally expected.


pbn4:
@bk3c I separated e2e tests from units in the https://github.com/bloom-housing/bloom/pull/574 which is a PR to your (this PR's) branch. I also added a command invocation to run e2e's. LGTM.

Issue #568, opened 2020-08-26
Show Address in Listings Page


slowbot:
**What is happening now:**
Currently the listings page does not show the address for each listing. 

![image](https://user-images.githubusercontent.com/13915/91332990-d0e9c480-e781-11ea-9dff-6f8b0d21b62e.png)

**What is should happen:**
Street address, city, zip should render beneath property name on each listing card

![image](https://user-images.githubusercontent.com/13915/91333066-ee1e9300-e781-11ea-8f5a-f361170f43c2.png)

bk3c:
@dominikx96 This should be pretty straight-forward to fix, I think the main thing is to try and figure out any edge cases (e.g. the address is blank from the backend) and make sure they're handled reasonably.

PR #575, opened 2020-08-28
Added missing address to listing element & updated story

dominikx96:
Issue: #568 

I also added state (it's visible on sample screenshot, but not in the description - let me know if the state should not be in the address line).

### DIFF
--- a/shared/ui-components/src/page_components/listing/ListingsList.tsx
+++ b/shared/ui-components/src/page_components/listing/ListingsList.tsx
@@ -30,11 +30,16 @@ const ListingsList = (props: ListingsProps) => {
       )
     }
 
+    // address as subtitle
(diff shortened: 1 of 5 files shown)


shared/ui-components/src/page_components/listing/ListingsList.tsx
```diff
@@ -30,11 +30,16 @@ const ListingsList = (props: ListingsProps) => {
       )
     }
 
+    // address as subtitle
+    const { street, city, state, zipCode } = listing?.buildingAddress || {}
+    const subtitle = `${street}, ${city} ${state}, ${zipCode}`
```
bk3c, line 35: I think the comma here should be between city and state rather than state and zipcode, but I want to defer to @slowbot before changing it.

PR #579, opened 2020-08-28
Disable checkboxes next to filled-in text boxes for Applications

### LINKED ISSUE #555: Disable "I don't have" checkboxes if value entered in the associated field
**What happens now**
When I enter a value into a field that has an associated "I don't have" checkbox the checkbox remains active, which would allow me to select it and delete the value entered

**What I expect to happen**
When I enter a value into a field that has an associated "I don't have" checkbox, the checkbox should be disabled
Removing the value from the field should re-enable the checkbox

image

jaredcwhite:
Fixes #555


jaredcwhite:
@slowbot I was noticing the disabled checkboxes are pretty dark in the current CSS… I almost added a 50% opacity but thought I'd run it by you first.

bk3c:
@jaredcwhite I think this is still waiting to incorporate the feedback items from @slowbot? Also, if you can fix the merge conflict whenever you get to it, that would be great.

jaredcwhite:
@bk3c Confirmed that there are some strange issues around page reloading going on. It works as @slowbot describes when you're navigating between forms, but if you refresh the browser, it gets messed up. Some oddness around default values and React hooks maybe. Investigating…

jaredcwhite:
OK, ready for a final review. Had to add in some funky client checks to get the checkboxes/disabled/dropdown states working across page reloads.

jaredcwhite:
@slowbot I identified and fixed a few more bugs around the email/phone checkboxes. Ready to try again!

jaredcwhite:
that sounds like a backend issue. something else going on.

bk3c:
I made it the whole way through the form without issue locally, so I agree that 500 error is probably something else unrelated.

Issue #528, opened 2020-08-14
[RFC] Front-end package and directory re-organization


bk3c:
Following recent discussions with @akegan and @james-wills-sf about making it easier to package common code for import rather than maintaining it in the reference implementation, I'd like to propose the following package and directory reorganization:

- Change the `apps/` top level directory to `sites/`, and generally change our naming to call those things sites rather than apps. The main reason for this is to avoid any potential confusion between "apps" and Applications, which will only get worse over time. I'm open to a name other than "sites", if anyone has a better idea.

- Rename the `@bloom-housing/core` package to be `@bloom-housing/site-core`. This will make it the clear home for common front-end code in addition to interfaces, and we can move logic that's shared between the public and partners apps there from ui-components (other than actual ui-components), plus aim to refactor as much as is reasonable there over time.

- Add `@bloom-housing/site-public` and `@bloom-housing/site-partners` packages to hold logic that we expect to be specific to one site or the other, but more general than the reference implementation. For example, the ApplicationConductor (or successor) will probably be unique to public, but it unlikely to be a point of customization so we're better of having it imported rather than have to maintain/diff/etc from the reference implementation to production sites. FWIW, I expect the partners package would grow to be substantially bigger than public over time, both in terms of code and dependency load size, so this gives us a way to keep the public site slimmed down but still have a place to abstract "core" code more cleanly from each reference ~~app~~ site.

I don't think this would change anything on the back-end in the short term.

Thoughts, @jaredcwhite, @dominikx96, @software-project, @bencpeters?

jaredcwhite:
@bk3c I like all of the suggestions. Would we move some of the non-visual pieces of `ui-components` over to `site-core`?

bk3c:
> @bk3c I like all of the suggestions. Would we move some of the non-visual pieces of `ui-components` over to `site-core`?

As many as we could.

james-wills-sf:
This all seems good to me.

I'm fine with everything going in site-core with separate public/partners/shared folders or splitting them at the package level. I don't think either make much difference to us.

jaredcwhite:
As discussed in Zoom earlier today, after trying out the move of some frontend-related code out of `shared/ui-components` and into `shared/core`, I ended up feeling like it was an increasing amount of complexity for no real gain in DX, and in fact I felt like maybe it's time to go in the opposite direction. So here's what I'm proposing:

[...]

PR #592, opened 2020-08-31
Partners: AG Custom Theme Setup

slowbot:
Sandbox to setup AG Grid theming options
https://www.ag-grid.com/javascript-grid-themes-customising/#ag-param 

Started configuring the table to match design

1. Implemented several ag-grid parameters to align with design
2. Suppressing default pagination in favor of custom nav which needs wiring up to be functional
3. Several of the data columns will need formatting to get rendered in each of the cells similar to what we do on the Review screen


bk3c (commented):
There are a bunch of functionality things that don't seem to work for me (e.g. the pagination and # to view control), but since it's all in Partners, I'm happy to merge it as a work-in-progress that's a huge improvement from a design perspective, and then we can go from there.

I did notice that the SCSS for ag-grid is in `ui-components` now, and I think I'd like to encourage us to maybe move it to `site-partners` once we get the directories re-arranged per #528, since I don't think we'll have ag-grid on the public site, and we need a way to keep the css-bloat under control until at least we get PurgeCSS working correctly.

PR #594, opened 2020-09-01
Use proper @types for markdown-to-jsx

### LINKED ISSUE #347: Contribute updated disableParsingRawHTML type to @types/markdown-to-jsx
Spun off from #342. It should be an easy 1-line + comment change, but might be a chunk of work to manage through the ceremonies as a new contributor to DefinitelyTyped.

It's certainly possible that someone else will do this before we get to it, but this seems like a pretty good way to contribute back to an upstream if someone has the time and energy.

bk3c:
They finally landed a PR to fix @types/markdown-to-jsx, so I went ahead and switched over to it before we forgot. Unrelated, other than it's another minor build cleanup in ui-components, but I also enabled `removeDeprecatedGapUtilities` for tailwind, which is officially a 2.0 deprecation, but we're not using it currently and the warning message at startup/build was annoying.

fixes #347

### DIFF
--- a/apps/public-reference/package.json
+++ b/apps/public-reference/package.json
@@ -37,6 +37,7 @@
   },
   "devDependencies": {
     "@next/bundle-analyzer": "9.5.2",
+    "@types/markdown-to-jsx": "^6.11.2",
     "@types/node": "^12.12.54",
     "@types/react": "^16.9.47",
     "concurrently": "^5.3.0",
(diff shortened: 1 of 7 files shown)


jaredcwhite:
@bk3c You could probably also remove it in `apps/public-reference/types/mdx.d.ts`

bk3c:
@jaredcwhite good catch; done. 

See also https://github.com/mdx-js/mdx/pull/811 -- it looks like they have types in mdx-js now, so we might be able to clean up some more of that.

jaredcwhite:
I'm not sure we'll be using MDX much longer…seems we're gradually shifting everything towards storing Markdown content elsewhere and parsing it on the fly with the m2jsx package.

Anyway, LGTM

PR #595, opened 2020-09-01
Use sessionStorage for site alerts

### LINKED ISSUE #570: Figure out a better way to pass alert messages than encoded in the URL
As @jaredcwhite perviously mentioned, putting those messages in the URL string makes it trivial to have arbitrary messages show up on the site, which doesn't seem like a good look. I also noticed that some messages aren't URLdecoded so they have `%20` characters in the user-facing text boxes.

I don't have an opinion on what the right/best solution is, so open to suggestions.

jaredcwhite:
Fixes #570


shared/ui-components/src/alerts/SiteAlert.tsx
```diff
@@ -1,36 +1,37 @@
...
-  } = router
+  const [message, setMessage] = useState("")
 
   useEffect(() => {
     let timeoutRef: number
-    if (message) {
+
+    setMessage(sessionStorage.getItem(`alert_message_${type}`) || "")
+
+    if (message != "") {
```
bencpeters, line 33: This effect is going to unnecessarily run at least twice and cause the component to render 3 times when we have a message:

1) triggered on first render, the `setMessage` call queues up a state update to message. However, `message` is still `""` on that render, so the rest of the logic doesn't fire.

2) The component re-renders in response to the state update. This time, `message` has changed based on the `setMessage` call in the previous render, and since it is in the effect dependency list, we got through the function again, including the `setMessage` call (fortunately, strings are immutable primitives in JS and React is smart enough to not blindly re-render, so the second call to `setMessage` with the same string doesn't trigger an infinite loop - if you were setting an object rather than a primitive this pattern would trigger an infinite re-render loop). Then we remove the item from `sessionStorage`, and `setOpen`, triggering a second re-render.

You can re-write this to avoid the unnecessary re-render:

```typescript
useEffect(() => {
  let timeoutRef: number
  const storedMessage = sessionStorage.getItem(...)
  if (storedMessage) {
    setMessage(storedMessage)
    setOpen(true)
    sessionStorage.removeItem()

    // timeout logic
  }
  return () => clearTimeout(timeoutRef)
}, [timeout])
```

Also, small note: use `!==`. To a first approximation, you never want to use `==` in JS - it performs type conversions automatically and way too aggressively, and has generally been the source of many a difficult-to-understand bug (not that I think it's causing an issue _here_ per se, but it's a good habit to get into writing JS code)

jaredcwhite: Sorry re: `!=`…old Ruby habits die hard! At any rate, I reworked it with your suggestion and that makes sense. Sounds like this is now RTM

bencpeters: > Sorry re: !=…old Ruby habits die hard! At any rate, I reworked it with your suggestion and that makes sense. Sounds like this is now RTM

@jaredcwhite Not just Ruby! Pretty much ANY sensible programming language other than JS...

jaredcwhite:
Oh right, I forgot that comment. Will look…

bk3c (changes requested):
Tested locally and I think something still isn't converted over correctly. If I go to the navbar and log in successfully, it redirects me back to the home page with the welcome message still in the URL, but also not displaying...

jaredcwhite:
@bk3c Confirmed that was an oversight. Try it again now.

bk3c (approved):
Tests locally well for me now.

Issue #582, opened 2020-08-31
Improve import of listings to the backend


bk3c:
Building on an idea from @pbn4, we should have a cleaner way to create multiple listings in the database from JSON files in the format we used for the old listings service. This probably means a combination of CRUD endpoints within backend/core, and then a command line tool similar to (or adapted from) the current `seed.ts` to read a directory full of JSON files and call the API to make the appropriate objects.

Some other things that will be needed to make this work:

- [ ] The new endpoints will have to be secured, and the calling script will need to authenticate without too much trouble
- [ ] Ideally there should be some way to check for existing objects and then update rather than create duplicates. This can be done later.

The other important thing to note is that the schema for listings will soon undergo some pretty fundamental changes as we break out Properties as their own entity (and likely keep going further from there), but once we have the basic scheme in place it seems like it should be straight-forward to evolve.

pbn4:
@bk3c Hey Ben I think it's done, isn't listings importer script exactly what we have here?

kathyccheng:
@bk3c followup on @pbn4's comment, is this good to close?

kathyccheng:
Closing

PR #599, opened 2020-09-02
Add /assets API and extend /listings with POST method.

pbn4:
Partially implements https://github.com/bloom-housing/bloom/issues/582

- attachments prop has been replaced with assets in Listing entity
- applicationsMethod prop has been added to Listing entity 
- `/assets` endpoint has been added, but still requires admin role checks from https://github.com/bloom-housing/bloom/pull/603
- `/` route has been moved to `/listings`
- 'POST /listings` methods has been added
- DTOs for units, preferences, applicationMethods, listings, assets have been added/refactored 

Missing parts:
- `GET,PUT /listings`
- `/applicatonMethods` endpoint
- `/preferences` endpoint
- `/units` endpoint


backend/core/src/units/units.dto.ts
```diff
@@ -0,0 +1,88 @@
...
+  @IsUUID()
+  id: string
+
+  @Expose()
+  @IsString()
+  amiPercentage: string
+
+  @Expose()
+  @IsString()
+  annualIncomeMin: string
```
bencpeters, line 16: This might not be the right time to fix this, but the data types of this model have always seemed off to me - it seems like these should be numeric?

backend/core/src/utils/utils.ts
```diff
@@ -0,0 +1 @@
+export const nameof = <T>(name: keyof T) => name
```
bencpeters, line 1: Is this being used? Maybe add a little docstring about what it does/is for?

Also, while I'm fine having a catch-all "utils" module for stuff that doesn't quite fit in other places, I'm not sure about the `utils/utils.ts` naming - it feels like sub files of `utils` should either be slightly focused/categorized (like `utils/strings.ts`, `utils/types.ts`, etc.) or just `utils/index.ts`.

bencpeters (changes requested):
Minor changes, but this looks great overall, good to finally swagger-ize everything!

bencpeters:
Also, now that we're moving the listings index action from `/` to `/listings` (an overdue change IMO), do we need to update the front end apps too to grab listings from the new path during build?

PR #603, opened 2020-09-03
Backend Authorization and Roles

bencpeters:
This PR adds the ability to do flexible authorization on API entities and a basic concept of roles.

### Roles
This adds the calculated property `User.roles: string[]`. Right now possible values are `["user"]` and `["user", "admin"]`, but the intention is to expand it to include other roles as we expand API capabilities. At the moment, admin is just stored as a boolean flag on the user model (`isAdmin`) for simplicity, but we could easily change that to a relation, and I'm imagining that for "domain-specific" (rather than app-wide) admin/manager roles (like a project manager, or a city manager), we will source this from join tables/models. Since `roles` is a dynamically calculated property, it can pull roles from wherever we want them to come from, as long as they are eager-loaded with the ActiveRecord model itself.

### Authorization
Authorization is implemented using [casbin](https://casbin.org/) to evaluate rules. Rules are defined in `src/auth/authz_policy.csv` - see the [application configuration](https://github.com/bloom-housing/bloom/compare/backend-authorization-and-roles?expand=1#diff-5e39420cc4eff03a469273486ffee9f4) for an example of what rules look like. This rule defines all permissions for `admin` role and requires an id match with the current user id for `user` role.

There are two methods for actually enforcing authorization. For high-level Role-Based Access Control (RBAC), a new Nest.js guard, `AuthzGuard` is provided. It can be defined on either a controller or individual request handler (method) scope, although the former is probably a more common use case. It should be used in in conjunction with the `@ResourceType` decorator to specify what type of entity (e.g. "application") this controller/handler will be requesting access to. It then checks access to all the requests based on the current loaded `req.user` (so it must run after a passport-based `AuthGuard` that loads the user onto the request object), the `ResourceType`, and the requested action. The action is either automatically inferred from the request (e.g. a `PUT` corresponds to `"edit"`, `GET` corresponds to `"read"`, etc.), or can be specified on a per-handler basis using the `@ResourceType("edit")` decorator.

The other method for enforcing authorization allows for per-object/attribute based access control (ABAC). In this mode, we are checking specific attributes about the resource access is requested on, so it must be checked in the body of the handler rather than as a guard (since the resource must be loaded from the DB). This is accomplished using the new `AuthzService.can` or `AuthzService.canOrThrow` methods.

The [Applications Controller](https://github.com/bloom-housing/bloom/compare/backend-authorization-and-roles?expand=1#diff-40580b04e54be4f162431bcf2010f445) has examples of both types of authorization - the `AuthzGuard` is applied at the controller level for RBAC permissions, and each of the actions that act on a specific application (edit, delete, and show) check access on that particular instance using `AuthzService.canOrThrow`.

Right now it seems like applications are the only resource in the API that really needs this logic applied, so that's all I've added, but it should be easy to extend this logic to other resources (and add domains to allow for different regions once it's relevant)

If we like this approach, I'll add the above descriptions w/ examples to a README before merge.


bk3c (commented):
I feel good about the overall approach here, though I agree that it's definitely just a starting point and there's plenty of interesting stuff ahead of us. 

My main comment about the implementation is that I'm not sure whether we want to use the file-based policy storage implemented here, or try to store it in the DB with https://github.com/node-casbin/typeorm-adapter (or equivalent). Part of me likes the file-based storage for being somewhat more immutable without anyone noticing, but overall I think we're on course to centralize things in the DB so that might be a better long-term answer. @bencpeters, I defer to you if you think that's better done in a separate PR (or a bas idea all together?)

backend/core/src/applications/applications.controller.ts
```diff
@@ -8,64 +8,92 @@ import {
...
   @ApiOperation({ summary: "Update application by id", operationId: "update" })
   @UseInterceptors(new TransformInterceptor(ApplicationDto))
   async update(
+    @Request() req,
     @Param("applicationId") applicationId: string,
     @Body() applicationUpdateDto: ApplicationUpdateDto
   ): Promise<ApplicationDto> {
-    return await this.applicationsService.update(applicationUpdateDto)
+    const app = await this.applicationsService.findOne(applicationId)
+    await this.authorizeUserAction(req.user, app, "edit")
```
pbn4, line 80: What do you think about make those "edit", "read", "delete", "create" strings an Enum?

pbn4 (changes requested):
This looks great to me conception wise. 

I made a couple of small comments about the code and there is this applications/list for an admin that needs resolution. I think we need to provide admins with a way of listing all apps and that can be either by extending /list or adding a separate endpoint. I prefer former and done at the controller level. What do you think?

bencpeters:
> My main comment about the implementation is that I'm not sure whether we want to use the file-based policy storage implemented here, or try to store it in the DB with https://github.com/node-casbin/typeorm-adapter (or equivalent). Part of me likes the file-based storage for being somewhat more immutable without anyone noticing, but overall I think we're on course to centralize things in the DB so that might be a better long-term answer. @bencpeters, I defer to you if you think that's better done in a separate PR (or a bas idea all together?)

I considered that. I think that would be 100% the way to to go **if** we were storing the mappings between users and roles/groups in the policies themselves. However, at this point, we're storing the roles _separately_  (on the user entity & potentially in related tables), so it's only the role definitions themselves and the individual policy directives that live in the files. Given this, I actually pretty strongly favor file-based definitions because they can be tracked in version control & PRs more easily and the policies essentially are "code", we need to view them to get a picture of what the permissions of any given API entity are. I worry that would be too error prone/difficult to keep track of in the DB, and I don't really see what the benefit would be using this approach - unless we dramatically shift the approach we're not ever going to end up with more than a few tens of lines of policies at most.

bencpeters:
>  I think we need to provide admins with a way of listing all apps and that can be either by extending /list or adding a separate endpoint. I prefer former and done at the controller level. What do you think?

Agreed with the need. The actual implementation gets a _little_ tricky though, since we don't want to issue a `can` call for every application listing. I think the best choice here is to add a new verb (`list_all`? `admin_read`?) and check for that permission within the controller:

```
  async list(
    @Request() req,
    @Query() params: ApplicationsListQueryParams
  ): Promise<ApplicationDto[]> {
    if (this.authzService.can(req.user, "application", "admin_read")) {
      // Request all apps from the service
    } else {
      // list just the user apps
    }
  }
```

That seems like a perfectly reasonable approach to me... in some ways having a separate controller or handler could make the permissions & results a bit cleaner too, but I don't feel strongly either way.

bencpeters:
@bk3c, @pbn4 I updated this based on the review comments & should be current with latest master - ready to merge? We could also add authorization permissions for any other routes (other than `applications`) if needed?

pbn4:
> @bk3c, @pbn4 I updated this based on the review comments & should be current with latest master - ready to merge? We could also add authorization permissions for any other routes (other than `applications`) if needed?

Application end-to-end tests need adjustment, but I can make it in a separate PR since I am the author of those and it will be faster this way.  

Routes /assets, /preferences, /applicationMethods, /units all should require admin permissions, /listings POST. PUT, DELETE too (except for GET), but all that requires tests adjustments so maybe it should be introduced as a of that mentioned separate PR?

pbn4 (approved):

Issue #515, opened 2020-08-12
QA: Consistently Hide Save and Finish Later 


slowbot:
Save and Finish Later shows inconsistently across application page footers. 

Let's remove the links for now until we have User Accounts

bk3c:
@slowbot Am I correct to think that the correct logic here (when we get to it), is that "Save and Finish Later" should take an already-logged-in user back to the home page or listing page (which would you prefer?) with a message that the application has been saved? And then for users that aren't logged in, it takes them to a login / create account screen with the application saved on success?

In terms of timing, I'm hoping that we can land the user account stuff soon enough that it's not worth the effort to remove the links before that's in place. Obviously if we have a pilot coming then we can remove them pretty quickly. 

On the other hand, once we land the results of #453, I expect the link will get removed from each screen anyway and be managed by whatever higher-level form component is handling overall form navigation. So maybe we could go ahead and take out the old way now? Thoughts, @jaredcwhite?

slowbot:
There is a whole flow documented for Save and Finish later.  https://github.com/bloom-housing/bloom/issues/246

slowbot:
@bk3c Can we get this in before pilot?

bk3c:
I think this is going to be borderline for the pilot since it depends on all of the edge-case account features (e.g. password reset flow) working correctly in addition to the flow outlined here. I'm happy to give it a try, but I think we should set expectations low.

bk3c:
@slowbot Wait, sorry, did I misunderstand the request? It'll be easy to *remove* the "Save and finish later" links before any pilot, and I agree we should do that for sure. What will be tough is actually making it all work right. Which did we want to aim for?

slowbot:
for Pilot simply remove the links since we wont have sign in @bk3c

bk3c:
@dominikx96 I think the preference would be to comment these out where they exist, since they'll be returning once we have all the login functionality ready for prime time.

PR #604, opened 2020-09-03
Hidden "save & finish later" button

dominikx96:
Issue: #515


slowbot (approved):
LGTM

bk3c (approved):
Points for simplicity here. I think let's merge this into core for the moment, and then we can uncomment it once we launch the listings pilot and get back to feature work, since we'll probably freeze syncing from upstream to housingbayarea at that point anyway.

Issue #537, opened 2020-08-17
Create a Print View of the Submitted Application Page


slowbot:
Goal:
As a user I want the option to be able to print a copy of he submitted application page if its difficult for me to get online or for my own security.

Acceptance Criteria: 
- If an applicant prints the page:
   - Links should render full URLs
   - Images should be minimized or hidden
   - Navigation should be hidden
   - Printed copy should print on as few pages as possible

bk3c:
@slowbot assigning back to you to work on the CSS and other design pieces, and then we can assign to an engineer to work on whatever punch list items still need to be done from there.

dominikx96:
@slowbot let me know if it's something else to do except changes provided by PR #605.

dominikx96:
@bk3c it has been implemented.

PR #605, opened 2020-09-03
Added origin to show full listing link on print view & hidden unneces…

dominikx96:
Issue: #537 

@slowbot, I noticed that few elements are unnecessary on "print view" like links. I also added a domain for listing URL to be full.

### DIFF
--- a/apps/public-reference/pages/applications/view.tsx
+++ b/apps/public-reference/pages/applications/view.tsx
@@ -7,7 +7,7 @@ import moment from "moment"
 import { FormCard, t } from "@bloom-housing/ui-components"
 import FormsLayout from "../../layouts/forms"
 import { AppSubmissionContext } from "../../lib/AppSubmissionContext"
-import { useContext, useMemo } from "react"
+import { useContext, useMemo, useState } from "react"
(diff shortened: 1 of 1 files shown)


slowbot (changes requested):
See comments @dominikx96

dominikx96:
@slowbot agree! The idea was to hide the only pager and add the domain to the listing URL. I fixed the issue by removing the hide class for this link.

apps/public-reference/pages/applications/view.tsx
```diff
@@ -18,13 +18,20 @@ export default () => {
     return moment().format(DATE_FORMAT)
   }, [])
 
+  const [origin] = useState<string>(typeof window !== "undefined" ? window.location.origin : "")
```
bencpeters, line 21: Is there a reason to wrap this in `useState`?

slowbot (approved):
LGTM

apps/public-reference/pages/applications/view.tsx
dominikx96: nope, it can be just a value, without useState, refactored.

Issue #589, opened 2020-08-31
add a feature flag (or similar) for user accounts / login


bk3c:
For our upcoming September application pilot, we're planning to disable all of the public applicant-facing features until we get them a little more QAed and some of the details implemented (e.g. the forgot password flow).

On one hand it seems like having a good way to to turn various features off and on is probably useful long-term, but I'm honestly not sure how much we'll use it, so it might be just as easy to disable the links from housingbayarea in the short-term. Once we have the login process stable, we might be able to use user groups / permissions to selectively expose features, but that obviously won't help with the login flow itself.

slowbot:
1. Main navigation sign in link
2. Already have an account on start/choose-language
3. Save and finish later on applications/household/preferred-units
4.

bk3c:
@dominikx96 (and cc: @jaredcwhite) I'm open for thoughts about whether having some sort of feature flag infrastructure is realistic in the next day or two. Otherwise, I think I'm fine with just commenting out the login elements / logic for the time being, presumably only in housingbayarea so we can just keep moving forward in core.

dominikx96:
@bk3c, for now, I just commented on these few elements, but the idea of using something like the flag is good for me.

PR #606, opened 2020-09-03
Disabled login flow elements

dominikx96:
Issue: #589 

Disabled:

"Sign In" link in main navigation
"Already have an account" section on start/choose-language
"Save and finish later" on applications/household/preferred-units and applications/review/demographics
"Create Account" screen on /applications/review/confirmation


slowbot (approved):
LGTM

Issue #297, opened 2020-05-15
Validation: Make inline error messages accessible to screen reader users by adding ARIA labels.


slowbot:
Most screen reader and browser combinations announce a form control as “required” when using the required attribute.

Prevent screen readers from announcing required form controls as invalid by default, by using aria-invalid="false".

Update the aria-invalid attribute to “true” if the current value (or lack of value) of a required control does not pass validation.

Wait to declare form controls as invalid and display inline error messages once the control has lost focus, or there has been an adequate delay in key presses (a few seconds)

Use aria-describedby on required form controls to point to an element that will contain any necessary inline error messaging.

Can’t always rely on browser’s client-side error validation to be accessible.
Create a custom client-side validation script and helpful messaging by suppressing the browser’s default validation message by adding the novalidate attribute to the  form element. 

**Code Example**:
Default Error Message Treatment
```
<form novalidate>
<label for="name"> Name:</label>
<input id="name" required aria-invalid="false">
</form>
 ```

Invalid Error Message Treatment
```
<form novalidate>
<label for="name">  Name:</label>
<input id="name" required aria-describedby="name_msg" aria-invalid="true">
<span id="name_msg">Please enter your name.</span>
</form>
```
```
<form novalidate>
<label for="Email">  Email:</label>
<input id="Email" required aria-describedby="emal_msg" aria-invalid="true">
<span id="email_msg">The email address entered is invalid.</span>
</form>
```

https://codepen.io/TPG/pen/3b68c22db120c0960503fe892bb2c068

dominikx96:
FYI: I think, before working on the main goal in this issue, we should do refactor (replace each form elements like input, select - with components Field, Select, etc.). After this, I'll start adding accessibility fix.

PR #607, opened 2020-09-03
Feature/errors accessibility

dominikx96:
Issue: #297 

This PR has a huge impact on the application form structure. I tested it but should be tested e2e by something else.


apps/public-reference/pages/applications/financial/income.tsx
```diff
@@ -63,6 +64,7 @@ export default () => {
   })
   const onSubmit = (data) => {
     const { income, incomePeriod } = data
+    console.log(data)
```
slowbot, line 67: Should this be cleaned up.

apps/public-reference/pages/applications/contact/address.tsx
```diff
@@ -266,18 +262,16 @@ export default () => {
...
+              name="sendMailToMailingAddress"
+              label={t("application.contact.sendMailToMailingAddress")}
+              register={register}
+              inputProps={{
+                defaultChecked: application.sendMailToMailingAddress,
+              }}
+            />
           </div>
 
           {(sendMailToMailingAddress || application.sendMailToMailingAddress) && (
```
slowbot, line 277: This doesn't need to be a label since the individual field now has a label. Ideally this would be a legend within a field set.

apps/public-reference/pages/applications/contact/alternate-contact-contact.tsx
```diff
@@ -88,49 +88,49 @@ export default () => {
...
+                <Select
+                  id="mailingAddress.state"
+                  name="mailingAddress.state"
+                  label={t("application.alternateContact.contact.stateFormPlaceholder")}
+                  defaultValue={application.alternateContact.mailingAddress.state}
+                  register={register}
+                  controlClassName="control"
+                  options={stateKeys}
+                  keyPrefix="application.form.options.states"
+                />
```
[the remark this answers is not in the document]
slowbot: Select produces an aria-describedby attribute with error suffix by default. Is this intentional? I'm curious if this is the reason the label is not being read.

apps/public-reference/pages/applications/household/member.tsx
```diff
@@ -169,239 +206,183 @@ export default () => {
...
+                        label="State"
+                        defaultValue={member.address.state}
+                        validation={{ required: true }}
+                        error={errors.address?.state}
+                        errorMessage={t("application.contact.stateError")}
+                        register={register}
+                        controlClassName="control"
+                        options={stateKeys}
+                        keyPrefix="application.form.options.states"
+                      />
```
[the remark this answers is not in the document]
slowbot: Select produces an aria-describedby attribute with error suffix by default. Is this intentional? I'm curious if this is the reason the label is not being read.

apps/public-reference/pages/applications/contact/address.tsx
```diff
@@ -125,262 +133,237 @@ export default () => {
...
+                    label={t("application.contact.state")}
+                    defaultValue={application.mailingAddress.state}
+                    validation={{ required: true }}
+                    error={errors.mailingAddress?.state}
+                    errorMessage={t("application.contact.stateError")}
+                    register={register}
+                    controlClassName="control"
+                    options={stateKeys}
+                    keyPrefix="application.form.options.states"
+                  />
```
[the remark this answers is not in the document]
slowbot: Select produces an aria-describedby attribute with error suffix by default. Is this intentional? I'm curious if this is the reason the label is not being read.

apps/public-reference/pages/applications/household/member.tsx
```diff
@@ -169,239 +206,183 @@ export default () => {
...
+                labelClassName={"field-label--caps"}
+                defaultValue={member.relationship}
+                validation={{ required: true }}
+                error={errors.relationship}
+                errorMessage={t("application.form.errors.selectOption")}
+                register={register}
+                controlClassName="control"
+                options={relationshipKeys}
+                keyPrefix="application.form.options.relationship"
+              />
```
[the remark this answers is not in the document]
slowbot: Select produces an aria-describedby attribute with error suffix by default. Is this intentional? I'm curious if this is the reason the label is not being read.

dominikx96:
> 1. State drop downs are producing an error and I was curious if this had to do with the current implementation of aria-describedby with an error suffix by default

I tried to investigate it and the error is because <span> element with an error shows only when the error exists. So during an initial render when there is no error element described by ARIA is missing. Is it an issue and <span> element should be always visible?

apps/public-reference/pages/applications/contact/address.tsx
```diff
@@ -125,6 +130,7 @@ export default () => {
               name="applicant.phoneNumberType"
               placeholder={t("application.contact.phoneNumberTypes.prompt")}
               label={t("application.contact.phoneNumberTypes.prompt")}
+              labelClassName={"sr-only"}
```
bencpeters, line 133: Totally pedantic, but just calling it out for perhaps cleaner code - you only need brackets `{..}` when you're evaluating JS - if it's just a basic string literal, JSX lets you do `labelClassName="sr-only"`. Similarly, if it's just a boolean flag, you can do just do `caps` (rather than `caps={true}`)

shared/ui-components/src/forms/Field.tsx
```diff
@@ -24,37 +26,56 @@ const Field = (props: FieldProps) => {
...
+  const isRadioOrCheckbox = ["radio", "checkbox"].includes(type)
+
+  const label = useMemo(
+    () => (
+      <label className={labelClasses.join(" ")} htmlFor={props.id || props.name}>
+        {props.label}
+      </label>
+    ),
+    [props.label, props.id, props.name, labelClasses]
+  )
```
bencpeters, line 50: Why is this memoized? I don't _think_ it'll cause problems, but it just seems to add unnecessary complication to me, it's not a particularly expensive operation...

shared/ui-components/src/forms/FieldGroup.tsx
```diff
@@ -1,57 +1,69 @@
...
+interface FieldGroupProps {
   error?: boolean
-  required?: boolean
   errorMessage?: string
   name: string
-  groupLabel: string
-  fields: CheckboxSingle[]
+  type?: string
+  groupLabel?: string
+  fields: FieldSingle[]
```
bencpeters, line 17: You could be passing `undefined` to this in a few of the components (when you do a conditional accessor check like `contactPreferenceKeys?.map` - if `contactPrefenceKeys` is `undefined` or `null`, then the variable will become `undefined`, rather than an empty list), I think the type should be `fields?: FieldSingle[]`. (the code works, since you've got a check in the actual render of this component, but I think they types should reflect that - I'm actually surprised Typescript isn't warning you about this/giving an error already...)

bencpeters (commented):
Looks good overall to me, although I didn't dive into the accessibility issues. Just a few minor cleaning suggestions.

shared/ui-components/styles/forms.scss
```diff
@@ -222,17 +227,18 @@ input[type=number] {
...
   @apply font-semibold;
   @apply text-tiny;
-  @apply mb-2;
+  @apply mb-3;
   @apply inline-block;
 }
 
 .field-label--caps + .field-note {
-  @apply mt-2;
+  // @apply mt-2;
```
jaredcwhite, line 241: Let's not leave commented styles checked in.

shared/ui-components/src/locales/general.json
```diff
@@ -150,7 +151,10 @@
       "provideAMailingAddress": "Provide an address where you can receive updates and materials about your application.",
       "doYouWorkIn": "Do you work in <REGION>?",
       "doYouWorkInDescription": "TBD",
-      "workAddress": "Work Address"
+      "workAddress": "Work Address",
+      "yes": "Yes",
+      "no": "No",
```
jaredcwhite, line 156: Yes and No are already in the "t" category, so let's not add them a second time. (Actually a third time, I'm noticing another duplication in the JSON.)

jaredcwhite (changes requested):
Looking good to me, other than previous comments from others. Added a couple of minor comments.

apps/public-reference/pages/applications/contact/address.tsx
dominikx96: yup, agree, I'll clean up it when I'll detect more places like this :)

shared/ui-components/src/forms/Field.tsx
dominikx96: hmm, in general, it's not an expensive operation, but for each situation when e.g user submit a form and there is an error state - each label is rendered again. useMemo prevents unnecessary re-render.

slowbot (commented):
See my latest updates. I normalized additional fieldsets and labels I had missed. @dominikx96

shared/ui-components/src/forms/Select.tsx
```diff
@@ -32,23 +32,19 @@ export const Select = ({
...
-
   return (
     <div className={"field " + (error ? "error" : "")}>
       <label className={labelClassName} htmlFor={id}>
         {label}
       </label>
       <div className={controlClassName}>
         <select
-          aria-describedby={`${idOrName}-error`}
-          aria-invalid={!!error || false}
```
dominikx96: was this "aria-invalid" and "aria-describedby" deletion intentional?

dominikx96 (commented):
Other changes LGTM.

shared/ui-components/src/forms/Select.tsx
slowbot: @slowbot No, I don't recall deleting that.

slowbot (approved):
LGTM. Will add these other items to another ticket so we can close this one. @dominikx96 @bk3c 

- Phone field does not have a default label which they should. We should also allow for sr-only and caps label class variants similar to default field component - STARTED
- Select and Fieldgroup fields should be aligned with standard field element variants including caps and sr only
- We could consider adding fieldset and legend to the fieldgroup component
- Still work to do to normalize all label conventions for proper vertical spacing - STARTED
- Empty label tags are still rendering for field component even when no label value is passed to the component.
- Can the terms checkbox use the single field component rather than fieldgroup? (It needs a primary variant)

shared/ui-components/src/locales/general.json
```diff
@@ -150,7 +151,8 @@
       "provideAMailingAddress": "Provide an address where you can receive updates and materials about your application.",
       "doYouWorkIn": "Do you work in <REGION>?",
       "doYouWorkInDescription": "TBD",
-      "workAddress": "Work Address"
+      "workAddress": "Work Address",
+      "radioError": "Please select an option."
```
jaredcwhite, line 155: Let's make sure this is a generic translation that's not specific to any one page/category,

shared/ui-components/src/locales/general.json
```diff
@@ -250,9 +253,11 @@
       "label": "ADA Accessible Units",
       "title": "Do you or anyone in your household need any of the following ADA accessibility features?",
       "subTitle": "If you are selected for a unit, the property will work to accommodate your need to the best of their ability. Should your application be chosen, be prepared to provide supporting documentation from your physician.",
+      "legend": "ADA Features",
       "mobility": "For Mobility Impairments",
       "vision": "For Vision Impairments",
-      "hearing": "For Hearing Impairments"
+      "hearing": "For Hearing Impairments",
+      "no": "No"
```
jaredcwhite, line 260: This isn't necessary, we can do `t.no`

jaredcwhite, line 260: Let's remove this in favor of using `t.no`

shared/ui-components/src/locales/general.json
jaredcwhite, line 155: Let's make sure this is a generic translation key and not tied to a particular page/category

jaredcwhite (changes requested):
LGTM but just a couple small translation key comments.

dominikx96:
@slowbot @bk3c @jaredcwhite added translations cleanup. I noticed a little more duplications so I think will be better to create a new issue for this.

bk3c:
Ok, I think we're ready to merge this in, if @dominikx96 you can resolve the conflicts?

(and then @jaredcwhite, you'll probably have a bunch of conflicts in #676 once we merge, so heads-up)

dominikx96:
@slowbot conflicts resolved.

slowbot (approved):
LGTM

PR #613, opened 2020-09-04
Listings complete API

pbn4:


backend/core/src/assets/assets.controller.ts
```diff
@@ -27,8 +38,8 @@ export class AssetsController {
     return this.assetsService.create(asset)
   }
 
-  @Put()
-  @ApiOperation({ summary: "Create asset", operationId: "create" })
+  @Put(`:assetId`)
+  @ApiOperation({ summary: "Update asset", operationId: "create" })
```
bencpeters, line 41: Maybe I'm misunderstanding the [swagger document for operationId](https://swagger.io/docs/specification/paths-and-operations/), but it seems to me that it's supposed to be unique among **all** operations in the API? so this would be something like `assetsUpdateById` or something uniquely identifying like that, not `create`? (affects the reset of the decorators too, I just noticed it here)

backend/core/src/assets/assets.service.ts
```diff
@@ -3,6 +3,7 @@ import { Asset } from "../entity/asset.entity"
 import { AssetCreateDto } from "./asset.create.dto"
 import { plainToClass } from "class-transformer"
 import { AssetUpdateDto } from "./asset.update.dto"
+import { BaseEntity } from "typeorm"
```
bencpeters: Is this being used?

bencpeters (changes requested):
Minor changes, but I think it mostly looks good for what we need now.

Later there might be a good discussion to have about whether we want "sub controllers" for each of the entities related to `listings` or if we just want to create/edit them using the main listing controller operations, but I don't think we need to worry about it too much for this release.

bencpeters (approved):

PR #620, opened 2020-09-08
Fix age validation for DOBField

jaredcwhite:
Age/year validation was broken in Safari but worked in Chromium-based browsers. This fix ensures it works consistently.


bk3c (approved):
LGTM in Safari

PR #635, opened 2020-09-10
Use assets for obtaining the listing image URL

### LINKED ISSUE #632: Application start/chooseLanguage page needs to read listings image from assets
I think this is just a matter of having a version of this function from the listing page:

```tsx
const imageUrlFromListing = (listing: Listing) => {
  return listing.assets.find((asset) => asset.label == "building")?.fileId
}
```

It's small enough that duplicating it isn't the end of the world, or anything more DRY would be great.

jaredcwhite:
Fixes #632


bk3c (approved):
LGTM after a couple minor fixes

PR #636, opened 2020-09-10
Add env var for backend API base URL, adjust the listings query accordingly

### LINKED ISSUE #633: Separate listings query string from base API URL
Right now we define the backend URL in an environment variable `LISTING_SERVICE_URL` which then gets read in to `apiUrl` in the ConfigContext on the frontend. This worked fine when we were only reading listings from that URL, but now it's also being used as the base for where we're submitting the application, and having the JSONPath query in there is confusing the backend.

Two potential solutions to this:

1. Split off into two variables like BACKEND_BASE_URL and LISTINGS_QUERY_STRING, change the listings call to use both, and then have application submission call just use the base URL + appropriate path.

2. Keep LISTING_SERVICE_URL the way it is, and add a new variable called APPLICATION_SUBMISSION_URL or similar, knowing that in the common case they will both have the same base hostname.

I'm fine with either option at this point, though I think that over the longer term the second option might offer us a little more flexibility, since it's easy to extend with additional URLs one per logical function. That said, more variables is more configuration and things that can go wrong, so I see the merits of option 1 until we actually have a need to split off one or more services. I suspect the SF DAHLIA case will need separate URLs eventually, but even that might be over time since they currently have a single backend.

jaredcwhite:
Fixes #633


bk3c (approved):
LGTM with fixes.

Issue #621, opened 2020-09-09
Phone Number Validation Bug 


kathyccheng:
On /applications/contact/address and //applications/contact/alternate-contact-contact after you've entered in a phone number, filled out the rest of the page and then clicked Next at the bottom, the last 4 digits of the phone number get stripped away and there's an error

bk3c:
Fixed in #645

PR #645, opened 2020-09-11
Changed phone mask package & input type to tel

dominikx96:
Issue: #621 

Changed mask package, because the previous has a lot of bugs and interfaces have been different than in the docs...
Now, autofill should work correctly.

Please see my issue explanation in the comment #621


shared/ui-components/package.json
```diff
@@ -73,9 +73,9 @@
     "react-dom": "^16.13.1",
     "react-focus-lock": "^2.4.1",
     "react-inlinesvg": "^2.0.1",
-    "react-input-mask": "^2.0.4",
     "react-map-gl": "^5.2.7",
     "react-media": "^1.10.0",
+    "react-text-mask": "^5.4.3",
```
bk3c, line 78: My main concern with switching is that react-text-mask seems to not be actively maintained anymore. It's still more popular than react-input-mask so I'm not sure it makes us any worse off, but it worries me as a long-term solution. Do we have any better ideas (cc: @jaredcwhite)

bk3c (changes requested):

bk3c (approved):
Needed a yarn.lock update, but I added that and looks good.

PR #651, opened 2020-09-11
Update household max size chart by unit type

software-project:


software-project:
@bk3c I've addressed most of @slowbot issues. Will just add <strong> thing tomorrow. But I can't make new field on units get sent to the frontend. I've tried regenerating client but it didn't help. Could you or @jaredcwhite help?

software-project:
@bk3c @kathyccheng @slowbot first column is strong now and subheader is fixed now. So should be ready to go.

backend/core/src/entity/unit.entity.ts
```diff
@@ -149,6 +149,11 @@ class Unit extends BaseEntity {
     onUpdate: "CASCADE",
   })
   listing: Listing
+
+  @Column({ type: "boolean", nullable: true })
+  @Expose()
+  @IsOptional()
+  bmrProgramChart?: boolean
```
pbn4, line 156: Please add `@IsBoolean` under IsOptional and change `barProgramChart?: boolean` to `bmrProgramChart: boolean | null`

services/listings/src/lib/unit_transformations.ts
```diff
@@ -32,16 +32,19 @@ const minMaxInCurrency = (minMax: MinMax): MinMaxCurrency => {
   return { min: usd.format(minMax.min), max: usd.format(minMax.max) }
 }
 
+const bmrHearders = ["Studio", "1 BR", "2 BR", "3 BR", "4 BR"]
```
pbn4, line 35: Headers*

pbn4 (approved):

bk3c (approved):
Ok -- I'm pretty sure that we're going to want to re-think which of these rules are on units and which are on listings in the future, but this works ok for me in the short-term, so let's go with it for now.

Issue #654, opened 2020-09-12
Do not expose passwordHash from API


bk3c:
When you create a new user the JSON that's returned includes the password hash. Needless to say we should never send that in response to this or any other API request. Let's make sure it's protected from view even in accidental cases.

PR #655, opened 2020-09-14
Fix passwordHash leak

pbn4:
Fixes https://github.com/bloom-housing/bloom/issues/654


backend/core/src/auth/user.dto.ts
```diff
@@ -0,0 +1,8 @@
+import { OmitType } from "@nestjs/swagger"
+import { User } from "../entity/user.entity"
+import { Expose } from "class-transformer"
+
+export class RegisterResponseDto extends OmitType(User, ["passwordHash", "applications"] as const) {
+  @Expose()
+  accesToken: string
```
bencpeters, line 7: spelling - should be `accessToken`

bencpeters (approved):
LGTM, minus a spelling change - we should see if that spelling error is just in the DTO, or if it also exists in the returned JSON?

bk3c (changes requested):
Looks ok from a code perspective, but when I run this I get the following error:

`Error: A circular dependency has been detected (property key: "user"). Please, make sure that each side of a bidirectional relationships are using lazy resolvers ("type: () => ClassType").`

backend/core/src/auth/user.dto.ts
bk3c: I agree we should fix this before merge.

pbn4: Fixed.

pbn4:
@bk3c  I don't see such warning. Isn't it maybe related to some unstaged code that you have locally?

bk3c (approved):
Previous errors were resolved after merging in the latest master.

Issue #656, opened 2020-09-14
User account email is not constrained to be unique 


pbn4:
As in the title: either are allowing to create multiple accounts for the same email or I'm missing something big.

PR #657, opened 2020-09-14
Fix email uniqueness for User entity

pbn4:


pbn4:
Fixes https://github.com/bloom-housing/bloom/issues/656

backend/core/src/entity/user.entity.ts
```diff
@@ -6,6 +6,7 @@ import {
   UpdateDateColumn,
   Index,
   OneToMany,
+  Unique,
```
bencpeters, line 9: Is this import necessary? It seems unused?

bencpeters (changes requested):
Looks good, but I think we might want to make email case insensitive unique.

Unfortunately, TypeORM [doesn't support this out of the box](https://github.com/typeorm/typeorm/issues/356), but I think it's worth doing. We can achieve it with a custom index:

Based on the [docs](https://github.com/typeorm/typeorm/blob/master/docs/indices.md#disabling-synchronization), I think we have to do the following:

`user.entity.ts`:
```typescript
@Index("user_accounts_email_unique_idx", { synchronize: false })
class User...
```

Then, in the migration:
```typescript
    public async up(queryRunner: QueryRunner): Promise<void> {
        await queryRunner.query(`CREATE INDEX user_accounts_email_unique_idx ON "user_accounts" (lower("email"))`);
    }

    public async down(queryRunner: QueryRunner): Promise<void> {
        await queryRunner.query(`DROP INDEX user_accounts_email_unique_idx`);
    }
```

bk3c (approved):
LGTM

bk3c:
I spoke too soon and totally agree with @bencpeters's comment. 

@pbn4 Can you do another PR that just modifies the existing migration? I won't run the current anywhere real until that's in...

PR #660, opened 2020-09-14
Render program rules and required documents with Markdown

### LINKED ISSUE #659: Allow Markdown for listing.requiredDocuments and listing.programRules
These two fields are all the way at the bottom of the listing page, and still in regular <div> elements instead of anything that supports Markdown.

This is immediately pressing since they want to do a link for the Jones pilot, but it's a good idea regardless since both fields tend to be long-ish, and can use bullets, etc.

jaredcwhite:
Fixes #659


bk3c (approved):
Tested locally and LGTM.

PR #662, opened 2020-09-14
Fix email uniqueness for User entity

### LINKED ISSUE #656: User account email is not constrained to be unique

pbn4:
Fixes #656

### DIFF
--- a/backend/core/src/entity/user.entity.ts
+++ b/backend/core/src/entity/user.entity.ts
@@ -10,7 +10,7 @@ import {
 import { Application } from "./application.entity"
 
 @Entity({ name: "user_accounts" })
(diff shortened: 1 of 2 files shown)


bencpeters (approved):
LGTM. We could consider removing the `unique: true` declaration on the `email` field on the User entity, but it doesn't really hurt anything (technically having both constraints probably _slightly_ slows performance & increases storage size for the `user_accounts` table, but it's really not a big deal, and it at least expresses the semantic intent nicely by having that decorator on the field).

pbn4:
@bk3c reverted previous changes and I branched off later so there is no `unique:true` on email anymore.

PR #664, opened 2020-09-14
Fix missing email confirmation

software-project:


software-project:
@bk3c I pushed a temp fix for email confirmation. It's just a application ID as discussed on the call. Tomorrow morning I'll push another one to pass it from the frontend.

bk3c (approved):
Made one minor fix to the template but otherwise looks great.

PR #669, opened 2020-09-15
Order unit types in tables

### LINKED ISSUE #666: Tables need to show unit types in logical size order
See https://github.com/housingbayarea/bloom/issues/152 for screen shots.

This could have been introduced with #651, but it seems like it's been around longer since the units array is being returned from the database in non-deterministic order rather than the order that they were hard-coded in the JSON file(s).

We can either order by unitType, and then have an pre-defined array or enum that puts them in the right order, or I guess we could order by numBedrooms after changing studios from `null` to `0`.

@dominikx96 and @software-project -- I'm assigning this to both of you but hoping that you can coordinate and it'll only take one person to code up a solution.

software-project:
Resolves #666

### DIFF
--- a/backend/core/src/listings/listings.service.ts
+++ b/backend/core/src/listings/listings.service.ts
@@ -16,9 +16,16 @@ export enum ListingsResponseStatus {
 @Injectable()
 export class ListingsService {
   public async list(jsonpath?: string): Promise<any> {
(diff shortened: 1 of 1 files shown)


bk3c (approved):
Tested locally and LGTM

Issue #667, opened 2020-09-15
Show the proper confirmation number on the confirmation screen.


bk3c:
Right now `applications/review/confirmation.tsx` has a hardcoded confirmation number with a TODO to replace it with the real thing. We're at that time, so we should replace it with the ID (UUID, for now) of the application object returned from the submittal in `terms.tsx`.

The UUID is long and awkward so there's a plan to replace it with a 10 character hashID soon, but we can just use the UUID for now.

PR #670, opened 2020-09-15
Added real confirmation ID from BE

dominikx96:
Issue: #667


bk3c (approved):
Tested locally and LGTM

PR #675, opened 2020-09-15
reduce use of the word "lottery" when not needed

### LINKED ISSUE #671: View Application still uses Lottery Number at the top
In `applications/view.tsx` we still have the key `application.confirmation.lotteryNumber` which uses the phrase "Lottery Number" instead of the more generic "Confirmation Number" or "Confirmation Code". I think we want to use something more generic in all cases. @slowbot?

bk3c:
Fixes #671.

Also fixes a bunch of other cases where I don't think we need to use the word "lottery" explicitly.


slowbot (approved):
All of this looks good.

Issue #526, opened 2020-08-14
[RFC] Maintain front-end interfaces separately from backend interfaces


bk3c:
I'll preface this by saying that I'm not at all convinced what the right answer is here, but I think I'm seeing some different tradeoffs after varying attempts to consolidate to a single entity/interface scheme this week, so I wanted to open up the discussion.

TL;DR: As much as it would be great to have key objects defined in one and only one place, I think the definitions will diverge just enough that it may not be a good idea. There will be enough times when we want different fields, different types, or different structure on the front-end compared to the back-end that hacking it all to work from one universal source will be more trouble than it's worth.

Some examples that are on my mind so far:

- There are definitely fields that we want on the backend but probably want to exclude from the front-end entirely. This seems pretty straight-forward with the swagger tools in basic cases (e.g. passwordHash), but seems to become way more difficult with nested/complex objects. For example, listing.applications is certainly a useful field for the backend, probably helpful to have for Partners (though I wouldn't want to load them all by default), but almost certainly not a field we would ever want or use in the public app. I guess we could make it optional?

- It seems reasonable to keep many of our front-end interfaces flatter than we probably want on the back-end (eventually). For example, we might want to have a LeasingAgent entity on the back-end but keep things flat as we currently do for listings on the front. I think we could do this by having the front-end based on the DTO rather than internal backend entity, but then we have the constraints of the DTO schema (e.g. IDs for things we do want as deep objects rather than the objects themselves for the front-end, like application.listing)

In general using DTOs separate from entities seems like a fine pattern regardless, since they never need to be considered further than the data-loading hooks on the front-end side, and it's nice to be able to optimize them for that purpose (e.g. including a status field or other metadata). To me it's really just a question of whether or not we want to re-assemble things on the front end in the exact same way as either the DTOs or how we're doing the back-end entities, and I'm not so convinced any more.

I recognize that having things with similar names and representing the same entities in slightly different ways is a potential maintenance nightmare, so I'm not convinced we should definitely split things up either, but I think it's worth a discussion before we build out the rest of the entity object model in the next few sprints.

jaredcwhite:
@bk3c It does seem risky to start maintaining two different sets of interfaces for all the "same" objects. I wonder if any deviations between backend storage and public-facing API could perhaps be handled programmatically in some fashion (aka merge a nested object's interface into the top-level interface).

kathyccheng:
Closed

PR #676, opened 2020-09-16
MVP for Application Form Config

### LINKED ISSUE #588: add a way to configure application languages
Right now the buttons in `choose-language.tsx` are hard-coded, but we want them to be configurable based on what languages are actually available with translations. This could probably be stored in the config object (cf. #453), or we can have some other interim way to turn languages off and on until that gets more sorted out.

jaredcwhite:
Closes #588 #453 #596


jaredcwhite:
Should come as no surprise that this PR is pretty massive, so we might want to do a quick walkthrough via Zoom or something to discuss the approach.

bencpeters (changes requested):
Some inline comments as well, but a few big-picture things:

- I think the types can & should be improved a bit - in particular, I think having a type for a step config (at the very least a placeholder `Record<string, any>` that can be easily improved later) and an enumerated set of keys for the config object would be good.

- As I said when we were discussing #596, I don't think we want to start a mini DSL that encodes a bunch of logic (like the skips here) into config files. I was thinking about ways to easily refactor this to remove that part of the config without needing a _ton_ of changes and came up with the following idea. At the top level, the config object defines an in-order list of steps that the form can render. Optionally, each step has an additional config section of its definition that is unconstrained in the "generic" version (e.g. can contain arbitrary keys/data). Each step has (or can have - maybe most of them can get away with a single generic version?) a `Definition` class that `implements StepDefinition` but can perform arbitrary logic to decide whether to render or go to the next step based on the config object and the current state of the application. The config object that goes along with this step gives us the ability to make it optional. In pseudo code:

```typescript
// Config object (eventually served by the backend)
steps: [
  {
    // These could be keys (e.g. "chooseLanguage"), or formatted titles as they are here.
    // Keys are probably better from a maintainability & localization standpoint though.
    name: "chooseLanguage",
  },
  ...
  {
    name: "alternateContactInfo",
    // In the actual implementation this config is probably unnecessary - the logic would just be assumed in
    // AlternateContactInfoStep class, but showing it anyway for illustration purposes.
    config: {
      requireAlternateDefined: true,
    },
  },
]

// Then, the front end provides a mapping between form step "names" and URLs & Definition classes:
const stepTypes = {
  chooseLanguage: {
    definition: StepDefinition,
    // TODO: localize this?
    title: "Choose Language",
    // URLs should be strictly the responsibility of the front-end, the backend shouldn't need to know or
    // care about how the URL mapping works IMO
    url: "/applications/start/choose-language",
  },
  alternateContactInfo: {
    definition: AlternateContactInfoStep,
    title: "Alternate Contact Info",
    url: "/applications/household/live-alone",
  },
}

// Finally, we have the implementation of the definition classes:
// This class is the "standard" one if special behavior isn't needed
class StepDefinition {
  conductor: ApplicationConductor
  step: StepConfig

  constructor(step: StepConfig, conductor: ApplicationConductor) {
    this.step = step
    this.conductor = conductor
  }

  get url(): string {
    return this.step.url
  }

  get skipStep(): boolean {
    // In this default implementation, the step is never skipped, but a step definition that inherits from this
    // might override this.
    return false
  }
}

class AlternateContactInfoStep extends StepDefinition {
  get skipStep(): boolean {
    return this.conductor.application.alternateContact.type !== "noContact"
  }
}
```

I don't think the individual step objects need to or should know anything about "next"/"prev" URLs - that's a property of the wider context (e.g. the order of steps), so I think it properly belongs to the conductor exclusively. I'd probably implement that by adding `nextUrl` and `prevUrl` helpers on the conductor that take the current StepDefinition as an argument and move through the ordered array of steps checking `definition.skipStep` on each until a suitable match is found.

Does that make sense?

jaredcwhite:
> * As I said when we were discussing #596, I don't think we want to start a mini DSL that encodes a bunch of logic (like the skips here) into config files.

My impression is that we *did* want that — aka, we want the steps to happen in (theoretically) any order with any number of them skipped arbitrarily according to a variety of criteria. The criteria itself may require a code change in order to add a new condition check (aka resolver), but hopefully once we have a bunch of them in the system we won't need to add new ones very often.

So encoding each step as an object on the frontend which determines what to do or where to go next wouldn't accommodate that flexibility from the backend. @bk3c am I getting that right, or do we want to lock things down a bit more like @bencpeters is suggesting?

bencpeters:
> My impression is that we _did_ want that — aka, we want the steps to happen in (theoretically) any order with any number of them skipped arbitrarily according to a variety of criteria. The criteria itself may require a code change in order to add a new condition check (aka resolver), but hopefully once we have a bunch of them in the system we won't need to add new ones very often.
> 
> So encoding each step as an object on the frontend which determines what to do or where to go next wouldn't accommodate that flexibility from the backend. @bk3c am I getting that right, or do we want to lock things down a bit more like @bencpeters is suggesting?

I think you're misunderstanding what I'm saying. I'm not advocating for the order to be strictly set by the front end, I'm trying to avoid basically creating a mini-DSL-in-config-files with `skipIf` & conditional field-type logic. If you look at my refactor suggestions I'm still proposing that the order of steps be strictly set by the config object (the order in which steps are specified is significant), it's just that the actual implementation details of what conditions to use for each step are defined in Javascript, with the actual step definition (e.g. in the front end) rather than creating an awkward DSL to try to encode the logic into a JSON config.

Basically:

- The order of steps is set by the order they are defined in the "master config" (that eventually will come from the backend)
- Any step-specific config options that are needed can also be defined in this config
- The actual business logic that the step needs to evaluate is defined by the front end, in Typescript.
- The URLs are considered an "implementation detail" and live with the front end, so the backend doesn't have to know/care about them.

Does that make sense?

jaredcwhite:
I understand what you're saying, and I think that still represents a shift from the original goal because in your example, the logic to skip a particular step is contained within a frontend JS object (your `AlternateContactInfoStep` example), rather than derived from the listing-specific setup saved in the backend. There's no way to switch to skipping some other step if needed (maybe Vouchers should be skipped if there's no alternate contact, for whatever reason).

james-wills-sf:
I agree with @bencpeters 's high level goals completely, I think the split between backend config, frontend config, and step implementation makes a lot of sense.

One thing I don't totally understand is why things like title and URL are defined in the Frontend config and not the frontend definition? Seems like the config could be replaced with a simple map from backend key to frontend step definition and the definition contains skipStep, title, url methods.

Another question, in Ben's version where does the actual render happen? It also seems like it makes sense for the step definition to have a function to return the step node

About @jaredcwhite 's statement:

> There's no way to switch to skipping some other step if needed (maybe Vouchers should be skipped if there's no alternate contact, for whatever reason).

I don't understand why that needs to be the case, couldn't the skip step implementation on VoucherStep definition take the current application state data and make its decision based on whether the alternate contact exists or not?

apps/public-reference/layouts/application.tsx
```diff
@@ -12,10 +12,19 @@ import {
...
+import { AppSubmissionContext } from "../lib/AppSubmissionContext"
+import { resolversLibrary } from "../lib/resolversLibrary"
 
 const Layout = (props) => {
-  const { profile, signOut } = useContext(UserContext)
+  const userContext = useContext(UserContext)
+  const { profile, signOut } = userContext
+  const applicationContext = useContext(AppSubmissionContext)
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
james-wills-sf: Are you all using eslint? Looks like issues like this could be caught (and autofixed!) with `react-hooks/exhaustive-deps`: https://www.npmjs.com/package/eslint-plugin-react-hooks#advanced-configuration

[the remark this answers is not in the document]
bencpeters: We are, but unfortunately warnings are not "enforced" in the pre-commit hooks, so if you don't manually run eslint checks, it's easy to miss.

bencpeters:
> One thing I don't totally understand is why things like title and URL are defined in the Frontend config and not the frontend definition? Seems like the config could be replaced with a simple map from backend key to frontend step definition and the definition contains skipStep, title, url methods.

They certainly could! I think my rationale there was that we don't necessarily _need_ a separate `StepDefinition` class for each type of step, if the "generic" logic would work fine (basically we only need to extend that class when custom logic is needed. For a "standard" non-skippable step I was imagining it being simpler to just have a single generic `StepDefinition` class customized by the config options rather than needing a custom class for each one. But if you don't like that convention, feel free to change that.

bencpeters:
> Another question, in Ben's version where does the actual render happen? It also seems like it makes sense for the step definition to have a function to return the step node

I think I was still imagining that Next.js would still take care of this with URL mappings. Basically the conductor takes care of loading the config from the backend, instantiating the `StepDefinition` objects, and figuring out what the `next`/`prev` urls are, but the actual rendering could still use Next.js routing.

jaredcwhite:
@bencpeters @james-wills-sf @bk3c Based on discussions regarding the feedback from Ben P, I've refactored the refactor, so now the individual step definitions are fairly self-contained with their frontend objects and the "backend" is only supplying the order. (The backend _could_ also supply config values as well, like min/max numbers or booleans or whatever, but we didn't seem to need that for this first pass.)

The other nice thing about this is we can consolidate form saving logic into the step definitions as well. I did some of that here, but left a few more complicated pieces as is. Probably best to save those for followup PRs.

Also added more TS types for the object graph here. `application` should also be typed, but that also feels like a followup PR (and is somewhat dependent on philosophical choices captured in issue #526).

BTW, the form "names" which get matched between backend and frontend may look like user-facing titles, but they're just labels for code comprehension. I'm on the fence if we should convert them to lower camel case keys or keep them as is.

Anyway, let me know what you think about this updated approach.

apps/public-reference/pages/applications/contact/alternate-contact-name.tsx
```diff
@@ -16,12 +16,16 @@ import {
...
 
 export default () => {
   const { conductor, application, listing } = useContext(AppSubmissionContext)
-  const currentPageStep = 1
+  const currentPageSection = 1
+
+  conductor.stepTo("Alternate Contact Name")
+  useEffect(() => {
+    conductor.skipCurrentStepIfNeeded()
+  }, [conductor])
```
bencpeters, line 28: Maybe combine this into a custom hook so that we don't have to add this manually to every form component that might be optionally skipped?

```typescript
function useFormConductor(stepName: string) {
  const submissionContextValues = useContext(AppSubmissionContext)
  
  conductor.stepTo(stepName)
  
  useEffect(() => {
    conductor.skipCurrentStepIfNeeded()
  }, [conductor])
  return submissionContextValues
}

// in the component
const { conductor, application, listing } = useFormConductor("Alternate Contact Name")
```

I know it'll be a no-op for most components (e.g. the non-optional ones), but I think it will be potentially error prone to have to remember to add this effect by hand all the time to get automatic skipping to work.

apps/public-reference/pages/applications/contact/alternate-contact-type.tsx
```diff
@@ -31,13 +36,10 @@ export default () => {
     application.alternateContact.type = data.type
     application.alternateContact.otherType = data.otherType
 
-    conductor.completeStep(1)
+    if (data.type === "noContact") conductor.completeSection(1)
```
bencpeters, line 39: I'm not 100% sure about our style guide on this matter, but most JS projects I've worked on don't like single line/block-less `if` statements - since whitespace isn't significant in JS, it's too potentially error prone. I know you're coming (mostly) from Ruby, so I understand the temptation for nice concise code ;-) tho...

apps/public-reference/pages/applications/household/live-alone.tsx
```diff
@@ -94,7 +83,7 @@ export default () => {
                 big={true}
                 className="w-full md:w-3/4"
                 onClick={() => {
-                  nextPageUrl = "/applications/household/members-info"
+                  if (application.householdSize == 1) application.householdSize = 0
```
bencpeters, line 86: `===` (also, potentially use a block around `if`)

bencpeters (approved):
This looks great overall, nice refactor! A few minor clean-up suggestions inline.

I'd also probably personally vote for changing the section "keys" to `camelCase` (or potentially another "code-friendly" format like `snake-case`) - it allows property access more easily (`stage.stageName`) if that ever becomes relevant to us, and it makes it abundantly clear to any developers down the line that they're meant as keys, not titles - my worry with leaving them as "human formatted" strings is that it's easy for someone else down the line to "cheat" and start using them as titles in the views, even though we're not intending them that way now...

apps/public-reference/lib/ApplicationConductor.ts
```diff
@@ -28,25 +33,127 @@ export const loadSavedListing = () => {
...
+
+  get listing() {
+    return this._listing
+  }
+
+  set config(newConfig) {
+    this._config = newConfig
+    this.listing.applicationConfig = { ...newConfig }
+    this.steps = this._config.steps.map((step) => {
+      const route = this.constructor["routes"][step.name] as StepRoute
```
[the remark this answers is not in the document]
jaredcwhite: I had to use this odd syntax to work around a TS typing error. I don't have the link handy but apparently there's an issue accessing properties off the constructor (aka class properties). Doing `this.constructor.routes` instead throws an error: "Property 'routes' does not exist on type 'Function'". Doing it this way provides the proper type to `route`.

apps/public-reference/pages/applications/contact/alternate-contact-type.tsx
jaredcwhite: I don't think we have a formal style guide other than to rely on the Prettier config. Did a quick regex search and there's one-line if statements a number of other places as well. I understand the rationale for wanting to standardize around multi-line blocks, but I would say I'm in favor of both options. (And yes, in Ruby the conditional is added to the end of the line where it belongs. 😉)

jaredcwhite:
Let's go with camelCase keys then. Will work on that and a couple of other tidbits and then I think we'll be in good shape for GTM…

jaredcwhite:
All right, third time's a charm? I like how much we were able to clean up the individual form pages with the latest config and the `useFormConductor` hook.

apps/public-reference/pages/applications/contact/alternate-contact-contact.tsx
```diff
@@ -46,19 +46,14 @@ export default () => {
     <FormsLayout>
       <FormCard header={listing?.name}>
         <ProgressNav
```
james-wills-sf, line 48: Could the progress nav, useFormConductor, back/next buttons etc all live in a shared applicationstep container component and the actual view rendering logic lives in unique components for each step? The only challenge I don't have an answer to right now is where the data setting logic in onSubmit should live...

Just thinking about how difficult it will be to test each of these step views, since we're mixing state connection and display logic

apps/public-reference/src/forms/applications/FormBackLink.tsx
```diff
@@ -0,0 +1,17 @@
+import Link from "next/link"
+import { OnClientSide, t } from "@bloom-housing/ui-components"
+import ApplicationConductor from "../../../lib/ApplicationConductor"
+
+const FormBackLink = (props: { conductor: ApplicationConductor }) => (
+  <p className="form-card__back">
+    {OnClientSide() && props.conductor.config.steps.length > 1 && (
```
james-wills-sf, line 7: How come `props.conductor.config.steps.length > 1` is the conditional here? Isn't steps the whole list, so if you're on the first step of a list of 5 steps, this would show the back button which is not what we want, right?

Could you just do `conductor.determinePreviousUrl() && ...`?

apps/public-reference/src/forms/applications/FormBackLink.tsx
```diff
@@ -0,0 +1,17 @@
+import Link from "next/link"
+import { OnClientSide, t } from "@bloom-housing/ui-components"
+import ApplicationConductor from "../../../lib/ApplicationConductor"
+
+const FormBackLink = (props: { conductor: ApplicationConductor }) => (
```
james-wills-sf, line 5: nit: To make this easier to test you might want to just pass in `backUrl` instead of the whole conductor.

james-wills-sf (approved):
@jaredcwhite This looks really good! Sorry for being unresponsive, context switching is hard.

I added a few non-blocking comments, up to you whether you want to make the changes or not

apps/public-reference/pages/applications/contact/alternate-contact-contact.tsx
jaredcwhite: Yeah it's been an uphill battle working on DRYing up the various form pages, but I think once we have the conductor step stuff fully in place, we can take most of the form saving logic out of there as well as beef up a dedicated shared layout (so for example the back link is pulled in once). Probably a bit much to chew off for this PR though.

apps/public-reference/src/forms/applications/FormBackLink.tsx
jaredcwhite: Good call. :+1:

jaredcwhite:
All right, I fixed test failures and got in changes addressing James' feedback. Shall we merge? 😱

jaredcwhite:
@bk3c Yay! One question: do we have an issue yet for serving the config from the backend and removing our fixture?

bk3c:
@jaredcwhite On my list to write up today

apps/public-reference/pages/applications/contact/alternate-contact-contact.tsx
james-wills-sf: Sounds good to me!

PR #677, opened 2020-09-16
Fix email confirmation when there is no email

software-project:
Fixes a bug that @pbn4 found. While applying with no email, email confirmation was throwing an error.


bk3c (approved):
LGTM

bk3c (changes requested):
Sorry, I lied. It still looks good to me code-wise and locally, but it also looks like tests are failing, at least in part to the alteration. Can you fix before we merge?

bk3c (approved):
Ok, tests passing now, thanks!

Issue #640, opened 2020-09-11
Phone Type should be Capitalized 


kathyccheng:
![Screen Shot 2020-09-10 at 9 12 37 PM](https://user-images.githubusercontent.com/3453407/92858458-efb7a000-f3aa-11ea-9b9d-8fe2a4720318.png)

On the review information screen, when the phone type is "Work" or "Cell" it shows as "work" or "cell" vs "Home" below

bk3c:
I defer to @slowbot if he'd prefer something else, but I'd expect that either they should either be capitalized or all be lowercase

slowbot:
@bk3c @kathyccheng @dominikx96 They should all be Capitalized

dominikx96:
Should be fixed now.

I also noticed that in the summary step is information about alternate contact phone type. Now it's applicant phone type, not alternate contact. In the alternate contact step is only the ability to put the phone number, without type.
Should we add <select> with phone type for "alternate contact"?
I temporarily removed this information from the summary card.

<img width="498" alt="Image 2020-09-18 at 10 00 56 AM" src="https://user-images.githubusercontent.com/16655908/93573278-36069380-f997-11ea-97ca-f08b224998e9.png">
<img width="524" alt="Image 2020-09-18 at 10 00 37 AM" src="https://user-images.githubusercontent.com/16655908/93573287-37d05700-f997-11ea-8546-e8561c6887e3.png">

slowbot:
@dominikx96 Let's not add a phone type to Alternate Contact Phone and lets remove it from the summary card

PR #683, opened 2020-09-18
Fix/phone type capitalized

dominikx96:
Issue: #640


slowbot (approved):
See my update. @dominikx96

PR #686, opened 2020-09-18
Add separate cleanup locale JSON for future integration

jaredcwhite:
We'll have to do a diff with the current file and update the entire app/components which is a Big Deal, so this can live in the repo separately at first.

### DIFF
--- a/shared/ui-components/src/locales/general_cleanup.json
+++ b/shared/ui-components/src/locales/general_cleanup.json
@@ -1,0 +1,724 @@
+{
+  "account": {
+    "accountSettings": "Account Settings",
+    "accountSettingsSubtitle": "Account Settings, email and password",
+    "createAccount": "Create Account",
+    "haveAnAccount": "Already have an account?",
+    "myApplications": "My Applications",
+    "myApplicationsSubtitle": "See lottery dates and listings for properties for which you've applied"
+  },
+  "application": {
+    "form": {
+      "general": {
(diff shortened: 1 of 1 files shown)


kathyccheng:
LGTM

kathyccheng:
@bk3c @jaredcwhite 

I did a review pass to make sure all of the keys from the old file are in the new one. I think they're all there, I just have one clarification: 
* firstnameFormPlaceholder, lastnameFormPlaceholder, [street/city/state/zip]FormPlaceholder's are all no longer explicitly there, but I assume you can use firstName, lastName, street, city, state, or zip as those placeholders instead 

And then I found a typo in race options, "Other/Mutliracial" instead of "Other/Multiracial" - should I file an issue to fix it currently? Can we just fix it in this file for the future? :)

bk3c:
I think this generally looks good to me in terms of organization, but I'd like to see if we can do a quick pass to remove keys that are unused anywhere before merging it in. I think there are some that were copied over from DAHLIA or some previous development version, but don't appear to be used in current Bloom at all. For example, both keys in the `lottery` section appear to be completely unused, so I think we can get rid of that whole thing.

Would there be an easy way to script this check? If so, I think it could have some use ongoing. If it's too big of a lift, then some combination of @kathyccheng and I can try and do a manual pass...

bk3c (approved):
Still have ambitions of some greater re-organization of some of the keys, but this is much better than before.

Issue #685, opened 2020-09-18
Additional household members address.state is being stored as "string"


bk3c:
Indeed we want the value to be stored as *a* string, but in this case the value is being stored as the string literal "string" where the other address fields (e.g. city, street) are blank.

Not sure why this is happening, but we obviously want to avoid it.

PR #700, opened 2020-09-21
Removed default value for member state

dominikx96:
Issue: #685 

Just removed default value "string" for the address.state.


bk3c (approved):
LGTM. I didn't realize it was this straight-forward.

PR #706, opened 2020-09-22
Add rentalAssistance field to listings

### LINKED ISSUE #682: Allow listing-specific text for rentalAssistanceSubtitle
Right now, rentalAssistanceSubtitle can't be customized per listing, but it should work like other fields where we use the generic text from `general.json` if there's no matching field on listing, but listings can define the field to have their own specific text.

As an aside, the naming of all of these fields with "Subtitle" seems a little inaccurate, since it's basically the body. Not strictly necessary, but we might want to rethink the names.

software-project:
Resolves #682


bk3c (approved):
Tested locally and LGTM.

Issue #679, opened 2020-09-17
Don't show Required Documents panel on Listing if no data


bk3c:
listing.programRules is correctly hidden if empty, but we need the same behavior for listings.requiredDocuments. Should be a trivial change.

In the unlikely event that both are empty (but it looks like it might happen for Monarch), we should probably hide the whole Additional Information section.

cc: @slowbot, in case you think we should force any of this to be shown.

PR #711, opened 2020-09-23
Hide section when requiredDocuments or programRules were not passed

dominikx96:
Issue: #679


bk3c (approved):
Tested locally and LGTM.

See also #716

PR #713, opened 2020-09-23
Order preferences by ordinal

### LINKED ISSUE #681: Store preference order as a number rather than an ordinal string
We're currently storing preference order as a string with an ordinal in it, but that makes it hard to sort for display and will make it more difficult when we need to actually evaluate preferences when running a lottery in the future. We should switch over to storing the preference as a number and just convert that to an ordinal with an easy JS function. (e.g. https://stackoverflow.com/questions/13627308/add-st-nd-rd-and-th-ordinal-suffix-to-a-number)

I realize that this strategy will break for languages other than english, so I filed #680 to track that case separately, but I think we should at least fix the base case now.

As part of this fix, we should also update the query in `backend/core/src/listings/listings.service.ts` to sort preferences appropriately on load.

software-project:
Resolves #681

### DIFF
--- a/backend/core/client.ts
+++ b/backend/core/client.ts
@@ -259,7 +259,7 @@ export class ApplicationsService {
   list(
     params: {
       /**  */
-      listingId?: string;
+      listingId: string;
     } = {} as any,
     options: IRequestOptions = {}
   ): Promise<Application[]> {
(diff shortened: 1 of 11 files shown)


bk3c (approved):
Tested locally and LGTM. Thanks for updating storyshots as well and making sure tests pass.

bk3c:
Note: we might want to consider renaming the column to something more descriptive like list_order, since ordinal is technically inaccurate now, but we can always do that later if needed.

PR #714, opened 2020-09-23
Fix listing rows

slowbot:
- Adding container to fix collapsing listing rows with no image present
- Adding background color and aspect ratio to fill image container to prevent collapsing image container


bk3c (approved):
LGTM

PR #715, opened 2020-09-24
Add Authorization Checks to API Endpoints

bencpeters:
This PR adds authorization checks to the following API endpoints:

- /assets (admin only for all)
- /preferences (admin only for all)
- /applicationMethods (admin only for all)
- /units (admin only for all)
- /listings (admin only for update/create/delete, anyone can read)


pbn4 (approved):
LGTM

bk3c (changes requested):
This looks good to me from a code-read perspective, but it's causing an infinite loop when running `yarn dev:all` locally, which is making it tough to test. It's looping on a HEAD request to /listings:
```
BACKEND_CORE [Nest] 25324   - 10/09/2020, 12:30:20 AM   [HEAD] 2020-10-09T07:30:20.299Z: /listings +24ms
BACKEND_CORE [Nest] 25324   - 10/09/2020, 12:30:20 AM   [HEAD] 2020-10-09T07:30:20.527Z: /listings +228ms
```

So maybe I'm thinking that we need to define that in addition to GET somewhere? that explanation doesn't entirely make sense, since master works fine without any explicit HEAD definition, so maybe I'm missing something else entirely?

bk3c:
@pbn4 Can you merge master in here to resolve the conflicts?

PR #719, opened 2020-09-24
Fix min max function

software-project:
Fix for https://github.com/housingbayarea/bloom/issues/160


bk3c (approved):
Tested locally and LGTM.

PR #723, opened 2020-09-25
Fix undefined join on string[] in csv builder

pbn4:
Fixes a 500 on CSV applications endpoint when application is a malformed JSON

### DIFF
--- a/backend/core/src/services/application-formatting-metadata.ts
+++ b/backend/core/src/services/application-formatting-metadata.ts
@@ -2,9 +2,10 @@ import {
   booleanFormatter,
   defaultFormatter,
   dobFormatter,
-  FormattingMetadataAggregateFactory, joinArrayFormatter,
+  FormattingMetadataAggregateFactory,
+  joinArrayFormatter,
   keysToJoinedStringFormatter,
-  streetFormatter
+  streetFormatter,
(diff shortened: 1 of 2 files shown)


backend/core/src/services/application-formatting-metadata.ts
```diff
@@ -2,10 +2,9 @@ import {
   booleanFormatter,
   defaultFormatter,
   dobFormatter,
-  FormattingMetadataAggregate,
-  FormattingMetadataAggregateFactory,
+  FormattingMetadataAggregateFactory, joinArrayFormatter,
```
bencpeters, line 5: Pedantic, but this dependency should be on its own line. The Prettier ESLint config should have taken care of that automatically, are you running pre-commit hooks?

bencpeters (approved):
LGTM, pending a small formatting update.

backend/core/src/services/application-formatting-metadata.ts
bk3c: Indeed, the linter was failing in CI too. I went ahead and fixed it so we can merge, but @pbn4, if you can see if you can get things working in your local environment (and feel free to make a PR with config changes, if needed), that would be appreciated.

Issue #609, opened 2020-09-03
Add pagination to /applications and /listings


pbn4:

kathyccheng:
Closed in #724

Issue #608, opened 2020-09-03
Make /applications (list) return only applications for a given auth 


pbn4:
As stated there is a need for this endpoint to return only user's applications. Now this endpoints returns the entire table.

software-project:
@pbn4 I think we need both. User's applications for user facing apps and full list for partners.

pbn4:
OK understood. I figured we also need https://github.com/bloom-housing/bloom/issues/609 for a case when we return the entire table.

kathyccheng:
Closed in #724

PR #724, opened 2020-09-25
Add NestJSx/crud to /applications /assets /units /applicationMethods

pbn4:
Closes https://github.com/bloom-housing/bloom/issues/739 https://github.com/bloom-housing/bloom/issues/608 https://github.com/bloom-housing/bloom/issues/609

In addition to removing duplication in mentioned routes I also introduced proof of concept  `ABACTypeOrmCrudService` (which is my extension to nestjsx/crud's `TypeOrmCrudService`) in order to abstract away ABAC stuff in the future for `/applications` route. It's not in use at the moment since all the routes modified are RBAC only (they are for admins only and admins can do whatever they want there, also there is no concept of ownership in those). 

@bencpeters I came up with an idea that we could get rid of `list_all` authzAction in case where we want listing to work for both admins and normal users by doing the following: 
1. Make frontend client responsible for specifying appropriate filters in the request e.g. userId == loggedInUserId (nestjsx/crud gives you nice out of the box filters exposed in the controller so you can filter by any entity prop on the frontend client)
2. Fetch from the DB using specified filters 
3. Apply AuthzService.can check separately for each item, if any triggers an exception then return 403. 

One thing I have not tested yet is whether ABACTypeOrmCrudService will work smoothly with `CrudRequest` being used with getOne in e.g updateOne method handler.


pbn4:
Superseded by https://github.com/bloom-housing/bloom/pull/744

PR #728, opened 2020-09-27
adding tab index to signed in state of navigation for accessibility focus

slowbot:
quick focus fix for keyboard accessibility


jaredcwhite (approved):
Looks good — just want to make a note though that technically we don't need to put the `0` in as an interpolated variable (i.e. `tabIndex={0}`), but since we'll likely want to have `tabIndex` values be dynamic in various components going forward, it's find to start using that pattern IMO.

PR #731, opened 2020-09-28
Feature/application form tests

dominikx96:
Hey @jaredcwhite @bk3c 

I created the first form test. I'd like to go further after your feedback.
Our form is veeeery long, so probably the better way will be to test for each form page, but...

- If each step is tested separately, the context is empty

What do you think about test context (if contains step values) in each step?
Probably to test the last steps like summary etc. we'll have to update context with dummy data in this approach.

- When the form is filled correctly, probably we should test redirect to the next step (check path after redirection), but we don't know exactly what form is next (not yet, but after form config implementation - yes)

I've got an idea to store config in the separated JSON file and just use values from the JSON file to test if after redirection form is exactly on the next step defined in config. What do you think or do you have other ideas?

- I also not sure if we want to test exact texts e.g errors - or just test if the form includes an error HTML element without checking the text?


apps/public-reference/cypress/integration/pages/application/contact/name.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  it("Should display initial form errors", function () {
+    // try to trigger form
+    cy.get("button").contains("Next").click()
+
+    // check errors
+    cy.get('span[id="applicant.firstName-error"]').contains("Please enter a First Name")
+    cy.get('span[id="applicant.lastName-error"]').contains("Please enter a Last Name")
+    cy.get(".error-message").contains("Please enter a valid Date of Birth")
+    cy.get('span[id="applicant.emailAddress-error"]').contains("Please enter an email address")
```
bencpeters, line 16: This seems pretty tightly coupled to the implementation to me (grabbing a `span` by the `id`) - if we change what the errors look like (either the format of the ID or the HTML element), these tests will fail. Maybe a better way would just be to have a selector for the root error block, and then search for the test within child elements? We could also make the text slightly less specific (e.g. just search for "First Name" instead of the full string in case we want to update the exact text later, but I don't have a strong feeling about that either way.

apps/public-reference/cypress/integration/pages/application/contact/name.ts
```diff
@@ -0,0 +1,50 @@
+describe("Form contact/name", function () {
+  it("Renders the /contact/name form", function () {
+    cy.visit("/applications/contact/name")
+    cy.get("form").should("be.visible")
+    cy.location("pathname").should("include", "/applications/contact/name")
+  })
+
+  it("Should display initial form errors", function () {
+    // try to trigger form
```
bencpeters, line 9: These tests are tightly coupled to the order in which they're defined inside this file - this isn't ideal, since it means that a test will fail if you try to run it individually, or if someone unwittingly modifies the order down the road (e.g. removes this first test).

A better way to structure this would be to make sure that each test ensures a clean slate - either start it with a `cy.visit` call, or potentially add a [`beforeEach`](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests.html#Hooks) hook to this `describe` block.

bencpeters (commented):
A few specific comments about the cypress code.

Overall, I think it's probably a good idea to specify a JSON config with test fixtures here rather than needing to do a bunch of verbose cypress commands for every single form field in the whole form. I think something like this might work:

```
{
  fields: [
    { name: "firstName", required: true/false, value: "Name" },
    ....
  ]
}
```
Then use that to generate the "should save form values" and "initial form errors" tests. "Custom" test logic more complicated than that (like the incorrect email address example in your code) should probably still exist in Cypress code rather than requiring super complicated JSON definitions though.

Does that make sense to everyone?

shared/ui-components/src/forms/Form.tsx
```diff
@@ -13,7 +13,7 @@ const Form = ({ id, children, className, onSubmit }: FormProps) => {
   }
 
   return (
-    <form id={id} className={className} onSubmit={onSubmit} onKeyPress={onKeyPress}>
+    <form id={id} className={className} onSubmit={onSubmit} onKeyPress={onKeyPress} noValidate>
```
jaredcwhite, line 16: I think I know why this is here, but I want to make sure this gets vetted by @slowbot.

apps/public-reference/cypress/integration/pages/application/contact/name.ts
jaredcwhite: If the `onBeforeLoad` callback for `cy.visit` could be used to set an initial state of the application data via session storage, maybe that would work?

jaredcwhite (changes requested):

apps/public-reference/cypress/integration/pages/application/contact/name.ts
jaredcwhite: Strange, I posted a comment and GitHub showed it twice, then I deleted one and both disappeared. Anyway…I was saying I think we could maybe leverage the `onBeforeLoad` callback for `cy.visit` so that session storage application data could be set for the context.

dominikx96: @bencpeters **beforeEach** sounds good to me, good idea!

shared/ui-components/src/forms/Form.tsx
dominikx96: Yea, it's definitely a thing we need to re-test whole the entire form. Not sure where, but @slowbot added info about this, and it's reasonable for me - e.g. custom phone validation doesn't work when noValidate attr is missing.

dominikx96:
@bencpeters @jaredcwhite thanks for your feedback! I've got few comments:

> Overall, I think it's probably a good idea to specify a JSON config with test fixtures here rather than needing to do a bunch of verbose cypress commands for every single form field in the whole form. I think something like this might work:
> 
> ```
> {
>   fields: [
>     { name: "firstName", required: true/false, value: "Name" },
>     ....
>   ]
> }
> ```
> 
> Then use that to generate the "should save form values" and "initial form errors" tests. "Custom" test logic more complicated than that (like the incorrect email address example in your code) should probably still exist in Cypress code rather than requiring super complicated JSON definitions though.
> 
> Does that make sense to everyone?

I don't think it helps us and be more intuitive than the current approach because in many cases we'll define custom rules.
If in the future we may want to update tests (change fields etc.) field definitions in one place are more clear for me.
I think we're talking about 2 other things.

By saying "JSON config" I mean "MVP for Application Form Config". So, if this config with form steps definition will be stored as JSON, we can use it in our test and after use:

`cy.get("button").contains("Next").click()`

we can check if the path changes correctly (because by using form config we know what's a next step).

Is it reasonable for you @bencpeters @jaredcwhite ?

apps/public-reference/cypress/integration/pages/application/contact/name.ts
dominikx96: Agree, I added HTML selectors, because I had a weird error... but I noticed it occurs because of ID's include dots.
I created 2 support commands to get element by ID & to click the next button - it should be better.

I think e.g. _applicant.lastName-error_ (generated by component automatically) is secure - I don't see reason, why we'll change it, but even - if we may want to change this behavior, we just can create prop for the error message ID in the future.

The best approach will be to create custom data- attribute, but it obliges us to update the current form so probably we don't want it.

I don't love the approach with search partial string, for me test should test the exact string or just test element visibility.

jaredcwhite:
> By saying "JSON config" I mean "MVP for Application Form Config". So, if this config with form steps definition will be stored as JSON, we can use it in our test and after use:
> 
> `cy.get("button").contains("Next").click()`
> 
> we can check if the location.pathname changes correctly (because by using form config we know what's a next step).

Yeah, I think a copy of the config could be used as a fixture to guide the tests through the form flow. That's what I was trying to say in: https://github.com/bloom-housing/bloom/pull/731#discussion_r496326952

The config is going to be loaded from the backend soon, so in order for that *not* to happen in the Cypress tests, you'll have to make sure the config gets preloaded into sessionStorage so it gets used by the context.

dominikx96:
@jaredcwhite @bencpeters I made few updates:

- created fixture to store data mocks
- added helper commands
- added context value test to make sure if all data is saved correctly

I think we can go further with the next steps and after #676 just add one scenario.

jaredcwhite (approved):
That was a ton of work! Great job @dominikx96 :+1: LGTM

jaredcwhite:
I can't seem to find Cypress tests in our CircleCI setup. Was that temporarily disabled?

jaredcwhite:
Testing locally…it required me to add `DATABASE_URL=postgres://localhost:5432/bloom` to my `.env` file. Didn't seem to need that previously.

jaredcwhite:
I got just a handful of errors locally but I think it was due to test data differences in my DB…I'll assume it's all green in actuality. Most of the form stuff worked just fine.

Issue #694, opened 2020-09-20
Storybook: Reorganize Components


slowbot:
See [Component Best Practices Outline](https://docs.google.com/document/d/1jLtpbDGFv_Lq_BNS61T6i_XGdhKyoBWZJ_UaWdkbuk8/edit?usp=sharing)

- **actions**
   - button
   - expandable text
   - link button
   - localized link
- **elements**
   - meta tags
- **footers**
   - site footer
   - exygy footer (rename brand footer)
- **headers**
   - site header
   - page header
   - hero
- **icons**
   - header badge
   - icon
- **forms**
   - checkbox
   - dob field
   - field
   - form
   - household member form
   - phone field
   - radio
   - select 
- **lists**
   - preference list
- **navigation**
   - footer nav
   - user nav
   - progress nav
   - tab nav
- **notifications**
   - alerts
   - message
   - error message
   - application status
   - status note
- **tables**
   - standard
      - basic
      - grouped
      - plain
   - unit tables
- **overlays**
   - modal
   - drawer
- **sections**
   - info card grid (rename info card section)
   - footer section
   - responsive content list (rename responsive content section)
   - dash blocks (rename dash block section)
   - responsive wrappers
   - content section
   - list section
   - markdown section
- **blocks**
   - app status item
   - form card
   - image card
   - info card
   - dash block
   - housing counselor
- **text**
  - flag
  - tag
  - description

@jaredcwhite wanting to get my head around how me might reorganize our components. Let me know your thoughts.

jaredcwhite:
@slowbot Looks like a good list to me. Not sure about the term "structures", but I can't admit anything better comes to mind. Widgets? Doodads? Thingamajigs? 😉

slowbot:
Maybe singular "structure" comes from Polaris. There is always the generic "blocks"

jaredcwhite:
@slowbot Yeah, Blocks seems like a pretty good term to me. And per our recent chat about the basic "Box" concept, maybe we could aim for adding something like that which could be utilized in higher-level forms where it makes sense. Like maybe a card is just a more complex version of a box…

jaredcwhite:
Or maybe a box isn't a component per se but a mixin you can include into components like cards… sorry, thinking out loud here…

jaredcwhite:
[...]

PR #742, opened 2020-10-03
Component cleanup folders

jaredcwhite:
Addresses #694


slowbot (changes requested):
@jaredcwhite See my comments

jaredcwhite:
@slowbot I've made the requested changes. Let me know if that looks good now.

PR #744, opened 2020-10-07
Feature/abstract service

### LINKED ISSUE #739: Remove duplication from controllers
We use a lot of duplicated code for the controllers definitions in routes /assets, /applicationMethods, /units and /preferences. I figured we could utilize nestjsx/crud for those endpoint easily and remove all that duplication + introduce nice out of the box pagination, joining and filtering.

pbn4:
Closes #739 for services related code

### DIFF
--- a/backend/core/src/application-methods/application-method.service.ts
+++ b/backend/core/src/application-methods/application-method.service.ts
@@ -1,39 +1,19 @@
+import { InjectRepository } from "@nestjs/typeorm"
+import { AbstractService } from "../shared/abstract-service"
 import { ApplicationMethod } from "../entity/application-method.entity"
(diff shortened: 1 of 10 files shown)


bencpeters (approved):
I like this! I think this should be a good change to remove a bunch of boilerplate...

One thought: a common usecase for `list` is to narrow the set of parameters in some way. It might just be easiest to implement this by overriding `list` for services that need this functionality, but we could also potentially have an optional "query parameters" argument to the generic version of `list` that passes along arguments to `find` - this probably won't work for more complicated query conditions, but it might avoid needing to override in a subclass for simpler queries...

bk3c (approved):
Tested locally and LGTM too.

Issue #738, opened 2020-09-29
Listing Sidebar Updates


slowbot:
**Goal**: As a housing applicant I want to know the date and time of events specific to a listing.

**Acceptance Criteria**:

**Open Houses**
- If there is an open house display the open house block in the listing sidebar
- For each open house display a date, time and url
- Display additional helper text for sidebar block that provides text on COVID policy
- Display block under Application Deadline and before Waitlist Details

![image](https://user-images.githubusercontent.com/13915/94619108-245ba080-0261-11eb-861e-4f69deb2f6af.png)

**Lottery**
- If there is a lottery display the open house block in the listing sidebar
- Display a date, time and url
- Display additional helper text for sidebar block that provides text on COVID policy
- Hide Lottery block after the lottery is over
- Display block under How to Apply section and before What to Expect

![image](https://user-images.githubusercontent.com/13915/94619398-b4014f00-0261-11eb-92ea-9b75aa474e92.png)

slowbot:
@kathyccheng My questions have to do with the URL.
- Is this recording or a livestream?
- If a recording do we leave the link up after the event
- If a live stream do we only display during the event and hide before and after event

slowbot:
@kathyccheng I see the email now. Perhaps we can update the CTA to read "Join the event live"

kathyccheng:
@bk3c can you take a look and make sure back end requirements are captured here?

kathyccheng:
For Jones specifically, the dates are: 

Open House 10/8@3-4PM 
https://teams.microsoft.com/l/meetup-join/19%3ameeting_NDRkYjUxZjktMWI4OS00NWNiLWE3OTktNTczMDNiYzE2M2Mw%40thread.v2/0?context=%7b%22Tid%22%3a%2215cb6c53-0a50-4876-a66c-9a753d760a7d%22%2c%22Oid%22%3a%223c3aa088-23f2-46a2-ad24-54dcddcd66a8%22%7d

Lottery 10/22@2PM 
https://teams.microsoft.com/l/meetup-join/19%3ameeting_NmYzZTZlZmUtYmU5ZC00ZTJmLWFhZjAtYjRjNWViYjFhMDdi%40thread.v2/0?context=%7b%22Tid%22%3a%2215cb6c53-0a50-4876-a66c-9a753d760a7d%22%2c%22Oid%22%3a%223c3aa088-23f2-46a2-ad24-54dcddcd66a8%22%7d

[...]

PR #746, opened 2020-10-07
Feature/sidebar updates

pbn4:
Closes https://github.com/bloom-housing/bloom/issues/738


apps/public-reference/pages/listing.tsx
```diff
@@ -3,41 +3,87 @@ import ReactDOMServer from "react-dom/server"
...
-  t,
 } from "@bloom-housing/ui-components"
 import Layout from "../layouts/application"
+import moment from "moment"
 
 interface ListingProps {
   listing: Listing
 }
 
+function EventDateSection(props: { event: ListingEvent }) {
```
jaredcwhite, line 42: @pbn4 To keep things consistent with other parts of the codebase, can we use the `const EventDateSection = (…) => {` syntax instead? (And similar for the next two functions)

bk3c (commented):
I'm having a hard time testing it because there's a loop running `yarn dev:all` locally. After some looking, it turns out that seems to be a problem with the authz-endpoints branch that this is based on, not anything specific to this issue. Master still works fine for me locally, and it's not just my machine since @slowbot `was having the same problem when he tried to review the visuals. Maybe take a look at that to unblock this?

My main feedback on the code itself here is that I think we'll probably want events to eventually be a first class object (maybe called listing_events?) so that we can do things like generate a calendar of events for a jurisdiction without having to iterate through the listings themselves. I hadn't even thought to genericize these two fields the way you did, but now that I see it, I think it's a great idea. I'm a little worried that moving to an independent object might be too much to take on for now, so fine to push that to a later issue if you agree.

The other thing I think we need here is a test listing json that demonstrates how the event should be represented for import. I realize we haven't yet done a core version of the listing_data directory that's in housingbayarea, and I'm open to other suggestions if there's a better place/way to house test listings

pbn4:
@bk3c @slowbot 

> I'm having a hard time testing it because there's a loop running yarn dev:all locally. After some looking, it turns out that seems to be a problem with the authz-endpoints branch that this is based on, not anything specific to this issue. Master still works fine for me locally, and it's not just my machine since @slowbot `was having the same problem when he tried to review the visuals. Maybe take a look at that to unblock this?

This should be fix right now with changes to root package.json using `wait-on` with `http-get://` instead `http://`. 

> My main feedback on the code itself here is that I think we'll probably want events to eventually be a first class object (maybe called listing_events?) so that we can do things like generate a calendar of events for a jurisdiction without having to iterate through the listings themselves. I hadn't even thought to genericize these two fields the way you did, but now that I see it, I think it's a great idea. I'm a little worried that moving to an independent object might be too much to take on for now, so fine to push that to a later issue if you agree.

I introduced a separate entity and a controller for it. Now listing events are available under standard REST API same like other resources.

>The other thing I think we need here is a test listing json that demonstrates how the event should be represented for import. I realize we haven't yet done a core version of the listing_data directory that's in housingbayarea, and I'm open to other suggestions if there's a better place/way to house test listings

Our API is discoverable thanks to Swaggerunder `/docs`. Example payload are there. So in order to add listing event to a listing you need to first create it using /listingEvents POST method and then find a listing of interest with GET `/listings/{listingId}` and update it with PUT accordingly. All payloads are described in the swagger. Bear in mind that in this branch your user must have `isAdmin` in `user_accounts` table set to true.

PR #752, opened 2020-10-09
Dependency updates for October/2020

bk3c:
This updates node slightly in addition to all the normal dependency suspects


jaredcwhite (approved):
Poked around on the public app and Storybook…all seems well. :+1:

PR #755, opened 2020-10-09
Breadcrumb component

### LINKED ISSUE #690: Component: Breadcrumb
**Goal**: As a user I want to see secondary navigation that helps me understand where I am on a website.

**Acceptance Criteria:**
Organization:
- Locate ui-components/src/navigation
- Product: Partners

Engineering Notes: 
- Stub with default data
- Should accept min two items
- Current page should not be a link item

Props:
- label - link label, _string_
- url - link destination, _string_
- current - indicates current page on link item

Accessibility:
- Use ARIA markup for additional context. Use aria-label="Breadcrumbs" on the main element and aria-current="page" on the current page.

Markup Prototype:
https://storybook.bloom.exygy.dev/?path=/story/prototypes-breadcrumbs--default

jaredcwhite:
Closes #690

I also added aria prop to LocalizedLink, as well as removed some old prototypes we no longer need.


slowbot (changes requested):
@jaredcwhite Is it possible to render the current item in breadcrumb as a span rather than a link. I think it will be confusing accessibility wise to have a blank link. We should instead use a span with list element aria current 

see: https://designsystem.digital.gov/components/breadcrumb/

jaredcwhite:
@slowbot I looked into the span vs. a thing, and it looks like from the [W3C example I was referencing](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html) that the `aria-current="page"` goes on the anchor tag, not the li tag. I think it would link to the same page you're on, not be a blank link. But if you think we can put it on the li tag instead and use a span, I can change that.

Issue #758, opened 2020-10-13
Short-term Authz tests + follow-ups


bk3c:
Filing these as a separate issue since I don't think any of them are strictly blocking a merge of #715 at this point, but I'm in favor of getting these items addressed as soon as is reasonable:

- Users can't be created with isAdmin set to true. I think it just needs to be added to the CreateDTO, but maybe there's more than that?
- User creation endpoint should probably be moved from auth/register to user/create. Seems to be more consistent that way, and I think it would make it easier to have authz set up on the user methods themselves, especially once we have editing
- We need an endpoint to update user accounts. Right now it looks like SQL is the only option once the user has been created (or maybe I'm missing it?)
- Need to build out test coverage to make sure authz continues to operate correctly as we keep moving with development. I know it might seem cumbersome, but ideally I think this will include a test of every endpoint with a) no token, b) an invalid/expired token, and c) the wrong permissions (e.g. isAdmin = false, for now) to make sure things fail every time, plus at least some spot checks that things are working correctly when they should. Obviously the priority is on making sure we don't accidentally allow inappropriate use.

pbn4:
> Users can't be created with isAdmin set to true. I think it just needs to be added to the CreateDTO, but maybe there's more than that?

@bk3c Isn't that going to allow users to register as admins? I think we should not include that in CreateDto and instead allow admins to edit any user profile -> make others admins too.

bk3c:
@pbn4 sorry for the delay in response. My machine hard-crashed yesterday and it must have eaten my reply. I agree that we definitely don't want to let arbitrary users sign up with isAdmin = true, but I still think we need a way for admin users to set permissions at user create time. 

I guess it's fine if we want to have it be two API calls from the front-end one right after the other (create, and then immediately set permissions), so long as it's transparent to a Partners user. Either way, I do think we should probably move the endpoint to users/create, if that still makes sense to you.

Also note that we should maybe go ahead now and add a second permission called `isPartner` now, so that we can separate users who can log in to partners at all from people who are truly admins (e.g. Exygy team for now). Obviously permissions will get more complicated from there, but I expect both of those globals will probably continue for the foreseeable future.

pbn4:
@bk3c I moved user creation to user controller in #770 but this is a branch off from env configuration fix since it was a blocker for this change to be introduced. Could you please fix ENVs on build systems so that jest-backend test passes? (It's related to ENV validation)

PR #762, opened 2020-10-14
Fix env configuration and auth/user controller updates

pbn4:
Addresses https://github.com/bloom-housing/bloom/issues/758:

- user creation moved to UserController,
- configuration refactored - ConfigService introduced,
- user dtos refactored,
- existing tests were fixed,
- introduces global ClassSerializerInterceptor and maps classes into response dtos in each controller.

While I was trying to move user creation over from auth module to user module I have hit a weird problem with process not starting and complaining about SendGrid Api Key not being defined (triggered when User module tries to import SharedModule). Turns out we were loading .env in ormconfig.ts and that introduced file import ordering dependency. 

I introduced ConfigService from `@nestjs/config` to make sure modules use env variables when they are loaded from `.env`. This is not applied for ormconfig.ts and ormconfig.test.ts because I don't currently see how to solve having ormconfig use ConfigService outside AppModule.


backend/core/src/auth/login.dto.ts
```diff
@@ -2,17 +2,16 @@ import { IsEmail, IsString } from "class-validator"
 import { Expose } from "class-transformer"
 
 export class LoginDto {
+  @Expose()
   @IsEmail()
   email: string
 
+  @Expose()
```
software-project, line 9: Password should be exposed?

backend/core/src/user/user.dto.ts
```diff
@@ -0,0 +1,43 @@
...
+  @ApiHideProperty()
+  createdAt
+
+  @Exclude()
+  @ApiHideProperty()
+  updatedAt
+
+  @Expose()
+  @IsString()
+  password: string
```
software-project, line 36: Same here with expose.

software-project (approved):
Couple small questions, but otherwise it looks good.

backend/core/src/app.module.ts
```diff
@@ -34,6 +45,15 @@ export function applicationSetup(app: INestApplication) {
 
 @Module({
   imports: [
+    ConfigModule.forRoot({
+      validationSchema: Joi.object({
+        PORT: Joi.number().default(3100).required(),
```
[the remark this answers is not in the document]
pbn4: Config module is loading .env file. This is a validation definition for those env variables. Default means that when there is no env var a 3100 will be used.

backend/core/src/auth/auth.controller.ts
```diff
@@ -33,24 +21,14 @@ export class AuthController {
...
-    return new LoginResponseDto({ accessToken })
-  }
-
-  @Post("register")
-  @ApiOperation({ summary: "Register", operationId: "register" })
-  async register(@Body() params: CreateUserDto) {
-    const user = await this.userService.createUser(params)
-    const accessToken = this.authService.generateAccessToken(user)
-    // noinspection ES6MissingAwait
-    this.emailService.welcome(user)
```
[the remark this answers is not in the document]
pbn4: It has been moved to UserController

backend/core/src/auth/login.dto.ts
pbn4: Yup it's a LoginDTO

backend/core/src/user/user.dto.ts
pbn4: It's UserCreate DTO it must have a password prop

PR #778, opened 2020-10-21
Upgrade to eslint 7, fix linting issues

### LINKED ISSUE #327: Upgrade to ESLint 7 and typescript-eslint 3
This looks to be relatively straight-forward, but there are a whole bunch of typescript rule changes so it might take some work to sort those out. Luckily the headline breaking change (Node 8 no longer supported) doesn't affect us.

See:

- https://eslint.org/docs/user-guide/migrating-to-7.0.0
- https://github.com/typescript-eslint/typescript-eslint/releases/tag/v3.0.0

The PR for this should include at least 2 commits -- one for changes to the tooling and any configuration changes we may want, and another with all of the code changes to comply with the new regime. I *think* the top-level `yarn lint` target should be helpful in testing here, but the details there might have to be adjusted as part of this work to make them more useful.

For extra credit: there are a **ton** of linter warnings (100+), almost all of which are either `no-explicit-any` or `no-named-as-default`. Not a blocker, but any progress (even partial) toward reducing those so we have less noise in the process would be much appreciated.

I recommend we don't start on this until we have relatively few other open branches so there aren't a ton of merge conflicts, but if anyone is feeling up to it soon then let's figure out how to get it done.

bencpeters:
This PR updates to ESlint 7 & makes corresponding updates in the typescript eslint plugins. It also makes a few configuration changes to the eslint configuration to deal with the fact that with the default configuration we have a large number of eslint warnings due to `any` types spread throughout the codebase.

- I've disabled the eslint rules designed to catch using variables typed as `any` (e.g. accessing a property or calling it). These rules would be useful if our code didn't have any `any` declarations in it, but create far too many errors as is.

- I've left `no-explicit-any` in as a warning. 

- I went through the other warnings & errors and updated the relevant files. There are a _lot_ of files affected by this, but the changes were mostly pretty simple.

- I've updated the precommit hook to fail with warnings as well as errors. This ensures that you actually see warnings created by your commits, not just errors - I think this was causing a lot of our warnings to slip through undetected before. Two caveats here: if you modify a file that has existing warnings, this check will fail unless you fix the existing warnings, since it checks on a per-file basis. This might mean that you need to update existing code to get the check to pass in those cases. The second is that if for some reason you _do_ want to just check in code without fixing all the eslint warnings, you can always disable the precommit check using the `--no-verify` flag on `git commit`.

Fixes #327


pbn4 (approved):
Good stuff.

bk3c (approved):
LGTM too!

PR #779, opened 2020-10-21
Updating partner status components

slowbot:
@jaredcwhite  In preparation for flagged duplicates and create applications I'm revisiting some of the components we need to track status and actions available. 

- **status bar** is a object level status indicator that on table based layouts will need to display the most recent status update
- **status aside** is actually multiple components I think that I'm prototyping here to get some input
   - **status messages** - collection of status messages
   - **button group** - collection of buttons with grid settings

image
image
image
image


jaredcwhite (approved):
LGTM!

PR #784, opened 2020-10-23
Fix frontend typings (make use of autogenerated interfaces)

pbn4:


jaredcwhite:
@pbn4 I look forward to trying this out! One question: can we drop the `Dto` suffix from all the type names? I think it would be cleaner on the frontend to simply reference `Listing`, `Application`, etc.

bk3c:
I'm not sure if it's too strange with expected nestjs naming conventions, but I think we could move DTOs into their own directory, and then you would potentially see `.../dto/ClassName.ts` in the backend import but you would just see `ClassName` in the swagger and front-end.

If we go down that road, it might also be good to have the structure be consistent with entities -- either all in the same directory how entities are now or all in module / feature directories like DTOs are now. I don't really have an opinion either way, just that I think it'll be a little easier to document and navigate if both of those kinds of objects had similar semantics.

pbn4:
> @pbn4 I look forward to trying this out! One question: can we drop the `Dto` suffix from all the type names? I think it would be cleaner on the frontend to simply reference `Listing`, `Application`, etc.

On the backend side Listing and ListingDto have different meanings and since we are living in a monorepo I believe they should have the same meaning on the frontend.

pbn4:
> I'm not sure if it's too strange with expected nestjs naming conventions, but I think we could move DTOs into their own directory, and then you would potentially see `.../dto/ClassName.ts` in the backend import but you would just see `ClassName` in the swagger and front-end.
> 
> If we go down that road, it might also be good to have the structure be consistent with entities -- either all in the same directory how entities are now or all in module / feature directories like DTOs are now. I don't really have an opinion either way, just that I think it'll be a little easier to document and navigate if both of those kinds of objects had similar semantics.

Yeah I didn't like the fact that entities are in the same directory too. I'd go with splitting them to the modules where they belong since modules should be isolated chunks of code anyway.

jaredcwhite:
> > @pbn4 I look forward to trying this out! One question: can we drop the `Dto` suffix from all the type names? I think it would be cleaner on the frontend to simply reference `Listing`, `Application`, etc.
> 
> On the backend side Listing and ListingDto have different meanings and since we are living in a monorepo I believe they should have the same meaning on the frontend.

That seems like an implementation detail the frontend shouldn't need to be aware of. Going from `Listing` to `ListingDto` isn't ideal, and for example `ListingEventType` changing to `EnumListingEventDtoType`…that is quite the mouthful. 😄

jaredcwhite (changes requested):
@pbn4 Replying here instead of Slack…so I think it would be worth the effort to figure out a way to map Dto interfaces to their preferred names, both for the API documentation as well as the frontend interfaces. When at some point we have `api.bloom.exygy.dev` or whatever, it'll be much cleaner to find references to `Listing`, `Unit`, `User`, etc., and I simply don't want to litter the frontend code with `*Dto` everywhere. It's unfortunate that https://github.com/nestjs/swagger/issues/981 isn't yet released—that seems perfect for our needs.

Let me know if you'd like to pair on this for further research.

pbn4:
@jaredcwhite I tried working around our problem with some post client generation processing. Once the client.ts is generated I run regex find and replace for every `Dto` suffix in the file. It seems to work without any negative consequences.

jaredcwhite:
@pbn4 That sounds like a good interim solution until Nest releases the official way to do it with Swagger. Thanks for taking a second look at this. :)

PR #791, opened 2020-10-30
Updated Button styles and reworked Modal to accept component-based actions

### LINKED ISSUE #626: Optimize Button Styles
- Update button classes
   - Replace is-filled with is-primary
   - Add types for:
      - secondary = is-secondary (default hollow)
      - tertiary = is-tertiary (hollow gray-750)
      - unstyled = is-unstyled
      - borderless = is-borderless
      - text-caps = is-text-caps
      - success = is-success
      - alert = is-alert
      - tiny = is-tiny
- Button styles should be accessible via the button component and not isolated in the modal component css.
   - Use button component styles and classes
      - primary = is-primary
      - success = is-success
      - failure = is-alert
      - cancel = is-borderless

jaredcwhite:
Closes #626


jaredcwhite:
Yeah I think we could add logic to make some of those exclusive…I'm wondering if it would just work like that under the hood or if it should issue a console warning if mutually-exclusive options are set.

jaredcwhite:
Another option might be to use a single prop for `type`, another for `size`, etc. Sort of like how Shoelace works for its buttons: https://shoelace.style/components/button

jaredcwhite:
@slowbot All right, I updated this with new `type`, `size`, and `border` props along with a mixin for a decent portion of all that. I'm using TypeScript enums for the particular values of each prop where the string of the key is the CSS class name itself, so I think that's a pretty cool pattern. Let me know what you think about the naming and how that's organized now.

jaredcwhite:
@slowbot Cool, I'll double-check we've fixed all the button calls across the app(s) and then we can merge…

Issue #730, opened 2020-09-28
Partners: Export Applications


slowbot:
**Goal**: As an agent I want to have access to bulk structured application data.

**Acceptance Criteria:** 
- Provide an export feature that allows the agent with a downloadable CSV file of applications for a listing.

Issue #404, opened 2020-07-09
Partners: View Application


slowbot:
**Goal**: As a leasing agent, I want to view a submitted application 

**Acceptance Criteria:**
- Application details should be grouped into logical sections, based on information collected from the application
   - Application Data
   - Primary Applicant
   - Alternate Contact
   - Household Members
      - Grouped as a table for high level viewing
   - Household Details
   - Application Preferences
   - Declared Household Income
- The application status should be noted at the top of the application
   - Draft
   - Submitted
   - Invalid
- The date last updated should be noted at the top of the application
   - By default this is the submission date
   - This date is updates if edits are made to the application or if the application was entered manually
- The leasing agent as the ability to edit a paper application but not an application that was submitted digitally

**Prototype Components:**
View Section
https://storybook.bloom.exygy.dev/?path=/story/sections-grid-section--default-three-columns
https://storybook.bloom.exygy.dev/?path=/story/sections-grid-section--address
https://storybook.bloom.exygy.dev/?path=/story/sections-grid-section--grid-section-table
Simple Table
https://storybook.bloom.exygy.dev/?path=/story/tables-minimaltable--default
Tag
https://storybook.bloom.exygy.dev/?path=/story/prototypes-tag--pill-and-success

**Screenshot:** 
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=965%3A90

![image](https://user-images.githubusercontent.com/13915/96766981-bfdbbf00-1390-11eb-9eff-4e87323c3ebb.png)

kathyccheng:
Closed in #792

Issue #392, opened 2020-07-04
Partners: Listing Applications


slowbot:
**Goal:**
As a partner I want to be able to view a list of applications for a listing that have been submitted electronically.

**Acceptance Criteria:**
- As a partner, I want to have access to a list of all applications for a specific listing
- Columns for applications include: - **STARTED**
   - Application Submission Date   
   - First Name
   - Middle Name
   - Last Name
   - Application Number
   - Household Size 
   - Declared Income
   - Subsidy or Voucher
   - Request ADA
   - Requested Unit Type
   - Date of Birth
   - Email
   - Phone (show phone type)
   - Additional Phone (show phone type)
   - Residential Street Address
   - Residential City
   - Residential State
   - Residential Zip
   - Mailing Street Address
   - Mailing City
   - Mailing State
   - Mailing Zip
   - Remaining fields as displayed here from X on as well if performance allows https://docs.google.com/spreadsheets/d/1zBTbBs1zhTnNlC2_LvIXmZ8FRrRfG82bgHsOT8Nx-So/edit?usp=sharing

- Individual table columns should be sortable - **DONE**
   - First Name
   - Last Name
   - Application Submission Date
- Table should be paginated - **STARTED**
   - Markup patterns in place
   - Custom next and previous buttons
   - Show number of rows dropdown
   - Jump to page dropdown
   - Show total number of applications
- Custom filter that sorts ALL table values 
   - https://www.ag-grid.com/javascript-grid-filter-quick/

Prototype:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=121%3A85

![image](https://user-images.githubusercontent.com/13915/96767487-750e7700-1391-11eb-87ae-04062cfa9c81.png)
![image](https://user-images.githubusercontent.com/13915/96767507-7b9cee80-1391-11eb-91d0-4ca49fd11150.png)

kathyccheng:
@slowbot can you audit this issue and make sure it captures remaining work for Dominik?

bk3c:
@dominikx96 Just a reminder: ag-grid is has now released 24.x, so you should upgrade that dependency first so you can base all development on the latest and greatest. It looks like there are just enough breaking / best practices changes to make upgrading first worthwhile, but it shouldn't be too crazy.

kathyccheng:
Closed in #792

PR #792, opened 2020-10-30
Feature/partners listing applications

dominikx96:
Resolves: #730, #404, #392

**Next steps (backend not ready yet, #811  ):**
#404:
- Replace hardcoded data:
    - Application Status
    - Application Language
    - Application Submission Type


dominikx96:
@slowbot 

Changes are implemented, also filtering in the applications table.
We still wait for backend fields.

dominikx96:
@slowbot header has been updated (added first, last name & confirmation ID), but I used the current text style in the header component, is it ok for you?

slowbot (commented):
@dominikx96 Still seeing the build errors in the screenshot above but otherwise Export and app submission are working locally. 

LGTM

jaredcwhite:
Hey @dominikx96, we have the ViewItem component in both `blocks` and `elements` now, so we should pick one category and make sure it's up to date.

jaredcwhite:
Looking more closely, the one in blocks is the one being exported out of ui-components, so I guess we could just remove the one in elements?

dominikx96:
@jaredcwhite yes, I removed the duplicated component, thanks for noticing me!

@slowbot before merge I have few questions about newly added fields:

- When the user does not check "Send my mail to a different address" in the application form: should we show the mailing address section with the same data as in "RESIDENCE ADDRESS" or show something like "Same as residence address"?

- A "Middle Name" field - should it be visible if is empty?

- We added fields for application language, submission type, application status, but what about old applications (existed before this change)?
  - LANGUAGE: should we hide this cell or show English as default (correct me if I'm wrong, but before all applications were in EN ?)
  - SUBMISSION TYPE: should be hidden?
  - STATUS: I think it should be "submitted" as the default

slowbot (commented):
@dominikx96 See my comments

Issue #317, opened 2020-05-27
Listings Service Object Model v2


bk3c:
As we move listings from JSON to proper database-backed storage, it seems like the right time to focus the object model on the longer-term needs, including what we'll need for all of the Partners App functionality. Some of this should focus on making sure the existing fields we have are reasonably normalized so that they can be grouped and edited reasonably from the Partners UI. For example, we may want to split off Leasing Agents into their own object (though I'm not 100% convinced in that particular case). We also want to make sure we can support versioning, as described in #286. 

One of the major expected changes is to separate the notion of a Property (or perhaps Building -- both names have downsides) which should represent a physical place of affordable housing (optionally composed of units) from a Listing, which represents the availability for lease or purchase for a particular period of time with a specified set of preferences, policies, explanatory text, etc.

The last big piece of this expanded object model is to have objects that group Properties, mostly as containers for permissions and policy rules. These PropertyGroups (desperate for a better name!) could represent jurisdictions (e.g. cities or counties) or property management companies (that work across multiple jurisdictions). Each property could be associated with zero, one, or multiple groups (though in the real world probably at most a few groups per property), and groups themselves might also be a part of a hierarchy (e.g. city -> county -> state). 

None of these changes need to affect the existing API to the front-end unless we think there's a net improvement that's worth the work, and we should balance the benefits of any potential changes with the desire to keep a reasonable migration path for SF DAHLIA when we get there later this year. To the extent that we can improve the API semantics to get closer to what we want our object model to be and also get closer to where DAHLIA is evolving, all the better.

Issue #747, opened 2020-10-07
[Meta] October 2020 Object Model Changes


bk3c:
This is meant as a meta-issue to track a host of object model changes that we'd like to get done this month, each of which should (eventually) be defined in their own issues, but I thought would be worth consolidating here so that we could manage it as a unified process since it's a **huge** amount of interrelated plumbing work to get done. 

Some other issues we do concurrently might also have minor object model implications (e.g. adding a field) that won't be tracked here, but this is meant to track the object model overhaul that will put us in good shape for the next several months of feature development.

- [x] Move to integer as DB primary keys and Hashids as public object IDs, instead of UUIDs (#348)
- [x] Split out Properties from Listings (#317) - PR 
- [x] Rationalize fields between Properties, Units, and Listings (#317)
- [x] Add PropertyGroups, which are mostly a container for permission assignment across multiple properties (#317)
- [x] Move AMI charts into the database (with a more general name) (#328) - PR
- [x] Split out* Applicants from Applications (#843) - PR
- [x] Add backend object for Application Form Config (1:1 with listings?) and rationalize with Listings as appropriate (#855) -PR
- [x] Migration of existing database 
- [x] Permissions based on property group 
- [ ] Add versioning of major objects (#286) 

-- 
- [ ] Build out object model for Preferences and Eligibility Rules that can be edited and machine-evaluated

Since this is such a substantial overhaul of the schema, I think it's acceptable to plan to rebuild production databases from scratch, so long as we can re-create the listings and have some manual process for the one-time export and re-import the Applications from the Jones pilot, which will certainly be over by the time we land this in production.

pbn4:
One thing to remember with ApplicationFormConfig is to create an entry for all the records once we get rid of the frontend embedded application form config. I marked it as a TODO in frontend code.

pbn4:
@jaredcwhite @software-project @netrabadami @dominikx96 For permissions management we are now using Casbin which defines a CSV backend side with all the rules for evaluation whether a specific user can or cannot call given action / access given resource. It looks like this:

```
p, admin, application, true, .*
p, anonymous, application, true, submit
p, anonymous, user, true, create
g, admin, user
g, user, anonymous
```

and is embedded in the code. It's quite nice because at controller level you only define what it the resource type and what is the action and based on a request Casbin evalutes whether given user is authorized to perform given action on a given resource or not. Unfortunately I haven't found a solution for casbin to easily share permissions between frontend and backend. 

[...]

PR #822, opened 2020-11-06
Properties and property groups

pbn4:
Closes https://github.com/bloom-housing/bloom/issues/317 (part of https://github.com/bloom-housing/bloom/issues/747)

This PR introduces:
- Property entity (which is basically a split up of what was previously in Listing),
- Property Group entity 
- refactored seeds Module (applies changes related to DB schema and introduces typed seeds)
- storyshots now use ArcherListing which is a typed Listing object (based on autogenerated Listing type), I removed untyped archer.json import

TODO:
- not sure about some properties that are still existent in types definitions but I couldn't find any references in the code
- Only 1 of 3 seeds.json listings was moved to a typed equivalent on the backend, listings e2e tests for  jsonpath queryParam are not completed because of this
- at the moment this PR introduces a breaking change to the DB schema, because of the foreign keys constraints being defined on Unit entity which is no longer in relation to Listing but to a Property. With the upcoming changes in https://github.com/bloom-housing/bloom/issues/747 there will be even more changes to the schema and I can hardly imagine how are we going to migrate this. @bk3c What was your initial idea for this?  Are we OK with spinning up another instance, copying data over with some scripts and switching the existing backend (talking about the one deployed in housingbayarea repo) to the new DB instance? 
- consider switching over to Data Mapper pattern instead of Active Directory and BaseEntity inheritance. It does not work well e.g. when backend tests want to use a shared ArcherListing object which is defined based on autogenerated interface (and this interface does not inherit from BaseEntity). There are also problematic for seed definitions (I needed to define seeds types using Omit<EntityClass, keyof BaseEntity> to get around that and have a proper typing).


jaredcwhite:
I don't have much to offer in the way of suggestions regarding the backend typing/data migration, but from a frontend perspective this looks good to me. I'm assuming the `listing.property` values will always be included in the response from the backend…i.e., we don't need to do anything special to request that data?

PR #844, opened 2020-11-18
Feature/application details email overflow

### LINKED ISSUE #828: Partners: Overflow View Item
As a leasing agent I want to see an indication of lang strings of text

- Add a variant to view item that truncates text and adds ellipsis
- Use ellipsis mixin to apply style to view item value

image

dominikx96:
Closes #828 

Added prop **truncated** to the viewItem value.

@slowbot applied to the email field only. Should I add it also to any other? Which?


slowbot (commented):
@dominikx96 Email is all we need for now I think.

LGTM

Issue #391, opened 2020-07-04
Partners: Sign In


slowbot:
**Goal:**
As a partner I want to be able to login securely to the portal in order to make updates to listings and applications.

**Acceptance Criteria:**
- Partner should be prompted for email and password 
- Both fields are required
- Email address must be formatted as a proper email address
- If the email and password is for a public user account, they should be denied access 
- If the partner has forgotten their password they can utilize a forgot password link and will be sent an email to start the reset process. (Process and components will be similar the the public experience See https://github.com/bloom-housing/bloom/issues/244)
- Once you sign in, you should be able to see all applications

![image](https://user-images.githubusercontent.com/13915/86629072-b2383000-bf7f-11ea-962d-4e86c86d41a8.png)

kathyccheng:
This issue should also encompass the back end work of actually making sure only partner accounts can log in

kathyccheng:
This also needs to include a deny state for if you don't have the proper permission

kathyccheng:
@dominikx96 I think this screen is mostly there, this is just making sure validations are in place

jaredcwhite:
@kathyccheng I talked with @dominikx96 and this looks like something I could help push forward now, unless there's something higher-priority you'd like me to tackle.

kathyccheng:
sounds good @jaredcwhite!

kathyccheng:
Closed in #854

PR #854, opened 2020-11-19
Enhancements to the partners Signin Page

jaredcwhite:
Addresses #391


jaredcwhite:
@slowbot This is ready to try out now. I think the error handling should be working pretty well now

jaredcwhite:
Sounds good to me 👍￼

jaredcwhite:
OK, ready for another review @slowbot

jaredcwhite:
FYI—I just noticed this comment by @kathyccheng on the original issue:

> This also needs to include a deny state for if you don't have the proper permission, e.g. if you try to log in with a public user account. @slowbot can you define an error message here?

kathyccheng:
@jaredcwhite cc @slowbot we just talked about using the same general looks like you don't have account message here. no need for a specific message per Jesse's suggestion above. Just want to know for testing purposes that public account won't be able to access the partner's portal and that part of our permissions has been hooked up properly.

jaredcwhite:
All right, so in that case are we good to merge this in now?

PR #855, opened 2020-11-19
Add application form config to DB schema

pbn4:
Addesses https://github.com/bloom-housing/bloom/issues/747 - Add the backend object for Application Form Config (1:1 with listings?) and rationalize with Listings as appropriate

### DIFF
--- a/apps/public-reference/lib/AppSubmissionContext.ts
+++ b/apps/public-reference/lib/AppSubmissionContext.ts
@@ -3,6 +3,10 @@ import ApplicationConductor from "./ApplicationConductor"
 import { ApplicationStatus, ApplicationSubmissionType, Language } from "@bloom-housing/core"
 
 export const retrieveApplicationConfig = () => {
+  // TODO https://github.com/bloom-housing/bloom/issues/747 Make sure all the
+  //  ApplicationFormConfig columns are filled in each listing row
+  //  once moving this config from frontend to the DB.
+
   // Note: this whole function will eventually be replaced with one that reads this from the backend.
   return {
     sections: ["You", "Household", "Income", "Preferences", "Review"],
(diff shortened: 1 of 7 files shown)


jaredcwhite:
@pbn4 This looks good to me. I feel like maybe we're missing a step though (pardon the pun 😉)…

My understanding is that there needs to be a "master" config somewhere that is essentially duplicated when a new listing is created. It was something we deliberated for some time back when this was all getting hammered out, so I don't think it ever got fully spec'd out. But we need some kind of template mechanism whether that is something living in the database or defined in a backend config file somewhere, so when a listing gets created it inherits that form config.

jaredcwhite:
Hey @pbn4, let me know if my comment above makes sense…I don't recall how detailed the original spec was (if at all) and we might need to double-check it's heading in a direction we still want…

jaredcwhite (changes requested):
Hey @pbn4, let me know if my comment above makes sense…I don't recall how detailed the original spec was (if at all) and we might need to double-check it's heading in a direction we still want…

pbn4:
@jaredcwhite Definitely makes sense, I switched over to other work but I will apply your comments shortly.

kathyccheng:
At this point, we'll want to support having multiple listings with applications running at the same time. These listings are likely to have different preferences. Also, we'll be starting our first San Mateo County application which will have 3 questions that are different than Alameda's so we want to make sure we're serving the right questions, it might be worthwhile to think about having a county default set of questions.

kathyccheng:
Front end refactoring addressed in #1130 and #1129

pbn4:
I'm closing this PR since it's long incompatible with existing codebase and current listings will address variable preferences with   metadata attached to listing model.

Issue #850, opened 2020-11-18
Partners: Single Application Cleanup


slowbot:
Update fields for single application view

**Primary Applicant:**
- Add Phone types to helper text for Primary and Secondary Phone
- Show "Work in Region" field
- Show Work address fields

**Alternate Contact:**
- Show Address fields

**Declared Income**
- Show Monthly income field

**Terms**
- Show Signature on Terms of Agreement

![image](https://user-images.githubusercontent.com/13915/99569575-a6289a00-2985-11eb-85f1-ca1d37189c51.png)

dominikx96:
- Alternate contact address fields are optional in the application form (so a user can type e.g only city), should we show available data and for missing just "n/a"?

- What about applications that don't have terms agreement data (we added this field 2-3 weeks ago)?

PR #856, opened 2020-11-19
Feature/application details cleanup

dominikx96:
Closes: #850 

@slowbot I'm not sure:

- Alternate contact address fields are optional in the application form (so a user can type e.g only city), should we show available data and for missing just "n/a"?

- What about applications that don't have terms agreement data (we added this field 2-3 weeks ago)?


slowbot (commented):
@dominikx96 See comments

dominikx96:
@slowbot I applied changes.

slowbot (approved):
@dominikx96  LGTM

PR #858, opened 2020-11-19
Temporarily bump max-warnings up for eslint

jaredcwhite:
~~I'm still a little foggy why husky seems to be linting everything, not just newly-staged files when trying to commit in VSCode, but at any rate I'm bumping allowed warnings up to 100 until we can fix the existing warnings we have in the codebase.~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,6 @@
     }
   },
   "lint-staged": {
-    "*.{js,ts,tsx}": "eslint --max-warnings 0"
+    "*.{js,ts,tsx}": "eslint --max-warnings 100"
   }
 }


jaredcwhite:
OK I think the issue is actually something very different…trying to update stories in ui-components is tripping up eslint and that's what the problem is, not that there are warnings across the whole codebase per se. There's actually an issue about this but it was closed without resolution so I'm not sure what to do about it: https://github.com/eslint/eslint/issues/12206

jaredcwhite:
Closing at this point, since there's no obvious fix for the *.stories.tsx bug.

PR #862, opened 2020-11-21
Fix scrolling issue with Overlay

### LINKED ISSUE #859: Scrolling ability isn't restored after dismissing modal/drawer
There seems to be an issue with the latest code…likely a bug from the recent refactoring work I did. Once you dismiss the modal or drawer, the body's still set to `overflow: hidden` and won't scroll.

jaredcwhite:
Fixes #859 (check out the first Model or Drawer story and scroll up/down to verify)


dominikx96 (approved):
LGTM

Issue #830, opened 2020-11-09
Partners: Persistent Sort Order


slowbot:
As a leasing agent i want the sort order applied to a column to persist

- Ensure the sort order on the applications table persists when paginating through results
- Ensure the sort order on the applications table persists when moving between applications table and single applications

PR #864, opened 2020-11-23
Feature/persistent sort order

dominikx96:
Closes: #830

Currently, sort order should be persistent on pagination change/after move to the details page, BUT sorting is only applied for active page - not for all table data (#866).

It's not an ideal implementation, because in the community license we can't use the server-side-model:
https://www.ag-grid.com/javascript-grid-server-side-model/

but applied some workarounds (we save column settings to the sesssionStorage after column sort/move and pagination change) and I think it's enough.


slowbot (approved):
LGTM @dominikx96

PR #875, opened 2020-11-30
"required" prop in PhoneField and DOBField

dominikx96:
The idea is to do not validate the phone field and DOB when the `required` prop is set to `false` and length = 0 or undefined.

Implemented for partners MVP purposes.


jaredcwhite:
LGTM :+1:

PR #877, opened 2020-11-30
Update translation files and keys

### LINKED ISSUE #869: Implement Translations
These are the translation files we got back 
[od-12714.zip](https://github.com/bloom-housing/bloom/files/5593412/od-12714.zip)

Given that we updated our general file and they did their translations based off of that, we'll have to update to reference the cleaned up keys and then implement the file. It might be good to define a plan for the general file integration work, open to how this can be broken down further

jaredcwhite:
Closes #869


software-project (approved):
LGTM, only some check are failing

jaredcwhite:
@software-project I think we need to hold off merging until you finish your current update of housingbayarea, right? So we don't "leak" the new translation keys into those sites just yet?

software-project:
@jaredcwhite these are different repos, so we should be save to merge now.

jaredcwhite:
@software-project but if we merge this into master, and then publish new packages off of master, wouldn't the new translations get into the package that's used by housingbayarea?

software-project:
@jaredcwhite I thought that we would need to first deploy the package with bumped version. But to be fair, I don't know the full process how that works, as Ben wasn't using the plugin for versioning that I prepared a while back.

software-project:
@jaredcwhite also some checks are failing on the build, maybe we should resolve them as well?

jaredcwhite:
Will work on resolving these merge conflicts.

Issue #806, opened 2020-11-03
Partners can add Terms 


kathyccheng:
As a partner
I want to indicate if the applicant has signed the Terms of Agreement or not 
When I'm on the New Application page
Then I'll see a Terms section with a yes/no
When I click Submit
I'll see a Terms section on the detail page with what I filled out 

![image](https://user-images.githubusercontent.com/13915/98025754-f8cb5900-1dbe-11eb-900b-1adb1d676cc0.png)
![image](https://user-images.githubusercontent.com/13915/98025785-02ed5780-1dbf-11eb-9352-a65aca60289b.png)

dominikx96:
Merged.

Issue #805, opened 2020-11-03
Partners can add Demographic Information 


kathyccheng:
As a partner
I want to add demographic information from paper application
When I'm on the new application page
I'll see a Demographics section with ethnicity, race, gender and how did you hear about us 
When I click submit
I will NOT see Demographics on the view (this data is saved but only viewable in aggregate)

![image](https://user-images.githubusercontent.com/13915/98025229-485d5500-1dbe-11eb-9e6a-b90922b04ed8.png)

slowbot:
@kathyccheng I'm removing this "Then I'll see the Demographics on the detail page" since I think we only want to provide demographics in aggregate

dominikx96:
Merged.

Issue #804, opened 2020-11-03
Partners can add Application Preferences 


kathyccheng:
As a partner
I want to add application preferences 
When I'm on the New Application page
I'll see an Application Preferences section with the ability to select the application's preferences 
When I click Submit
Then I'll see an Application Preferences section on the detail page 

![image](https://user-images.githubusercontent.com/13915/98025107-2368e200-1dbe-11eb-8f51-fb20a7189271.png)
![image](https://user-images.githubusercontent.com/13915/98025129-29f75980-1dbe-11eb-8c29-239ca4085f54.png)

dominikx96:
Merged.

Issue #803, opened 2020-11-03
Partners can add Application Data


kathyccheng:
As a partner
I want to add Application Data 
So that I can mark when I got the application
When I'm on the New Application page
Then I'll see an Application Data section
Where I can input Date Submitted, Time Submitted and Language Submitted In 

![image](https://user-images.githubusercontent.com/13915/98024879-cff69400-1dbd-11eb-9c34-af4fc99b5d71.png)
![image](https://user-images.githubusercontent.com/13915/98024913-dab12900-1dbd-11eb-8ec7-5d313b9bd89b.png)

dominikx96:
Merged.

PR #891, opened 2020-12-07
Feature/paper application form

### LINKED ISSUE #799: Partners can add alternate contact
As a partner developer
I want to add an applicant's alternate contact information (name, relationship, agency, email, phone, phone type) 
When I'm on the new application page
I'll see a section for Alternate Contact 
When I click submit
Then I'll see a section for Alternate Contact with the fields I filled in 

image
image

dominikx96:
Closes #799, #800, #801, #803, #804, #805, #806

I noticed some differences in the current implementation and designs:

- Alternate Contact:
    - Phone Type (field)

Currently, we don't have this field in the form in the public reference, should we also add this to the public app or remove from partners?

- Declared Household Income

I added radio buttons to select period - to prevent incorrect field values. In the backend, we can store only one income. Only one field can be active at one time. Is it ok?

- Demographic Information
    - added "sexual orientation" field - because this is visible in the form in the public app
    - changed "How did you hear..." field to checkbox group - the user has the ability to select multiple elements

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Operational configuration of each app and service is read from environment varia
 
 ### Setting up a test Database
 
-The new `backend/core` uses a postgres database, which is accessed via TypeORM. Once postgres is sety up and a blank database is initialized, yarn scripts are available within that package to create/migrate the schema, and to seed the database for development and testing. See backend/core/README.md for more details.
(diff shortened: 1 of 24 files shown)


dominikx96:
@slowbot changes applied. I'm not sure about:

> Declared Household Income is the same as Household Income. We just adjust the field label for Partners

Image 2020-12-08 at 5 38 16 PM

So I added the radio buttons for selecting period and to activate the proper field to put income.

slowbot (approved):
See comments

PR #900, opened 2020-12-09
Inline style button

### LINKED ISSUE #829: Refactor: Back Button
As an engineer I want to use a single back button throughout both applications

- Repurpose actions/inline-button in both partners and public applications

jaredcwhite:
Closes #829


jaredcwhite:
@slowbot I got the "inline button" style/variant merged into the Button component for the back usage, but I'm sort of wondering how best to label the differences between "unstyled" and "inline"…technically they're both "inline" and only slightly different, so it's not clear at first glance which to use for what.

slowbot (changes requested):
@jaredcwhite What do you think about referencing the icon in the name

- inline-icon (options left and right)
- inline-icon--left

jaredcwhite:
@slowbot Sure…it leaves room for further tweaking down the road, but for now, sounds goods.

jaredcwhite:
@slowbot Incorporated your suggestion, let me know how this looks.

slowbot (approved):
@jaredcwhite LGTM

PR #901, opened 2020-12-09
Fix/rename native btn property

dominikx96:
Currently **type** prop in the <Button /> component is used for style purposes. Natively it describes button type. For partners MVP purposes we have to add <Button **type="button"** /> prop (we have button to open Drawer inside form).

I renamed type -> styleType and replaced this in each component instance.


jaredcwhite:
@slowbot I was chatting with @dominikx96 about this…I suggested leaving `type` as-is for `Button` and adding something like `submitsForm={false}` as an optional prop, but Dominik thought it would be better to shadow the HTML attributes for `button` tag as much as possible. Any thoughts?

slowbot (commented):
@dominikx96 @jaredcwhite See my comment above

dominikx96:
I updated this branch to be up to date with the master. I used styleType name also inside another PR, so can we keep this name? @jaredcwhite

I want to merge those things as soon as possible to focus on paper form integration.

jaredcwhite (approved):
LGTM :+1:

Issue #908, opened 2020-12-14
Time Component


dominikx96:
The goal is to create new components to store time.

<img width="254" alt="Image 2020-12-15 at 10 25 45 AM" src="https://user-images.githubusercontent.com/16655908/102196249-ed8c4280-3ebf-11eb-8301-2d3c56232391.png">

PR #912, opened 2020-12-15
Feature/date time field

dominikx96:
Closes: #908


software-project (approved):
Looks good. Validation working nicely. I'm only wondering if we should fix accessibility warnings?

dominikx96:
> Looks good. Validation working nicely. I'm only wondering if we should fix accessibility warnings?

I'm not sure how accesslint tries to test this code, but field labels are available. In a storybook, a11y tests have been passed.

jaredcwhite:
> > Looks good. Validation working nicely. I'm only wondering if we should fix accessibility warnings?
> 
> I'm not sure how accesslint tries to test this code, but field labels are available. In a storybook, a11y tests have been passed.

Ah, I think I get why we keep seeing this…it's because the JSX-specific markdown in the React component uses `htmlFor` instead of `for` in the label tag to point to the input tag, and the linter isn't understanding that.

PR #917, opened 2020-12-15
Add StatusAside component and place on the New/Edit Application form

### LINKED ISSUE #696: Component: Status Aside
**Goal**: As a user I want to see summary updates 

**Acceptance Criteria:** 
Organization:
- Locate at ui-components/src/notifications

Engineering notes:
- Stub with default data

Props:
- timestamp - date and time of note. _string_
- children
   - button group - include one or more actions

Markup Prototype: https://storybook.bloom.exygy.dev/?path=/story/prototypes-statusaside--status-aside-last

jaredcwhite:
Close #696

@dominikx96 I added the new component to the sidebar, let me know what you think. We'll probably need to adjust for adding vs. editing.

@slowbot Right now this is a single column, but it supports two-column button placements as well for more advanced layouts.

### DIFF
--- a/shared/ui-components/src/notifications/StatusAside.tsx
+++ b/shared/ui-components/src/notifications/StatusAside.tsx
@@ -18,10 +18,7 @@ export const StatusAside = (props: StatusAsideProps) => (
     </div>
 
     {React.Children.count(props.children) > 0 && (
-      <div className="status-aside__messages">
-        <h3 className="status-aside__title">Status History</h3>
-        {props.children}
-      </div>
+      <div className="status-aside__messages">{props.children}</div>
     )}
   </div>
 )
--- a/shared/ui-components/src/notifications/StatusAside.tsx
+++ b/shared/ui-components/src/notifications/StatusAside.tsx
@@ -1,0 +1,24 @@
+import React from "react"
+import { GridSection } from "../sections/GridSection"
+
+import "./StatusAside.scss"
+
+export interface StatusAsideProps {
+  actions: React.ReactNode[]
+  children: React.ReactNode
+}
+
+export const StatusAside = (props: StatusAsideProps) => (
+  <div className="status-aside">
+    <div className="status-aside__buttons">
+      <GridSection columns={2} tightSpacing={true}>
+        {props.actions}
+      </GridSection>
+    </div>
+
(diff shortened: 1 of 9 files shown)


slowbot (approved):
@jaredcwhite This LGTM with a small update to spacing.

shared/ui-components/src/notifications/StatusAside.tsx
```diff
@@ -0,0 +1,27 @@
...
+  <div className="status-aside">
+    <div className="status-aside__buttons">
+      <GridSection columns={props.columns || 2} tightSpacing={true}>
+        {props.actions}
+      </GridSection>
+    </div>
+
+    {React.Children.count(props.children) > 0 && (
+      <div className="status-aside__messages">
+        <h3 className="status-aside__title">Status History</h3>
```
dominikx96, line 22: Can we use translated strings only?

dominikx96 (changes requested):
@jaredcwhite noticed some things:

- StatusAside.tsx & StatusMessage.tsx have hardcoded strings, can we use translated strings?
- The StatusAside component doesn't support single children with e.g.: "Last updated: XXX", "Status History" is always visible. @slowbot do we plan to keep only this version or "minimal" variant should be available as well?

jaredcwhite:
> @jaredcwhite noticed some things:
> 
> * StatusAside.tsx & StatusMessage.tsx have hardcoded strings, can we use translated strings?
> * The StatusAside component doesn't support single children with e.g.: "Last updated: XXX", "Status History" is always visible. @slowbot do we plan to keep only this version or "minimal" variant should be available as well?

Both good points, thanks. I'd gotten both variants working for StatusMessage already, I think I just need to enhance StatusAside to support those.

jaredcwhite:
I made some changes so that the "Status History" title is part of the StatusMessages component, rather than the StatusAside component. I think this makes more sense, but let me know if you think there's a better way to do things.

dominikx96 (approved):
@jaredcwhite noticed hardcoded "Last Updated:" string inside StatusMessage.tsx - if we fix it LGTM and I think we can merge it :)

slowbot (approved):
LGTM @jaredcwhite

PR #923, opened 2020-12-21
Prevent fetch data in client-side mode

dominikx96:


dominikx96:
Just for test purposes

PR #927, opened 2021-01-04
Rename and relocate packages

### LINKED ISSUE #528: [RFC] Front-end package and directory re-organization

jaredcwhite:
Resolves #528 

* Remove shared/core and services/listings and move ui-components to top-level
* Put type definitions for frontend in `backend/core/types`
* Rename apps/\*-reference to sites/\*


jaredcwhite:
FYI this is likely not going to build because of all the directory changes. Also, if folks feel like the autogenerated frontend types should be located somewhere other than `backend/core/types`, feel free to suggest alternatives.

kathyccheng:
@jaredcwhite do you need reviews from @software-project @pbn4 @dominikx96 to move this forward? I know this is one of those we don't want to leave hanging too long

software-project (changes requested):
Nice work and a positive change. I did a quick tests and only little thing I found was a reference to old listings service in `dev:all` yarn task. 👍

jaredcwhite:
I got jest tests working in ui-components, and I made the package.json changes you pointed out @software-project.

jaredcwhite:
Still some CircleCI tests failing like for the public/partners apps…will investigate.

pbn4:
@jaredcwhite I'll more backend/core/types to ui-component in a separate PR

jaredcwhite:
All tests passing now! (except for Netlify)

pbn4:
@jaredcwhite What is the ETA for this?

PR #930, opened 2021-01-05
Move frontend client generation to shared/core and add README.md

pbn4:


jaredcwhite:
@pbn4 Have you taken a look at the other PR #927 where all the packages are redone? We don't have a shared/core package anymore, so we'll need to discuss what the approach should be here.￼￼￼

jaredcwhite:
Will be addressed in a different PR.

PR #940, opened 2021-01-12
Applications service refactor (remove listing join) 

pbn4:
It's not captured by any issue, but this is somewhat a bug that some of the frontend code was retrieving listing properties through application object. Doing that introduces repeating listing in data for each retrieved application.  I decided to remove that problem early.


dominikx96 (approved):
LGTM 👍

PR #941, opened 2021-01-12
Fix PUT /listings

pbn4:
It's not caputed by any issue, removes a bug causing PUT /listings to throw 500 internal server error when trying to modify a listing through the API.


dominikx96 (approved):
Tested /PUT Listings, I can successfully update listing data, and get listing, where "property" object still exists. 👍

PR #946, opened 2021-01-13
Fix/update waitlist component 707

pbn4:


pbn4:
@dominikx96 
```
  1) applications/household/preferred-units
       Should save form values and redirect to the next step:
     AssertionError: Timed out retrying: Expected <div.alert-box.alert.invert> not to exist in the DOM, but it was continuously found.
      at Context.eval (http://localhost:3000/__cypress/tests?p=cypress/support/index.js:224:26)
```

Is that something new?

dominikx96:
> @dominikx96
> 
> ```
>   1) applications/household/preferred-units
>        Should save form values and redirect to the next step:
>      AssertionError: Timed out retrying: Expected <div.alert-box.alert.invert> not to exist in the DOM, but it was continuously found.
>       at Context.eval (http://localhost:3000/__cypress/tests?p=cypress/support/index.js:224:26)
> ```
> 
> Is that something new?

I run the test on circleCI again and it works. A potential fix is to increase the cypress timeout limit, but let's resolve it in a separate issue.

PR #955, opened 2021-01-18
Change JSX.Element to React.ReactNode

### LINKED ISSUE #450: Cleanup: swapping JSX.Element out for ReactNode

jaredcwhite:
Resolves #450


dominikx96 (approved):
LGTM! 👍

Issue #860, opened 2020-11-20
Partners: Applications Table Household Member Details


slowbot:
As a leasing agent I want to know the most valuable data for a household member via the applications table

1. Show First Name, Last Name, DOB, Relationship for 6 stubbed out household members
2. Show "Same Address as Primary" Yes/No and "Work in Region" Yes/No fields for each household member
3. Hide Residence and Work Address fields from table
4. Show a max number of household members columns based on the applications submitted (nice to have)

![image](https://user-images.githubusercontent.com/13915/100916724-fa09a780-348a-11eb-9ec5-f0cce9472c94.png)

kathyccheng:
@slowbot can you add a screenshot here

dominikx96:
"Hide Residence and Work Address fields from table" 

Do you mean - from a primary applicant?

PR #956, opened 2021-01-19
Household Columns Update

dominikx96:
Closes: #860

@slowbot I'm not sure about Residence and Work Address (primary applicant, not HH member) - Is this information should be visible?

Also added n/a inside HH columns if the information is not set. Do we want to apply this for primary applicant columns as well?


slowbot (approved):
@dominikx96 LGTM. Added small spacing styles

PR #959, opened 2021-01-20
Tab nav component

### LINKED ISSUE #697: Component: Tab Nav
**Goal**: As a user I want a horizontal secondary navigation 

**Acceptance Criteria:** 
Organization:
- Locate at ui-components/src/navigation

Engineering notes:
- Stub with default data

Accessibility:
- aria label for secondary navigation

Props:
- label - link text _string_
- url - destination, _string_
- current - applied to the current page link, _boolean_
- count - includes a tag with a number

Markup Prototype: https://storybook.bloom.exygy.dev/?path=/story/prototypes-tabnav--default

jaredcwhite:
Closes #697


jaredcwhite:
Demo: https://deploy-preview-959--amazing-davinci-96c1cd.netlify.app/?path=/story/navigation-tab-nav--default

ui-components/src/navigation/TabNav.tsx
```diff
@@ -0,0 +1,33 @@
...
+  href: string
+  current?: boolean
+  children: React.ReactNode
+}
+
+const Tab = (props: TabProps) => {
+  return (
+    <li className="tab-nav__tab">
+      <LocalizedLink
+        className={props.current ? "is-active" : undefined}
```
dominikx96, line 33: I think we can simplify it to just: `props.current && "is-active"`

dominikx96 (changes requested):
When I opened the storybook, I saw 3 errors, e.g:

`ARIA role tablist is not allowed for given element` - it means that we can't use this role for `<nav>` element.

Can you re-check this part?

jaredcwhite:
Thanks @dominikx96 for catching that! I think I got it figure out now.

Also I totally goofed and forgot to implement that extra tag pill count as a feature in the new component so I'll implement that now…

ui-components/src/navigation/TabNav.tsx
jaredcwhite: Well that only works on the "true" condition…on the false condition the expression literally returns false, rather than undefined

slowbot (approved):
@jaredcwhite LGTM. I may update the nested pill style later.

Issue #948, opened 2021-01-14
Preferences Page in Application 


kathyccheng:
* Pull in 4 SoHay preferences 
* For the first 2 displaced preferences, if applicant selects them, then it'll open asking for Name of Displacee and Address 
* It would be great to store name and address separately, but open to using a general text field or more structured Name + Address fields  
* Check that preference opt in / out and additional name + address stored in the back end 
* Make sure preferences stored in CSV export 

@slowbot to add mockup 

**Preferences Copy**

Displacee Tenant Housing Preference: At least one member of my household was displaced from a residential property due to redevelopment activity by the Hayward Housing Authority, the Redevelopment Agency or the City of Hayward. 

Mission Boulevard Corridor (SR 238) Displacee Tenant Housing Preference: At least one member of my household is an SR 238 Program Participant as required under the terms of the SR 238 Settlement Implementation Project. Program Participants are tenants of previously sold Caltrans properties along the SR 238 Mission Boulevard Corridor and are given occupancy preference when new below-market housing is constructed in the Corridor. 

For more information on the SR 238 Settlement Implementation Project, please visit: 
https://www.hayward-ca.gov/238/background

Live in City of Hayward Preference
At least one member of my household lives in the City of Hayward.

Work in City of Hayward Preference
At least one member of my household works or has been offered work in the City of Hayward.

slowbot:
https://www.figma.com/file/oU5mI1afPd9zjt9vNxcLFa/dahlia_short-form?node-id=152%3A2243
https://www.figma.com/file/oU5mI1afPd9zjt9vNxcLFa/dahlia_short-form?node-id=153%3A94
https://www.figma.com/file/oU5mI1afPd9zjt9vNxcLFa/dahlia_short-form?node-id=153%3A188

![Shortform Preference_ Sohay](https://user-images.githubusercontent.com/13915/105745232-e90d6d80-5ef2-11eb-9860-9508da837216.png)
![Shortform Preference_ Sohay_ More](https://user-images.githubusercontent.com/13915/105745248-edd22180-5ef2-11eb-8c2b-a395de5b850e.png)
![Shortform Preference_ Sohay_ Active](https://user-images.githubusercontent.com/13915/105745261-f296d580-5ef2-11eb-84b8-f09f59a89f6f.png)

slowbot:
@jaredcwhite @pbn4 @kathyccheng 

[...]

PR #960, opened 2021-01-21
Add dynamic preferences type to Application entity's ApplicationPreference property

pbn4:
Addresses https://github.com/bloom-housing/bloom/issues/948


pbn4:
I decided to tackle the tech debt we have in Application preferences (backend side for now, frontend tbd). Based on the requirements:

R1. Preference related to a Listing is an option an applicant can choose in Application submission form e.g. a `liveOrWorkIn` preference, which is defined on a Listing as follows:
```
  "preferences": [
    {
      "ordinal": 1,
      "title": "Live or Work in Hayward",
      "subtitle": "",
      "description": "At least one member of my household lives in City of Hayward. At least one member of my household works in the City of Hayward",
      "links": []
    },
...
```
R2. We want to easily correlate user chosen preferences (those existing in ApplicationPreference entity) to preferences living in Listing entity. 

I would like the Application form page `preferences/select.tsx` to be dynamically created based on the aforementioned "preferences" array, not on application form config. Here's how:

- Each item in the array should generate a Form input on `preferences/select.tsx`, and the input object schema should be determined based on a new property 'type' added to Preference:
   ```
   export enum PreferenceType {
     liveOrWork = "liveOrWork",
     displaced = "displaced",
   }
   ```
   so now the "preferences" array looks as follows: 

   ```
     "preferences": [
       {
         "ordinal": 1,
         "type:" "liveOrWorkIn",
         "title": "Live or Work in Hayward",
         "subtitle": "",
         "description": "At least one member of my household lives in City of Hayward. At least one member of my 
   household works in the City of Hayward",
         "links": []
       },
   ...
   ```
- an interface for Application.preferences has changed to an *array* of type ApplicationPreference, example value:
   ```
   [{
       "preference": {"id": listing.preference.id},
       "data": {"type": "liveOrWorkIn", "liveIn": true, "workIn": true}
   }]
   ```
   where "preference" is a reference to a specific Preference from a corresponding Listing, `data` is of union type that is determined based `type` property (of type `PreferenceType`, the same used in `Preference.type` property, so this is easy to determine from Listing data). Union types are: `LiveOrWorkPreference` and `DisplacedPreference`: 

   ```
   export class BasePreference {
     type: PreferenceType
   }

   export class LiveOrWorkPreference extends BasePreference {
     liveIn: boolean
     workIn: boolean
   }

   export class DisplacedPreference extends BasePreference {
     name: string
     address: AddressCreateDto
   }
   ```

Pros:
1. Form is dynamically created based on Listing data, single source of truth
2. Everything is typed, less errors on both backend and frontend
3. Validators are working for this union type too (based on 'type' property), less corrupted data inserted by users
Cons:
1. Requires reimplementing `preferences/select.tsx` page

@jaredcwhite Since you are the author of `preferences/select.tsx` page could you help me and take a look at what should be done here frontend-wise? What do you think about the approach? 

@Jesse I see some upcoming problems with this dynamic form creating e.g. how "preferences" are now defined in Sohay listing: 

```
  "preferences": [
    {
      "ordinal": 1,
      "title": "Live or Work in Hayward",
      "subtitle": "",
      "description": "At least one member of my household lives in City of Hayward. At least one member of my household works in the City of Hayward",
      "links": []
    },
    {
      "ordinal": 2,
      "title": "Displaced Tenant Housing Preference",
      "subtitle": "",
      "description": "At least one member of my household was displaced from a residential property due to redevelopment activity by the Hayward Housing Authority, the Redevelopment Agency or the City of Hayward.",
      "links": []
    },
    {
      "ordinal": 3,
      "title": "Mission Boulevard Corridor (SR 238) Displaced Tenant Housing Preference",
      "subtitle": "",
      "description": "At least one member of my household is an SR 238 Program Participant as required under the terms of the SR 238 Settlement Implementation Project. Program Participants are tenants of previously sold Caltrans properties along the SR 238 Mission Boulevard Corridor and are given occupancy preference when new below-market housing is constructed in the Corridor.",
      "links": [
        {
          "title": "Read More",
          "url": "https://www.hayward-ca.gov/238/background"
        }]
    }
  ],
```

if we want to dynamically create an Application form for those then we would have 3 separate items (because there are 3 preferences on a Listing). IIRC we treat 2nd and 3rd items in the array  as related (displacement) because we want to retrieve new data if *either of them* is selected. Can we merge those two preferences to be a single item? The way the are displayed on Listing page will change, but we will have our dynamic form nicely working. Note that this Sohay preferences I copy-pasted does not contain 'type' becaue it's from 'housingbayarea' fork and I didn't update it yet.

software-project (approved):
I think it's a good approach. Nice work.

pbn4:
I redefined schema after on call discussion. Preferences on a Listing Entity define metadata for application form:
```
  preferences: [
    {
      description: "Description",
      links: [
        {
          title: "Link title",
          url: "",
        },
      ],
      ordinal: 1,
      subtitle: "Subtitle",
      title: "Title",
      formMetadata: [
        {
          type: InputType.boolean,
          label: "Live In",
          key: "liveIn",
        },
        {
          type: InputType.text,
          label: "Some note input",
          key: "note1",
        },
        {
          type: InputType.address,
          label: "Address",
          key: "displaceeAddress",
        },
      ],
    },
  ],
```

and the corresponding form output in ApplicationPreference array (Application entity) is now: 

```
  preferences: [
    {
      description: "Description",
      links: [
        {
          title: "Link title",
          url: "",
        },
      ],
      ordinal: 1,
      subtitle: "Subtitle",
      title: "Title",
      formMetadata: [
        {
          type: InputType.boolean,
          label: "Live In",
          key: "liveIn",
        },
        {
          type: InputType.text,
          label: "Some note input",
          key: "note1",
        },
        {
          type: InputType.address,
          label: "Address",
          key: "displaceeAddress",
        },
      ],
    },
  ],
```

jaredcwhite:
@pbn4 I have to admit, I'm confused about what the application data should be as it comes from the frontend. Right now, the preferences are saved to the Application like this:

```
"preferences": {
  "liveIn": false,
  "workIn": true,
  "none": false
}
```

So it' just key=true/false for each preference. I was imagining it would simply be expanded with additional metadata, somewhat like this:

```js
"preferences": {
  "liveIn": {
    "enabled": false
  }
  "workIn": {
    "enabled": "true",
    "address": {
      "street": "12345 Anywhere St.",
      // etc.
    }
  },
  "none": false
}
```

Similarly, the form right now is configured just with a basic key array: `["liveIn", "workIn"]`

I would assume it gets expanded to something like this:

```js
[
  {
    "key": "workIn",
    "formMetadata": {
      key: "displaceeAddress",
      type: InputType.address
    },
  },
  // etc.
]
```

In all cases, no labels are required, only keys so we can use the translation files.

Maybe there's something else you're trying to account for with a more advanced schema that I'm not aware of, but I think anything more complex than what's presented above may take quite a while to implement.

pbn4:
Yeah I see a copy pasted wrong application output. Sorry for that.This is a preference defined in the Listing (I removed labels):
```
  preferences: [
    {
      description: "Description",
      links: [
        {
          title: "Link title",
          url: "",
        },
      ],
      ordinal: 1,
      subtitle: "Subtitle",
      title: "Title",
      formMetadata: [
        {
          type: InputType.boolean,
          label: "Live In",
          key: "liveIn",
        },
        {
          type: InputType.text,
          label: "Some note input",
          key: "note1",
        },
        {
          type: InputType.address,
          label: "Address",
          key: "displaceeAddress",
        },
      ],
    },
  ],
```

it means that for this preference we want an an applicaton form with a boolean input, a text input and an address type input. The corresponding application data would be:

```
    body.preferences = [
      {
        data: [
          {
            type: InputType.boolean,
            key: "liveIn",
            value: true,
          },
          {
            type: InputType.text,
            key: "note1",
            value: "some text",
          },
          {
            type: InputType.address,
            key: "displaceeAddress",
            value: {
              street: "",
              street2: "",
              city: "",
              state: "",
              zipCode: "",
            },
          },
        ],
        preference: {
          id: listing[0].preferences[0].id,
        },
      },
    ]
```

So for each listing  preference there is a corresponding application data preference in `application.preferences` array and for each formMetadata item from Listing.preferences[].formMetadata array there is a corresponding item in 'data' field of application preference. 


So make the example complete here is what I think Sohay preference will look like in final version. Listing.preference: 
```
  "preferences": [
    {
      "ordinal": 1,
      "title": "Live or Work in Hayward",
      "subtitle": "",
      "description": "At least one member of my household lives in City of Hayward. At least one member of my household works in the City of Hayward",
      "links": [],
      "formMetadata": [
        {
          "type": InputType.boolean,
          "key": "liveIn",
        },
        {
          "type": InputType.boolean,
          "key": "workIn",
        },
        {
          "type": InputType.boolean,
          "key": "none",
        }
      ]
    },
    {
      "ordinal": 2,
      "title": "Displaced Tenant Housing Preference",
      "subtitle": "",
      "description": "At least one member of my household was displaced from a residential property due to redevelopment activity by the Hayward Housing Authority, the Redevelopment Agency or the City of Hayward.",
      "links": [],
      "formMetadata": [
        {
          "type": InputType.text,
          "key": "name",
        } ,
        {
          "type": InputType.address,
          "key": "displaceeAddress",
        } 
      ]
    }
  ],
```

and the corresponding output:

```
    body.preferences = [
      {
        data: [
          {
            type: InputType.boolean,
            key: "liveIn",
            value: true,
          },
          {
            type: InputType.boolean,
            key: "workIn",
            value: true,
          },
          {
            type: InputType.boolean,
            key: "none",
            value: false,
          }
        ],
        preference: {
          id: correspondingLiveWorkPreferenceId
        },
      },
      {
        data: [
          {
            type: InputType.text,
            key: "name",
            value: "name",
          },
          {
            type: InputType.address,
            key: "displaceeAddress",
            value: {addressObject},
          },
        ],
        preference: {
          id: correspondingDisplacedPreferenceId
        },
      }
    ]
```

From what I understand you said about translations keys I think that we need an additional key called "descriptionTranslationKey" and the label also needs to be somehow determined from the data e.g. a labelTranslationKey? Is that clear now?

pbn4:
Form metadata for latest schema
```
"preferences": [
    {
        "ordinal": 1,
        "title": "Live or Work in Hayward",
        "subtitle": "",
        "description": "At least one member of my household lives in City of Hayward. At least one member of my household works in the City of Hayward",
        "links": [],
        "formMetadata": {
            "key": "liveWork",
            "options": [
                {
                    "key": "live",
                },
                {
                    "key": "work",
                },
            ]
        },
    },
    {
        "ordinal": 2,
        "title": "Displaced Tenant Housing Preference",
        "subtitle": "",
        "description": "At least one member of my household was displaced from a residential property due to redevelopment activity by the Hayward Housing Authority, the Redevelopment Agency or the City of Hayward.",
        "links": [],
        "formMetadata": {
            "key": "displacedTenant",
            "options": [
                {
                    "key": "general",
                    "extraData": [
                        {
                            "key": "name",
                            "type": InputType.text,
                        },
                        {
                            "key": "name",
                            "type": InputType.address,
                        }
                    ]
                },
                {
                    "key": "missionCorridor",
                },
            ] 
        }
    }
],
```

jaredcwhite (commented):
@pbn4  You might still be working on this…I just wanted to make note that the `blankApplication` starting point for `preferences` now is just a blank array. There isn't an object now and we don't need a `none` key for example.

Looking good otherwise I think!

PR #962, opened 2021-01-21
Improve validators

pbn4:
It's not related to any existing issue but it's just a quick improvement to our API, stricter control over incoming that so someone talking directly to the API e.g. with Postman is not able to send huge blobs of data and basically trash our DB.


dominikx96 (approved):
LGTM 👍

PR #966, opened 2021-01-22
Routing updates

dominikx96:
Fixed "not found" error on the netlify partners instance.

Basically, Next.js support 2 main modes: SSR (server side) and SSG(static generated content). We can't use SSR, because in Netlify it's not possible.

For admin management purposes, we need to fetch the data dynamically and create routes. e.g application details dynamically without build (SPA).

We have 2 possibilities here:

**1) (Implemented in this PR, recommended): Change paths to query params:**

[OLD]:
/listings/[ID]/applications
/listings/[ID]/applications/add
/applications/[ID]
/applications/[ID]/edit

[NEW]:
/listings/applications?listing=XXX
/listings/applications/add?listing=XXX
/application?id=XXX
/application/edit?id=XXX

2) Use SPA routing system

We can also add e.g. react-router library to detect all routes and show requested page.
Cons:

- Problem with next/link, instead of this we need to use react-router methods
- Netlify updates - we need to create redirect to index.html to use react-router

I think we should keep 1 approach if URL structure is fine for admin purposes.


jaredcwhite:
@dominikx96 Hmm, this isn't working on Netlify as-is. Saw an unstyled "Loading…" screen, routes weren't acting right on clicks, then I got logged out.

pbn4 (approved):

slowbot (approved):
@dominikx96 LGTM

jaredcwhite:
I'm still getting weird behavior here, but I'll file an issue and meanwhile merge this in so it's more demoable.

Issue #808, opened 2020-11-03
Partners can edit applications 


kathyccheng:
As a partner 
I want to edit an existing paper application 
When I'm on the application detail page for a submitted application
I expect to see an Edit button
When I click on the Edit button 
Then I'll see the edit details page with a Save and Exit option and a cancel option 
When I made edits and click Save and Exit
Then I'll be directed to the detail view
And I'll the last updated field updated 
When I click cancel
Then I'll be taken back to the details page 
If I had edited any of the fields and clicked cancel
Then it will not be saved 

![image](https://user-images.githubusercontent.com/3453407/101066937-c5a9f000-354b-11eb-9816-d3ab8f028335.png)
![image](https://user-images.githubusercontent.com/3453407/101066989-d5293900-354b-11eb-8999-c37d601c8a0b.png)

Notes 
* We'll want to track who made the edits and when

PR #968, opened 2021-01-25
Edit Application

dominikx96:
Closes: #808

Should be merged after #966


pbn4 (approved):
Looking good. A nitpick: "yes" "no" fixed string could be replaced with YesNoAnswer enum.

dominikx96:
@pbn4 fixed!

dominikx96:
@slowbot 1,2 is ready.

3 - I think we need to modify an alert component to show in the "fixed" position. Can we create a different story for this?
4 - you mean, additional button to save the application as "draft", correct? - I think we also create a separate story and resolve it in the different PR.

dominikx96:
@slowbot header styles updated

slowbot (approved):
@dominikx96 I think this is good now. 

There is another issue with back button routing behavior but I think we can resolve ina future story.

Issue #974, opened 2021-01-28
Show Application Type Column in Applications List 


kathyccheng:
As a housing developer
I want to see Application Type in the list of all applicaitons - Digital, Paper

PR #979, opened 2021-02-01
Added application type column

dominikx96:
Closes: #974

Instead of "Digital" (like in the issue description), I added "Electronic" as in the design.
@slowbot let me know if we should use "Digital" instead.


pbn4 (approved):

slowbot (commented):
@dominikx96 Functionally I think this looks good. Can we make this column sortable?

@kathyccheng What do you think about column placement? Initially I had intended this to be further to the right of declared income or subsidy but seeing it here might make sense during the application entry process. Either way we can treat the pilot as a test and ask the developer their preference.

dominikx96:
@slowbot made this column sortable. It has a default arrow icon like the other cols.

slowbot (approved):
@dominikx96 Looks good! I added the sort none icon state to distinguish sortable from unsortable columns

PR #985, opened 2021-02-03
Preferences screen/schema reworked, Displaced Tenent added

### LINKED ISSUE #948: Preferences Page in Application

jaredcwhite:
Resolves #948

There some gnarly bits of code and typing I'd like to clean up, but this should prove useful for review and for similar backend/partners-related work.


jaredcwhite:
@kathyccheng I think this is more-or-less the general/English translation keys we'll need for the Alameda app. As far as I know, we don't have translations in the other 3 languages. Is that something we'd be able to roundtrip before our go-live date?

```
    "preferences": {
      "liveWork": {
        "live": {
          "label": "Live in City of Hayward Preference",
          "description": "At least one member of my household lives in the City of Hayward.",
        },
        "work": {
          "label": "Work in City of Hayward Preference",
          "description": "At least one member of my household works or has been offered work in the City of Hayward.",
        }
      },
      "displacedTenant": {
        "whichHouseholdMember": "Which household member is claiming this preference?",
        "whatAddress": "What address was the household member displaced from?",
        "general": {
          "label": "Displacee Tenant Housing Preference",
          "description": "Displacee Tenant Housing Preference: At least one member of my household was displaced from a residential property due to redevelopment activity by the Hayward Housing Authority, the Redevelopment Agency or the City of Hayward."
        },
        "missionCorridor": {
          "label": "Mission Boulevard Corridor (SR 238) Displacee Tenant Housing Preference",
          "description": "At least one member of my household is an SR 238 Program Participant as required under the terms of the SR 238 Settlement Implementation Project. Program Participants are tenants of previously sold Caltrans properties along the SR 238 Mission Boulevard Corridor and are given occupancy preference when new below-market housing is constructed in the Corridor. For more information on the SR 238 Settlement Implementation Project, please visit:",
          "link": "https://www.hayward-ca.gov/238/background"
        }
      }
```

slowbot (commented):
Bulk of this looks good.

1. Looks like there is an error on the displaced screen for mission corridor the title is rendering incorrectly.
2. We can update the copy for opt out on both screens to say "I don't want these preferences"
3. Typo on first displaced preference. Currently reads "Displacee"

kathyccheng:
@jaredcwhite we do have the translations, I currently have them captured here: 
https://github.com/housingbayarea/bloom/issues/203

jaredcwhite:
OK cool. I think I need to do most of that in the housingbayarea repo rather than the bloom repo.￼￼ Any sense of when the new Alameda branch will be ready for me to work on?￼

sites/public/pages/applications/preferences/displaced.tsx
```diff
@@ -0,0 +1,307 @@
...
+                    }, 1)
+                  }}
+                />
+                <label htmlFor={option} className="font-semibold uppercase tracking-wider">
+                  {t(`application.preferences.displacedTenant.${option}.label`)}
+                </label>
+              </div>
+
+              <p className="ml-8 -mt-3">
+                <button
```
dominikx96, line 122: I think will be good to create a separate component for this or add a variant to the existing component (we have the same button in 2 steps).

dominikx96 (changes requested):
- I'm not sure, but when a user selects checkbox and address fields are visible - currently they're not validated (user can fill partially), should we require all fields in this situation?

- I noticed address information from preferences are missing in the summary

+ some suggestions about code :)

jaredcwhite:
> * I'm not sure, but when a user selects checkbox and address fields are visible - currently they're not validated (user can fill partially), should we require all fields in this situation?
> * I noticed address information from preferences are missing in the summary

Good comments all, I'll work on those. Types from backend would be good as well, but I don't have them yet. Will wait until the final sync-up with @pbn4's work.

jaredcwhite:
All right, I addressed some feedback and fixed some bugs, so this should be ready for a backend merge event. I would suggest that if we still have any glaring gaps in functionality and code quality we should perhaps file those as new issues after merge so we're on a good QA cadence. FYI @slowbot I'll fix the translation string issue with the mission corridor option in the housingbayarea repo once we copy this downstream.

jaredcwhite:
Oh I forgot to mention — I've added in a configurable idle timeout. So you don't have to see your application form get blown away every 5 minutes, forcing you to start over when testing locally! 🤯

jaredcwhite:
All right I'll work out the merge conflicts and then get this in￼.

jaredcwhite:
It looks like this will block the public Next site from building due to a mismatch in the application save typing, so will have to wait for @pbn4's updates.

pbn4:
@jaredcwhite Yes I will fix your branch as soon as a I merge content from a branch where Dominik is working on partners portal fuctionality

pbn4:
@jaredcwhite Backend part is done for this PR. It seems that we are missing some translation keys and cypress tests rewrite for the new page. Once this is done it's ready to be merged.

jaredcwhite:
Thanks @pbn4! I'll try to fix up the Cypress tests and then we can merge in. :+1:

jaredcwhite:
OK, Cypress ran and pass all except one, but the one was maybe just a weird timeout because that test works for me locally. So I think we're good. (Also Cypress is crazy slow on CircleCI…mentioned on Slack…)

PR #994, opened 2021-02-08
Fix Listing query performance by making some of the foreign relations…

pbn4:
… a jsonb column

Relations: ListingEvent, Assets, ApplicationMethods and AmiChartItem.
Since ListingEvent, Assets and ApplicationMethods are no longer tables this commit also removes their corresponding controller from the API. Now all changes to those data structures should be done using PUT /listings


backend/core/src/ami-charts/entities/ami-chart-item.entity.ts
```diff
@@ -1,50 +1,16 @@
...
-} from "typeorm"
-import { Expose, Type } from "class-transformer"
-import { IsDate, IsNumber, IsString, IsUUID } from "class-validator"
-import { AmiChart } from "./ami-chart.entity"
+import { Expose } from "class-transformer"
+import { IsNumber } from "class-validator"
 import { ValidationsGroupsEnum } from "../../shared/validations-groups.enum"
 
-@Entity()
 export class AmiChartItem {
```
software-project, line 5: Shouldn't this entity be deleted as well?

software-project (approved):

Issue #972, opened 2021-01-27
Preference Fields to Paper Application Form in Partners


slowbot:
As a leasing agent, I want to be able to add Displaced Preference fields for Name and Address to the Add Application in Partners. 

Relates to https://github.com/bloom-housing/bloom/issues/948

kathyccheng:
Closed in #997

PR #997, opened 2021-02-08
Dynamic Partners Preferences

dominikx96:
Closes: #972

Created dynamic preferences depend on listing preferences schema.

@slowbot I suppose you need to run `db:reseed` before testing.


slowbot (changes requested):
@dominikx96 Mission corridor also needs to prompt user for Name and Address same as General Displaced.

dominikx96:
@slowbot 1 and 4 are related to backend changes. I talked with @pbn4 and applied updates.
Applications should be editable and mission corridor option should have name and address fields.

2, 3 - should be applied for both **general** and **mission corridor** options?

slowbot (commented):
@dominikx96 See comments

sites/partners/lib/formatApplicationData.ts
```diff
@@ -232,10 +288,42 @@ export const parseApplicationData = (applicationData: ApplicationUpdate) => {
 
   const phoneNumber = applicationData.applicant.phoneNumber
 
+  const preferences = (() => {
```
jaredcwhite, line 291: @dominikx96 Is there a way this block of code could be expressed in a more English-like readable manner, or at the very least provide some comments? I'd have to read this over a number of times before generally understanding what it's trying to accomplish.

sites/partners/src/applications/PaperApplicationForm/sections/FormPreferences.tsx
```diff
@@ -1,66 +1,130 @@
...
+  )
+
+  return (
+    <GridSection title={t("application.details.preferences")} separator grid={false}>
+      <GridSection columns={2}>
+        {preferences?.map((preference) => (
+          <GridCell key={preference.id}>
+            <ViewItem label={preference.title}>
+              <fieldset className="mt-4">
+                {preference.formMetadata?.options?.map((option) => {
```
jaredcwhite, line 82: Could this section and the memo/callback blocks above be refactored into regular functional components? It might clean up some of the procedural nature of the code here.

jaredcwhite (changes requested):
@dominikx96 Code review looks good, just had a couple suggestions for places where some refactoring for clarity might be warranted. Let me know what you think.

sites/partners/src/applications/PaperApplicationForm/sections/FormPreferences.tsx
dominikx96: Definitely makes sense! 👍  I think we can store this component in the same file (logic is highly connected to this component).

sites/partners/lib/formatApplicationData.ts
dominikx96: @jaredcwhite agree! I refactored function names to be more intuitive, renamed variables, and added some comments. I think will be easier when we'' have backend types for this, I temporary typed it as:

`const preferencesApiData = applicationData.preferences as Record<string, any>`

because `object` is not a valid type. Talked with @pbn4 and we should change it after production data migration to a proper format.

Hope the current version of the code is more readable :)

dominikx96:
@slowbot I applied changes.

Can we create separated ticked for this? 

> Can name choose form existing household members via dropdown? (lower priority)

It's a little bit related to the backend implementation because listing preference provides us 3 types of fields (text, boolean, address). Depends on this type I can render the proper field. So I think to do this in an acceptable way, we need to update the backend as well. Do we want it now or post-launch is still okay?

pbn4:
@dominikx96 @slowbot Please lets put any other changes outside of this PR because I would like to it be merged as soon as possible.

slowbot (approved):
@dominikx96 LGTM

I'll create a new ticket for the last item
- Can name choose form existing household members via dropdown? (lower priority)

PR #1000, opened 2021-02-10
Fix importer and styles for release

### LINKED ISSUE #998: Fix listing importer
Fix the importer

software-project:
Resolves style issues and fixes #998

### DIFF
--- a/backend/core/listings-importer.ts
+++ b/backend/core/listings-importer.ts
@@ -30,10 +30,11 @@ const instance = axios.create({
 
 serviceOptions.axios = instance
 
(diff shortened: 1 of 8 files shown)


pbn4 (approved):

jaredcwhite:
@software-project Hey just a few stylistic things I noticed:

* React props that are booleans typically don't use `isSomething={true}`, but just `something={true}`.
* CSS class names that modify an element appearance should probably be in the `.is-something` format, rather than `.something`.
* There's a typo in Hero–it's setting `is-cetered` rather than `is-centered`.

PR #1001, opened 2021-02-11
tests/fix errors in ui-component test suite

emilyjablonski:
I went to start writing the new unit test suite, and noticed the current snapshot suite was generating a lot of warnings and errors when it was being run.

It was hard to parse the output, so I just fixed all the warnings before moving onto writing new tests. Most of the issues were either a lack of a unique key in a component, or props that had just changed.

Would love @jaredcwhite your input here! Open to feedback / new to this bit and not sure if there are criteria for how stories need to display, and since this fixed some of them there are some new ones displaying.

I also had to touch one file in /core for typing issues and want to make sure there are no other implications.

GIPHY


emilyjablonski:
Holy batman that's a lot of diffed characters. I'm thinking it's just because we have a lot of snapshots?

software-project:
It looks nice. Happy to see less warnings and errors ;) I'm just wondering regarding null vs undefined thing. What's the reason why that makes a difference? And what's real life data that we are providing there? I just want to make sure we are fixing real problem, not just tests.

emilyjablonski:
Yeah sure thing that's my concern w the typing change as well. For context this is the error that was being produced for each null: 
`Unexpected error while loading ./forms/TimeField.stories.tsx: TSError: ../backend/core/types/src/Member.ts:9:3 - error TS2416: Property 'birthMonth' in type 'Member' is not assignable to the same property in base type 'HouseholdMemberUpdate'.`
`Type 'null' is not assignable to type 'string | undefined'.`
`Line 9: birthMonth = null`

My struggle was understanding why TimeField was triggering this, but since the `Member` class implements the `HouseholdMemberUpdate` class, and those fields are not typed as `string | null` just as `string`, then setting those values to null is invalid. The alternative I think would be to change the `HouseholdMemberUpdate` class to type things explicitly as `string | null` but that isn't as safe from a typing perspective.

jaredcwhite:
@emilyjablonski Looks good to me too. I think I know why you had to remove the `javascript:` hrefs for the TabNav stories, but it would be nice to have some kind of visual feedback when a tab is activated, rather than just trigger the # anchor.

emilyjablonski:
Yeah it's considered "unsafe" (which it is to use in production) but it definitely shouldn't matter in a story. I just couldn't find a way to make it happy about it / ignore it. My thought was unit testing that portion in the new suite, is that okay? If not I can keep poking for a different visual indicator solution :)

emilyjablonski:
I've got a solution for the above using Storybook actions, will push it up soon!

jaredcwhite:
Nice work @emilyjablonski! LGTM :+1:

emilyjablonski:
Touched based with Jared, waiting until after the pilot launch to merge 👍

emilyjablonski:
Spoke with Jared offline and in the testing call this morning, consensus was good to merge! :)

PR #1005, opened 2021-02-12
Skip Live Alone page when there are already multiple household members

### LINKED ISSUE #999: QA: Household Members Need to be Re-added When Editing Public App
What happens now:
As a user if I edit a field on the application prior to Household Members the application prompts me to choose live alone or household members

What I expect to happen:
As a user, if Ive added household members and I go back to a page in the app before Household Members section I would expect to skip the "Live Alone" screen and see the Household Members screen with all household members listed.

image

image

jaredcwhite:
Resolves #999 

**Note:** this also updates recent migrations so the jsonb columns default value is an array: `[]`


slowbot (approved):
@jaredcwhite  LGTM

PR #1010, opened 2021-02-12
Fix public lottery and add multiple open houses

software-project:
Fixes date formatting issue for public lottery and adds option to add multiple open house events.


emilyjablonski:
LGTM apart from the missing photo, but may just be a consequence of testing it here and not in housingbayarea

software-project:
> LGTM apart from the missing photo, but may just be a consequence of testing it here and not in housingbayarea

Missing photo is expected as it's added on the housing bay repo. Thanks for the review

PR #1011, opened 2021-02-12
tests/ui-components new unit test suite

emilyjablonski:
In this PR I set up react-testing-library for ui-components, and write one test suite (actions) as an example. I didn't want to continue to write the rest of the unit tests in case the team doesn't agree with this implementation for any reason.

I personally have found functional unit tests like these to be more valuable than snapshot tests as you can pinpoint specific functional cases you want to test and automate them, vs updating static output changes in a snapshot & testing the functional bit manually. It's more obvious more quickly if you've broken something, and the test name should be a nice specific indicator as to what is broken, which is why I'd like to add them! :)

If you run `yarn test:coverage` it will run all but the snapshot tests and produce coverage reports that indicate exactly which lines and branches have been hit by the unit tests.

Coverage report for these tests:
src/actions                  |     100 |      100 |     100 |     100 |
     Button.tsx                  |     100 |      100 |     100 |     100 |
     ExpandableText.tsx   |     100 |      100 |     100 |     100 |
     LinkButton.tsx           |     100 |      100 |     100 |     100 |
     LocalizedLink.tsx      |     100 |      100 |     100 |     100 

Would love some input & everyone's blessing to move forward with the new test suite, super open to feedback.

GIPHY

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,6 @@ yarn-error.log
 
 # VS code debugger config
 launch.json
+
+# Jest unit test coverage
+test-coverage/
--- a/ui-components/.jest/setup-tests.js
+++ b/ui-components/.jest/setup-tests.js
@@ -1,4 +1,5 @@
 import registerRequireContextHook from "babel-plugin-require-context-hook/register"
+import '@testing-library/jest-dom/extend-expect'
 
 import { configure } from "enzyme"
 import Adapter from "enzyme-adapter-react-16"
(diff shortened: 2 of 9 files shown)


jaredcwhite:
Thanks for taking this on @emilyjablonski, nicely done! Seems like a good approach.

I think there are a few considerations we'll want to discuss as a team about how we want to manage these new component tests.

1. Are we aiming for complete coverage, or just a minimum amount of coverage?
2. For simplest of cases is the maintenance overhead of unit tests worth it, considering we'll always want to write stories for the visual preview? (The question then is why would I want to write two different tests for one component?)
3. Do we block acceptance of a PR for a new component if it doesn't have a unit test?
4. Do we block acceptance of a PR for a change to an existing component if it didn't previously have a unit test?
5. If I were to go in and make a change to a component and the unit test breaks, I may not immediately understand why. What is the procedure for reviewing or communicating with unit test authors?

I don't have any particular opinions to offer, other than on (1): I don't think complete coverage is a realistic goal for all of our components or all future components. Possibly we'd want to focus mainly on the ones with higher complexity or multiple interactive elements.

Issue #763, opened 2020-10-14
CSV Exporter that includes demographics data 


kathyccheng:
Can we have a separate CSV exporter that includes everything in the current applicant export but also includes all of the optional demographics fields on the last page of the application, the fields below: 

* ethnicity
* race 
* gender
* sexual orientation
* how did you hear about this property - comma separated list of all options selected is fine

kathyccheng:
@pbn4 an you also prep a script to run an export for everything including the demographics above? We'll need to run this once all of the applications are in, but last time we missed the racial demographic field so wanted to test this script with sample data

pbn4:
@kathyccheng What do you mean exactly by "separate"? Right now we have an export button in partners portal, should there be a new button the partners frontend?

kathyccheng:
Nope! This shouldn't be user facing. We just need a manual way to pull all of the data + the optional demographics. I use that data to then do some manual data reporting for the developers + cities. We don't show the person by person demographics data in the Partners Portal because it could bias them so we obfuscate it and right now, we aggregate it manually for them. 

We didn't run this CSV export until after the applications closed on Jones. But then we realized that we were missing a column of data for demographics. So I'd like to run this on staging before we launch just to make sure we have all of the data and then it's a trial run before we have to do it in a month.

kathyccheng:
@pbn4 can you do a sample pull of this for me now with the sample data we have on staging tied to SoHay just so I can make sure all the fields are there?

pbn4:
Done on slack

kathyccheng:
Closed in #1031

PR #1031, opened 2021-02-22
Add includeDemographics query param to CSV exporter

pbn4:


pbn4:
Closes https://github.com/bloom-housing/bloom/issues/763

PR #1037, opened 2021-02-24
Improve preferences UI functionality

### LINKED ISSUE #1009: QA Save and return to review link flashes on last page of preferences
What is happening now
Save and return to review displaying on final page of preferences.

What I would expect too happen
No save and return to review action would display

image

jaredcwhite:
Fixes #1009 

Also fixes I bug I saw where depending on how you selected various preferences, it wouldn't advance the progress nav.


slowbot (approved):
LGTM, not sure if I tested all of the edgecases

jaredcwhite:
It's works better than before at least, so let's rolllll 😄

Issue #931, opened 2021-01-06
Translate listings household size header 


kathyccheng:
From @jaredcwhite 
OK I think I got all the issues here resolved except for the headers on the HMI table…it's a little funky in that the summary data is coming from the backend so we need to (a) change the backend text to programmatic labels instead of "regular" copy, and then (b) add the translation strings for those to all 4 language files.

![image](https://user-images.githubusercontent.com/3453407/103720426-2af46500-4f80-11eb-9711-09a51e6c95d6.png)

kathyccheng:
@jaredcwhite these are the translations 

English:
"household_size": "Household Size",
"max_income_month": "Maximum Income / Month",
"max_income_year": "Maximum Income / Year",

Spanish
"household_size": "Tamaño del hogar",
"max_income_month": "Ingresos máximos / Mes",
"max_income_year": "Ingresos máximos / Año",

Vietnamese
"household_size": "Quy mô Hộ Gia đình",
"max_income_month": "Thu nhập Tối đa / Tháng",
"max_income_year": "Thu nhập Tối đa / Năm",

Chinese
"household_size": "家庭人數",
"max_income_month": "最高收入 / 月",
"max_income_year": "最高收入 / 年",

PR #1051, opened 2021-03-04
Fix backend side translation string

pbn4:
Closes https://github.com/bloom-housing/bloom/issues/931 Previous PR I opened was an incorrect implementation.


dominikx96 (approved):
LGTM

Issue #951, opened 2021-01-14
Partners Listing Header and Local Navigation


slowbot:
As a leasing agent I want to be able to see the listing that I am on and navigate between all incoming applications and flagged applications. 

- Include listing name
- Include tab navigation (component in [storybook](https://storybook.bloom.exygy.dev/?path=/story/prototypes-tabnav--default))
   - Tab navigation should default to applications
   - Flags tab include a table of applications that have been marked as potential duplicate applications (stub for now)
   - Flags tab should include a count of number of unresolved flagged sets  (stub for now)

![image.png](https://images.zenhubusercontent.com/5ed544f1e9914efa39671dda/54b59295-76f2-42db-88e6-dd76b4965009)

jaredcwhite:
Dependency: #697 (tabnav component)

dominikx96:
Merged to `feature/777-duplicate-application-wip`

Issue #777, opened 2020-10-21
Partners: Flagged Applications Table


slowbot:
**Acceptance Criteria**:
- As a partner, I want to be made aware of a potential set of duplicate applications
- Alongside the applications tab should be a tab to house flagged application sets
   - The flagged applications tab should have a count indicating the number of unresolved flagged application sets there are
   - When the number of flagged application sets is 0 the count in the tab is not present
- Duplicate flags should reflect a status of "Flagged" in the flagged applications table
- A duplicate flag should be attached to a set of applications if the meeting the following criteria
   - They have a the same email address 
   - They have an applicant or household member with the same name or DOB
   - They have the same residential address - **removing from requirements**

![image](https://user-images.githubusercontent.com/13915/96772570-2c0df100-1398-11eb-8bcb-44487b60ef4f.png)

slowbot:
@netrabadami I want to add a "+n" to names if there is more than one primary name on an application set
![image](https://user-images.githubusercontent.com/13915/106058439-18b0a700-60a6-11eb-995a-3119e6edb3de.png)

netrabadami:
@slowbot Updates based on our conversations

PR #1078, opened 2021-03-17
Added flags navigation

dominikx96:
Closes: #951
It's highly connected with #777 (this branch includes necessary routes).

@netrabadami few things that I noticed during development:

- Your branch is not up to date with the master and you need to resolve conflicts.

- I had to use --no-verify flag to push changes because there are a lot of linter/type issues. You need to focus on this element and maybe check your IDE settings? Happy to support you with this element, just let me know on Slack :)
 
In the PR scope is tab navigation only. We need to connect flags number with Tag in the "Flags" tab when we'll have the final version of data in the frontend. Currently, it's just hardcoded 0. I added _TODO_ notes in these places.

### DIFF
--- a/sites/partners/pages/listings/applications/index.tsx
+++ b/sites/partners/pages/listings/applications/index.tsx
@@ -4,7 +4,6 @@ import moment from "moment"
 import Head from "next/head"
 import {
   Field,
-  PageHeader,
   MetaTags,
   t,
   Button,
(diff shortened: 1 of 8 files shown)


slowbot (commented):
@dominikx96 See comment, that may be coming later. Otherwise LGTM

dominikx96:
@slowbot that's because in the @netrabadami branch is hardcoded listing id - for testing purposes. I think @netrabadami will resolve it before merge :)

PR #1088, opened 2021-03-24
fixed mobile view of open houses section in public reference

netrabadami:
Added the OpenHouses component to Mobile view of Public reference site.


slowbot (changes requested):
@netrabadami See comment

netrabadami:
@kathyccheng @jaredcwhite @software-project  as per Mobiles components design specified by @slowbot , its not a good 
 idea to group the OpenHouse, ApplicationSection & Lottery components together. I believe its better that we keep the way it is and just add the missing OpenHouse component to Mobile view.

Would like to hear the comments/recommendations.

jaredcwhite:
@netrabadami Yeah I guess I didn't realize we already showed some of those components down in the "Process" section on mobile. So your original plan just to add open houses in the mobile section makes sense. Sorry about all the confusion!

jaredcwhite:
Wondering what the status is of this issue. @slowbot did we end up in the right place in terms of what's where on mobile vs. desktop?

kathyccheng:
@slowbot can you take a look and see what work still needs to be done?

PR #1089, opened 2021-03-24
Account audit

### LINKED ISSUE #726: Create Account Audit
See: https://github.com/bloom-housing/bloom/issues/242

What happens now: 
- Create account with an email already in use, nothing happens upon submit of form
- Create account form has one email input and one password input
- Create account automatically logs the user in and stays on create account screen

What should happen:
- Create account with an email already in use, should present error message: "Email is already in use."
- Create account should have re-enter email and password inputs
   - re-enter inputs should prevent copy and paste
   - Password fields should have helper text under label that reads “Must be at least 8 characters and include at least one number" 
- Create account should redirect to sign in screen with a modal that directs the user to confirm via email or resent email
- Successful confirmation via email should redirect to Dashboard
- If the user tries to submit the form without providing the required fields a validation message should appear at the top of the form as well as incline messages on each field
- If the applicant tries to login to their account before confirming their email address they will get the same Confirmation modal

image
image
image

software-project:
Resolves #726


backend/core/src/seed.ts
```diff
@@ -20,7 +20,11 @@ const newSeed = (): ListingSeed => {
 
 async function bootstrap() {
   const app = await NestFactory.createApplicationContext(SeederModule.forRoot({ test: argv.test }))
+  const userService = app.get<UserService>(UserService)
   const listing1 = await seedListing(app, newSeed())
+  listing1.leasingAgents.forEach((agent: User) => {
+    void userService.confirm({ token: agent.confirmationToken })
```
pbn4, line 26: confirm is an async action and doing it in a foreach loop will start the execution of each confirmation but will not await for it to complete, use `await Promise.all([])` syntax for it.

backend/core/src/user/user.service.ts
```diff
@@ -79,20 +81,47 @@ export class UserService {
     return savedPasswordHash === verifyPasswordHash
   }
 
+  public async confirm(dto: ConfirmDto) {
+    const user = await this.repo.findOne({ confirmationToken: dto.token })
+    if (!user) {
+      throw new HttpException(USER_ERRORS.TOKEN_MISSING.message, USER_ERRORS.TOKEN_MISSING.status)
+    }
+    user.confirmedAt = moment().toDate()
```
pbn4, line 89: It's just a nitpick but, that could be just `new Date()`

backend/core/src/user/user.service.ts
```diff
@@ -79,20 +81,47 @@ export class UserService {
...
+    user.confirmationToken = null
+
+    try {
+      await this.repo.save(user)
+      return user
+    } catch (err) {
+      throw new HttpException(USER_ERRORS.EMAIL_IN_USE.message, USER_ERRORS.EMAIL_IN_USE.status)
```
pbn4, line 96: Why the exception says EMAIL_IN_USE here?

backend/core/src/user/user.controller.ts
```diff
@@ -48,12 +49,19 @@ export class UserController {
...
   @UseGuards(OptionalAuthGuard, AuthzGuard)
   @ApiOperation({ summary: "Create user", operationId: "create" })
-  async create(@Body() dto: UserCreateDto): Promise<UserDtoWithAccessToken> {
+  async create(@Body() dto: UserCreateDto): Promise<UserDto> {
     const user = await this.userService.createUser(dto)
-    const accessToken = this.authService.generateAccessToken(user)
     // noinspection ES6MissingAwait
-    void this.emailService.welcome(user)
-    return mapTo(UserDtoWithAccessToken, { ...user, accessToken })
+    void this.emailService.welcome(user, dto.appUrl)
```
pbn4, line 55: I know this is not the case now in the codebase, but in the future I would like to move all the business logic inside services. So for example that emailService.welcome  would go inside userService. It's just something I'll try to clean up soon up and mentioning it here to spread the info.

backend/core/src/user/user.controller.ts
```diff
@@ -48,12 +49,19 @@ export class UserController {
...
+    return mapTo(UserDto, user)
+  }
+
+  @Put("confirm")
+  @ApiOperation({ summary: "Forgot Password", operationId: "forgot-password" })
```
pbn4, line 60: The summary should be "Confirm email", operationId "confirm"

pbn4 (changes requested):

backend/core/src/user/user.service.ts
```diff
@@ -79,20 +81,47 @@ export class UserService {
...
+    }
+  }
+
   public async createUser(dto: UserCreateDto) {
+    let user = await this.findByEmail(dto.email)
+    if (user) {
+      if (!user.confirmedAt) {
+        return user
```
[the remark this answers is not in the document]
software-project: This is only in case the account isn't confirmed, so we can resend email confirmation. We can either change controller to return text like `ok` or create a separate controller action to resend email confirmation. Which do you feel like is better?

backend/core/src/user/user.controller.ts
software-project: Good catch

backend/core/src/user/user.service.ts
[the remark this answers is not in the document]
pbn4: @software-project I think we will need resending a confirmation email anyway, so maybe let's go that way. I think resending should be for admins only right?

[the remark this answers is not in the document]
software-project: @pbn4 nope, it's part of that ticket to allow user to send email confirmartion again. There is a button to do it, but also if user just fills out the form again the way I had it prepared will reassure that account is not created twice and just email confirmation will be sent. So IMHO we just should restrict whats returned by the controller.

PR #1092, opened 2021-03-30
Add better support for applications search

### LINKED ISSUE #969: Application filter doesn't work properly
Filter by application data works only when a search phrase is exact as inside database e.g:

Existing application:
Name: Dominik Barcikowski

A user should be able to search by putting e.g. "Domi", not fully phrase.

Current behavior:
A user has to put the exact phrase.

Image 2021-01-25 at 6 04 14 PM

bpsushi:
Plain user input is now converted to tsquery with prefix support
Add tests for improved search

resolves #969


pbn4 (changes requested):
Hey @bpsushi this change stops searching e.g. the email address of applicant. Also I think the problem was that `john` wouldn't get a match when data contains `John` and actually `Joh` would be a successful partial match so I think the test also requires an adjustment to consider lower case.

PR #1101, opened 2021-04-02
Remove the Storyshots functionality from ui-components

jaredcwhite:
This follows up on #1050 by removing Storybook's Storyshots functionality from the test suite. We will still write stories to provide human-centric UI demos in Storybook, but it won't be connected to the test suite.

A worthwhile follow up from _this_ would be to redo some of the stories with knobs and other mechanisms so they work better as visual demos (now that automated testing isn't part of the equation).


emilyjablonski:
Looks great! It think the only remaining change is removing `--testPathIgnorePatterns storyshots.test.tsx` from the `yarn test:coverage` script in `ui-components/package.json`

jaredcwhite:
@emilyjablonski Done!

emilyjablonski (approved):

Issue #889, opened 2020-12-05
Add rate limiting to `POST /applications/submit`


pbn4:
We need to add rate limiting to prevent malicious users/bots from flooding our database with trash data.

emilyjablonski:
@pbn4 Do you have input on what strategies we should use for rate limiting so we are all on the same page?

bpsushi:
@emilyjablonski after my discussion with @pbn4 we see two possible approaches:

Approach 1: Rate limits based on user IP
This will be pure backend approach where rate limits are calculated for each IP address that sends the request. Frontend will only need to show proper message in case of reaching the limits. This approach will require from us to set up Redis - we cane use free plan of heroku addon for this (for now it should be enough and can be scaled in the future: https://elements.heroku.com/addons/heroku-redis). @pbn4 mentioned that setting up Redis is in backlog anyway.

Approach 2: Google reCaptcha v3
Classic approach where we implement Google reCaptcha. A little more frontend work will be required, but no need of any additional work like setting up Redis. Captcha will then take care of any flood tries.

dominikx96:
@kathyccheng we discussed with Bart about showing an error message on a frontend side.
"Rate limit exceeded, try again later." is okay? Do we want to add translations?

kathyccheng:
@dominikx96 Sure error message sounds good, we will want to add translations. I'll add that into the next batch for the translators

PR #1103, opened 2021-04-06
Feature/master applications rate limit

bpsushi:
resolves:  #889


backend/core/test/applications/applications.e2e-spec.ts
```diff
@@ -158,6 +159,11 @@ describe("Applications", () => {
         ListingsModule,
         ApplicationsModule,
         TypeOrmModule.forFeature([Application, HouseholdMember]),
+        ThrottlerModule.forRoot({
+          ttl: 60,
+          limit: 5,
+          ignoreUserAgents: [/^node-superagent.*$/],
```
pbn4, line 165: Why ignore node-superagent?

backend/core/.env.template
```diff
@@ -1,6 +1,7 @@
 PORT=3100
 NODE_ENV=development
 DATABASE_URL=postgres://bloom-dev:CHANGE-ME@localhost:5432/bloom
+REDIS_URL=redis://127.0.0.1:6379/0
```
pbn4, line 4: I just noticed that Heroku added two URLs to the container config for Redis and one of them is SSL, so I think the connection to redis will be over a public internet and we should use SSL in that case.

pbn4 (changes requested):

backend/core/test/applications/applications.e2e-spec.ts
bpsushi: This agent is used by all tests. Without this all tests hits rate limit and fails. Note that the test for rate limits i added in this PR is forcing different agent.

backend/core/.env.template
bpsushi: @pbn4 redis over TLS applied

PR #1106, opened 2021-04-06
feature/account settings

emilyjablonski:
Allows signed-in users to visit My Account --> Account Settings, where they will be able to edit their name, dob, email, and password.

Eventually, the email field will trigger a confirmation email flow, but that will be handled in a separate ticket.

Linked issue: https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/bloom-housing/bloom/729


sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+} from "@bloom-housing/ui-components"
+import Link from "next/link"
+import FormsLayout from "../../layouts/forms"
+
+type AlertMessage = {
+  type: AlertTypes
+  message: string
+}
+
+const Edit = () => {
```
jaredcwhite, line 28: This file is pretty lengthy with several different forms. Any way we could split this up into subcomponent files?

jaredcwhite (approved):
Looks good. I had one comment about breaking up a long file, but that doesn't seem like a blocker.

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+                  label={t("account.settings.newPassword")}
+                  placeholder={t("authentication.createAccount.mustBe8Chars")}
+                  validation={{
+                    minLength: MIN_PASSWORD_LENGTH,
+                    pattern: passwordRegex,
+                  }}
+                  error={errors.password}
+                  errorMessage={t("authentication.signIn.passwordError")}
+                  register={register}
+                  className={"mb-1"}
```
dominikx96, line 303: Brackets are not necessary (noticed in a few places) :)

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+            )}
+            <div className="form-card__group border-b">
+              <p className="field-note mb-4">{t("account.settings.passwordRemember")}</p>
+              <div className={"flex flex-col"}>
+                <Field
+                  caps={true}
+                  type="password"
+                  name="currentPassword"
+                  label={t("account.settings.currentPassword")}
+                  placeholder="Current password"
```
dominikx96, line 277: We should avoid hardcoded texts, instead, you can create a translated string.

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+        body: { ...profile, password, currentPassword },
+      })
+      setPasswordAlert({
+        type: "success",
+        message: `${t("account.settings.alerts.passwordSuccess")}`,
+      })
+    } catch (err) {
+      const { status } = err.response || {}
+      if (status === 401) {
+        setPasswordAlert({ type: "alert", message: `Invalid current password. Please try again.` })
```
dominikx96, line 112: We should avoid hardcoded texts, instead, you can create a translated string.

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+    }
+  }
+
+  return (
+    <RequireLogin signInPath="/sign-in" signInMessage={t("t.loginIsRequired")}>
+      <FormsLayout>
+        <FormCard>
+          <div className="form-card__lead text-center border-b mx-0">
+            <Icon size="2xl" symbol="settings" />
+            <h2 className="form-card__title">Account Settings</h2>
```
dominikx96, line 126: We should avoid hardcoded texts, instead, you can create a translated string.

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+                  register={register}
+                  defaultValue={profile ? moment(new Date(profile.dob)).utc().format("DD") : null}
+                />
+                <Field
+                  name="birthYear"
+                  placeholder={`${t("account.settings.placeholders.year")}`}
+                  error={errors.birthYear}
+                  register={register}
+                  defaultValue={profile ? moment(new Date(profile.dob)).utc().format("YYYY") : null}
+                />
```
dominikx96, line 208: I'm not sure, but can't we just use DOBField (or refactor this component to have more generic name / logic inside)?

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+              </div>
+              <div className="text-center mt-5">
+                <Button
+                  onClick={() => {
+                    //
+                  }}
```
dominikx96, line 214: This case is in a few places in the application - I think we should avoid using an empty function, instead of we can create a condition in the <Button /> component (most of the button usage cases don't require an onClick event.

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+                errorMessage={t("errors.firstNameError")}
+                register={register}
+                defaultValue={profile ? profile.firstName : null}
+              />
+
+              <Field
+                name="middleName"
+                placeholder={`${t("application.name.middleNameOptional")}`}
+                register={register}
+                defaultValue={profile ? profile.middleName : null}
```
dominikx96, line 154: The `middleName ` property in the profile is also optional, so I think we need to use: `profile?.middleName`

dominikx96 (changes requested):

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,349 @@
...
+  }) => {
+    const { birthDay, birthMonth, birthYear } = data
+    setDobAlert(null)
+    try {
+      await userService.update({
+        body: { ...profile, dob: new Date(`${birthYear}-${birthMonth}-${birthDay}`) },
+      })
+      setDobAlert({ type: "success", message: `${t("account.settings.alerts.dobSuccess")}` })
+    } catch (err) {
+      setDobAlert({ type: "alert", message: `${t("account.settings.alerts.generic")}` })
```
dominikx96, line 71: This string does not exist, you need to use `genericError`

dominikx96 (commented):
I noticed, e.g when you enter an invalid date of birth - frontend validation doesn't work, but the backend returns 400.

Before backend request, we should validate this form using react-hook. I see you're passing `error` prop into each `<Field />` component, but `errorMessage` is also required in this case (because inside Field component we're checking if this prop has been passed).

backend/core/src/shared/password-regex.ts
```diff
@@ -0,0 +1 @@
+export const passwordRegex = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9]).{8,64}$/
```
dominikx96, line 1: I noticed, on the frontend - the helper message does not contain any information about at least one capital letter.

emilyjablonski: Great find, I'll ask Jesse/Kathy which one should change

emilyjablonski: Looks like we want to lower the regex requirements and keep the copy, I updated the regex 👍

sites/public/pages/account/edit.tsx
jaredcwhite: Just checking if you saw this, thanks.

emilyjablonski: Sorry I thought I replied! IMO I'm worried that might actually decrease readability to need to shuffle between four different files for this one view, but I definitely see where you're coming from and don't feel particularly strongly about this and could go either way :) Happy to discuss further

kathyccheng:
@jaredcwhite one final look please!

dominikx96:
@emilyjablonski when I tried to update my password and entered a different confirmation than the password - I don't have any error (PUT to the backend doesn't work - and it's good, but I should see error text) :) 

Image 2021-04-16 at 5 45 00 PM

sites/public/pages/account/edit.tsx
```diff
@@ -0,0 +1,323 @@
...
+              <DOBField
+                id="dateOfBirth"
+                name="dateOfBirth"
+                register={register}
+                error={errors?.dateOfBirth}
+                watch={watch}
+                defaultDOB={{
+                  birthDay: profile ? moment(new Date(profile.dob)).utc().format("DD") : null,
+                  birthMonth: profile ? moment(new Date(profile.dob)).utc().format("MM") : null,
+                  birthYear: profile ? moment(new Date(profile.dob)).utc().format("YYYY") : null,
```
dominikx96, line 192: Ideally would be creating a helper function to do not repeat this 🐍 (not necessary of course!)

sites/public/pages/applications/household/current.tsx
```diff
@@ -56,14 +56,7 @@ export default () => {
...
           <div className="text-center mt-6">
-            <Button
-              styleType={AppearanceStyleType.primary}
-              onClick={() => {
-                //
-              }}
-            >
-              Next
-            </Button>
+            <Button styleType={AppearanceStyleType.primary}>Next</Button>
```
dominikx96, line 59: Missing translation :)

sites/public/pages/applications/household/preferred-units.tsx
```diff
@@ -89,14 +89,7 @@ export default () => {
...
             <div className="form-card__pager-row primary">
-              <Button
-                styleType={AppearanceStyleType.primary}
-                onClick={() => {
-                  //
-                }}
-              >
-                Next
-              </Button>
+              <Button styleType={AppearanceStyleType.primary}>Next</Button>
```
dominikx96, line 92: Missing translation :)

sites/public/pages/applications/reserved/units.tsx
```diff
@@ -52,14 +52,7 @@ export default () => {
...
           <div className="text-center mt-6">
-            <Button
-              styleType={AppearanceStyleType.primary}
-              onClick={() => {
-                //
-              }}
-            >
-              Next
-            </Button>
+            <Button styleType={AppearanceStyleType.primary}>Next</Button>
```
dominikx96, line 55: Missing translation :)

sites/public/pages/applications/review/demographics.tsx
```diff
@@ -156,14 +156,7 @@ const Demographics = () => {
...
             <div className="form-card__pager-row primary">
-              <Button
-                styleType={AppearanceStyleType.primary}
-                onClick={() => {
-                  //
-                }}
-              >
-                Next
-              </Button>
+              <Button styleType={AppearanceStyleType.primary}>Next</Button>
```
dominikx96, line 159: Missing translation :)

sites/public/pages/sign-in.tsx
```diff
@@ -92,14 +92,7 @@ const SignIn = () => {
...
             <div className="text-center mt-6">
-              <Button
-                styleType={AppearanceStyleType.primary}
-                onClick={() => {
-                  //
-                }}
-              >
-                Sign In
-              </Button>
+              <Button styleType={AppearanceStyleType.primary}>Sign In</Button>
```
dominikx96, line 95: Missing translation :)

dominikx96 (changes requested):
I found one issue with password repeat error, put one suggestion about moment date, and found some hardcoded "Next" button labels. I think when we resolve the main problem it's ready to merge! 👍

emilyjablonski:
All of the missing labels were existing, I would prefer to fix those in a different ticket given the priority of this one if thats ok?

emilyjablonski:
@dominikx96 The error handling on that doesn't kick in until submit, do you see the error on submit or does that also not display for you?
Screen Shot 2021-04-16 at 10 44 39 AM

dominikx96:
@emilyjablonski I found what's the issue! :D 

Image 2021-04-16 at 6 58 42 PM

So the problem is with error alert for confirmation length ;D

emilyjablonski:
@dominikx96 Thanks for taking another look! When I test that on my end as well, I get the following validation errors on submit:
Screen Shot 2021-04-16 at 11 05 08 AM

dominikx96:
@emilyjablonski I tried to create a step-by-step guide on how to reproduce it. I also noticed another problem (2):

**Problem 1:**

1. Refresh the page
2. Fill current password field (should be correct so min. 8 sings and number)
3. Fill a new password (following password rule)
4. Fill a password confirmation (but shorter than allowed)

_Result:_
I don't see any error alert/input error

**Problem 2:**

1. Refresh the page
2. Keep password fields empty

_Result:_
An incorrect PUT request to the backend

Solution:
It should be caught by frontend validation

emilyjablonski:
Gotcha, thank you so much for the steps! I think we're miscommunicating here a little bit. I was relying on backend validation (ie you have to hit submit before you'll see any errors) and you were expecting frontend validation. I've moved the validations to the frontend.

jaredcwhite:
We should have backend validation of passwords, no? Or is it that both frontend and backend share validation logic?

emilyjablonski:
@jaredcwhite If you check out the diff, we have both :)

dominikx96 (changes requested):
@emilyjablonski password-repeat field validation works as expected!

but I still see this behavior:

1. Refresh the page
2. Keep password fields without change (empty)
3. Click update

**Result:**

It produces not necessary PUT request (with bad request and generic error)

**Solution:**

I think it should be also caught by frontend validation e.g show red border and focus 'current password' field. (we should still keep backend validation). If you think differently - I'm also okay to merge it! 💪

dominikx96 (approved):

Issue #1041, opened 2021-02-25
Lock Name, DOB and Email for Autofill 


kathyccheng:
As an applicant 
I don't want to be able to change my name, DOB or email 
So that my account information stays clean 

When I'm on the What's your Name screen 
Then I will see a locked icon and disabled form fields on Name, DOB and Email

Issue #249, opened 2020-05-07
Autofill Application


slowbot:
**Goal**
As a housing applicant I want to be able to fill out my applications quickly and accurately

**Acceptance Criteria**

- If an applicant is signed in and has previously submitted an application
   - Fields on the first page of the application are auto-filled and locked based on account details
      - Name
      - DOB
      - Email
   - There is a notification on the first page of the application that reads
      _These fields are editable via your [Account Settings](url)_
   - Account settings is a link to the account settings page
   - There is an inline message under the email field that reads
      _Your email address is editable via your [Account Settings](url)_
   - Account settings is a link to the account settings page
   - After the Name page the applicant to presented with a summary of their previously submitted application and asked if they want to reuse those application details
      **Save time by using the details from your last application.**
      We'll simply pre-fill your application with the following details, and you can make updates as you go.
- Autofilled Fields does not include
   - Preference fields
   - Income fields (after 30days of previous submission)
- Options at the bottom of the auto fill page include:
   - Start with these details (primary)
      - Take the user through the rest of the application with details autofilled
   - Reset and start from scratch (secondary)
      - Starts the user on the next page of the application with no auto fill

**Documentation**
![image](https://user-images.githubusercontent.com/13915/81344667-12881f80-906c-11ea-8f68-b2941591f7b3.png)

![image](https://user-images.githubusercontent.com/13915/81330844-8bc84800-9055-11ea-8bf6-479f50703457.png)

![image](https://user-images.githubusercontent.com/13915/81330337-ced5eb80-9054-11ea-9082-b8bca550d2f5.png)

**Engineering Notes**

PR #1111, opened 2021-04-08
Autofill applications for signed in users

### LINKED ISSUE #1040: Autofill Application
As an applicant 
I want to be able to reuse my old application data 
So that I can more easily apply to new opportunities 

When I am signed in and starting an application 
Then after the What to Expect screen, I expect to see a screen letting me know that I can save time by using my old application data 
And it will show me the details of my last application data 
When I click Start with these Details
Then I will be taken to the Contact Information screen with my information pre-filled 
When I click Reset and start fresh 
Then I will proceed to the contact information screen with no information pre-filled 

Notes 
* Applicant must be signed in to use autofill 
* Should auto-fill everything except for preferences 
* Applicants will still have to step through every page 

Designs 
https://www.figma.com/file/i3IAkbjv4nVPKg1d1E1g1d/bloom-accounts?node-id=24%3A306

jaredcwhite:
Resolves #1040 #1041 

There's also an old duplicate here: #249

This also encompasses some fixes to the application form flow I noticed when signed in.


jaredcwhite:
@kathyccheng Here's the rundown of all the stuff that will get changed or deleted when the application is loaded up:

* **deleted**: ids, timestamps, etc.
* language: "en" (I could be wrong but it looks like all applications are being submitted as English? I don't see anywhere we're actually capturing this data)
* acceptedTerms: false
* status: "submitted"

And then various other sections being deleted/reset:

* preferences
* applicant.workInRegion
* applicant.workAddress

Let me know if that makes sense to you

jaredcwhite:
@slowbot Here's that issue I mentioned regarding the final page link:

Screen Shot 2021-04-08 at 3 34 09 PM

Also, on the first form step when somebody's already logged in, I wasn't sure what to do there so there's just a bit of white space. Let me know if you have any thoughts.

Screen Shot 2021-04-08 at 3 37 58 PM

kathyccheng:
@jaredcwhite all the things you removed makes sense to me!

kathyccheng:
@slowbot will let you take a look at whitespace. for the weird No thanks, I'm Done thing, were you able to figure it out @jaredcwhite?

jaredcwhite:
I still need guidance on the "No thanks" language.

kathyccheng:
Can we do "I'm done" or just remove it? @jaredcwhite @slowbot

jaredcwhite:
I'm happy to change it to "I'm done" but I don't know how to change the translations.

kathyccheng:
Like we don't have the translations, or you can't find it in our translation file? @jaredcwhite

jaredcwhite:
Oh, I mean I don't know what the new translations should be.

kathyccheng:
Gotcha! That's no problem, I'll add it to the next wave of translations with our contractors!

jaredcwhite:
Sounds good. And I just realized I have some strings which need translating on the new autofill page(s), so let me move those to the english JSON and I'll send to you for reference.

jaredcwhite:
@kathyccheng Here's what I've got so far:

```json
{
  "application": {
    "autofill": {
      "saveTime": "Save time by using the details from your last application",
      "prefillYourApplication": " We'll simply pre-fill your application with the following details, and you can make updates as you go.",
      "start": "Start with these details",
      "reset": "Reset and start fresh"
    },
    "household": {
      "addMembers": {
        "doubleCheck": "Please double-check the information for each household member.",
      },
    },
    "review": {
      "confirmation": {
        "imdone": "I'm done."
      }
    }
  }
}
```

kathyccheng:
Perfect, thanks!

sites/public/pages/applications/contact/name.tsx
```diff
@@ -45,6 +50,20 @@ export default () => {
   const emailPresent: string = watch("applicant.emailAddress")
   const noEmail: boolean = watch("applicant.noEmail")
   const clientLoaded = OnClientSide()
+  if (!autofilled && clientLoaded && application.autofilled) setAutofilled(true)
+
+  const AutofillIcon = () => {
```
emilyjablonski, line 55: (Non-blocking) I find this name to be a lil confusing, maybe something like DisabledIcon or LockIcon?

sites/public/pages/applications/review/confirmation.tsx
```diff
@@ -72,27 +74,31 @@ export default () => {
...
-            <Button
-              styleType={AppearanceStyleType.primary}
-              onClick={() => {
-                void router.push("/create-account").then(() => window.scrollTo(0, 0))
-              }}
-            >
-              {t("account.createAccount")}
-            </Button>
-          </div>
+          {initialStateLoaded && !profile && (
```
emilyjablonski, line 90: Just curious what this change is for?

sites/public/src/forms/applications/FormSummaryDetails.tsx
```diff
@@ -261,40 +261,43 @@ const FormSummaryDetails = ({ application, editMode = false }) => {
...
-                        }
-                      >
-                        {t(`application.preferences.${preference.key}.${option.key}.label`)}
-                      </ViewItem>
-                    ))
-                )}
-            </>
-          )}
-        </div>
+        {!hidePreferences && (
```
emilyjablonski, line 264: Curious about the need behind this change?

ui-components/__tests__/forms/DOBField.test.tsx
```diff
@@ -28,4 +29,9 @@ describe("<DOBField>", () => {
     expect(getByLabelText("Day")).not.toBeNull()
     expect(getByLabelText("Year")).not.toBeNull()
   })
+
+  it("can toggle all internal fields to be disabled", () => {
```
emilyjablonski, line 33: 🎈 🎊

ui-components/src/forms/HouseholdMemberForm.tsx
```diff
@@ -2,11 +2,19 @@ import React from "react"
...
 
-const HouseholdMemberForm = (props: { member: HouseholdMemberUpdate; type: string }) => {
+const HouseholdMemberForm = (props: {
+  member: HouseholdMemberUpdate
+  type: string
+  editMode?: boolean
+}) => {
   const { member, type } = props
   const router = useRouter()
+  let editMode = props.editMode
```
emilyjablonski, line 16: Can we just use a default prop value here?

jaredcwhite: Man, this must have been some weird debugging thing. Bad pattern! 😬 Will fix…

sites/public/src/forms/applications/FormSummaryDetails.tsx
jaredcwhite: We shouldn't show any preference information on the autofill preview since we're blanking them all out.

sites/public/pages/applications/review/confirmation.tsx
jaredcwhite: It should only show an option to create a new account if the logged-in state has loaded and the user _isn't_ logged in.

sites/public/pages/applications/contact/name.tsx
jaredcwhite: That makes sense.

jaredcwhite:
OK I figured out the infinite render issue and got that fixed…but I'm not super happy with the code. 50 LoC just to check an endpoint for data and either redirect or provide a form handler feels excessive. Open to any feedback on how to improve.

ui-components/src/forms/HouseholdMemberForm.tsx
```diff
@@ -2,11 +2,18 @@ import React from "react"
...
 
-const HouseholdMemberForm = (props: { member: HouseholdMemberUpdate; type: string }) => {
+const HouseholdMemberForm = (props: {
+  member: HouseholdMemberUpdate
+  type: string
+  editMode?: boolean
+}) => {
   const { member, type } = props
   const router = useRouter()
+  const editMode = props.editMode !== false // undefined should default to true
```
emilyjablonski, line 16: Was hoping to use a default prop of false here but this is def non-blocking!

jaredcwhite: I think the weirdness is that the default state should actually be true, so only if editMode prop is present _and_ set specifically to false should it be set false.

sites/public/lib/appAutofill.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.
+      addDefaults().
+      removeAdditionalKeys().
+      removeLiveWorkAddresses()
+
+    return this.application
+  }
+
+  addDefaults() {
+    ;["id", "createdAt", "updatedAt", "deletedAt", "listing", "submissionDate"].forEach((key) => {
```
dominikx96, line 27: `;` <- I suppose it's not necessary :D

sites/public/lib/appAutofill.ts
```diff
@@ -0,0 +1,93 @@
...
+      delete obj.id
+      delete obj.createdAt
+      delete obj.updatedAt
+    }
+
+    unsetIdentifiers(this.application.accessibility)
+    unsetIdentifiers(this.application.applicant)
+    unsetIdentifiers(this.application.mailingAddress)
+
+    if (this.application.alternateAddress) unsetIdentifiers(this.application.alternateAddress)
```
dominikx96, line 54: Are sure that it works? I think `alternateAddress` is loaded from the blackApplication.ts and it's an object, so it always has values (?) NOT SURE - just to check :) 

The similar situation below with `alternateContact `

sites/public/lib/appAutofill.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.application.applicant.workInRegion = null
+    this.application.applicant.workAddress = {
+      street: "",
+      street2: "",
+      city: "",
+      state: "",
+      zipCode: "",
+      county: "",
+      latitude: null,
+      longitude: null,
```
dominikx96, line 86: I think could be better to load defaults from the `blankApplication.ts`.

sites/public/pages/applications/start/autofill.tsx
```diff
@@ -0,0 +1,126 @@
...
+          editMode={false}
+          hidePreferences={true}
+        />
+        <Form onSubmit={handleSubmit(onSubmit)}>
+          <div className="form-card__pager">
+            <div className="form-card__pager-row primary">
+              <Button
+                styleType={AppearanceStyleType.primary}
+                onClick={() => {
+                  useDetails = true
```
dominikx96, line 102: Shouldn't be processed by React and `useState`?

dominikx96 (changes requested):
@jaredcwhite probably it's not in this scope, but when I filled out an application, I saw no primary applicant data:

Same situation with "Do you work in <Region> question.

Image 2021-04-19 at 5 10 13 PM

Image 2021-04-19 at 5 12 37 PM

sites/public/lib/appAutofill.ts
jaredcwhite: Prettier is adding it…must have thought there could be a JS error otherwise.

sites/public/lib/appAutofill.ts
jaredcwhite: Everything's coming directly from the backend in an autofill scenario (which is only when this code gets run), so I'm trying to strip out or change data to fit that flow.

sites/public/pages/applications/start/autofill.tsx
jaredcwhite: Well it's not a state change, it's just a runtime flag that lets the submit handler know which action to take. I specifically wouldn't want to trigger any re-render.

sites/public/lib/appAutofill.ts
jaredcwhite: Good point, I could load that in and copy the object over.

jaredcwhite:
@slowbot I think I identified and fixed a low-level bug regarding object references that was causing some data loss…the bug had been present before the autofill PR, but never triggered because we'd never needed to perform a wholesale swap of application data mid-flow before. Can you do one more QA pass?

jaredcwhite:
Looking good testing on Netlify preview, got the thumbs up from @slowbot…so plan to merge after fixing the conflict.

jaredcwhite:
Getting some Cypress timeouts but I'm assuming those are spurious. 😬 LGTM

PR #1115, opened 2021-04-09
test/flaky cypress tests

emilyjablonski:
* Changes the Cypress job to run in production mode instead of dev mode (cuts the time in half from 20 min to 10 min and removes the flakiness)
* Stores videos and screenshots for tests that fail in CircleCI, which has been critical for debugging the failing tests
* Removes tests that depended on the name of a listing, as the ordering isn't stable and the seed data could change (fixes the broken `navigation` issues)
* Comments out the `terms` suite tests that are broken on master - there is an incoming fix in #1155

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,6 +87,16 @@ jobs:
       - restore_cache:
           key: build-cache-{{ .Environment.CIRCLE_SHA1 }}
       - run: yarn build:app:partners
+  cypress\run:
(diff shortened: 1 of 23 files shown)


jaredcwhite (approved):
Just realized I hadn't formally added my review here. LGTM! :+1:

dominikx96 (approved):
@emilyjablonski thanks for your effort, LGTM! 🚀

PR #1087, opened 2021-03-24
Upgrade the public app to Next v10 and new architectural patterns

jaredcwhite:
Part of the work for #1081 

**Cypress test issue on CircleCI**: see https://github.com/bloom-housing/bloom/pull/1087#issuecomment-844555411

PR #1117, opened 2021-04-09
Remove next router from LanguageNav component

james-wills-sf:
This is a duplicate of https://github.com/SFDigitalServices/bloom/pull/11

We're basing the branch off of the bloom-housing master branch to see if circleCI tests run

Currently the LanguageNav component does two things that make it not work in our webapp repo:
1.  The component relies on next-router to handle the language changes
2.  The component expects there to be a "config.prefix" translation string to determine the current locale route prefix

We can make the component more agnostic to the environment around it by passing an on click handler and prefix string in as props.

In this case, I made `<LanguageNav />` more general and created a more specific `<RoutedLanguageNav />` that populates the props in the next-routing way.

In the webapp repo we'll just call the more general `<LanguageNav />` directly.

### DIFF
--- a/ui-components/__tests__/navigation/LanguageNav.test.tsx
+++ b/ui-components/__tests__/navigation/LanguageNav.test.tsx
@@ -1,32 +1,47 @@
 import React from "react"
-import { render, cleanup } from "@testing-library/react"
-import { LanguageNav } from "../../src/navigation/LanguageNav"
-import * as useLanguageChange from "../../src/helpers/useLanguageChange"
+import { render, cleanup, fireEvent } from "@testing-library/react"
+import { LanguageNav, LangItem } from "../../src/navigation/LanguageNav"
+
+const ENGLISH_LANG_ITEM: LangItem = {
+  prefix: "",
+  label: "English",
+}
+
+const SPANISH_LANG_ITEM: LangItem = {
+  prefix: "es",
+  label: "Spanish",
+}
 
 afterEach(cleanup)
 
 describe("<LanguageNav>", () => {
   it("renders without error", () => {
-    const useLanguageChangeSpy = jest.spyOn(useLanguageChange, "useLanguageChange")
+    const onChangeLanguageSpy = jest.fn()
     const { getByText } = render(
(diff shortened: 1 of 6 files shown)


jaredcwhite:
Thanks James. I like this approach. I'll have to look at #1087 again because I might have started to streamline this a bit anyway, but if that doesn't get merged in shortly we should be good to go with this.

james-wills-sf:
Hey @jaredcwhite, thanks for taking a look so quickly! I'll look at #1087 too. 

The meta purpose for this PR is to test-run our process for contributing changes to ui-components. The plan so far is outlined [in a comment from the bloom-sf version of this pr](https://github.com/SFDigitalServices/bloom/pull/11#issuecomment-818026470), but basically we're thinking we'll manage a fork of ui-components that we'll deploy on our own schedule. It will hopefully be totally in sync with bloom's, just maybe a couple commits ahead so we're unblocked.

Let's talk about it in our sync meeting tomorrow!

jaredcwhite:
That makes sense to me…at least in the short term as you get everything spun up. Looking forward to diving into this further tomorrow￼.

jaredcwhite:
So it looks like in that Next v10 branch I mentioned, the only reference to anything specific to Next is a single `router.push` call. https://github.com/bloom-housing/bloom/blob/next-10-upgrades/ui-components/src/navigation/LanguageNav.tsx

If that component were abstracted out to take a callback prop, sort of like what you did here, then we could handle the router function within the Next layer and the component would be agnostic. That seems like a good pattern.

jaredcwhite:
Our `LinkButton` component will be a bit tricker since it relies on the `Link` Next component which gets imported in. Probably should file a new issue to address that specifically.

jaredcwhite:
Sorry about the lack of progress on this. Hoping to get the Next v10 update finalized this week, along with related ui-components changes that'll affect your integration work (hopefully to make it easier!).

seanmalbert:
The Next updates are in and per @jaredcwhite , this can be closed. Expect a new release/publish today or tomorrow.

PR #1121, opened 2021-04-12
bug/my applications loading state

emilyjablonski:
My Applications will now no longer flash the empty state as the page is loading, and will instead show the loading overlay.


jaredcwhite (approved):
LGTM 👍

Issue #1118, opened 2021-04-10
Confirm Email Page Loading Bug


slowbot:
Confirmation email seems to work but when I am redirected to the site there is a weird flash 

1. Homepage loads with success message for a moment  
2. Redirects to user dashboard

What I expect is to happen is that I would be immediately directed to the user dashboard

ludtkemorgan:
Closing as it was marked as fixed in #1122

PR #1122, opened 2021-04-13
Fix message after account confirmation

### LINKED ISSUE #1119: Format Confirm Account Email
Is it possible to add line breaks to the email to make the confirm link easier to locate

What we have now
image

What I would expect
image

software-project:
Resolves #1119 and #1118

### DIFF
--- a/backend/core/src/shared/email.service.spec.ts
+++ b/backend/core/src/shared/email.service.spec.ts
@@ -51,9 +51,7 @@ describe("EmailService", () => {
       expect(sendMock.mock.calls[0][0].to).toEqual(user.email)
       expect(sendMock.mock.calls[0][0].subject).toEqual("Welcome to Bloom")
       // Check if translation is working correctly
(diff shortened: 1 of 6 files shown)


software-project:
> @software-project Ideally the user wouldn't see the homepage at all. They would be directly to the dashboard immediately. Is that possible? If not we can merge this since its better than what we have and refine. @kathyccheng

@slowbot unfortunately it's not possible as we are pinging frontend server that needs to first render a page, run a validation call to backend and then it can redirect to dashboard. I suppose initial call could be to a page with a loading spinner, with would make the experience maybe a bit better. Maybe with Nextjs update it will be possible, but I'm not sure. @jaredcwhite ?

dominikx96 (changes requested):
#1118 I was redirected to the dashboard and I saw a success message, it looks good!

#1119 I don't see expected formatting, see screenshot:

Image 2021-04-15 at 10 24 40 AM

PR #1123, opened 2021-04-13
Remove self package reference in ui-components

james-wills-sf:
While exploring creating a npm package fork, I ran into an issue where I couldn't change the @bloom-housing org name in package.json because there was an `@bloom-housing/ui-components` import from _within_ ui-components.

This seems like a mistake, so this change switches it to a local import instead. This might be something we can lint for?

### DIFF
--- a/ui-components/src/blocks/AppStatusItem.tsx
+++ b/ui-components/src/blocks/AppStatusItem.tsx
@@ -3,7 +3,7 @@ import "./AppStatusItem.scss"
 import moment from "moment"
 import { Application, Listing } from "@bloom-housing/backend-core/types"
 import { LocalizedLink } from "../actions/LocalizedLink"
-import { t } from "@bloom-housing/ui-components"
+import { t } from "../helpers/translator"
 
 interface AppStatusItemProps {
   application: Application


jaredcwhite:
Absolutely. Not sure what we'd use to lint this but happy to get something like that set up. Was it just this once instance, or are there other cases lurking about?

jaredcwhite (approved):

james-wills-sf:
@jaredcwhite I could only find the one instance. It looks like there are @bloom-housing/ui-components in ui-components/.storybook but when I changed those to relative imports storybook failed to run. I don't really know much about how storybook gets its assets and they don't break anything so I'm fine with keeping those imports as-is.

james-wills-sf:
Looked into linting this, you could use something like [import-blacklist](https://www.npmjs.com/package/import-blacklist-extended) but that would require some refactoring since you'd need linting rules specific to each package directory.

Not necessary for now, maybe something to think about if it comes up again

james-wills-sf:
@jaredcwhite looks like tests are passing. I'll leave it to you to do the merge since idk bloom's merge and deploy processes

jaredcwhite:
LGTM!

PR #1132, opened 2021-04-15
Flagged Applications Table

dominikx96:
# Pull Request Template

## Issue

Closes: #777 , #1243

Also:

- [x] Created useMutation hook to easily manage POST/PUT request states (see: `[flagId]/index.tsx`).
- [x] Created pagination component to do not repeat this element (we had one in the `/applications`)
- [x] Added SHOW_DUPLICATES=FALSE environment variable to control flags navigation visibility

Addresses # (issue)

## Description

A user is able to view and resolve duplicates.

## Type of change

Create applications with e.g the same name and/or email address, then open the flags tab.

- [x] Desktop View
- [x] Mobile View

## Checklist:


jaredcwhite (commented):
Just one comment…shouldn't be a blocker but recent conversations with SFDS are pushing us in the direction of keeping any page routes out of ui-components. Ideally if we need deal with particular routes within the component layer, they should either come through props or a context. Also I'd suggest `ListingSecondaryNav` be moved somewhere in `page_components/listing` since it's not really a generic component.

dominikx96:
> Just one comment…shouldn't be a blocker but recent conversations with SFDS are pushing us in the direction of keeping any page routes out of ui-components. Ideally if we need deal with particular routes within the component layer, they should either come through props or a context. Also I'd suggest `ListingSecondaryNav` be moved somewhere in `page_components/listing` since it's not really a generic component.

Agree! I also renamed the component, because it is inside applications, not listing.

seanmalbert:
Hey @dominikx96 ,
What's the status of this? Is this waiting on backend changes and if so, is there a related issue?

dominikx96:
> Hey @dominikx96 ,
> What's the status of this? Is this waiting on backend changes and if so, is there a related issue?

Hey, it's ready for the next QA round by @jaredcwhite. I also added `SHOW_DUPLICATES` to temporarily hide this navigation until other duplicates parts are ready.

kathyccheng:
@slowbot I think we should do the tag size clean up and then save the rest. Let's talk about the nav stuff when we discuss Partners nav more broadly. What do you think? 

This one: 
** Can we reduce the size of the tag used in the flag tab to display the count?

dominikx96:
@slowbot @jaredcwhite I implemented missing elements, so we can review it again :)

slowbot (approved):
@dominikx96 I think this looks good. 

One small nitpick is the back button behavior from single application flag brings me back to the flags table when I expect it to take me to the flag set for that single application. but this can wait until a followup

Additionally, I think we might add a "view application" button because there is tendency to double click the application number when attempting to highlight that row, but this can wait until a followup

ui-components/src/headers/PageHeader.tsx
```diff
@@ -27,9 +39,24 @@ const PageHeader = (props: PageHeaderProps) => {
...
         <h1 className="page-header__title">{props.title}</h1>
         {props.subtitle && <p className="page-header__lead"> {props.subtitle}</p>}
-      </hgroup>
+        {props.children}
 
-      {props.children}
+        {!!props?.tabNav?.length && (
+          <TabNav className="page-header__secondary-nav">
+            {props.tabNav?.map((tab) => (
+              <Tab
```
jaredcwhite, line 47: FYI, this was renamed to `TabNavItem` in #1305…also might have a conflict with your type above.

ui-components/src/headers/PageHeader.tsx
```diff
@@ -27,9 +39,24 @@ const PageHeader = (props: PageHeaderProps) => {
       <hgroup className="page-header__group">
         <h1 className="page-header__title">{props.title}</h1>
         {props.subtitle && <p className="page-header__lead"> {props.subtitle}</p>}
-      </hgroup>
+        {props.children}
 
-      {props.children}
+        {!!props?.tabNav?.length && (
+          <TabNav className="page-header__secondary-nav">
+            {props.tabNav?.map((tab) => (
```
jaredcwhite, line 46: I'm wondering if we could use a different pattern here: passing items directly in as components rather than using a JSON intermediary. An example would be the `StatusAside` component, which accepts buttons via the `actions` prop. What do you think?

ui-components/src/headers/PageHeader.tsx
dominikx96: Fixed, thanks.

ui-components/src/headers/PageHeader.tsx
dominikx96: Agree, it's a good idea, especially we have to get rid of any nextJS elements from ui-components (now it's still in the package but in the page_components dir).

jaredcwhite:
Maybe we can relocate the `ui-components/src/page_components/application/ApplicationSecondaryNav.tsx` component to somewhere inside `sites/partners`? Otherwise, looks good to me?

PR #1135, opened 2021-04-15
Fix CircleCi Redis typo

emilyjablonski:
It looks like master is broken after https://github.com/bloom-housing/bloom/pull/1103 was merged (it also looks like the tests were broken there as well, not sure why it was merged?) - I'm not sure how to fix these issues but we did notice a typo in the CircleCi config (it doesn't seem to have solved the issue)


emilyjablonski:
@pbn4 Do you have more context here?

PR #1138, opened 2021-04-16
fix/broken Cypress tests on master

emilyjablonski:
The navigation suite was broken by this PR https://github.com/bloom-housing/bloom/pull/1124 (It's not flake, the suite is failing consistently / broken after the change - looks like changing the listing name changes the URL slug)


emilyjablonski:
@pbn4 Just as a heads up

pbn4:
@emilyjablonski I just fixed the test checks for where slugs have changed

emilyjablonski:
@pbn4 Thanks for taking a look! This was the approach I tried first and I couldn't get the tests to pass locally. I have re-seeded, is there something else I need to do locally for the tests to pass?

Issue #1097, opened 2021-04-01
Figure out a way to customize emails for SMC 


kathyccheng:

pbn4:
Requirement: being able to determine which public site user is browsing in order to e.g. change an appropriate email template.

Options that I see we have are:

1. Add custom HTTP header to each request - can be easily done with https://docs.netlify.com/routing/headers/ : 
   ```
   ```
2. Reuse existing HTTP headers like e.g. `Origin`, but there are some caveats: 
   ```
   Basically, browsers add the Origin request header to:

   all cross origin requests.
   same-origin requests except for GET or HEAD requests (i.e. they are added to same-origin POST, OPTIONS, PUT, PATCH, 
   and DELETE requests).
   ```

dominikx96:
For me, 1st solution makes sense! (after testing, but I'm pretty sure it should work as expected).

software-project:
I also like the headers solution. But it's only the first part of this issue. Second is translations. Should we store the in the DB in file, for each county? Commiting translation files as jsons on the long run to a "universal" backend might not be the best idea. We will have an entity to hold counties in so having just a json translation field there would be plenty enough. As it's going to be mostly for emails.

kathyccheng:
Closing, covered with footer updates for now

PR #1145, opened 2021-04-20
Add translations module and county codes for frontend

pbn4:
Addresses https://github.com/bloom-housing/bloom/issues/1097


backend/core/src/translations/entities/translation.entity.ts
```diff
@@ -0,0 +1,36 @@
...
+import { AbstractEntity } from "../../shared/entities/abstract.entity"
+import { Expose } from "class-transformer"
+import { IsEnum, IsString, MaxLength } from "class-validator"
+import { ValidationsGroupsEnum } from "../../shared/validations-groups.enum"
+import { ApiProperty } from "@nestjs/swagger"
+import { CountyCode } from "../../shared/types/county-code"
+import { Language } from "../../shared/types/language-enum"
+
+@Entity()
+@Index(["countyCode", "language", "key"], { unique: true })
```
software-project, line 11: I don't think we need `key` here. To get keys one by one, would require many requests to db. If we use just countyCode and language as keys and have a json field holding all the translations for county I think that would be enough. One call to get all of the needed translations.

software-project (changes requested):
I like the enums refactoring. That's really nice. Translations I for language/ county I would make just as a 'bucket'. So basically one row in the db would hold sth like a whole json file on the frontend. I think that is going to be ok as I don't expect huge amount of keys per county.

backend/core/src/translations/entities/translation.entity.ts
pbn4: Applied comments, thanks

backend/core/src/migration/1619094998058-alter-translations-table.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export class alterTranslationsTable1619094998058 implements MigrationInterface {
+    name = 'alterTranslationsTable1619094998058'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`DROP INDEX "IDX_4c17bc3d8eefdb8702bb24a2c5"`);
+        await queryRunner.query(`ALTER TABLE "translation" DROP COLUMN "key"`);
+        await queryRunner.query(`ALTER TABLE "translation" DROP COLUMN "text"`);
+        await queryRunner.query(`ALTER TABLE "translation" ADD "translations" jsonb NOT NULL`);
+        await queryRunner.query(`CREATE UNIQUE INDEX "IDX_ada354174d7f8a8f3d56c39bba" ON "translation" ("county_code", "language") `);
```
software-project, line 11: I think we also need to add language to user_accounts so we know what language to use.

software-project (changes requested):

backend/core/src/migration/1619094998058-alter-translations-table.ts
pbn4: It will be a part of a separate PR, too much stuff in those one already.

PR #1155, opened 2021-04-21
Add env variables to control rate limits

bpsushi:
Add two env variables THROTTLE_TTL and THROTTLE_LIMIT to control global rate limits. Variables are not required. By default it is set to almost infinite.

This resolves the issue where Cypress tests hits rate limits.


emilyjablonski:
It looks like the `terms` Cypress suite is still failing in CI on this branch, is that expected / is some other change needed?

pbn4:
@bpsushi Why did you remove throttle for /submit, is it not globally configured? Also tests are red at the moment

bpsushi:
> It looks like the `terms` Cypress suite is still failing in CI on this branch, is that expected / is some other change needed?

@emilyjablonski im not 100% sure, but it looks like now Cypress is failing on other error, that is not related to the rate limit. Can you please double check the reason of failing?



> @bpsushi Why did you remove throttle for /submit, is it not globally configured? 
It is now configured globally. Decorator used before is no more needed.

> Also tests are red at the moment
Yup i see. Will try to reproduce locally

emilyjablonski:
The error you're getting now in the `terms` suite is the same error we get with the rate limiting issue, so I think it's failing for the same reason

bpsushi:
@pbn4 lint and all test fixed, only Cypress is failing but it looks like totally unrelated issue.
I had to change CircleCi configuration a little - i have added Redis container for Cypress executor, since it is using backend app.

@emilyjablonski please take a look at cypress tests, they are failing on navigation.ts with timed out error.

emilyjablonski:
Yup that suite is unrelated, looks great!

emilyjablonski:
Just a heads up that the terms suite was commented out on master until this fix goes in 👍

PR #1163, opened 2021-04-22
(fix): broken test suite on master

emilyjablonski:
This PR aims to fix the `navigation` Cypress suite in the public app. I think our suite should be flexible around the listings data so I removed one of the tests that checked specifically the name of a listing, because that feels like testing an implementation detail, and the seed data could easily change. In another test, I removed looking for the specific name of the listing, but we still check that we navigate to the listing specific page.


emilyjablonski:
Wrapping this change up with the flaky work, will close this one

Issue #1200, opened 2021-05-03
Localize birth date correctly.


software-project:
When I register with birth date being 01.01.1999 and then go to settings I get an unexpected date of 31.12.1998. Seems like date is localized incorrectly either on creation or on display.

emilyjablonski:
How might I test this locally? When I do account creation, an e-mail doesn't send, so I'm not sure how to create the account in order to check that the birthdate is localizing.

PR #1202, opened 2021-05-04
fix: birthdate localizing issues

emilyjablonski:
Fixes the issue in #1200 
Here is a screenshot from the account settings page after creating an account with a birthdate of 01/01/1999
Screen Shot 2021-05-10 at 1 55 09 PM


jaredcwhite:
Makes sense. LGTM :+1:

Issue #1065, opened 2021-03-09
Translation issue with bmrProgramChart set to true.


software-project:
When listing units have `bmrProgramChart` set to true translations coming from backend are already text instead of keys and that causes frontend to raise translation missing errors.

PR #1203, opened 2021-05-04
fix: bmr translation issue

emilyjablonski:
Addresses #1065 

When a listing had `bmrProgramChart` set to true, the console was reporting to translation errors.
The issue was that `unit-transformations` was sometimes passing a translation key (i.e. `t.key`) and sometimes (in the case of BMR) passing an english string (i.e. `20% AMI units`). This produced the console errors because `StandardTable` was passing every incoming string as if it were a key, so there were "missing keys". So it was both untranslated, and not a key.

The tricky part about this fix is that `X% AMI Units` can't just be passed as a key as it has an argument, and I didn't see a precedent for that anywhere. Following the format `t.key,argumentName:argumentValue` I'm allowing the headers of `StandardTable` to be passed header key strings that include optional argument information so that the string can still be translated.

Below is a screenshot of it working properly, passing "Unit Type" as just the key and parsing "20% AMI Unit" with its argument. No console errors are produced and it's all translated. To test this locally you may need to import a listing with `bmrProgramChart` set to true.
Screen Shot 2021-05-04 at 12 31 33 PM

### DIFF
--- a/ui-components/src/helpers/getTranslationWithArguments.ts
+++ b/ui-components/src/helpers/getTranslationWithArguments.ts
@@ -1,0 +1,14 @@
+import { t } from "../.."
+
+// translationKey must be in the format `key,argumentName:argumentValue,argumentName:argumentValue`
+// any number of arguments can be passed, returns a fully translated string
+export const getTranslationWithArguments = (translationKey: string) => {
+  if (translationKey.indexOf(",") >= 0) {
+    const [key, ...stringArguments] = translationKey.split(",")
+    const argumentsObject = stringArguments.reduce((obj, arg) => {
+      const [key, value] = arg.split(":")
+      return { ...obj, [key]: value }
+    }, {})
+    return t(key, argumentsObject)
+  } else return t(translationKey)
+}
--- a/ui-components/src/helpers/getTranslationWithArguments.ts
+++ b/ui-components/src/helpers/getTranslationWithArguments.ts
@@ -1,4 +1,4 @@
-import { t } from "../.."
+import { t } from "./translator"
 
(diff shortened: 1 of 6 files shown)


ui-components/src/tables/StandardTable.tsx
```diff
@@ -36,9 +36,18 @@ export interface StandardTableProps {
 export const StandardTable = (props: StandardTableProps) => {
   const { headers, data, cellClassName } = props
 
+  // Translation strings are optionally passed in with arguments with format `key,argumentName:argumentValue`
+  const getTranslatedHeader = (headerKey: string) => {
+    if (headerKey.indexOf(",") >= 0) {
+      const translation = headerKey.split(",")
+      const argument = translation[1].split(":")
```
jaredcwhite, line 43: Seems like a good approach for the arguments syntax. Could we make it more flexible so multiple pairs supported by commas are allowed?

`t.key.path,val1:Value 1,val2:Value2`

jaredcwhite (approved):

jaredcwhite:
Just had one suggestion. Otherwise LGTM!

pbn4 (approved):

emilyjablonski:
Pulled it into a helper method, destructured the arguments, and made it flexible for any number of arguments 👍

ui-components/src/helpers/getTranslationWithArguments.ts
```diff
@@ -0,0 +1,14 @@
+import { t } from "../.."
```
jaredcwhite, line 1: Cool that this works, but it sure looks odd! 😅

jaredcwhite (approved):
Looks good to me!

seanmalbert (approved):
Looks good @emilyjablonski . Nice work!

PR #1208, opened 2021-05-05
pull_request_template proposal

seanmalbert:
Here's a pull request template for us to iterate on. We may want more or less items depending on the team's preferences and we may want separate templates for front and backend changes, if they're typically separate, but it's easy enough for a dev to remove the list items that aren't relevant on PR creation.

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,0 +1,37 @@
+# Pull Request Template
+
+## Description
+
+Please include a summary of the change and which issue is addressed. Please also include relevant motivation and context. List any dependencies that are required for this change.
+
+Addresses # (issue)
+
+## Type of change
+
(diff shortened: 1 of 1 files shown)


emilyjablonski:
This looks great!

jaredcwhite:
Agree, looking good @seanmalbert! Maybe we could add one other pair of questions along the lines of:

- [ ] This change addresses all aspects of the related issue
- [ ] This change addresses only certain aspects of the related issue

Not sure of the exact wording there, but anyway I think it's helpful to know if a PR is fully intended to close an issue or if there's something left over still to address elsewhere.

seanmalbert:
Thanks for the feedback @jaredcwhite. I moved the issue section up and added your suggestions and two more checks.

PR #1209, opened 2021-05-06
WIP: Add a Changelog

jaredcwhite:
Based on discussions we've been having within our core team, it's our intention to adopt a more regular and consistent release cycle, and to document what's going into each release via the Changelog.

Because there are several layers to the Bloom architecture reflected in the repo itself being a monorepo with several packages, I've broken out change entries into General, Backend, Frontend, and UI Components. We should discuss if those delineations make sense or if there's a better way to frame it.

I also included two different formats for how the content structure is established: **CHANGELOG.md** has the layer categories first and then Added, Changed, etc. **CHANGELOG alt.md** flips the order of the hierarchy. I personally prefer the initial structure, but I'm fine with the alt version if it's preferred by the team overall.

----

It becomes pretty obvious after looking at this that we'll definitely want to be more deliberate with how we title PRs and hence what will go into the changelog…in addition, it's clear too much time currently elapses between official releases, so its hard to tell what a release "is" as well as when the next one will/should be.

@seanmalbert's proposed PR template #1208 should greatly aid us in this endeavor. That + a regular release cadence will be a major leap forward.


jaredcwhite:
Seems like we got consensus on this from earlier conversations, so will now merge.

PR #1219, opened 2021-05-10
Feature/NextJS 10 Partners Update

### LINKED ISSUE #1211: Translate Next Update into Partners
Reference: 

https://github.com/bloom-housing/bloom/issues/1081
https://github.com/bloom-housing/bloom/issues/1087

dominikx96:
Closes #1211

- Bumped NextJS version to 10.1
- Updated all paths to support listing/application URL params instead of query params
- Removed unnecessary MetaTags component
- Fixed issue in the `DetailsMemberDrawer.tsx` component - after fixing it shows live/work address details only if they are custom (if the same address as a primary applicant is not YES or workInRegion is YES)

@jaredcwhite after your acceptance I think we can resolve conflicts in your branch and merge this one to yours.

I think we should merge https://github.com/bloom-housing/bloom/pull/1190 first, then do a PROD deploy, and next resolve conflicts and finally merge it to master.


seanmalbert:
> Got a weird thing happening where if I'm viewing a specific application or adding one, then click on either the logo or the Listings link in the nav bar, I get a React error: TypeError: undefined is not a function (near '...rowData.forEach...')

@jaredcwhite and @dominikx96 , this is interesting and I don't know how this hasn't been an ongoing issue. I'm able to reproduce the issue and the first thought that something was getting cached somehow on the listing route. I played with the cache in the backend and could still get the issue. So then I looked at swr's documentation on useSWR, and from the documentation @ https://www.npmjs.com/package/swr#ssr-with-nextjs, it states `The key is a unique identifier of the request, normally the URL of the API`. So I made it so the key for useSWR in useSingleListingData and the one used in useListingsData were different and I was no longer able to get the issue.

So, in short, the keys for useSWR(key, fetch) just need to be different.

sites/partners/src/LinkComponent.tsx
```diff
@@ -0,0 +1,16 @@
+import { PropsWithChildren } from "react"
```
seanmalbert, line 1: This is identical to the one in public, so if we go the route of merging this into @jaredcwhite public, should this be moved into ui-components so they can use the same one?

seanmalbert (changes requested):
@dominikx96 ,
First pass, I think this looks good, except for the issue @jaredcwhite called out (which should be an existing issue and an easy fix), and also how we want to handle the duplicate LinkComponent.

sites/partners/src/LinkComponent.tsx
jaredcwhite: I don't like the duplication either, but the reason for this at all is the line `import Link from "next/link"`

Through discussions with SFDS, we were tasked to ensure ui-components doesn't have any dependency on Next (as in it we literally remove it from its `package.json`), so that's why we wrap Link here and push it down the stack via a context. This does bring to mind something the team discussed quite a long time ago: that maybe we need a shared package for bits of Next-specific view/helper code we need in both public+partners that is a separate thing entirely from ui-components, but at the time we decided that extra complexity wasn't worth it.

dominikx96: Yup, exactly as @jaredcwhite said that's the reason why I duplicated this component.
I also agree we should think about a shared package for both websites.

dominikx96:
> > Got a weird thing happening where if I'm viewing a specific application or adding one, then click on either the logo or the Listings link in the nav bar, I get a React error: TypeError: undefined is not a function (near '...rowData.forEach...')
> 
> @jaredcwhite and @dominikx96 , this is interesting and I don't know how this hasn't been an ongoing issue. I'm able to reproduce the issue and the first thought was that something was getting cached somehow on the listing route. I played with the cache in the backend and could still get the issue. So then I looked at swr's documentation on useSWR, and from the documentation @ https://www.npmjs.com/package/swr#ssr-with-nextjs, it states `The key is a unique identifier of the request, normally the URL of the API`. So I made it so the key for useSWR in useSingleListingData and the one used in useListingsData were different and I was no longer able to get the issue.
> 
> So, in short, the keys for useSWR(key, fetch) just need to be different.

@seanmalbert thanks! I updated keys to be unique and it seems to be working now 👍

jaredcwhite (approved):
LGTM (into the other PR)!

seanmalbert (approved):
Updates look good to me! Thank you @dominikx96

PR #1221, opened 2021-05-10
fix: waitlist open showing for every listing

emilyjablonski:
# Pull Request Template

## Issue

No pre-existing issue

## Description

`Open Waitlist` was showing on the listings overview page for every listing regardless of whether or not it had a waitlist. New logic flow is if there is a current size and max size of a waitlist, and there are spots, show `Open Waitlist` otherwise nothing. You can see the bug here on staging: https://ala.bloom.exygy.dev/listings

## Type of change

Checked out the /listings page on both desktop and mobile. I locally have listings with and without a waitlist.
- [x] Desktop View
- [x] Mobile View

## Checklist:

### DIFF
--- a/ui-components/src/page_components/listing/ListingsList.tsx
+++ b/ui-components/src/page_components/listing/ListingsList.tsx
@@ -51,7 +51,11 @@ const ListingsList = (props: ListingsProps) => {
           />
         </div>
         <div className="listings-row_content">
-          <h4 className="listings-row_title">{t("listings.waitlist.open")}</h4>
+          {listing.waitlistMaxSize !== null &&
+            listing.waitlistCurrentSize !== null &&
+            listing.waitlistCurrentSize < listing.waitlistMaxSize && (
+              <h4 className="listings-row_title">{t("listings.waitlist.open")}</h4>
+            )}
           <div className="listings-row_table">
             {unitSummaries && (
               <GroupedTable
--- a/ui-components/src/page_components/listing/ListingsList.tsx
+++ b/ui-components/src/page_components/listing/ListingsList.tsx
@@ -51,11 +51,9 @@ const ListingsList = (props: ListingsProps) => {
           />
         </div>
         <div className="listings-row_content">
-          {listing.waitlistMaxSize !== null &&
-            listing.waitlistCurrentSize !== null &&
-            listing.waitlistCurrentSize < listing.waitlistMaxSize && (
-              <h4 className="listings-row_title">{t("listings.waitlist.open")}</h4>
-            )}
+          {listing.showWaitlist && (
+            <h4 className="listings-row_title">{t("listings.waitlist.open")}</h4>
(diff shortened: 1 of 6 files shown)


ui-components/src/page_components/listing/ListingsList.tsx
```diff
@@ -51,7 +51,11 @@ const ListingsList = (props: ListingsProps) => {
           />
         </div>
         <div className="listings-row_content">
-          <h4 className="listings-row_title">{t("listings.waitlist.open")}</h4>
+          {listing.waitlistMaxSize !== null &&
+            listing.waitlistCurrentSize !== null &&
+            listing.waitlistCurrentSize < listing.waitlistMaxSize && (
+              <h4 className="listings-row_title">{t("listings.waitlist.open")}</h4>
+            )}
```
seanmalbert, line 58: I'm wondering if it would be useful to have a computed/virtual key on listings for this case, something like listing.showWaitlist.

If we wanted to do this, we could update the listing entity to use TypeOrm's AfterLoad decorator, and we would add it to listing.entity with something like:
```
// declare the computed field
showWaitlist?: boolean

// set computed values
@AfterLoad()
setComputed() {
  this.showWaitlist = this.waitListMaxSize !== null && this.waitlistCurrentSize !== null && this.waitlistCurrentSize < this.waitlistMaxSize
}
```

emilyjablonski: Having not yet touched the backend, curious about @pbn4 's thoughts?

pbn4: Yes doable, do you want me to implement it backend side? Seems pretty straightfoward and I agree backend is a better place for such conditional logic.

pbn4: @emilyjablonski @seanmalbert

emilyjablonski: I'll give this a first pass! Thanks 👍

pbn4: @emilyjablonski I don't think there is need for @AfterLoad, we could make it with such defining getter method on a property and @Expose that, checkout of how generating a slug property is handled because it's the same concept.

seanmalbert: That's a good point @pbn4 , it's the same affect, and not putting all of our eggs in TypeOrm's basket is probably a good idea.

emilyjablonski:
@pbn4 Hoping for your feedback here

pbn4 (approved):
No comments from my side. It's correct. @emilyjablonski What did you refer to when you said "at least one weird thing there"?

emilyjablonski:
Nothing in particular, I just was out of my element and figured there'd be something weird :) Thanks!

Issue #820, opened 2020-11-05
Images on Storybook are missing


jaredcwhite:
I was trying to track down why the close icon wasn't showing up for modals in Storybook, and upon further investigation I discovered that *none* of the files in `ui-components/public/images` are working anymore…404s on all of them. Not sure yet if this happened due to a recent upgrade to Storybook, or some other configuration that got hosed.

jaredcwhite:
Strange, it's working for me locally on the `updated-button-styles` branch. Is it just an issue on Netlify then? (look at https://storybook.bloom.exygy.dev/?path=/story/headers-hero--with-background — no background image is showing up!)

emilyjablonski:
At the current moment, images appear to be working just fine, but icons are not - focusing the effort there

PR #1225, opened 2021-05-11
Add spinner to the Application Form terms page upon submit

jaredcwhite:
## Issue

Addresses #1193 
- [x] This change addresses the issue in full
- [ ] This change addresses only certain aspects of the issue
- [ ] This change is a dependency for another issue
- [ ] This change has a dependency from another issue

## Description

The submit button on the Terms page for the Application was allowing multiple clicks before the form was submitted. This PR uses the "loading" state of the button component to provide visual feedback as well as preventing additional clicks.

In addition, I:
* Refactored the loading overlay components to use the spinner in the icons SVG
* Changed the spinner SVG so it uses CSS animation (apparently SMIL is deprecated, also buggy in Safari)
* Equalized the spinner icon SVG across our three packages (previously the public site was out of date)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This change requires a SQL Script

## How Can This Be Tested/Reviewed?

I manually tested the Terms page of the Application Form, and wrote an example in Storybook for `LoadingOverlay` and `Button` with the loading state.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have reviewed the changes in a desktop view
- [x] I have reviewed the changes in a mobile view
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have assigned reviewers

PR #1230, opened 2021-05-11
fix: storybook svgs don't appear

emilyjablonski:
# Pull Request Template

## Issue

Related to but separate from #820 

## Description

Icons (SVGs) were not working on Storybook, locally or at our published Storybook URL, outside of the Icon story.
After lots of research, the way we were doing SVGs (basically as "external SVGs", symbol tags squished into one svg file and then indexed by ID) is deprecated and not supported by some tools (like Storybook). The preferred way of doing SVGs in a React application seems to be by component-tizing them, so that's what I've done here. This comes with multiple benefits:
1) Our SVGs can now take props and are thus more dynamic/extensible
2) We get Typescript auto-complete on icon types
3) Works in Storybook!
4) Not deprecated :)

## Type of change

I would mainly check stories w Icons like the Referral Application, you can also check out any icons in the application.

## Checklist:


jaredcwhite:
@emilyjablonski Thanks for digging into this! FWIW, Storybook has always worked for me locally, no problems with icons or images which is why the breakage on Netlify was puzzling.

My only concern with this approach is embedding SVG data inside of JS files (SVG-in-JS?). If we move away from the single bundle/use/symbol approach, can we just save individual SVG icon files to a folder somewhere and then import them into React icon components? Maybe there's a semi-automated way to do that? I feel like icons and other images are assets that should remain usable outside of React only.

emilyjablonski:
@jaredcwhite Do SVGs work for you locally on Storybook right now? When I added the new referral component and the icons didn't appear and I had to add it into the application to test it, Jesse seemed to indicate that was expected and that SVGs have never worked in Storybook.

This is the approach that seems to be recommended following research online (component-tizing them). I don't think it's unusual to find SVG data inside of a js file after what I've seen :) There is a slightly different approach available where we would have to pull in a 3rd party package in order to keep them as SVG files and then import them as "components" but I didn't feel like the benefits of that we were worth pulling in the package. Do we anticipate using our SVGs outside of a React environment?

seanmalbert:
I know it to be a common practice at this point to have componentized SVGs (the popular Material UI library does this for example) and I think it makes sense for this project. If we find that we want to use them outside of react/Bloom, we could keep them in separate files and import them and use with these components so that it doesn't affect this project and others could use them as plain SVGs, but to that point the rest of the ui component library is in React anyway.

jaredcwhite:
I must admit I'm still not convinced, but perhaps for now it's a moot point due to the merge of #1235?

emilyjablonski:
That PR was just for images and SVGs within the Icon story. Without this change SVGs still don't work elsewhere in Storybook locally or at the published site.

jaredcwhite:
Can you share a link where it's broken? It's definitely working somewhere outside just the icon stories, for example here:
https://deploy-preview-1235--amazing-davinci-96c1cd.netlify.app/?path=/story/notifications-alert-box--alert-box-alert

and again, I haven't personally seen any problems using the icons in Storybook locally.

emilyjablonski:
Can you check out the Referral story? That's an example from this PR description of it not working. Outside of having it work in Storybook though, I also still think this is a valuable refactor. Let me know if you're not able to repro and we can touch base tomorrow and pair!

jaredcwhite:
Oh, right. I think that's because it needs `<SVG src="/images/icons.svg" />` added to the story wrapper, otherwise the iframe is missing the SVG data. See `Modal.stories.tsx` for example.

emilyjablonski:
Makes sense! Feels weird that it's also in assets and needs that included as well, but I'm following that. I still think we should merge this change and that it's cleaner, let me know what you think! :)

jaredcwhite:
Well, just in general I'm wary of adding a bunch of JS/JSX overhead for an icon library, due to performance/bundle size/memory concerns…came across a [similar sentiment from Jason Miller of Preact](https://twitter.com/_developit/status/1382838799420514317?s=21) (which is funny because he's also at Google. He and the Material UI folks better get talking!) He seems to think the `use` method we've been doing thus far is preferable.

jaredcwhite:
If we can figure out how to avoid duplicating the SVG file in three places though, that would be ace!

seanmalbert:
Correct me if I'm wrong @emilyjablonski  and @jaredcwhite , it appears we have two arguments here, one for usability and dev experience and the other for performance. To test performance, I thought I would create a test page with the current way icons are used and the way they're used in this branch. 

Add this under /public/pages:
```
import { Icon } from "@bloom-housing/ui-components"

const Test = () => {
  return (
    <div>
      <Icon size="medium" symbol="asterisk" />
      <Icon size="medium" symbol="bed" />
      <Icon size="medium" symbol="browse" />
      <Icon size="medium" symbol="cross" />
      <Icon size="medium" symbol="down" />
      <Icon size="medium" symbol="phone" />
      <Icon size="medium" symbol="profile" />
      <Icon size="medium" symbol="search" />
      <Icon size="medium" symbol="settings" />
      <Icon size="medium" symbol="spinner" />
    </div>
  )
}

export default Test

```
And add `"/test": { page: "/test" },` to the next.config.js paths.

The first thing that stands out, is that on master, the icons don't render, similar to how they don't render in storybook. Switching to this branch, they do. On top of that, with ten icons at least (not much so feel free to test with more), whether or not they render on master, the size and time are relatively the same. I get what you're saying @jaredcwhite , but this is easier and cleaner to use from a dev perspective, and given that we're not SVG heavy, I'm not sure performance concerns come into play, or at least there are bigger fish out there within this app.

To me, this is better than what's there now and any performance hits we'll see from this appear to be negligible for the apps' use cases. I would love to be at the point where we're concerned with this level of optimization.

jaredcwhite:
>  on master, the icons don't render, similar to how they don't render in storybook

I think Emily already resolved the rendering issue @seanmalbert, so at this point all we're discussing is the ergonomics of embedding SVG data in a large collection of React components vs using an SVG sprite file. From a consumer point of view, I believe calling the `Icon` component would remain the same.

ui-components/src/icons/Icons.tsx
```diff
@@ -0,0 +1,541 @@
+import * as React from "react"
+
+export interface IconProps {
+  fill?: string
+}
+
+export const Application = (props: IconProps) => {
+  return (
+    <svg viewBox={"0 0 24 32"} fill={props.fill ?? "#000000"}>
```
dominikx96, line 9: I think would be better to use `currentColor` instead of `#000000`, what do you think? (noticed multiple places where we can change it)

ui-components/src/icons/Icon.tsx
```diff
@@ -1,28 +1,137 @@
...
+}
+
+export type IconTypes = keyof typeof IconMap
+
+export type IconFill = "white" | "primary"
+
+const IconFillColors = {
+  white: "#ffffff",
+  primary: "#0077DA",
+}
```
dominikx96, line 116: I think we also should define `black` or give the ability to pass a custom color.

dominikx96 (commented):
I like this component approach! I just put a few thoughts.

jaredcwhite:
FYI, the spinner SVG was just updated in #1225. The `animateTransform` tag can be removed.

seanmalbert:
> > on master, the icons don't render, similar to how they don't render in storybook
> 
> I think Emily already resolved the rendering issue @seanmalbert, so at this point all we're discussing is the ergonomics of embedding SVG data in a large collection of React components vs using an SVG sprite file. From a consumer point of view, I believe calling the `Icon` component would remain the same.

Yea, that's all I was pointing out, that it works and that I'm good with the approach.

emilyjablonski:
Incorporated Dominik's feedback, and the recent Spinner changes and made sure those work. Should be good to go again!

seanmalbert (approved):
This looks good, @emilyjablonski . A coliseum json file made it into this PR. Can you remove that? Other than that, I think this is good to go!

emilyjablonski:
Whoops! Just removed it 👍  Thanks

dominikx96 (approved):
Looks good to me! 👍

PR #1234, opened 2021-05-12
Fix/infinite loop

dominikx96:
## Issue

It's not covered by any issue.

Addresses # (issue)

## Description

This PR solves 2 problems:

1)
A language from the `choose-language` step is not saved by a conductor. I also refactored autofill to do not reset context (language is selecting in the first step and autofill is a secondary step - so we need to know what language the user selected)

2)
url: http://localhost:3000/applications/start/choose-language without `listingId` parameter occurs infinity loop with GET /listings call to the API.

## Type of change

1) Log in
2) Choose language different than EN (it was default previously so we want to test different case)
3) Submit application
4) Open public app in a new card (to reset context)
5) Log in
6) Post application with different language
7) See on the partners if an application includes selected language

- [x] Desktop View
- [ ] Mobile View

## Checklist:


dominikx96:
@jaredcwhite I also noticed, `ApplicationForms` test doesn't have any sense after infinite loop fix. I need to figure out how to open this step with a correct listingId, then update the test. I think for now we can test all except this - I'll prepare fix soon

jaredcwhite (changes requested):
Hi @dominikx96, good catch on the application language not being saved. (I think there might have been a problem before where we never actually saved it.)

I have some concerns about this PR altering subtle behavior I fixed previously after discovering bugs. Can we basically just include the

```
     conductor.currentStep.save({
       language,
     })
```

change, and then for Autofill step just pass a language code directly as the second argument to the constructor?

Regarding the missing listing ID param issue, it looks like the fix in this PR is sort of broken:
https://deploy-preview-1234--clever-edison-cd22c1.netlify.app/applications/start/choose-language

I know in the past we had sort of hack in place which just pulled the first listing out of the result set. Maybe what we should just do instead is redirect to the home page if there's no listing ID param?

dominikx96:
> Hi @dominikx96, good catch on the application language not being saved. (I think there might have been a problem before where we never actually saved it.)
> 
> I have some concerns about this PR altering subtle behavior I fixed previously after discovering bugs. Can we basically just include the
> 
> ```
>      conductor.currentStep.save({
>        language,
>      })
> ```
> 
> change, and then for Autofill step just pass a language code directly as the second argument to the constructor?
> 
> Regarding the missing listing ID param issue, it looks like the fix in this PR is sort of broken:
> https://deploy-preview-1234--clever-edison-cd22c1.netlify.app/applications/start/choose-language
> 
> I know in the past we had sort of hack in place which just pulled the first listing out of the result set. Maybe what we should just do instead is redirect to the home page if there's no listing ID param?

@jaredcwhite agree I updated PR to just redirect the user to the `/`.

Unfortunately, I don't understand the first problem, could you explain it more with code samples?
Currently, I decided to pass a current application from the context as a second `AutofillCleaner` parameter in the constructor - it gives us more flexibility.

sites/public/pages/applications/start/autofill.tsx
```diff
@@ -37,9 +36,8 @@ export default () => {
       setSubmitted(true)
       if (previousApplication && useDetails) {
         conductor.application = previousApplication
-      } else {
-        conductor.application = blankApplication()
```
jaredcwhite, line 41: I had this here so if somebody went back in their browser history and changed to *not* autofilling the application, it'd be rewritten as a blank application. Otherwise they could never undo their change.

jaredcwhite (changes requested):
OK, I think there was just one specific place where a bug might have crept in again. Otherwise looks good to me—although will need a update for Next v10. Should we get that merged in and then this? Or merge this first?

sites/public/pages/applications/start/autofill.tsx
seanmalbert: @jaredcwhite , how close do you think we are to being able to merge the Next branch? If you're thinking next week, then can we get this in first? We're in a tough spot of not wanting to delay the Next updates any longer, but we also don't want to stack up a lot of dependencies.

jaredcwhite: @seanmalbert Oh I want to get Next v10 ready to merge today, so I think this could follow suit.

dominikx96:
@jaredcwhite I realized we can just create a new application object in the autofill step, even if autofill is available - the first step is choose-language, so it works as expected in both situations. Could you check it again? :)

sites/public/lib/appAutofill.ts
```diff
@@ -3,13 +3,14 @@ import {
...
   ApplicationSubmissionType,
-  Language,
 } from "@bloom-housing/backend-core/types"
 import { blankApplication } from "@bloom-housing/ui-components"
 
 class AutofillCleaner {
   application: Application = null
 
+  // provide context value to override current application language choosen by user on the first step
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jaredcwhite, line 13: I don't see anything untyped here…do we still need the disable comment?

jaredcwhite (approved):
@dominikx96 Approach looks good to me! :+1:

sites/public/lib/appAutofill.ts
dominikx96: nice catch! Removing it...

Issue #1210, opened 2021-05-06
Listings Management: States Prototype 


kathyccheng:
The most barebones flow that would be helpful to see is a listing with minimum data that can move through the basic states 

**Listing** 
* Title 
* Address
* Housing Developer
* Application Deadline Date 
* Apply Online button 

**Actions** 
* Publish, pushes to open 
* Preview, during draft 
* Edit and Delete actions could be optional, but might be interesting to double check for pushing updates to preview or open listings 

**States** 
* Draft 
* Open - live on the public site with a URL, included in the list of listings  
* Closed - closed on the public site, no application data, this part should be automatic based on timing   

**Notes**
* I think the publish action can be manual vs scheduled, but I think the closing should be automatic (open to discussing this)

seanmalbert:
This is complete here https://github.com/bloom-housing/bloom/pull/1247. I can perform a demo on Monday. As part of this I created a base listing form and detail page with some updates to the listing grid as well, so we can link to the listing detail and the applications. A new listing can actually be created (with some hardcoded values) with name and application due date and status respecting the form values. Please see the PR for more notes. As stated in those notes, this includes a lot of legwork necessary for listings management in general, and if we wish to use it (with some adjustments required), it gives us a framework to add the new components and build out the rest of the features.

Issue #1168, opened 2021-04-23
Listings Management Fieldset: Listing Intro


slowbot:
**Goal**: Leasing agent should be able to add basic information about the listing.

**Fields**:
Field Name | Entity Model | Type | Current State
------------ | ------------- | --- | ----- | 
Listing Name | name | string | Yes
Developer | developer | string | Yes
County | county | string | Yes

**Acceptance Criteria**:
- User can enter content in all fields

**Questions**
- Should developer be an object in the future?

**Figma**:
Default:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2721%3A125812

Filled:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2721%3A125836

pbn4:
**Fields**:
Field Name | Entity Model | Type | Current State | Backend review comment | 
------------ | ------------- | --- | ----- | --- |
Listing Name | name | string null | Yes | |
Developer | developer | string null | Yes | `developer` is a field on a `Property` model which in turn is a field in `Listing` model |

@slowbot 
Q: What is the definition of a "property" in future state? Should we remove it at the moment?  If we want to flatten it and remove it then developer would just be a  string column on a Listing. If we are planning for having a UI where admin can add developers to the system then this should probably be a separate DB entity now (then I imagine in listings management form there is a dropdown with developers instead of a string input).

seanmalbert:
I think it makes sense for developer to be a top level object and selectable on the listing form.

The basic hierarchy would be:
- developer
    - property
        - listing

The dropdown's options would be the developers the leasing agent has access to, but this step can probably be MVP+1, if we can get away with listing all developers as the options to begin with.

pbn4:
@seanmalbert could you elaborate on the idea of this hierarchy taking into consideration https://github.com/bloom-housing/bloom/issues/1174#issuecomment-833876531

seanmalbert:
@pbn4 ,

Per our conversation, the hierarchy in my comment above is what we may end up working towards. I think it makes sense for this task to move `developer` to the listing along with the rest of the relevant property fields.

To further elaborate, at some point we may want to add a property management feature, where developers or leasing agents could create properties under which they would manage their listings. Having a copy/clone feature for listings, and the training to use it, is really important in this case, because it would ease in the migration of moving properties back into a property model (with hopefully minimal duplicates).

[...]

PR #1247, opened 2021-05-16
Listing Management Prototype

seanmalbert:
# Pull Request Template

## Issue

Addresses #1210 

## Description

This is a POC for changing the state of a listing (active, pending, closed), but it does go beyond that and includes a lot of the legwork  required for Listings Management in general. That being said, there were decisions made to speed up this process. For example, you'll notice the property ID is hardcoded if the listing doesn't have one. You can create a new listing, but the address won't be reflected, since it's using the set property. If we were going to continue forward with this, that would need to be updated. We are planning on moving property fields to listings directly, so that would actually make this easier.

There is also a lot of components between listings and applications that could be reused, but I didn't want to worry about creating breaking changes for applications with this work, so I kept it isolated.

The preview button is there, but doesn't do anything. And again, changing the address or other property info in the form doesn't cascade to the property, but changing the name and application due date do get reflected.

In doing this POC it became apparent that the applications caching strategy has issues. In particular, the backend listings module requires different strategies for public and partner uses. We don't require the same level of caching for listings in partner as we do with public.

You can test this by either editing a current listing or creating a new one. The name and application due date are the only fields that really matter (or actually get updated for this POC). If the due date is before now, the listing will appear as closed, regardless if it's active or pending. If the due date is in the future, clicking "publish" on the form will set the listing to active and clicking "save as draft" will set the listing's status to pending.

## Type of change

- [x] POC

- [x] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


pbn4:
https://github.com/bloom-housing/bloom/issues/1168#issuecomment-843019600

pbn4:
> In doing this POC it became apparent that the applications caching strategy has issues. In particular, the backend listings module requires different strategies for public and partner uses. We don't require the same level of caching for listings in partner as we do with public. 

I think our caching is configurable with http headers, so we could manipulate that on frontend request. What do you think?

backend/core/src/listings/entities/listing.entity.ts
```diff
@@ -338,6 +340,16 @@ class Listing extends BaseEntity {
   @IsEnum(CountyCode, { groups: [ValidationsGroupsEnum.default] })
   @ApiProperty({ enum: CountyCode, enumName: "CountyCode" })
   countyCode: CountyCode
+
+  applicationCount: number
+
+  @AfterLoad()
```
pbn4, line 346: I don't think we need an extra method decorated with @AfterLoad here, see  how property.entity.ts has implemented `getUnitsSummarized`:
```
  @Expose()
  @ApiProperty({ type: UnitsSummarized })
  get unitsSummarized(): UnitsSummarized | undefined {
    if (Array.isArray(this.units) && this.units.length > 0) {
      return transformUnits(this.units)
    }
  }
```

backend/core/src/listings/listings.service.ts
```diff
@@ -15,6 +15,7 @@ export class ListingsService {
       .leftJoinAndSelect("listings.leasingAgents", "leasingAgents")
       .leftJoinAndSelect("listings.preferences", "preferences")
       .leftJoinAndSelect("listings.property", "property")
+      .leftJoinAndSelect("listings.applications", "applications")
```
pbn4, line 18: With current TypeORM performance issues this will mostly likely cause container run OOM on heroku. Probably a workaround for that would be doing a separate query with `count`.

backend/core/src/listings/entities/listing.entity.ts
seanmalbert: Cool, I'm good with using getters instead of AfterLoad.

seanmalbert:
@pbn4 ,

I cleaned this up with the Next updates and addressed the issues you had called out, plus removed some hardcoded hacks I had in place. I also removed the add listing button and links directly to listing detail, so this can be merged without affecting anything. You can still get to /listings/add and listings/[id] directly via the URL. I didn't go with a feature flag here, since we're going to handle that in general with another issue.

If this looks good to you, we can merge this in, and give us a good start on the listing form and edit pages. Let me know if you have any questions or concerns.

Issue #1236, opened 2021-05-12
Preferences Update for Applications 


kathyccheng:
Right now our preferences are all on one page for the application. Moving forward, we want one page for each "parent" level preference. So for example: for Coliseum listing coming up, there would be one page for the "Live/Work" type with subpreferences Live and Work, and then a separate page for the Oakland Housing Authority "preferences" with sub-preferences for Residency, Family, etc.

In this issue, don't worry about hiding a preference from the listing page or how to handle the None of these apply to me. We just want to split out the application pages by the high level "parent" preference and show sub-preference per page.

PR #1252, opened 2021-05-19
feature: preferences on multiple pages

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1236 

## Description

This change allows preferences to be split into a one-per-page flow. I added a new column in the preference table `page` that starts at 1, which indicates which page a preference should live on. The data migration auto-adds every preference to have a page of 1 if it wasn't listed.

If there are multiple preference pages, the form handles it by staying on one URL/form step and just keeping track of the page using state. Adjustments had to be made in order to handle the "back" flow properly, both between preference pages and from the following page. My solution to this was to add a `navigatedThroughBack` flag on the `ApplicationConductor` class and to set the flag whenever we hit the back button. It also meant I had to update the `FormBackLink` component so that it doesn't always change the URL (in the case that we have multiple preference pages).

## Type of change

I recommend puling the branch, running db migrations, and trying the normal flow on the 2-preference listing (they will all be one page) to ensure it works that way still (I initially broke this).

Then I recommend re-seeding which will get the 2-preference listing to have the preferences on multiple pages and trying this out.

Move forward and backward, make updates then move throughout the flow, etc.

## Checklist:


emilyjablonski:
Rebasing broke this PR, under construction again

emilyjablonski:
It doesn't work on the Netlify preview deploy but after tons of testings it seems like that doesn't automatically pick up seed data / migration changes, which I didn't know

seanmalbert (changes requested):
Hi @emilyjablonski ,
This looks good and appears to be working well, nice work! 

I have a couple of minor things around the FormBackLink updates. Aside from that, it feels like there's a lot of redundancy around the onClicks with both the FormBackLink and Button components (always calling setNavigateBack), though I'm not sure there's necessarily a problem with that, since doing anything about it would probably mean making FormBackLink and Button less reusable/more complicated.

emilyjablonski:
@seanmalbert Thanks for the review! I've made the two requested changes 👍

seanmalbert (approved):
Thanks for the updates, looks good to me.

PR #1255, opened 2021-05-20
Add information about Next 10 upgrade to Changelog

jaredcwhite:
@seanmalbert Let me know if you think this provides enough context. cc @pbn4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ All notable changes to this project will be documented in this file. The format 
 
 - Changed:
 
+  - Upgrade the public + partners apps to Next v10 along with new architectural patterns [#1087](https://github.com/bloom-housing/bloom/pull/1087))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ All notable changes to this project will be documented in this file. The format
 
 - Changed:
 
+  - Upgrade the public + partners apps to Next v10 along with new architectural patterns [#1087](https://github.com/bloom-housing/bloom/pull/1087))
```
seanmalbert, line 19: Can we start separating the changes with release headers, so if we can agree the new version should be v1.0.0, to indicate major breaking changes can we add at the top of this section a header like `## [1.0.0] - 2021-05-20`? So maybe replace `### General` with that? What do you think @jaredcwhite and @pbn4 ?

seanmalbert (approved):
Looks good to me. Thanks @jaredcwhite

Issue #1199, opened 2021-05-03
Oakland Housing Authority Application Questions 


kathyccheng:
As a developer, 
I want applicants to know that they can also apply to Housing Authority preferences on the same property 
So that they can maximize their chances of getting housing 

Implementation Notes 
* Add a screen after the Alameda Live Work preference specifically for Oakland Housing Authority 
* The Oakland Housing Authority "preferences" should show up in the Partners Portal table and CSV export. It could either be a column for each OHA preference + don't consider me,  or a column for OHA and the options are (each preference, None, Don't Consider Me) 
* The Oakland Housing Authority Preferences should NOT show up on the listings page (we'll need to add something custom to hide it just in this instance)
* The title, subtext, and select text are all different here 
* We can use our current preferences model 
* We need a flag for the none of these preferences apply to me 

![image](https://user-images.githubusercontent.com/3453407/116940876-9bfd0e00-ac3c-11eb-91c1-ef1a2499383b.png)

seanmalbert:
From conversation during planning we're split on how to resolve this. @jaredcwhite and I seem to be in agreement that this should be a one off for Alameda for the time being, because we don't have enough information to build a core solution that all clients will benefit from. There is the recognition that the backend core would need an update to even support this. And there is the idea that adding categories to preferences, so we can use those on the frontend to build out where these preferences should appear.

I'm not opposed to a core solution, as long as we can come to an agreement on a solution that is outside of this specific case. @pbn4 has some ideas on how to handle this, and I think if we can get that in writing, we can review these approaches.

seanmalbert:
After talking with @emilyjablonski , here's an update to how we're going to handle this. Per @pbn4 's recommendation, we can handle this with preferences as they currently are with some small adjustments:

1. Add a column to preferences we can use to indicate which preferences should show. If we were to go with something like "page": integer, and have it default to 0 or 1 for the preferences that should show up on the current preferences page, and for each iteration of "page" we display a new set of preferences. This makes it more flexible for future implementation and for this case we don't necessarily have to implement the "pages" aspect in the UI for the first pass. You can just check if it's page 1 or 2.
2. Update sites/public/pages/applications/preferences/all.tsx to display the preferences according to the page they're on. Again for this first pass, if it's relatively much easier to just map them based off of page === [int] then let's go with that, and we can enhance it better in core at a later date.
[...]

PR #1267, opened 2021-05-25
fix: coliseum preferences issues - allow optional description & generic decline

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1199 

## Description

The new preference request out of Coliseum requires that we be able to have preferences that don't have a description, that we don't automatically always show the "I don't want these preferences" option, and that we can have custom select text.

Below are two screenshots of the Coliseum OHA preference before this change and after this change. You can see in the before, the issues are that "read more" shows for the `None of these preferences apply to me...` and `I don't want to be considered...` options but there actually isn't a description to show for those, that the `I don't want these preferences` still shows even though we don't want it to, and that it reads "If you have one of these housing preferences...".

I added three new optional fields on preference to set the description to false for an individual preference, to hide the generic decline text from a preference set, and to add custom select text.

Screen Shot 2021-05-25 at 3 07 57 PM
Screen Shot 2021-05-25 at 4 15 29 PM

## Type of change

You will need to import the Coliseum listing. You can find it [here](https://github.com/housingbayarea/bloom/blob/0.3_alameda/backend/core/listing_data/coliseum.json).

## Checklist:


sites/public/pages/applications/preferences/all.tsx
```diff
@@ -234,22 +234,24 @@ const ApplicationPreferencesAll = () => {
...
-                                  <br />
-                                  {preference?.links?.map((link) => (
-                                    <a key={link.url} className="block pt-2" href={link.url}>
-                                      {link.title}
-                                    </a>
-                                  ))}
-                                </p>
-                              </ExpandableContent>
-                            </div>
+                            {!(option.description === false) && (
```
jaredcwhite, line 237: Maybe use `option.description !== false`?

jaredcwhite (approved):
Looks good to me!

sites/public/pages/applications/preferences/all.tsx
```diff
@@ -184,132 +184,138 @@ const ApplicationPreferencesAll = () => {
 
         <Form onSubmit={handleSubmit(onSubmit)}>
           <>
-            <div className="form-card__group px-0 pb-0">
```
emilyjablonski, line 187: Big diff! The only thing I did was move this into the map - nothing else :)

Issue #1229, opened 2021-05-11
Flag for none of these preferences apply to me


kathyccheng:
For the OHA preference, we want to introduce an option to include this None of these preferences apply to me. This will automatically uncheck any of the other options the applicant has already checked. 

When an applicant is on the OHA page 
And they check Residency or Family sub-preferences 
And then they check None of these preferences apply to me but I would still like to be considered
Then their previous checked sub preferences will become unchecked 

When an applicant has checked none of these preferences apply to me, but I would like to be considered 
And then they click another one of the sub-preferences
Then the None of these preferences apply to me will become unchecked 

When an applicant has any of the subpreferences or None of these preferences apply to me checked
And then they check I don't want to be considered for this opportunity 
Then any of the previous checks should become unchecked 
( this should be current behavior, just want to make sure it's still working

PR #1272, opened 2021-05-26
feat: new checkbox behavior

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1229 

## Description

The new OHA preference set has new "exclusive" options, meaning options that when checked should un-check all other boxes, and if checked must be the only checked box.

This logic somewhat already existed (although through this PR I found it was buggy) with the generic "no thanks" option that was default added to the end of every preference set. The bug on master was that if option D is the generic-exclusive-no-thanks checkbox, if you click A B C then D it will correctly uncheck A-C, but if you then click any of A-C with D selected both will be checked, which is incorrect. This PR solves that issue.

In a separate PR before this which is merged, I added the ability to not automatically include the default exclusive box (through the `hideGenericDecline` flag), since that isn't part of OHA.

This PR adds the `exclusive` flag on a preference option which indicates it has the exclusive behavior outlined above. There can be any number of exclusive options. I had to abstract out some of the functions I wrote/updated so that the logic could be shared between public and partners (we needed the logic in parters in the piece where you can edit applications manually).

## Type of change

You will need to pull in the [Coliseum listing](https://github.com/housingbayarea/bloom/blob/0.3_alameda/backend/core/listing_data/coliseum.json), and check it out in public by going through the application process and checking/unchecking allllll the boxes. Then submit the application, head over to partners and edit it, and look at the checkboxes there for the same exclusive functionality on the last two options for OHA.

## Checklist:

** I would love to be testing this change but we don't really have a way to do that yet. Just another push to add a unit test suite to the public and partners site.


sites/partners/src/applications/PaperApplicationForm/sections/FormPreferences.tsx
```diff
@@ -34,7 +36,7 @@ const FormPreferences = ({ county, preferences, hhMembersOptions }: FormPreferen
     const keys = []
     preferences?.forEach((preference) =>
       preference?.formMetadata?.options.forEach((option) =>
-        keys.push(getPreferenceOptionName(preference?.formMetadata.key, option.key))
+        keys.push(getPreferenceOptionName(option, preference?.formMetadata.key, option.key))
```
jaredcwhite, line 39: Just noticing the order of arguments here seem to be reversed from code in the public app. Was that intentional?

jaredcwhite (commented):
Just had one question regarding the refactor, otherwise LGTM.

On an infinite timescale, I'd love to discuss refactoring how the frontend is now handling preferences overall. My OOP-loving brain is really struggling with the verbosity and ceremony of all these functions passing low-level values back and forth as well as conditionals everywhere. I fear it won't be straightforward to maintain over the long term.

sites/partners/src/applications/PaperApplicationForm/sections/FormPreferences.tsx
emilyjablonski: Great catch! 😅  I checked all other uses of the fxn just to be sure, this one was just flip-flopped.

emilyjablonski:
I think in the long term we have hopefully a new application section for something like these "Programs" since these aren't really preferences. We kind of planned to work with what we already have in the short term but I almost wish we had just gone for a bigger change anyway. I definitely agree with you.

jaredcwhite:
> We kind of planned to work with what we already have in the short term but I almost wish we had just gone for a bigger change anyway.

Oh yeah, totally understandable to try to utilize what we have already set up and seems reasonable to me in the near term. I just think we'll want to be aware of the tech debt involved…

sites/partners/src/applications/PaperApplicationForm/sections/FormPreferences.tsx
```diff
@@ -43,21 +45,71 @@ const FormPreferences = ({ county, preferences, hhMembersOptions }: FormPreferen
...
+        inputProps={{
+          onChange: (e) => {
+            if (e.target.checked) {
+              setExclusive(true, setValue, exclusiveKeys, noneKey, preference)
+            }
+          },
+        }}
+      />
+    )
+  }
```
seanmalbert, line 71: I feel like I'm missing something, but does it make sense to start saving the value of the "none" checkboxes? In the case of Coliseum, and maybe future cases, wouldn't it be useful to know the distinction? Then the only real difference between these "none" types and the rest, would be with setExclusive, and a lot of this could be simplified.

Same goes for the public site's handling of this.

sites/public/pages/applications/preferences/all.tsx
```diff
@@ -150,6 +158,106 @@ const ApplicationPreferencesAll = () => {
...
+            id={getPreferenceOptionName(option, option.key, preference.formMetadata.key)}
+            name={getPreferenceOptionName(option, option.key, preference.formMetadata.key)}
+            type="checkbox"
+            label={t(`application.preferences.${preference.formMetadata.key}.${option.key}.label`, {
+              county: listing?.countyCode,
+            })}
+            register={register}
+            inputProps={{
+              onChange: () => {
+                setTimeout(() => {
```
seanmalbert, line 215: Is the setTimeout necessary?

seanmalbert (commented):
@emilyjablonski ,

I'm not necessarily requesting changes, but I do have questions, that depending on the answer will require changes. Please see my comments in the code. My main question is around the handling of "normal" vs "exclusive" options. Can't the exclusive options be treated the same as normal options, aside from the "checking/unchecking" logic? With multiple exclusive options, won't we want to know which one of those was selected and even if there was just one exclusive option, does it hurt to save that as a normal option with this new setup?

sites/public/pages/applications/preferences/all.tsx
emilyjablonski: Not sure, it was existing, I can remove

sites/partners/src/applications/PaperApplicationForm/sections/FormPreferences.tsx
emilyjablonski: The distinctions between a normal and exclusive option if you look at the public implementation is the onChange, the validation, and the optional existence of both a description and a household member dropdown. I feel like it's cleaner to separate them out into two fxns than have one with an "if exclusive" in four places, but I don't feel strongly about this. And the values still get sent through the form, so I'm not sure what you mean by starting to save the values?

seanmalbert: When I check "None of these preferences apply to me, but I would like to be considered" or "
I don't want to be considered for [housing authority] project-based voucher units" those values are not saved in the application.

emilyjablonski: Taking offline (online?)

seanmalbert (approved):
Updates are looking good and not seeing any issues, thanks @emilyjablonski

Issue #1157, opened 2021-04-21
Component: Tabs


slowbot:
**Goal** Navigation and container for showing and hiding panels of content

**Acceptance Criteria:**
Organization:
- Locate ui-components/src/navigation

Engineering Notes: 
- Stub with default data
- Be sure to address the accessibility items listed

Props:
- children: navigation items 
- current: current nav item, _boolean_
- disabled: disabled nav item, _boolean_
- tabpanels: panel content 

**Accessibility:**
Keyboard:
   Tab: 
   - When focus moves into the tab list, places focus on the active tab element
   - When the tab list contains the focus, moves focus to the next element in the tab sequence, which is the tabpanel element.

   Right or Left Arrows:
   - Moves focus to the next  or previous tab.
   - If focus is on the last tab, moves focus to the first or last tab.
   - Activates the newly focused tab.

Attributes:
role: tablist- Indicates that the element serves as a container for a set of tabs.
   - aria-label
      - Provides a label that describes the purpose of the set of tabs.
role: tab - (button) Indicates the element serves as a tab control.
   - aria-selected 
      - Indicates the tab control is activated and its associated panel is or is not displayed.
      - Set when a user activates a tab.
      - Does not change when focus moves in the tablist.
   - tabindex=-1
      - Removes the element from the page Tab sequence.
      - Set when a tab is not selected so that only the selected (active) tab is in the page Tab sequence.
      - Since an HTML button element is used for the tab, it is not necessary to set tabindex=0 on the selected (active) tab element.
     - aria-controls
        - Refers to the tabpanel element associated with the tab.
role: tabpanel - Indicates the element serves as a container for tab panel content.
   - aria-labelledby
      - Refers to the tab element that controls the panel.
   - tabindex=0
      -  Puts the tabpanel in the page Tab sequence.

Figma Mockup:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2776%3A0

jaredcwhite:
So far I'm impressed by https://reactcommunity.org/react-tabs/ and think we'll be able to use this for our tabs component.

[...]

PR #1305, opened 2021-06-01
Add new React Tabs-based component

jaredcwhite:
# Pull Request Template

## Issue

Addresses #1157 
- [x] This change is a dependency for another issue

## Description

This PR adds the [React Tabs dependency](https://github.com/react-component/tabs) and wraps those components so we get default styling. The previous `Tab` subcomponent of `TabNav` has been renamed to `TabNavItem`.

## Type of change

- [x] Storybook (Navigation / Tabs)

## Checklist:


jaredcwhite:
@slowbot Good thing we have a checklist! I just looked at the Storybook example in a narrow (aka mobile) window size and the tabs look pretty funky. Wonder what you think we should do. Truncate and add elipsis… ? Make them horizontally scrolling?

slowbot (approved):
@jaredcwhite Added a couple of small style updates. Otherwise LGTM

emilyjablonski:
@software-project what do you think about Jared's comment concerning the tabs on mobile?

ui-components/__tests__/navigation/Tabs.test.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react"
+import { render, cleanup, fireEvent } from "@testing-library/react"
+import { Tabs, TabList, Tab, TabPanel } from "../../src/navigation/Tabs"
+
+afterEach(cleanup)
+
+describe("<Tabs>", () => {
+  it("displays the right CSS", () => {
```
emilyjablonski, line 8: This is just testing implementation details - is there a way we can test component or prop functionality without looking at class names?

ui-components/src/navigation/Tabs.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <TabList>
+        <Tab>Title 1</Tab>
+        <Tab>Title 2</Tab>
+        <Tab>Long Tab Title 3</Tab>
+        <Tab disabled>Disabled Tab</Tab>
+      </TabList>
+
+      <TabPanel>
+        <h2>Any content 1</h2>
+        <p>Weee</p>
```
emilyjablonski, line 21: weeeeeee

ui-components/src/navigation/Tabs.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const className = ["tabs__tab"]
+  if (props.className) className.push(props.className)
+  return (
+    <ReactTab {...props} className={className}>
+      {children}
+    </ReactTab>
+  )
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
emilyjablonski, line 54: What is this doing / why the ignore?

jaredcwhite: Since I'm directly wrapping the React Tabs components, I have to set the `tabsRole` property on some of the components as described in docs here: https://github.com/reactjs/react-tabs#custom-components

But that line wasn't getting along with TS, not sure why.

ui-components/src/navigation/Tabs.stories.tsx
jaredcwhite: Haha, I forgot I'd left those in there 😅

ui-components/__tests__/navigation/Tabs.test.tsx
jaredcwhite: I tried to at least test the left/right arrow keys like on TabNav but was unsuccessful. Spent a while trying to get the events to fire but something was off. Basically we'd be testing a third-party component anyway, so not sure how much we need to dig into that. Maybe write a test to simulate clicking on the different tabs and double-check if that works?

emilyjablonski: Fair point on the 3rd party component :) I'll take a quick peek

jaredcwhite:
@slowbot That sounds good. I'll work on a stacked presentation for mobile.

ui-components/__tests__/navigation/Tabs.test.tsx
emilyjablonski: Looks like it's really well tested on their end. I took a peek just to see [how they did it](https://github.com/reactjs/react-tabs/blob/main/src/components/__tests__/Tabs-test.js) bc I also couldn't get the tabs to change, and it looks like they're pulling in another RTL sub-package called userEvent (🤷‍♀️) I would still recommend we change the first test though. I don't think we should rely on any of their class names, as we could pull a new version of this and it would break the test. Maybe we can just test our own prop? Something like:
` it("can pass a custom class name", () => {
    const { container } = render(
      <Tabs defaultIndex={1}>
        <TabList>
          <Tab className="other-tab">Other</Tab>
          <Tab>Default</Tab>
        </TabList>
        <TabPanel className="other-panel">OtherPanel</TabPanel>
        <TabPanel>DefaultPanel</TabPanel>
      </Tabs>
    )
    expect(container.getElementsByClassName("other-tab").length).toBe(1)
    expect(container.getElementsByClassName("other-panel").length).toBe(1)
  })`

emilyjablonski:
One comment on the test, otherwise LGTM! I personally think the mobile piece could be a separate issue if you want to break it out. Up to you.

ui-components/__tests__/navigation/Tabs.test.tsx
jaredcwhite: Thanks @emilyjablonski, I'll try that…good call on not hardcoding their internal class names. I'll see if I can remove one from our stylesheet in lieu of a custom one as well.

jaredcwhite:
I got the mobile tab stacking implemented and improved the resiliency of the tab/panel CSS and testing.

emilyjablonski (approved):

slowbot (approved):
@jaredcwhite LGTM

Issue #1297, opened 2021-06-01
Units out of order for Coliseum 


kathyccheng:
I expect it to sort by 1BR, 2BR, 3BR

![image.png](https://images.zenhubusercontent.com/5f0361f3c34cb3aabaf096f2/b291706f-f754-4d9a-b330-8675942ea26a)

PR #1306, opened 2021-06-01
fix: unit tables multiple rows for same rent type, sorting issues

emilyjablonski:
# Pull Request Template

## Issue

Addresses HBA#247 #1297 

## Description

This PR tackles two issues. The first is that when I made the change to show multiple rows for different rent types (% income vs static number), it also ended up showing multiple rows for the same type if the values were different. For example, Coliseum appeared as below with the bug. There should be two rows for 2 BR, not three. The different static rent numbers should appear as a range.
Screen Shot 2021-06-01 at 5 31 23 PM

With this fix, it now appears as below.
Screen Shot 2021-06-01 at 5 31 31 PM

The second issue this solves is unit summaries coming through unsorted (see the linked issues).

I also just tried my best to clean up the fxns, they were confusing.

## Type of change

Import a listing with rent ranges, like [Coliseum](https://github.com/housingbayarea/bloom/blob/0.3_alameda/backend/core/listing_data/coliseum.json). Check out the unit summaries table, the occupancy ranges, and the floor ranges.

## Checklist:


backend/core/src/shared/units-transformations.ts
```diff
@@ -115,58 +111,79 @@ const hmiData = (units: Units, byUnitType: UnitSummary[], amiPercentages: string
...
-  const summary = existingSummary ?? ({} as UnitSummary)
-  summary.unitType = unit.unitType
-  if (!summary.totalAvailable) {
-    summary.totalAvailable = 0
-  }
-  const minIncomeRange = minMaxValue(
-    {
-      min: parseFloat(summary.minIncomeRange?.min),
-      max: parseFloat(summary.minIncomeRange?.max),
+const getCurrenyString = (initialValue: string) => {
```
jaredcwhite, line 114: ⬆️ typo

jaredcwhite (commented):
Just saw one typo. Otherwise, changes make sense. I'm wondering if we can further break this file out into more discrete functions/constants? LOC is creeping ever higher…

backend/core/src/shared/units-transformations.ts
emilyjablonski: fixed, thanks!

emilyjablonski:
Curious what you would suggest breaking out? (Also does LOC stand for lines of code?)

jaredcwhite:
Yeah, Lines of Code. So basically anything that isn't directly performing the data transformations specific to unit sumamries/hmi/etc. I would recommend breaking out (and this isn't anything specific to your work, looks like the file's been touched by a bunch of people and there might be some stuff lurking underneath I originally worked on).

Suggestions:

* We can remove `AnyDict` and just use TS's `Record`
* `usd` could be extracted for use elsewhere. also `minMax`, `minMaxCurrency`, `getRoundedNumber`, etc.
* `bmrHeaders`, `UnitTypeSort`, etc. feel like constants that could be all defined in one centralized location
* `UnitMap` move to a types folder?

emilyjablonski:
That all makes a ton of sense! Thank you for taking a look and giving me context :)

jaredcwhite:
Sure, np :)

backend/core/src/shared/units-transformations.ts
```diff
@@ -115,58 +111,79 @@ const hmiData = (units: Units, byUnitType: UnitSummary[], amiPercentages: string
...
+  summary.areaRange = minMax(summary.areaRange, parseFloat(unit.sqFeet))
+
   return summary
 }
 
 type UnitMap = {
   [key: string]: Unit[]
 }
 
+const UnitTypeSort = ["studio", "oneBdrm", "twoBdrm", "threeBdrm"]
```
seanmalbert, line 185: This should probably be moved to an enum that `unitType` also uses on the unit entity. I don't think we need to do this right now with this task though, so I created this issue https://github.com/bloom-housing/bloom/issues/1335

backend/core/src/shared/units-transformations.ts
```diff
@@ -205,20 +229,18 @@ const summarizeUnitsByType = (
...
-        finalSummary.rentRange = minMaxInCurrency(finalSummary.rentRange)
-      }
-
       return finalSummary
     }
   )
-  return summaries.filter((item) => Object.keys(item).length > 0)
+  return summaries.sort((a, b) => {
+    return (
+      UnitTypeSort.indexOf(a.unitType[0]) - UnitTypeSort.indexOf(b.unitType[0]) ||
```
seanmalbert, line 240: Isn't the value of unitType a string here, like "threeBdrm"? If that's the case, then a.unitType[0] would be "t" in this example.

seanmalbert (commented):
Hey @emilyjablonski , this looks good. I generally agree with @jaredcwhite 's comments, though I'm not sure how much of it we want to do here. For UnitTypeSort for example, I created a separate issue to convert unitType to an enum and have UnitTypeSort be replaced with that.

I have a question on the summaries sort too.

backend/core/src/shared/units-transformations.ts
emilyjablonski: It's the index of the string in the array, so the value is a number

emilyjablonski: Should be all set now 👍  I think the reason this typo didn't create any visible issues is because the income is also tied pretty deeply to unit size, and we don't have any units right now where unit would be a tie breaker for income

seanmalbert (approved):

Issue #1170, opened 2021-04-23
Backend: File upload, add generating S3 presigned urls for PUT and GET 


pbn4:
Blocker for https://github.com/bloom-housing/bloom/issues/1153

seanmalbert:
Hey @pbn4 ,

Can you flesh out your thoughts on this, so we can attempt to size it? Were you thinking a direct upload solution like outlined here https://devcenter.heroku.com/articles/s3-upload-node#direct-uploading? What about bucket strategy (i.e. one bucket per client or can we get away with one bucket for all?) and file naming conventions?

From what I can tell, we currently aren't using the aws-sdk package to interact with s3 storage, so would this be the first implementation of this?

pbn4:
> Were you thinking a direct upload solution like outlined here https://devcenter.heroku.com/articles/s3-upload-node#direct-uploading? 

Yes this is exactly what I wanted us to implement. 

> What about bucket strategy (i.e. one bucket per client or can we get away with one bucket for all?) and file naming conventions?

I'd go with file naming conventions and 2 buckets (one for public read and one for private reads). Right now there are two use cases I know of:

- listing hero image https://github.com/bloom-housing/bloom/issues/1177 which could follow a pattern `/listing_<listing_id>/images/hero-image-name.jpg`, this is a public read, private write bucket (pattern does not really matter much as long as it's unique in this case),
- uploading paper applications per listing (cc: @slowbot) which could also follow the same pattern but for reading those you would need a presigned URL too (same as for writing)

> From what I can tell, we currently aren't using the aws-sdk package to interact with s3 storage, so would this be the first implementation of this?

yes first implementation of this. 

There are react libraries for frontend that handle a lot of what we need e.g. https://react-dropzone-uploader.js.org/ see s3 uploads [manual](https://react-dropzone-uploader.js.org/docs/s3). Backend could return a standardized message for a presigning request. This is for frontend folks to evaluate if the tool suits our needs though.

seanmalbert:
[...]

PR #1307, opened 2021-06-02
Add assets module

pbn4:
# Pull Request Template

## Issue

Addresses # (https://github.com/bloom-housing/bloom/issues/1170)

## Description

Assumption: frontend will handle file upload via a presigned URL directly to cloudinary and only return fileId to the backend for saving in `assets` table. 

Flow:
1. Frontend calls create POST `/assets/upload-url` and specifies `key` in the payload.
2. Frontend uploads a file to cloudinary using the presigned URL
3. Frontend calls POST `/assets` and provides `fileId` and `label` as input

This change adds assets module and `/assets` controller to the API. At the moment there are 2 endpoints `create` and `createUploadURL`, both are admin only (TBD what should be the permissions @seanmalbert). 

## Type of change

- [ ] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


seanmalbert (approved):
@pbn4 ,

This looks good, nice work! Don't forget to generate and commit the migration file, otherwise creating an asset will fail.

For asset endpoint access, I think it depends on how granular we want to get with listings management. It doesn't seem like there will be a need for asset specific permissions, since the only UI access to upload an asset will be via other modules, like listings. So it seems like we'll need an inheritance model of sorts, where assets are bound to other entities, and if a user has write access to any of those entities, then they will be allowed to generate an upload signature and create an asset. This is all for a future state of access control of course, so for now admin will do.

Issue #1158, opened 2021-04-21
Component Update: Money Field


slowbot:
**Goal**: Users have the ability to enter and edit any currency value

**Acceptance Criteria:**
Organization:
- Locate ui-components/src/forms

Engineering Notes: 
- Leverage work done for field and income 

Props:

Accessibility:

Figma Mockup:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2777%3A128809

emilyjablonski:
@slowbot I don't think we need to do any work here? For the existing money field we have, income, we used the Field component and pre-prend it with "$".

Is there anything wrong with that? Seems like that should extend itself to any money field.

slowbot:
@emilyjablonski I think we want to make it more general than "Income" and make it a generic "money" field since we have "rent" and other currency values

emilyjablonski:
It already is generic though, since we are using the Field component and pre-pending the $? We have no Income field.

slowbot:
@dominikx96 Do you want to add your thoughts here?

emilyjablonski:
@dominikx96

dominikx96:
@emilyjablonski @slowbot currently we have <Field /> component - and it's generic (we use it for text, numbers, money).

It includes following props that can be helpful for the "money type":

`prepend="$"` - we can set text before input
`validation={{ required: true, min: 0.01 }}` - we can control min. value

but I noticed something like:

`inputProps={{ step: 0.01, onBlur: formatValue }}`

and

```
 const formatValue = () => {
    const { income } = getValues()
    const numericIncome = parseFloat(income)
    if (!isNaN(numericIncome)) {
      setValue("income", numericIncome.toFixed(2))
    }
  }
```

in the _financial/income_ step.

it means, for each <Field /> instance we need to re-declare this function to parse money value.

We have 2 possibilities:

1. We can extend <Field /> component - e.g add an additional prop to detect 'money' type and move FormatValue()
2. Create a separated component

In the <Field /> component we have the necessary field setup and accessibility features, so maybe would be better to go with option 1? Any thoughts @emilyjablonski?

emilyjablonski:
Sounds good thanks for the context! Went with option 1 :)

PR #1356, opened 2021-06-11
1158/currency field component

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1158 

## Description

There's good context / discussion in the issue. Currently we only have a currency field in once place, the income step during the application. But with listings management we will have a number of different currency fields, so this issue was to consolidate the currency logic we had in the income step so that it could be reusable.

I pulled the currency pieces out of that individual file and put them into the Field component itself so now you can pass "currency" as the type and the Field component will handle all the currency things.

Also fixed a UI bug for the currency field in an error state, added some new stories, and retyped some form pieces that were typed as `any`.

## Type of change

Check out the new stories / the code :)

## Checklist:

### DIFF
--- a/ui-components/src/forms/Field.tsx
+++ b/ui-components/src/forms/Field.tsx
@@ -22,6 +22,7 @@ export interface FieldProps {
   register: UseFormMethods["register"]
   validation?: Record<string, any>
   disabled?: boolean
+  prepend?: string
   inputProps?: Record<string, unknown>
   describedBy?: string
   getValues?: UseFormMethods["getValues"]
@@ -45,7 +46,7 @@ const Field = (props: FieldProps) => {
 
   const formatValue = () => {
     if (props.getValues && props.setValue) {
-      const currencyValue = props.getValues()[props.name]
+      const currencyValue = props.getValues(props.name)
       const numericIncome = parseFloat(currencyValue)
       if (!isNaN(numericIncome)) {
         props.setValue(props.name, numericIncome.toFixed(2))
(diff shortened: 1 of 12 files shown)


ui-components/src/forms/Field.tsx
```diff
@@ -69,7 +84,7 @@ const Field = (props: FieldProps) => {
       {!isRadioOrCheckbox && label}
       {note}
       <div className={controlClasses.join(" ")}>
-        {props.prepend && <span className="prepend">{props.prepend}</span>}
+        {props.type === "currency" && <span className="prepend">{`$`}</span>}
```
dominikx96, line 87: hmm, maybe would be better to keep a prop for the "prepend" text to do not hardcode this value?

ui-components/src/forms/Field.tsx
```diff
@@ -41,7 +43,20 @@ const Field = (props: FieldProps) => {
     controlClasses.push(props.controlClassName)
   }
 
-  const type = props.type || "text"
+  const formatValue = () => {
+    if (props.getValues && props.setValue) {
+      const currencyValue = props.getValues()[props.name]
```
dominikx96, line 48: I think it's okay, but I suppose simpler would be: `props.getValues(props.name)`

ui-components/src/forms/TextField.stories.tsx
```diff
@@ -0,0 +1,63 @@
+import React from "react"
+import { useForm } from "react-hook-form"
+import { Field } from "./Field"
+
+export default {
+  title: "Forms/Text Field",
+  decorators: [(storyFn: any) => <div style={{ padding: "1rem" }}>{storyFn()}</div>],
+}
+
+export const TextFieldDefault = () => {
```
dominikx96, line 10: What do you think about using just `Field` as a story name? I think this field is not just for text, but also for numbers, currency, etc. 🤔

dominikx96 (approved):
Just a few comments, but overall LGTM! 👍

Issue #1362, opened 2021-06-14
Bug - Application flagged sets endpoint returns incorrect flagged items quantity


dominikx96:
I noticed that the `/applicationFlaggedSets` endpoint returns an empty array with duplicated sets, but the `totalFlagged` property value is `2`. Probably it's a backend bug.

This property is used in the flags navigation (blue tag).

Tested in: https://partners.housingbayarea.bloom.exygy.dev/listings/fdafca36-981a-4858-a1dd-6a32635957f3/flags

![Image 2021-06-14 at 6 14 06 PM](https://user-images.githubusercontent.com/16655908/121925278-0d040b00-cd3d-11eb-88eb-9c9aa4a07d19.png)

PR #1366, opened 2021-06-15
Issue 1362/total flagged bug

pbn4:
# Pull Request Template

## Issue

Addresses # (#1362)

## Description

This PR addresses two problems:

- incorrectly displaying totalFlagged (backend ignored listingId query param when counting totalFlagged in AFS table)
- a race condition when two applications were added in similar time, resulting in one of them not being captured by an AFS table (e.g. a duplicate set contained one application less than it should). The new implementation now runs the entire application insert and AFS creation as a transaction with serializable isolation level. This now leads to potential read/write dependencies between transactions and PSQL aborting one of the ongoing transactions (in a case where two users submit an application in the same time). I have implemented basic expontential backoff retry logic with `async-retry` package, but the maximum number of retries is 6, with 200ms backoff as a starting base and with expontent factor equal to 2. So it's a maximum wait of `200+400+800+1600+3200 + 6400 = 12600ms` otherwise backend will throw a 500 error and abort both application and AFS creation. 

@seanmalbert Maybe we should consider returning something like `429 too many requests` instead of a 500 for the frontend to display an appropriate message then such transaction dependency happens? 

## Type of change

- [ ] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/backend/core/src/application-flagged-sets/application-flagged-sets.service.ts
+++ b/backend/core/src/application-flagged-sets/application-flagged-sets.service.ts
@@ -3,7 +3,14 @@ import { PaginatedApplicationFlaggedSetQueryParams } from "./application-flagged
 import { AuthzService } from "../auth/authz.service"
 import { ApplicationFlaggedSet } from "./entities/application-flagged-set.entity"
 import { InjectRepository } from "@nestjs/typeorm"
-import { Brackets, DeepPartial, Repository, SelectQueryBuilder } from "typeorm"
+import {
+  Brackets,
+  DeepPartial,
+  Repository,
+  SelectQueryBuilder,
+  getManager,
+  EntityManager,
+} from "typeorm"
 import { paginate } from "nestjs-typeorm-paginate"
 import { Application } from "../applications/entities/application.entity"
 import { ApplicationFlaggedSetResolveDto } from "./dto/application-flagged-set.dto"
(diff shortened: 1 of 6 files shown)


dominikx96 (approved):
Tested it locally (manually) and I don't see any issues, focused just on the frontend part.

pbn4:
@seanmalbert Can we merge or do you still plan to review?

backend/core/src/application-flagged-sets/application-flagged-sets.service.ts
```diff
@@ -56,63 +66,87 @@ export class ApplicationFlaggedSetsService {
...
-    await this.afsRepository.save(afs)
-    return afs
+      for (const appNotToBeResolved of appsNotToBeResolved) {
+        appNotToBeResolved.markedAsDuplicate = false
+        await transApplicationsRepository.save(appNotToBeResolved)
+      }
+      appsToBeResolved.forEach((app) => (app.markedAsDuplicate = true))
+      await transAfsRepository.save(afs)
+      return afs
+    })
```
seanmalbert, line 102: I know this logic already existed, but these separate iterations could probably be condensed down into one. I'm not sure how large these record sets in this case get though, so not sure if it's worth any savings.

seanmalbert (approved):
@pbn4 ,
This looks good. A 429 may be better than a 500 in this case, but I don't know if that message should be getting back to the user. Ideally the applicant doesn't lose any data and they're told to try again in a few seconds unless the problem persists. Also, do you have a good sense on how frequent this issue has been?

pbn4:
> Also, do you have a good sense on how frequent this issue has been?

In a test sending two requests in parallel it occured almost 100% of the time. 

I added 429 throwing and collapsed DB writes where you have pointed out.

Issue #1418, opened 2021-06-24
Fix Publish and some Prototype Clean Up 


kathyccheng:
Fix Publish functionality 
Add button + link to detail and then turn on on staging 
Fix cache control so we can see changes

PR #1422, opened 2021-06-25
1418/cache and cleanup

seanmalbert:
# Pull Request Template

## Issue
#1418 

Addresses # (issue)

## Description

This adds cache manager to the listing controller, adds in link/button for detail/add listings behind a flag, fixes issues with creating a new listings without units and some cleanup.

## Type of change

In partners, you should be able to add/edit listings and see the changes immediately without having to restart.

- [x] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


backend/core/src/listings/listings.controller.ts
```diff
@@ -82,7 +87,15 @@ export class ListingsController {
   @Post()
   @ApiOperation({ summary: "Create listing", operationId: "create" })
   async create(@Body() listingDto: ListingCreateDto): Promise<ListingDto> {
-    return mapTo(ListingDto, await this.listingsService.create(listingDto))
+    const listing = await this.listingsService.create(listingDto)
+    /**
+     * clear list caches
+     * As we get more listings we'll want to update this to be more selective in clearing entries
+     */
+    for (const key of this.cacheKeys) {
```
pbn4, line 95: To clarify, this is clearing the entire cache whena new listing is created, right?

backend/core/src/listings/listings.controller.ts
```diff
@@ -44,7 +44,13 @@ import { defaultValidationPipeOptions } from "../shared/default-validation-pipe-
 @UseGuards(OptionalAuthGuard, AuthzGuard)
 @UsePipes(new ValidationPipe(defaultValidationPipeOptions))
 export class ListingsController {
-  constructor(private readonly listingsService: ListingsService) {}
+  cacheKeys: string[]
+  constructor(
+    @Inject(CACHE_MANAGER) private cacheManager,
+    private readonly listingsService: ListingsService
+  ) {
+    this.cacheKeys = ["/listings", "/listings?filter[$comparison]=%3C%3E&filter[status]=pending"]
```
pbn4, line 52: What is the meaning of this expression?

backend/core/src/listings/listings.controller.ts
```diff
@@ -100,12 +113,25 @@ export class ListingsController {
...
+    }
+    await this.cacheManager.del(`/listings/${listingId}`)
+    return mapTo(ListingDto, listing)
   }
 
   @Delete(`:listingId`)
   @ApiOperation({ summary: "Delete listing by id", operationId: "delete" })
   async delete(@Param("listingId") listingId: string) {
     await this.listingsService.delete(listingId)
+    for (const key of this.cacheKeys) {
```
pbn4, line 132: This deletion of cacheKeys + listingId related key could be a private method because it's exactly the same code in every handler using it.

backend/core/src/listings/listings.service.ts
```diff
@@ -27,6 +27,8 @@ export class ListingsService {
       .leftJoinAndSelect("units.amiChart", "amiChart")
       .leftJoinAndSelect("listings.jurisdiction", "jurisdiction")
       .leftJoinAndSelect("listings.reservedCommunityType", "reservedCommunityType")
+      .leftJoin("listings.applications", "applications")
```
pbn4, line 30: What is the performance overhead of this join? Because it's probably joining ~1.5k rows for some listings and might result in the memory problem we faced some time ago. Please take a look how `totalFlagged` is implemented here https://github.com/bloom-housing/bloom/blob/master/backend/core/src/application-flagged-sets/application-flagged-sets.service.ts#L44 and maybe consider going with COUNT query over applications with specified listing_id foreign key.

pbn4 (changes requested):

backend/core/src/listings/listings.controller.ts
seanmalbert: The cache entry keys are the route path by default, so the first one is from partners and the second is coming from public. We can also specify the cache key with a decorator per operation, if we don't want it based off the route.

backend/core/src/listings/listings.controller.ts
seanmalbert: Correct. Instead of clearing the cache we could update the cache entry to have the new listing.

backend/core/src/listings/listings.controller.ts
seanmalbert: I agree. Before I create any methods though, I want us to come to an agreement on how we should handle posts/puts for the list route caches. Since we have such few listings I think it's fine to clear those routes from the cache, but it's easy enough now to add in handling it by updating the entries. In the case of `post`, this is relatively easy, we get the current entry, add the new listing and set it. In `put` there's an extra step of making sure we filter out the old listing from the cached route before adding it. This all becomes more involved when paging and filtering get added, where we'd want to only clear/update the routes that the listing is in, which we could also accommodate now, or just iterate when we need to.

backend/core/src/listings/listings.service.ts
seanmalbert: Ah, yes, I like this approach better. I'll update.

seanmalbert:
@emilyjablonski ,
If you find yourself blocked by updates here regarding the fixes to getting a listing displaying without units, you can either pull in the changes here from ListingView, UnitTables and StandardTable, or I can split those out into a separate PR and merge in sooner.

emilyjablonski:
LGTM 👍

backend/core/src/listings/listings.service.ts
seanmalbert: I removed this and added logic to listing service to fetch the application counts and map them to the listings. I also pulled in the origin header, so that we only do this on partners, so there is a new env variable for partners URL.

backend/core/src/listings/listings.controller.ts
seanmalbert: Also, rather than setting the keys like this, we could get the keys via `cacheManager.keys()` and filter out the keys for individual listings (i.e. /listings/[id]), so we only modify the "list" entries.

backend/core/src/listings/listings.controller.ts
pbn4: > Since we have such few listings I think it's fine to clear those routes from the cache 
I'm OK with that too.

pbn4 (approved):

seanmalbert:
Added cacheLib with clearCacheKeys function, now used by listing endpoints.

Issue #1381, opened 2021-06-17
Listings: Application Address MVP Fields 


kathyccheng:
**Goal**: Leasing agent should be able to specify details about how to pickup and submit paper applications. 

**Fields**:
Field Name | Entity Model | Type | Current State | Required
------------ | ------------- | --- | ----- | -----
Leasing Agent Address | leasingAgentAddress | address| Yes | No
Mailing Address | applicationAddress | address| Yes | Yes
Picked Up | ApplicationMethod | boolean | Yes | Yes
Pick Up Address | applicationPickUpAddress | address | Yes| If enabled
Pick Up Office Hours | applicationPickUpAddressOfficeHours | string | Yes | No
Dropped Off | ApplicationMethod | boolean | New Field | Yes
Dropped Off Address | | address | New Field | If enabled
Dropped Off Office Hours | | string | New Field | No
Postmarks Considered | | boolean | New Field | Yes
Postmark by Date | postmarkedApplicationsReceivedByDate | date | Yes | If enabled
Additional Application Submission Notes | | string | No | No

**Acceptance Criteria**
- If application can be Picked Up, ask where application will be picked up
   - If submission address use previously entered Submission Address
   - If leasing agent address use previously entered Leasing Agent Address
   - If another address provide fields for Pick Up Address and Office Hours
- If application can be Dropped Off, ask where application will be dropped off
   - If submission address use previously entered Submission Address
   - If leasing agent address use previously entered Leasing Agent Address
   - If another address provide fields for Dropped Off Address and Office Hours
- If postmarks are considered, provide field for Postmark by Date

**Engineering Notes**
- Remove PO Box application method
- Additional Application Submission Notes will be added to application submission notes 

**Public**
- Will be removing PO Box application method and revise how we render application submission dates note

![image.png](https://images.zenhubusercontent.com/5f0361f3c34cb3aabaf096f2/df3f5e74-0262-4086-a7e6-927bdb7b904b)

**Figma**:
Default:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2726%3A126111

Address:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2851%3A19

Postmarks Yes:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2726%3A126796

Picked Up Yes:
[...]

PR #1425, opened 2021-06-25
1381/listings management, application addresses

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1381

## Description

Adds the application addresses section to the listings management prototype. This section is probably 10x the work and size of any other section so far so it might be harder to review 😓.

The new backend fields we needed are: `applicationMailingAddress` `applicationDropOffAddress` `applicationDropOffAddressOfficeHours` `applicationPickUpAddressType` `applicationDropOffAddressType` `additionalApplicationSubmissionNotes`

Here are the [Figma](https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2726%3A126111) states. Visually my sections are not arranged in exactly the same way because our form grids are not responsive to mobile in the exact layout in the designs. You have a leasing agent address section, a mailing address checkbox that toggles on the mailing address field, a radio button for application pickup that toggles an application pickup location radio group wherein if you select another address it toggles on a new address field, and the same as for application drop off. Lastly you have a postmark radio that if selected toggles on a postmark date field. Application submission notes is new and not currently used anywhere in the public app.

The issue included removing the PoBox application method and I went ahead and removed the LeasingAgent and PaperPickup methods as well since they were used identically to the PoBox method and seemed really unnecessary.

## Type of change

Oh boy. I would create a few listings from listings/add in partners doing a number of different combinations of possible inputs and then checking them still within partners on the details page to ensure the data is showing appropriately w all the toggleable options. Mailing address should only show as a radio option if you have indicated that a mailing address should exist with the checkbox.

The biggest thing I would test is then re-editing a listing you just made and making sure all the default values load in appropriately with what you just submitted (I had a number of bugs here). Include toggling sections on and off again after having the form loaded with default values to make sure all is working appropriately.

I would also test adding all the addresses and then removing them and making sure they are actually removed (had bugs here).

From the public side, I would create a custom leasing agent address, mailing address, pickup address, and dropoff address and make sure they all appear. I would create one with no addresses and make sure that looks alright. And lastly I would select the dropoff and pickup to connect to the leasing and mailing addresses to make sure those pull through.

## Checklist:


seanmalbert:
@emilyjablonski and @pbn4 ,

Since I don't think we have time before MVP demo (which is what we get done this week), a decent and re-usable solution is that we can create two new columns for dropoff/pickup address references. If these values are the "type" of address (e.g. leasingAgentAddress), then on the frontend we can check that value and use it accordingly. When we move to the permanent address table solution, we can either keep the fields referencing the type or we can update it to just reference the addressId.

pbn4:
@sean @emilyjablonski I added two previously discussed Listing columns to this PR.

backend/core/src/listings/entities/listing.entity.ts
```diff
@@ -157,6 +163,19 @@ class Listing extends BaseEntity {
   })
   applicationPickUpAddressType?: ListingApplicationPickUpAddressType | null
 
+  @Column({ type: "jsonb", nullable: true })
+  @Expose()
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @ValidateNested({ groups: [ValidationsGroupsEnum.default] })
+  @Type(() => AddressDto)
+  applicationDropOffAddress: AddressDto | null
```
pbn4, line 176: One comment is that at some point I realised that our client generation library will only create proper typing for optional properties marked with `?` sign.

pbn4: But that probably means we should fix it everywhere, not a part of this PR.

emilyjablonski:
Looking into the unit test and build issues 👀

sites/partners/src/listings/PaperListingDetails/sections/DetailApplicationAddress.tsx
```diff
@@ -0,0 +1,208 @@
+import React, { useContext } from "react"
+import { t, GridSection, ViewItem, GridCell } from "@bloom-housing/ui-components"
+import { ListingContext } from "../../ListingContext"
+
+const DetailApplicationAddress = () => {
+  const listing = useContext(ListingContext)
+
+  const getAddressString = (addressType: string): string | undefined => {
+    if (addressType === "leasingAgent") return t("listings.leasingAgentAddress")
+    if (addressType === "mailingAddress") return t("application.contact.mailingAddress")
```
dominikx96, line 10: Can we use `ListingApplicationPickUpAddressType` enum from the backend-swagger instead of `string`?

sites/partners/src/listings/PaperListingForm/index.tsx
```diff
@@ -148,7 +169,29 @@ const ListingForm = ({ listing, editMode }: ListingFormProps) => {
...
+        applicationPickUpAddressType:
+          ListingApplicationPickUpAddressType[data.whereApplicationsPickedUp] ?? null,
+        applicationDropOffAddress:
+          data.canApplicationsBeDroppedOff && data.whereApplicationsPickedUp === "anotherAddress"
+            ? data.applicationDropOffAddress
+            : null,
+        applicationPickUpAddress:
+          data.canPaperApplicationsBePickedUp && data.whereApplicationsPickedUp === "anotherAddress"
+            ? data.applicationPickUpAddress
+            : null,
```
dominikx96, line 189: What do you think about creating const/enum for `anotherAddress` string? It could prevent string repetition and potential problems :)

sites/partners/src/listings/PaperListingForm/index.tsx
```diff
@@ -148,7 +169,29 @@ const ListingForm = ({ listing, editMode }: ListingFormProps) => {
       const typedData: FormListing = {
```
dominikx96, line 169: Maybe we can create a separate function to create the request body in a good shape (move all typedData logic). It could reduce the amount of `onSubmit` function complexity.

ui-components/src/page_components/listing/listing_sidebar/Apply.tsx
```diff
@@ -66,11 +64,26 @@ const Apply = (props: ApplyProps) => {
     return method.type == ApplicationMethodType.FileDownload
   })
 
+  const getDropOffAddress = (addressType: string | undefined) => {
```
dominikx96, line 67: I think we can use `ListingApplicationDropOffAddressType` as `addressType` type, what do you think?

ui-components/src/page_components/listing/listing_sidebar/Apply.tsx
```diff
@@ -66,11 +64,26 @@ const Apply = (props: ApplyProps) => {
...
+    return addressType === ListingApplicationDropOffAddressType.leasingAgent
+      ? listing.leasingAgentAddress
+      : addressType === ListingApplicationDropOffAddressType.mailingAddress
+      ? listing.applicationMailingAddress
+      : listing.applicationDropOffAddress
+  }
+
+  const getPickUpAddress = (addressType: string | undefined) => {
```
dominikx96, line 75: I think we can use `ListingApplicationPickUpAddressType ` as `addressType` type, what do you think?

dominikx96 (changes requested):

jaredcwhite:
@emilyjablonski Please don't hate me! 😅 It looks like the `Apply.stories.tsx` Storybook file will need get updated as well regarding the changes to application methods and the new listings fields. There might be something lurking too in another listing-specific story.

emilyjablonski:
No worries at all, can do!

backend/core/src/listings/entities/listing.entity.ts
seanmalbert: Added new issue #1444 to address this.

backend/core/src/listings/types/listing-application-drop-off-address-type.ts
```diff
@@ -0,0 +1,4 @@
+export enum ListingApplicationDropOffAddressType {
+  leasingAgent = "leasingAgent",
+  mailingAddress = "mailingAddress",
+}
```
seanmalbert, line 4: This and the one below can be shared with `ListingApplicationAddressType` or just `AddressType` is probably fine.

ui-components/src/page_components/listing/listing_sidebar/Apply.tsx
```diff
@@ -66,11 +64,26 @@ const Apply = (props: ApplyProps) => {
...
+  }
+
+  const getPickUpAddress = (addressType: ListingApplicationPickUpAddressType | undefined) => {
+    return addressType === ListingApplicationPickUpAddressType.leasingAgent
+      ? listing.leasingAgentAddress
+      : addressType === ListingApplicationPickUpAddressType.mailingAddress
+      ? listing.applicationMailingAddress
+      : listing.applicationPickUpAddress
+  }
+
```
seanmalbert, line 82: Once you combine the address types, then you should be able to have one function for this that takes the address key whose value would be `applicationDropOffAddress` or `applicationPickUpAddress`

seanmalbert (approved):
This looks good @emilyjablonski . There are a couple of minor updates to make noted in the comments, and aside from that the address layout differs a little form the mocks @ https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2717%3A158654, where some fields should be next to each other. I'm not sure that piece needs to block this from getting merged in, so you can merge this once the address type changes are made.

emilyjablonski:
Sounds good thanks! In the PR description I talk about why the layout is not the exact same as the designs: `Visually my sections are not arranged in exactly the same way because our form grids are not responsible to mobile in the exact layout in the designs.`

emilyjablonski:
Working through what I think is an issue w the database migration for the type listing address type, then should be g2g

seanmalbert:
@emilyjablonski ,
After trying a bunch of workarounds, including duplicating the branch to ensure it wasn't using the cache, I think we may have to update the migration script to handle dropping listings_application_address_type_enum_old as if it were there, because for some reason, it is.

seanmalbert:
@emilyjablonski ,
Got the migration file updated and the backend tests are passing in circleci. It seems that the migration runner doesn't like it when more than one column shares the same enum, so creating two with the same values works (why it worked before). This doesn't seem to cause an issue in interacting with the db either. The only thing we'll want to watch out for is if/when we alter the enum values, we want to make sure that both are altered, and not the one. I think the migration generation should account for that though.

Issue #1382, opened 2021-06-17
Listings: Application Dates MVP 


kathyccheng:
**Goal**: Leasing agent should be able to add events to a listing

**Fields**:
Field Name | Entity Model | Type | Current State | Required
------------ | ------------- | --- | ----- | -----
Application Due Date | applicationDueDate | date | Yes | Yes
Application Due Time | | date | No | Yes

**Acceptance Criteria**
- Applicants should be able to add application due date 

**Notes**
* Title: Application Dates
* Subtitle: Tell us about important dates related to this listing.
* For the MVP we're just going to start with the application due date so we won't be tackling start date or open house for now, see screenshot below 

![image.png](https://images.zenhubusercontent.com/5f0361f3c34cb3aabaf096f2/d466f876-c922-4f76-ab42-c83109827f8e)

**Figma**:
Default:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2726%3A127598

Filled:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2726%3A127818

Original Issue: https://github.com/bloom-housing/bloom/issues/1182

PR #1432, opened 2021-06-25
1382/listings management, application dates

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1382 

## Description

Adds the application dates section to the listings management MVP. Note: our existing time fields has a default `seconds` input which I don't think we always need, but moving forward with it as is for now.

## Type of change

Create a new listing from partners, check out the new section at the bottom, save the listing and checkout the detail page, re-edit the listing and ensure the data is properly loaded. Preview the listing and see the new time field at the top of the right hand panel.

## Checklist:


sites/partners/src/listings/PaperListingForm/sections/ApplicationDates.tsx
```diff
@@ -0,0 +1,69 @@
+import React from "react"
+import { useFormContext } from "react-hook-form"
+import { t, GridSection, DateField, TimeField } from "@bloom-housing/ui-components"
+import { FormListing } from "../index"
+import moment from "moment"
```
dominikx96, line 5: Do we still want to use `moment`? Maybe it's a good moment to add `dayjs` and then replace other moment instances in a separated PR?

sites/partners/src/listings/PaperListingForm/sections/ApplicationDates.tsx
```diff
@@ -0,0 +1,69 @@
...
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const { register, watch } = formMethods
+
+  return (
+    <div>
+      <GridSection grid={false} separator>
+        <span className="form-section__title">{t("listings.sections.applicationDatesTitle")}</span>
+        <span className="form-section__description">
+          {t("listings.sections.applicationDatesSubtitle")}
+        </span>
```
dominikx96, line 23: I think it would be better to use the `title` prop and create a `description` prop (currently we have `subtitle`, but it has different text styles).

ui-components/src/notifications/ApplicationStatus.tsx
```diff
@@ -30,10 +30,13 @@ const ApplicationStatus = (props: ApplicationStatusProps) => {
...
   } else {
-    const date = listing.applicationDueDate
     if (listing.applicationDueDate) {
-      const dueDate = moment(date)
-      formattedDate = dueDate.format("MMM. DD, YYYY") + " at " + dueDate.format("h:mm A")
+      const dueDate = moment(listing.applicationDueDate)
+      const dueTime = moment(listing.applicationDueTime)
+      formattedDate = dueDate.format("MMM. DD, YYYY")
+      if (listing.applicationDueTime) {
+        formattedDate = formattedDate + " at " + dueTime.format("h:mm A")
```
dominikx96, line 38: Instead of hardcoded `at` we can use `t("t.at")` :)

sites/partners/src/listings/PaperListingForm/index.tsx
```diff
@@ -137,6 +152,35 @@ const ListingForm = ({ listing, editMode }: ListingFormProps) => {
     }
   }
 
+  console.log(listing)
```
dominikx96, line 155: 🧹

sites/partners/src/listings/PaperListingForm/sections/ApplicationDates.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { t, GridSection, DateField, TimeField } from "@bloom-housing/ui-components"
+import { FormListing } from "../index"
+import moment from "moment"
+
+type ApplicationDatesProps = {
+  listing?: FormListing
+}
+
+const ApplicationDates = ({ listing }: ApplicationDatesProps) => {
+  const formMethods = useFormContext()
```
dominikx96, line 12: hmm, not sure, but probably we have an access to the form data from the `useFormContext()`?
If yes - the additional prop `listing` is not necessary :)

`const { register, watch, formState } = formMethods`

https://react-hook-form.com/api/useform/formstate

dominikx96 (changes requested):
Reviewed frontend part, overall looks good! I put just a couple of comments :)

ui-components/src/notifications/ApplicationStatus.tsx
emilyjablonski: OOPS! Thanks! I copy-pastad this from somewhere else, I'll fix it in both places 👍

sites/partners/src/listings/PaperListingForm/sections/ApplicationDates.tsx
emilyjablonski: I think a change like that should probably be handled in a separate PR so it stays consistent if that's ok

sites/partners/src/listings/PaperListingForm/sections/ApplicationDates.tsx
emilyjablonski: Added a new description prop and made sure it has the right styles & updated all the sections 👍 thanks!

sites/partners/src/listings/PaperListingForm/sections/ApplicationDates.tsx
emilyjablonski: I don't see the listing here as part of form state when I restructure it and log it in the console as a heads up

seanmalbert: getValues should have it.

seanmalbert (changes requested):
This looks good. Only a couple of relatively minor updates needed. The main thing to change is the consistency of using moment utc for displaying the dates and you might be able to use getValues instead of passing in listing to the form.

sites/partners/src/listings/PaperListingForm/sections/ApplicationDates.tsx
emilyjablonski: so getValues has the form fields but the form fields don't match directly to the listing, so I can get `applicationDueTimeField: { hours: "",  minutes: "" , period: "am" , seconds: ""  }` but not the actual time, for now I'm still passing the listing as a prop but let me know if you'd prefer a different approach

seanmalbert: I'm ok with this for now. I think at some point we may want to revisit the date and times fields to work in a way that we don't have to specify a separate "field', as in `applicationDueDateField` vs `applicationDueDate`, though I understand the way it is now makes sense, because it doesn't tie a date library to the component.

seanmalbert (approved):
@emilyjablonski , thanks for the updates!

Issue #1184, opened 2021-04-26
Listings Management Fieldset: Listing Units


slowbot:
**Goal**: Leasing agent should be able to add units to a listing

**Fields**:
Field Name | Entity Model | Type | Current State | Required
------------ | ------------- | --- | ----- | -----
Unit Display |  | boolean | ? | Yes | 
Unit Number | ? | number | Yes | No | 
Unit Type | unitType | string | Yes | Yes | 
Number of Bathrooms | numBathrooms | number | Yes | Yes | 
Unit Floor | floor  | number | Yes | No | 
Square Footage | sqFeet | string | Yes | Yes |
Minimum Occupancy | minOccupancy | number | Yes | Yes | 
Maximum Occupancy | maxOccupancy | number | Yes | Yes | 
Unit Status | ? | string | Yes | Yes | 
AMI Chart | ? |  | Yes | Yes | 
Percentage of AMI | amiPercentage | number | Yes | Yes | 
Unit Rent Type |  | string | New field | Yes | 
Fixed Monthly Rent | Currently "Monthly Rent"  | number | Yes | If selected | 
Percentage of Income Rent |  | number | ? | If selected | 
Minimum Monthly Income | monthlyIncomeMin | string | Yes | Yes | 
Accessibility Priority Type | priorityType  | string | Yes | Yes | 

**Acceptance Criteria**
- Selecting Unit Display "unit types" Public front end display of unit type accordions
- Selecting AMI Chart and Percentage of AMI should reference an existing AMI chart. On the Public front end this should determine the Monthly Income Max and Annual Income Max by Household Size
- If Unit Rent Type is Fixed, display Monthly Rent and Minimum Monthly Income Fields
- If Unit Rent Type is % of Income, display Percentage if Income Rent

**Questions**
- Should min and max occupancy be set by the unit type and owned by the jurisdiction? 
- Should we figure out a way to "preview" Household Max Income and Occupancy tables other than Listing Preview so Agents can see generated unit summary values?

**Figma**:
Default:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2721%3A128025

Types No:
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2742%3A127835

Modal
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2742%3A127862

Modal Fixed:
[...]

PR #1439, opened 2021-06-29
1184/unit type rent

pbn4:
# Pull Request Template

## Issue

Addresses # (https://github.com/bloom-housing/bloom/issues/1184)

## Type of change

- [ ] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


backend/core/src/units/dto/unit.dto.ts
```diff
@@ -15,6 +22,27 @@ export class UnitDto extends OmitType(Unit, ["property", "amiChart"] as const) {
...
+  @ValidateNested({ groups: [ValidationsGroupsEnum.default] })
+  @Type(() => IdDto)
+  unitRentType?: IdDto
+
+  @Expose()
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @ValidateNested({ groups: [ValidationsGroupsEnum.default] })
+  @Type(() => IdDto)
+  unitAccessibilityPriorityType?: IdDto
```
seanmalbert, line 45: I think we should be consistent in the naming conventions for these reference IDs. This convention doesn't seem to be followed anywhere, but it makes sense to me to name these fields based off of what they are, the id, so it would be unitTypeId, for example.

backend/core/src/units/entities/unit.entity.ts
```diff
@@ -156,6 +160,27 @@ class Unit {
   @IsOptional({ groups: [ValidationsGroupsEnum.default] })
   @IsBoolean({ groups: [ValidationsGroupsEnum.default] })
   bmrProgramChart?: boolean | null
+
+  @ManyToOne(() => UnitType, { eager: true, nullable: true })
+  @Expose()
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @ValidateNested({ groups: [ValidationsGroupsEnum.default] })
+  @Type(() => UnitType)
+  unitTypeRef?: UnitType | null
```
seanmalbert, line 169: Shouldn't this be unitType, not unitTypeRef?

seanmalbert: Also, `unitType` is still defined here as a string, which I don't think we want anymore with this update. With that in mind, wouldn't we also want a script to create a the unit types based off of these strings and assign them accordingly?

backend/core/src/units/entities/unit.entity.ts
```diff
@@ -156,6 +160,27 @@ class Unit {
...
+  @ValidateNested({ groups: [ValidationsGroupsEnum.default] })
+  @Type(() => UnitRentType)
+  unitRentType?: UnitRentType | null
+
+  @ManyToOne(() => UnitAccessibilityPriorityType, { eager: true, nullable: true })
+  @Expose()
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @ValidateNested({ groups: [ValidationsGroupsEnum.default] })
+  @Type(() => UnitAccessibilityPriorityType)
+  unitAccessibilityPriorityType?: UnitAccessibilityPriorityType | null
```
seanmalbert, line 183: I think these have similar issues to the handling of unit type. For example, isn't unitAccessibilityPriorityType replacing priorityType, or rather, shouldn't it?

seanmalbert (changes requested):
Hey @pbn4 ,

I think we may want more updates here to make the best use of adding these new references. Please see my comments and let me know if you have any questions.

backend/core/src/migration/1625825502154-refactor-unit-types-rent-types-and-priority-types.ts
```diff
@@ -0,0 +1,91 @@
...
+    await queryRunner.query(
+      `ALTER TABLE "units" ADD CONSTRAINT "FK_6981f323d01ba8d55190480078d" FOREIGN KEY ("priority_type_id") REFERENCES "unit_accessibility_priority_types"("id") ON DELETE NO ACTION ON UPDATE NO ACTION`
+    )
+    for (const accessibilityPriorityType of [
+      "Mobility and Hearing & Visual",
+      "Mobility and Mobility with Hearing & Visual",
+      "Mobility and hearing",
+      "Mobility",
+      "mobility",
+    ]) {
```
seanmalbert, line 46: Rather than have these hardcoded, can we get the values in the db with distinct select of priority types and then loop over them. We can do the same with unit and rent type.

seanmalbert (changes requested):
Hey @pbn4 ,

Thanks for all the updates here, I like the changes to seeds. I would like the migration script to replace the hardcoded values with selecting the distinct values from the tables and using those. 

There are also issues with the partner's site with units, which I'm looking into. I can fix that up, if you want to focus on the updates to the migration script.

seanmalbert:
@pbn4 ,

I found an issue, priorityType.name was missing from some places in partners, so I updated that. There are still updates needed to the unit form:

1.  to fetch unit type and priority type options
2. update the fields to use these

I'm not sure if you or @dominikx96 has time to make these updates. I'm going to jump to issues and see if I can circle back on those two items by the end of day.

backend/core/src/migration/1625825502154-refactor-unit-types-rent-types-and-priority-types.ts
pbn4: Improved in latest commit @seanmalbert

Issue #1443, opened 2021-07-01
Listings - Unpublish 


kathyccheng:
As a leasing agent, 
I might want to unpublish a listing 
So that I can remove an error or correct a mistake 

**Notes**
* Unpublish will move a listing back to draft form 
* It should no longer appear publicly 

Figma 
![image.png](https://images.zenhubusercontent.com/5f0361f3c34cb3aabaf096f2/f0df9cc8-98ad-4398-9ecc-b24ab319bd6a)

https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=2521%3A0

kathyccheng:
Looks good!

Issue #1442, opened 2021-07-01
Listings - Manual Close


kathyccheng:
As a leasing agent, 
I might want to close a listing manually
So that I have flexibility in case there is no application due date or something happens 

Notes
* When a leasing agent closes a listing manually, then it should moved to the Closed state. It should over-ride application due date

kathyccheng:
@slowbot I think we should support a manual close? For First Come First Serve listings with no due date, otherwise there's no way to close it. But maybe we want an are you sure alert? What do you think?

slowbot:
@kathyccheng  I think we can use the same modal we use for delete 

![image](https://user-images.githubusercontent.com/13915/124192192-bf480c00-da79-11eb-8e78-af47d0c926ed.png)

kathyccheng:
@dominikx96 the Close action is working on the partners app, it marks the listing as closed. But I don't think it's translating to the public app quite right. The listing that I closed still appears open on the public app. I waited a long time for the cache and refreshed a lot so I don't think that's the issue, and it worked fine for the unpublish. I wonder if we need to change the application due date when we close it, or update the public side to use status vs a due date to filter? cc @seanmalbert @pbn4

pbn4:
@kathyccheng Changing dueDate on close is an information loss, you won't be able to check what was the dueDate after it's closed, so I think we should fall back to using status only.

kathyccheng:
Last piece resolves in #1576

PR #1498, opened 2021-07-09
Listings close & unpublish

dominikx96:
# Pull Request Template

## Issue
Closes: #1443, #1442

Addresses # (issue)

## Description
Added `close` and `unpublish` buttons to the `<Aside />` component and modal to close a listing.

Current action buttons which are depending on the listing status:
- closed:
    - View listing
    - Save & exit
    - Publish
- pending:
    - View listing
    - Save & exit
    - Publish
- open:
    - View listing
    - Save & exit
    - Unpublished
    - Close

## Type of change

## Checklist:


backend/core/src/listings/listings.controller.ts
```diff
@@ -133,4 +133,18 @@ export class ListingsController {
...
+    await this.listingsService.close(listingId)
+    await clearCacheKeys(this.cacheManager, [...this.cacheKeys, `/listings/${listingId}`])
+  }
+
+  @Put(`:listingId/unpublish`)
+  @ApiOperation({ summary: "Unpublish listing by id", operationId: "unpublish" })
+  async unpublish(@Param("listingId") listingId: string) {
+    await this.listingsService.unpublish(listingId)
+    await clearCacheKeys(this.cacheManager, [...this.cacheKeys, `/listings/${listingId}`])
+  }
```
seanmalbert, line 149: @dominikx96 and @pbn4 , do we really want to maintain separate endpoints for closing, unpublishing, etc? If it does make sense to have endpoints to change the status, can we have one with a status param, like `@Put(':listingId/:status')`?

seanmalbert (changes requested):
@dominikx96 ,

I think this looks great and works well. There's one minor styling issue compared to the comps where the close button is red. Aside from that there are two issues with this that I see:

1. I'm not sure we should have the "View Listing" button here, or at least not have it work as it does now. If a user has made a bunch of edits and then clicks "View Listing" they will lose all their work.
2. Similarly with "Close", the changes aren't saved, and though probably not as big of deal, a user may choose to make some updates before setting it to closed.

dominikx96:
@seanmalbert 

I fixed both of the points from your list, agree that the "view listing" button doesn't have sense in this case. Also, fixed the problem with form data after the "close" action.

Not sure if I understand correctly, but I decided to keep the same color for the close buttons (in the aside and modal), because 'unpublish' is red so it makes a little confusion.

seanmalbert (approved):
Thanks for the updates, @dominikx96 . They look good.

Issue #1480, opened 2021-07-07
Double-submit issue with Listings forms


jaredcwhite:
In `sites/partners/src/listings/PaperListingForm/index.tsx`, the `onSubmit` handler for the `Form` component was running through a flow in an `onSubmit` function to save data and redirect to the listings detail page. However, in the the `sites/partners/src/listings/Aside.tsx` component, clicking either Publish or Save as Draft triggers `setStatusAndSubmit` which adds a bit of additional logic and then runs the `onSubmit` function. Essentially the two end up running simultaneously and thus perform a double-submit.

I believe the event handlers for the buttons in Aside run before the form's submit handler, so likely the best outcome here is to do any necessary state setup from Aside and then let the form's submit handler run automatically. Otherwise, we can provide an empty function to the form's submit handler directly and only rely on the buttons performing full submits.

I'm also noticing a similar pattern for the PaperApplicationForm so we should double-check that as well to ensure we don't have double-submits.

Thoughts?

seanmalbert:
@jaredcwhite ,
I opened a PR which addresses this issue @ https://github.com/bloom-housing/bloom/pull/1501, which removes the need for the setStatusAndSubmit function.

PR #1508, opened 2021-07-13
1480/ Fix double submit

seanmalbert:
# Pull Request Template

## Issue
#1480 

Addresses # (issue)

## Description

This is a resubmission of the PR I had out yesterday, but with updates to address potential race conditions. I didn't want to just void out the onSubmit function on the form, though harmless, it would create a detached form warning. triggerSubmit now calls a setState hook instead of onSubmit directly and there is now a useEffect hook that looks for submitData and status, so when they're both valid, the form submits.

## Type of change

Edit or create a new listing, submit should not be firing twice and there should be no warning about the form being detached.

- [x] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


jaredcwhite (changes requested):
Seems like a good approach. It looks like there's a call site for Aside though that is missing the `setStatus` prop so there's a Next.js build failure. https://github.com/seanmalbert/bloom/blob/9583338af82f9249c98be2b4c08aa2821e2ff03e/sites/partners/pages/listings/%5Bid%5D/index.tsx#L128

seanmalbert:
Thanks, @jaredcwhite , setStatus is now optional and build passes.

PR #1571, opened 2021-07-26
Remove San Francisco email from ui-components general.json

elenm:
This file is used as a template for UI strings across all site implementations. 

By removing the San Francisco email and replacing it with a generic string, we ensure that this email is not accidentally surfaced for sites other than San Francisco.

Note: I used Google translate to generate the translations. Is there a process I need to follow to get human generated translations?


emilyjablonski:
I'll flag these strings for future translation, but google translate is fine for now. Thank you!

emilyjablonski (approved):

elenm:
Thank you!

Issue #1545, opened 2021-07-21
Listings Mgmt: Lat/Long Map 


kathyccheng:
As a leasing agent, 
I want to see the map location and be able to edit the pin
So that I can be sure the map is pointing to the right location especially because for new construction, it might be on the maps properly yet 

**Acceptance Criteria** 
* When I edit an existing listing that has an address already entered, then I will see the Map preview in the Building Details section with Automatic option defaulted 
* When I try to interact with the map with the Automatic option selected, then I can't do anything between it fixes the address to the pin 
* When I select Custom, then I can move the pin around. When I move the pin around, I expect the address to remain fixed based on what I had entered 
* When I move the pin in the edit view with Custom selected and then click Save or Publish, then I expect to see the updated pin on the preview and public listing if it's already published
* When I select Automatic after I have selected Custom, then I expect the pin to go back to the default based on the address selection 

Notes
* We should probably use Mapbox's own library 
* There's no preview on the summary view of the listing 
* We will need to make a back end update to add a boolean to know whether it's custom or automatic 

**Designs** 
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4017%3A167126
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4112%3A170185

akegan:
This could be useful for SF too! We just recently had a listing called 290 Malosi, but that street doesn't exist yet, so they had to put in another address to get the pin to show up correctly. 

Seems tough to do though!

emilyjablonski:
@kathyccheng I'm not sure what `There's no preview on the summary view of the listing` means?

kathyccheng:
@emilyjablonski when you save a listing as a draft, there's a page that renders all of the information with the preview action. I just mean you don't need to show the map on that page

kathyccheng:
This looks good! Closing

PR #1577, opened 2021-07-27
1545/lat-long draggable map

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1545

## Description

Adds the map to listings management. When a user enters a full address, the map will render with automatic pin positioning. If the user wants to use a custom pin position they can select the custom radio button and the pin becomes draggable. When the listing is saved, the lat/long of the pin is saved as with the address.

## Type of change

Add a new listing in partners and play around with toggling automatic/custom positioning and previewing the listing.

## Checklist:


ui-components/src/page_components/listing/listing_sidebar/SidebarAddress.tsx
```diff
@@ -1,13 +1,12 @@
 import * as React from "react"
 import ReactDOMServer from "react-dom/server"
 import { Icon, IconFillColors } from "../../../icons/Icon"
-import { Address } from "@bloom-housing/backend-core/types"
```
emilyjablonski, line 4: These changes were not necessary, just removing references to our backend from ui-components when I'm touching these components

sites/partners/src/listings/PaperListingForm/sections/BuildingDetails.tsx
```diff
@@ -8,13 +8,116 @@ import {
...
+      buildingAddress.zipCode.length >= 5
+    )
+  }
+
+  let geocodingClient
+  if (process.env.mapBoxToken || process.env.MAPBOX_TOKEN) {
+    geocodingClient = GeocodeService({
+      accessToken: process.env.mapBoxToken || process.env.MAPBOX_TOKEN,
+    })
+  }
```
seanmalbert, line 82: Can you wrap this with a useEffect, so we only create the client once. I think you'll either have to move `let geocodingClient` above the component or use useState to set it.

sites/partners/src/listings/PaperListingForm/sections/BuildingDetails.tsx
```diff
@@ -8,13 +8,116 @@ import {
...
+
+  const getNewLatLong = () => {
```
seanmalbert, line 84: I think we may want to [debounce](https://lodash.com/docs/#debounce) this function, so we're not calling it on every input change.

seanmalbert (changes requested):
Hey @emilyjablonski ,
This looks great and thank you for the UI Component updates too! The only requests I have are around creating the map client and how often we call forwardGeocode. Please see my comments on the code.

sites/partners/src/listings/PaperListingForm/sections/BuildingDetails.tsx
emilyjablonski: debouncing 👍  had lots of bugs w the lodash flavor so just went with a set/clear timeout

seanmalbert (approved):
Thanks for the updates, @emilyjablonski . I bumped the timeout up to 1 second, since I think we can get away with that.

PR #1578, opened 2021-07-27
Add filters, pagination, and tests to listings endpoint

avaleske:
# Pull Request Template

## Issue

Addresses CityOfDetroit/bloom#192

## Description
Adds filters, pagination, and tests to listings endpoint.

Sorry this is long, the backend changes resulted in lots of small frontend changes too.

Rather than expose arbitrary fields to the frontend, which requires the frontend to know about the database schema, we opted to keep the map from filter name to field name in the backend. This keeps the frontend simpler and the abstraction from leaking.

We also move the `WHERE` clauses added to the listings querybuilder to a subquery, so that we can get all info about a listing that matches the filter. (If we were to filter for units with 3 bedrooms on the fully joined outer table, the resulting entity would only have 3 bedroom units in the units table. Instead, we want to show all the units in buildings that have 3 bedroom units, which is what this PR enables.) This also improves performance by only doing all the joins against the 10 or so listings on a given page.

Also, we had to build out the pagination logic by hand because the nestjs-typeorm-paginate package used elsewhere in the backend had [bugs](https://github.com/nestjsx/nestjs-typeorm-paginate/issues/6) when used with joined tables, and using TypeORM’s `take()` and `skip()` directly did not work with the `orderBy()` clause or our subquery logic. More details are in this [issue](https://github.com/CityOfDetroit/bloom/issues/56).

Additionally:
- Moves the ListingFilterParams into a ListingsQueryParams object, which includes pagination and jsonpath info.
- Changes the `filter` param to `addFilter()` to `ListingFilterParams` instead of `ListingFilterParams[]`, to match the actual type read in by express/qs.
- Tightens the types of `$comparisons` and `value` in `addFilters()`.
- Makes the `WHERE` clauses added to the query case-insensitive.
- Applies pagination and filtering to a subquery, with appropriate pagination metadata.
- Adds `ListingFilterKeys` and `filterTypeToFieldMap`, to list allowed filters and map filters to database fields.
- Replaces direct references to the `Listings` entity in the listings service with references to `this.listingRepository`, so the db can be mocked in the tests.
- Renames `addFilter()` to `addFilters()` to clarify that it adds all filters to the query.
- Adds tests for the listings service.
- Adds a `PaginationAllowsAllQueryParams` type that accepts `all` as an option for the limit.

## Type of change

Test with paths like:
- All six matching listings not in Foster City (Check that limit is carried through to metadata):  
`/listings?filter[$comparison]=%3C%3E&filter[neighborhood]=Foster%20City&page=1&limit=90`
- Page 2 of listings not in Foster City, 2 per page:  
`/listings?filter[$comparison]=%3C%3E&filter[neighborhood]=Foster%20City&page=2&limit=2`
- Show the 1 listing in Foster City:  
`/listings?filter[$comparison]==&filter[neighborhood]=Foster%20City&page=1&limit=90`
- All listings:  
`/listings?limit=all`

## Checklist:


backend/core/src/listings/listings.service.ts
```diff
@@ -38,29 +48,61 @@ export class ListingsService {
...
 
-    qb.orderBy({
-      "listings.id": "DESC",
-      "units.max_occupancy": "ASC",
-      "preferences.ordinal": "ASC",
-    })
+    const paginate = params.limit !== "all" && params.limit > 0 && params.page > 0
+    if (paginate) {
+      // Calculate the number of listings to skip (because they belong to lower page numbers).
+      const offset = (params.page - 1) * (params.limit as number)
```
avaleske, line 72: If the validation is changed to allow string values other than "all", this type assertion would cause an error. I'm open to suggestions for how to type guard this, since TypeScript isn't clever enough to realize that the type is narrowed by the `paginate` variable.

backend/core/src/shared/dto/pagination.dto.ts
```diff
@@ -64,3 +64,41 @@ export class PaginationQueryParams {
...
+    toClassOnly: true,
+  })
+  page?: number
+
+  @Expose()
+  @ApiPropertyOptional({
+    type: "number | 'all'",
+    example: 10,
+    required: false,
+    default: 10,
```
avaleske, line 88: Alternatively, we could default to all if no limit is specified.

backend/core/test/listings/listings.e2e-spec.ts
```diff
@@ -32,33 +32,60 @@ describe("Listings", () => {
...
+      .get("/listings" + params)
+      .expect(200)
+    expect(res.body.items.length).toEqual(allSeeds.length - 1)
+  })
+
+  it("should return the last page of paginated listings", async () => {
+    // Make the limit 1 less than the full number of listings, so that the second page contains
+    // only one listing.
+    const page = "2"
+    const limit = (allSeeds.length - 1).toString()
```
avaleske, line 54: In our fork, `allSeeds` still existed when we wrote this test. What should we use now to get the list of seeded listings?

backend/core/types/src/backend-swagger.ts
```diff
@@ -3319,13 +3328,16 @@ export interface JurisdictionUpdate {
...
-  name?: string
+  name?: string;
 
   /**  */
-  status?: EnumListingFilterParamsStatus
+  status?: ListingStatus;
+
+  /**  */
+  neighborhood?: string;
 }
```
avaleske, line 3341: Something about my annotations on the listings controller GET endpoint are incorrect. I had to manually include these types because Swagger didn't.

backend/core/types/src/backend-swagger.ts
```diff
@@ -4998,13 +5018,13 @@ export enum UserRole {
...
-export enum EnumListingFilterParamsStatus {
-  "active" = "active",
-  "pending" = "pending",
-  "closed" = "closed",
+export enum ListingFilterKeys {
+  'status' = 'status',
+  'name' = 'name',
+  'neighborhood' = 'neighborhood',
 }
```
avaleske, line 5028: Ditto for these. Something about my annotations on the listings controller GET endpoint are incorrect. I had to manually include these types because Swagger didn't.

avaleske:
This is ready for an initial review, but there's some issues I'll need help resolving. I noted them with inline comments, but to restate them here:
- I'm having trouble getting Swagger to include the ListingFilterParams types. I had to include them manually, which I'm guessing is not preferred. Is something wrong with my annotations on the listings GET endpoint?
- The listings e2e tests are broken because I don't know how to get a list of all the seeded listings, the way we used to be able to by referencing `allSeeds`. It doesn't seem to be there anymore. What should I use instead?
- Do we want to default the listings pagination to `limit=10` or `limit=all`? defaulting to 10 is safer for sites with lots of listings, but required updating more the querystrings of listings requests.

(@anders-schneider, @abbiefarr, @willrlin for visibility as well)

backend/core/src/listings/listings.service.ts
```diff
@@ -72,22 +114,29 @@ export class ListingsService {
...
     }
 
-    if (jsonpath) {
-      listings = jp.query(listings, jsonpath)
+    // TODO(https://github.com/CityOfDetroit/bloom/issues/135): Decide whether to remove jsonpath
+    if (params.jsonpath) {
+      listings = jp.query(listings, params.jsonpath)
+    }
+
+    const paginatedListings = {
```
pbn4, line 122: Mapping to a DTO is usually done at the controller level and services operate on DB models.

backend/core/src/listings/listings.service.ts
```diff
@@ -72,22 +114,29 @@ export class ListingsService {
       })
     }
 
-    if (jsonpath) {
-      listings = jp.query(listings, jsonpath)
+    // TODO(https://github.com/CityOfDetroit/bloom/issues/135): Decide whether to remove jsonpath
+    if (params.jsonpath) {
```
pbn4, line 118: @seanmalbert I think we would like to get rid of this too, but that requires adding a filter for listing.property.buildingAddress.county value.

backend/core/test/listings/listings.e2e-spec.ts
pbn4: @avaleske There is still a `allSeeds` variable but it's hidden inside `backend/core/src/seed.ts:seedListings`, feel free to expose it through a function

pbn4:
> I'm having trouble getting Swagger to include the ListingFilterParams types. I had to include them manually, which I'm guessing is not preferred. Is something wrong with my annotations on the listings GET endpoint?

Please add  `@ApiExtraModels(ListingFilterParams)` to ListingsController, and let me know if that works (it worked for me locally). It looks like a bug in nestjs/swagger

backend/core/src/listings/listings.service.ts
avaleske: I noted this on our open issue to remove jsonpath:
https://github.com/CityOfDetroit/bloom/issues/135

backend/core/test/listings/listings.e2e-spec.ts
avaleske: Done, thanks!

backend/core/src/listings/listings.service.ts
avaleske: Done.

backend/core/types/src/backend-swagger.ts
avaleske: Updated based on @pbn4's comments below

backend/core/types/src/backend-swagger.ts
avaleske: Updated based on @pbn4's comments below

backend/core/src/listings/dto/listing.dto.ts
```diff
@@ -708,6 +711,24 @@ export class ListingUpdateDto extends OmitType(ListingDto, [
...
+// @ApiExtraModels and picked by the backend-swagger.ts codegen.
+export class ListingsFilterKeysTransporter {
+  @Expose()
+  @ApiProperty({
+    enum: Object.keys(ListingFilterKeys),
+    required: false,
+  })
+  listingFilterKey?: ListingFilterKeys
+}
+
```
avaleske, line 731: @pbn4 in order to get the ListingsFilterKeys to be picked up by the swagger codegen, I needed to wrap them in a model. which is a bit of a hack.

A better alternative would probably be to put this enum in a util file that both the backend and frontend import. Is there a place that would be?

avaleske:
@pbn4 per [your comment](https://github.com/bloom-housing/bloom/pull/1578#issuecomment-889090129) I added the ListingFilterParams to ApiExtraModels, but I ran into another issue I noted [here](https://github.com/bloom-housing/bloom/pull/1578/files#r679509555), which is that ApiExtraModels only takes classes, and I have an enum (of filter names) that I want to access from the frontend code.

backend/core/src/listings/dto/listing.dto.ts
pbn4: You can always put it in somewhere in `types/src` and export it through `types/src/index.ts` or do the hacky way the same way as in `backend/core/src/applications/application-preference-api-extra-models.ts`. BTW between latest release and now I think I used IDE to refactor some of the models and I didn't notice that `backend/core/index.ts` exports nothing. Could you add `export * from ./types` to your PR, please? Without it the next package release will contain no types. cc: @seanmalbert

backend/core/src/listings/listings.service.ts
pbn4: This will work with typescript 4.4, but for now I don't have any ideas for you.

avaleske: Ah good to know! I added a todo explaining when we can remove the extra assertions.

backend/core/src/listings/dto/listing.dto.ts
avaleske: Where do I need to add `export * from ./types`? in `types/src/index.ts`?

seanmalbert: I believe he means to `backend/core/index.ts`

backend/core/src/shared/dto/pagination.dto.ts
seanmalbert: I prefer this over defaulting to all.

backend/core/src/listings/dto/listing.dto.ts
avaleske: Done.

avaleske:
I moved the ListingFilterKeys enum to a file in backend/core/src/listings/types and fixed an issue with the pagination validator (I didn't realize the validation happens after transformation, not before).

avaleske:
@pbn4 It looks like there's new relic errors in a lot of the tests. Is that something on my end or is the cicleci config missing environment variables? (There's also some legitimate failures I'll fix)

avaleske:
@seanmalbert and @pbn4 all the tests pass! The linter seems to be failing because it doesn't have permission to run on a fork, but everything else passes and VS Code isn't warning of any linter errors on my local.

avaleske:
Bumping this, @seanmalbert and @pbn4. I can't tag as "Ready for Review", but it's ready, and the failing linter is a permissions issue because the branch is on a fork.

seanmalbert (approved):
Hey @avaleske ,
This looks great, can you make two minor updates real quick? 

1. Can you move the changelog entries up to the unreleased section 
2. In listing.controller, there's a cachekey entry for `"/listings?filter[$comparison]=%3C%3E&filter[status]=pending"`, can you please update that to be `"listings?limit=all&filter[$comparison]=<>&filter[status]=pending"`. We have an issue to update how we're handling caching, but this will work in the meantime.

avaleske:
Yeah, doing that now @seanmalbert

avaleske:
@seanmalbert in that cachekey, do you want the `<>` to stay as `<>` or be escaped?

seanmalbert:
> @seanmalbert in that cachekey, do you want the `<>` to stay as `<>` or be escaped?

Oh, escaped please. Thanks for checking, @avaleske!

avaleske:
Done! @seanmalbert you'll need to merge it once the tests pass since I don't have write permissions on the repo.

Issue #1588, opened 2021-07-29
Adding Unit Summary Entity - Discussion


seanmalbert:
# Overview
Detroit has a need to enter and store unit summaries directly. On a longer time scale, Bloom does as well, but there is a more immediate desire to move the processing of transformUnits to upserting a listing and storing the data in a table, so we can read from that directly instead of processing it on request.

Ideally, we can align on the same structures, so we'll need to see what Detroit's requirements are.

# Columns
This is a little difficult to pinpoint what would make sense for Bloom, since transform units returns six different structures:
- byUnitTypeAndRent
- byUnitType
- byReservedType
- byAMI
- hmi

I have more investigation to do to figure out a structure that would best work for all of these and I'll add that to a comment soon, but don't want to block Detroit.

# Implementation
@willrlin mentioned the idea of using a derived table for unit summary. That could possibly work for Bloom's case of using the units table to create the derivatives, but how would that work for Detroit directly writing to a table?

We can create a feature flag feature to determine whether the unit summary table should be directly writable or synced from units. In both cases though, we want to read from the the unit summary table(s).

# Current Considerations
Our team is currently creating a way to override Ami Chart values, by annual income percentage. These values will be taken into account when processing the unit summary data.

willrlin:
@abbiefarr @anders-schneider please use this issue to discuss our needs and align on this work.

anders-schneider:
Hey @seanmalbert, thanks for starting a thread about this. This all sounds good to @abbiefarr and me: always read from the new unit summary table, and have a feature flag to determine whether that table is writable directly or is derived from the units table.

If we can agree on the structure of the new unit summary table, I think that would unblock us so we can create the new table and start on a couple different workstreams. Abbie and I poked around [`UnitsSummarized`](https://github.com/bloom-housing/bloom/blob/3fd11f849cc84f133d544b52037661560f57c2df/backend/core/src/units/types/units-summarized.ts#L11) and the [`UnitSummary`](https://github.com/bloom-housing/bloom/blob/3fd11f849cc84f133d544b52037661560f57c2df/backend/core/src/units/types/unit-summary.ts#L9)’s that are included within it, and have ideas about a possible structure for the new unit summary table that would offer the same granularity as the existing unitsSummarized field. Proposed columns:

- Unit type (string)
- Rent (string)
- Rent as percentage of income (number)
- Reserved type (string)
- AMI percentage (string)
- Minimum income minimum (string)
- Minimum income maximum (string)
[...]

PR #1607, opened 2021-08-03
Create units summary table

abbiefarr:
## Issue
https://github.com/CityOfDetroit/bloom/issues/238

Addresses # (issue)

## Description

Creates a UnitsSummary entity and dto, and the corresponding migration.

**Caveat**: I wasn't really sure how to structure the dto, would love some feedback there.

The table structure is described [here](https://github.com/bloom-housing/bloom/issues/1588#issuecomment-892034445), plus a `property` column which holds a `property.id` foreign key.

Current primary keys for the new table are `unitType`, `property`, and `monthlyRent`, but I expect we may want to add more fields to the composite key.

## Type of change

Run the migration and view the newly created table. I manually added a few rows to verify the primary keys.

## Checklist:


abbiefarr:
I think the lint failure (403) could be fixed by adding `pull_request_target` to Bloom's lint.yml, based on [these docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target).

abbiefarr:
@seanmalbert this is ready for another set of eyes, could someone on your team take a look?

backend/core/src/units-summary/dto/units-summary.dto.ts
```diff
@@ -0,0 +1,6 @@
+import { OmitType } from "@nestjs/swagger"
+import { UnitsSummary } from "../entities/units-summary.entity"
+
+export class UnitsSummaryDto extends OmitType(UnitsSummary, [] as const) {}
+export class UnitsSummaryCreateDto extends OmitType(UnitsSummaryDto, [] as const) {}
+export class UnitsSummaryUpdateDto extends OmitType(UnitsSummaryDto, [] as const) {}
```
seanmalbert, line 6: I think these can be fairly basic. If you take a look at `unit-types/dto/unit-type.dto.ts`, you can follow what was done there.

seanmalbert (commented):
@abbiefarr ,
This is looking good. Please see my comment regarding the dto, it should be straight forward. Can you also please add an entry into the changelog for this work. 

One thing I'd like to think about, which I don't think we should cover here, is how do we foresee working with the service? In our initial case, we'll call it in the listing service on listing upsert, so we shouldn't need a separate controller for it. I imagine we'll want to do similar logic with directly saving the unit summary. So on upsert, we would check if the user has selected to use individual units to compute the unit summary or to save the unit summary directly. Does that make sense to you?

backend/core/src/units-summary/dto/units-summary.dto.ts
abbiefarr: The `UnitsSummaryDto` doesn't extend `AbstractEntity` like `UnitTypeDto` does, and none of the fields are generated. So I _think_ we may not need to omit any fields and the file is ok like this??

abbiefarr (commented):
> So on upsert, we would check if the user has selected to use individual units to compute the unit summary or to save the unit summary directly. Does that make sense to you?

Yes, that makes sense to me, I agree we shouldn't need a separate controller, since this can be a flag in the `listing` object (or we can repurposed the existing `disableUnitsAccordion` flag). 

Toggling between the two options may be disruptive, so we may also want to consider warnings when attempting that. E.g. if you've been using a summary but want to switch to individual units, and you don't have any data in your units table, that's pretty bad. So perhaps a warning that a data upload or refresh may be required? In the edit listing page (where I'm assuming this option is set), we could also display the contents of the table when you select each option as a visual warning too (our site current does this, with a mock summary table for now).

backend/core/src/units-summary/dto/units-summary.dto.ts
seanmalbert: @abbiefarr , looks like you are right. I think we're good then!

seanmalbert:
>Yes, that makes sense to me, I agree we shouldn't need a separate controller, since this can be a flag in the listing object (or we can repurposed the existing disableUnitsAccordion flag).

I think we can only repurpose it if we're sure that a user will never want to individually add units and disable the individual unit accordion. @slowbot , can you weigh in on this case?

> Toggling between the two options may be disruptive, so we may also want to consider warnings when attempting that. E.g. if you've been using a summary but want to switch to individual units, and you don't have any data in your units table, that's pretty bad. So perhaps a warning that a data upload or refresh may be required? In the edit listing page (where I'm assuming this option is set), we could also display the contents of the table when you select each option as a visual warning too (our site current does this, with a mock summary table for now).

I've discussed this with @slowbot and we agree that some type of warning message should display on the form. I think the plan was to start with helper text explaining that the user could incur data loss by switching between the options.

seanmalbert:
@abbiefarr , I spun off #1633 to continue the discussion around handling the partner frontend use of the toggle.

seanmalbert (approved):

Issue #1621, opened 2021-08-05
Create View for /listings endpoint


seanmalbert:
From discussion in #1563 , we want to create a way to specify a "view" as a parameter to the GET listings endpoint (e.g. `\listings?page=2&limit=10&view=simpleListings`). This view should return the bare minimum needed for the main listings page, which uses the ListingsList component. These fields are:

- id
- name
- urlSlug
- showWaitList
- reservedCommunityType
  - name
-  applicationDueDate
- applicationDueTime
- applicationOpenDate
- buildingAddress
  - city
  - state
  - street
  - zipCode
 - unitsSummarized.byUnitTypeAndRent
   - unitType
   - minimumIncome
   - rent  

I'm thinking that the listings service can have map of views, with each view being a function that takes a query builder like:
```
{
  simple: (qb) => {
    qb.select([
      "listings.id",
      "listings.name",
      [...]
    ])

    // leftJoins
  },
}
```

There are likely DTOs that need updating, because they probably require fields that won't be returned from this query.

Since we will only need byUnitTypeAndRent from unitsSummarized, the call for summarizedUnitsByTypeAndRent should be abstracted from transformUnits so we don't have to calculate all of it.

avaleske:
This is essentially what I was picturing, yeah!

How often do you think a view might be a superset of another view? For example, the listing page will be a superset of the view used for the listings list, right? In that case, we could have something like
```
  listingView = {
    listView: (qb) => {
      qb.select([
        "listings.id",
        "listings.name",
        [...]
      ])
  
      // leftJoins
    },
    pageView: (qb) => {
      listingViews.listView(qb)
        .select([
          "listing.extraField",
        ])
        ... // joins and such
    },
  }
```
right? I'm not sure if that would override the original `select()` or add to it though.

Another way I had been considering (and I don't know if this is better) would be:
Instead of taking a querybuilder, the map could return an object shaped like
```
queryComponents = {
  selects: ["listings.id", "listings.name", ...],
  leftJoins: ["property", ...]
}
```
and then the querybuilder just uses it like `qb.select(queryComponents.selects).leftJoin(queryComponents.leftJoins)...`

In that world, the "inheritance" would be something like
```
  listView: { selects: ["listings.id", "listings.name"] ... },
  pageView: { selects: [...listView.selects, "listings.extraField"], ...},
```
[...]

PR #1626, opened 2021-08-09
1621/Listing Views

seanmalbert:
# Pull Request Template

## Issue
#1621 

Addresses # (issue)

## Description

This adds a concept of "views" discussed within the issue.

Things to consider while reviewing:
- You can see that DetailView isn't used here. Specifying that many fields is tedious so I created FullView, which uses leftJoinAndSelect. I decided to keep DetailView in place, at least for this, so you can see how the configs extend each other.
- I also thought about how the configs extend each other. Instead of doing it by merging the fields, we could do it in `getView`, by calling super.getView(), which would add the extended views selects and joins to `qb`. Then we would add on other "view" configs. This may be a better approach
- Between the leftJoin and leftJoinAndSelect definitions, you can see they're handled differently and I think we should choose one for consistency. 
- I thought about having the views themselves create the query builder, instead of passing it in, but that made it much more difficult to test, and the services already have the repository injected, so it's easy to pass in.
- I didn't create a view factory here to generate the views, since we don't have many use cases at the moment, but that is something we could do, as @avaleske suggested in the issue

Both public and partners should work as before, but the list view should be considerably smaller.


abbiefarr:
@seanmalbert thanks for putting this together!

I'm running into a problem in #1670, where the left joins I add overwhelm the service and cause the site functionality to seriously degrade. If I remove a few other left joins it works just fine, so I think we're just pushing the limit of the query.

I think the views drafted here would unblock the units summary progress. Is there an expected timeline for the remainder of this work?

seanmalbert:
> @seanmalbert thanks for putting this together!
> 
> I'm running into a problem in #1670, where the left joins I add overwhelm the service and cause the site functionality to seriously degrade. If I remove a few other left joins it works just fine, so I think we're just pushing the limit of the query.
> 
> I think the views drafted here would unblock the units summary progress. Is there an expected timeline for the remainder of this work?

Hi @abbiefarr ,
Thanks for checking in on this, I haven't had time to get back to it. I think the remainder of this piece is to make sure the extension model works well by creating a view for getFullyJoinedQueryBuilder, which I can get to today. If your team can plan on reviewing this tomorrow to make sure it meets your needs as well, we should be able to get this in this week. Does that work for you?

abbiefarr:
> If your team can plan on reviewing this tomorrow to make sure it meets your needs as well, we should be able to get this in this week. Does that work for you?

Sounds great! @avaleske is out of office this week, so feel free to add me and @willrlin as reviewers.

seanmalbert:
@abbiefarr and @willrlin , this is ready for you to look at. I updated the PR description so please have a look there. For some reason I was only able to assign Abbie to this PR and Will didn't show up as an option 🤷

backend/core/src/listings/listings.service.ts
```diff
@@ -77,19 +54,24 @@ export class ListingsService {
       // join on the listings we want to show.
       innerFilteredQuery.offset(offset).limit(params.limit as number)
     }
-
-    let listings = await this.getFullyJoinedQueryBuilder()
+    const view = new BaseView(this.listingRepository.createQueryBuilder("listings"))
+    // let listings = await view
```
emilyjablonski, line 58: Leftover comment

backend/core/src/listings/listings.service.ts
```diff
@@ -173,7 +155,8 @@ export class ListingsService {
   }
 
   async findOne(listingId: string) {
-    const result = await this.getFullyJoinedQueryBuilder()
+    const view = new FullView(this.listingRepository.createQueryBuilder("listings")).getView()
+    const result = await view
       .where("listings.id = :id", { id: listingId })
       .orderBy({
         "preferences.ordinal": "ASC",
```
emilyjablonski, line 162: Ordering by preferences here feels weird?

backend/core/src/listings/views/config.ts
```diff
@@ -0,0 +1,180 @@
+import { Views } from "./types"
+
+function getBaseAddressSelect(schemas: string[]): string[] {
+  const fields = ["city", "state", "street", "zipCode", "latitude", "longitude"]
```
emilyjablonski, line 4: We also have `street2` just want to me sure this was a cognizant omission

backend/core/src/listings/views/index.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    it("should call getView qb select and leftJoin", () => {
+      const view = new BaseView(mockListingsRepo.createQueryBuilder())
+
+      view.getView()
+
+      expect(mockQueryBuilder.select).toHaveBeenCalledTimes(1)
+      expect(mockQueryBuilder.leftJoin).toHaveBeenCalledTimes(6)
+    })
+
+    it("should cal getView qb leftJoinAndSelect", () => {
```
emilyjablonski, line 65: `call`

abbiefarr:
Looks good from Detroit's perspective, thanks!

willrlin:
I haven't been following the View discussion too deeply, so I trust @abbiefarr's review.

seanmalbert:
@abbiefarr ,
Heads up, I realized I didn't include a way to pass the view as a query parameter, so I updated this so that ?view=<view> can get passed in for both the listing list and by id endpoints. I'm just making sure the pipeline passes and I'll merge in.

Issue #1589, opened 2021-07-29
Tie Nest Cache Manager to Redis


seanmalbert:
Right now the backend cache is stored in memory and moving it to Redis can have benefits, especially when we scale to more dynos (i.e. each dyno can use the same cache).

We do want to do some investigation on the feasibility of  tying the two together, vs rolling a more generic approach with Redis, which could give us more control on when and we update the cache.

seanmalbert:
Looks like we can easily specify the cache store https://docs.nestjs.com/techniques/caching#different-stores

seanmalbert:
This is merged and dev appears to be building fine now with drastically reduced build times. There is one remaining issue that seems to be intermittent, but only causes an issue in circleci:

I had to listen for onModuleDestroy in the backend so when the app created for the seed script closed, we could trigger the disconnect to redis (without it the build will hang). This seems to be an intermittent fix since it can still hang sometimes, so if you see jest-backend and cypress/run hang on waiting for the seed script to finish, this is the cause.

PR #1637, opened 2021-08-09
1589/Sets Redis to handle backend cache

seanmalbert:
## Issue

Addresses # #1589 

## Description

This sets NestJS cache manager to use Redis instead of in memory.

## Type of change

Spin up both partner and portal apps. You can monitor Redis with `redis-cli monitor`. When you edit a listing in partners you should see the cache entry get cleared. When you go to listings on public, you should see the entry get added.

- [x] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


emilyjablonski:
Overall this seems to be working as expected when testing locally with the monitor, I think there’s just some very slight delays that I wouldn’t expect. If I visit an individual listing I can see it be cached in through the monitor, and if I go back to listings and then back to the individual listing I can see us using the cache, but there’s still a small delay in starting to load the page and it seems like it should be nearly instantaneous? I might be expecting too much from the speed here.

backend/core/.env.template
```diff
@@ -5,6 +5,7 @@ TEST_DATABASE_URL=postgres://localhost/bloom_test
 REDIS_TLS_URL=
 REDIS_URL=redis://127.0.0.1:6379/0
 REDIS_USE_TLS=0
+REDIS_USE_TLS=0
```
emilyjablonski, line 8: Looks like a duplicate of the line above

PR #1642, opened 2021-08-10
Fix backend core tests

### LINKED ISSUE #369: Handle the case where a listing string is null
Right now the build has problems dealing with listing values that return `null`. Even worse, it seems to work ok if you do a dev build, but it fails when you do an `export` when trying to call `.replace` in next.config.js.

I'm open to solving this by adding tooling to make sure we never get into this situation, but I think we're probably better off making the evaluation less brittle so that it can still build with just the one value blank if there's not a string to fill in there.

willrlin:
# Pull Request Template

## Issue

Closes #369

## Description

Hardcodes the length of the list for the pagination calls to avoid flakiness in the backend e2e tests.

## Type of change

backend/core e2e tests were flaky prior to this; after this they were no longer flaky

## Checklist:


willrlin:
Ack, wrong comparison again.

Issue #872, opened 2020-11-25
Fix how unitsSummarized uses AMI charts 


pbn4:
When changing AMI chart entity to be a relation of Unit instead of Property I noticed that we are actually not using amiChartId per unit when computing this, but we are retrieving an AMI chart for first Unit in a collection and apply that for the rest. 

Notes
* Unit summarized logic only used AMI of first unit and used that to compute summary table 
* Use range if we have multiple AMI charts

pbn4:
Related code https://github.com/bloom-housing/bloom/blob/974f5d0b71081f8e5b75aedb95e2017d978da7a6/services/listings/src/lib/unit_transformations.ts#L43

software-project:
For now at least that works as we don't really have multiple charts per listing. If we did there is a matter how would we even show it (just wider income ranges? or something more?). Another question is (I don't remember the answer) if we really need multiple ami charts per one listing. If we will have new property model, maybe one chart per each listing for the property will be enouch?

pbn4:
> If we did there is a matter how would we even show it (just wider income ranges? or something more?). 

OK that I don't know and it probably needs to be discussed more broadly. 

> Another question is (I don't remember the answer) if we really need multiple ami charts per one listing. 

@slowbot If I recognize correctly we want to be able to configure ami chart at unit level, right?

slowbot:
@software-project @pbn4 

There are and will be use cases of multiple ami charts per building so we should allow for ami chart to be assigned at the unit level.

emilyjablonski:
For QA, ensure that existing HMI tables look as we expect them to

slowbot:
@seanmalbert @emilyjablonski Can you help me get the new multiple AMI charts seeds onto dev so I can QA?

slowbot:
@emilyjablonski Why is there a range expressed for max income per ami level? 

I think we can simplify this table and only display a single max max income for each hh size and ami level.

emilyjablonski:
@slowbot We have ranges in the case that a listing is using multiple ami charts across different units, so we don't have dedicated columns for each chart in this case. Do you still want to not use ranges here and just take the max of all possible values?

PR #1575, opened 2021-07-27
Remove AmiChart object from UnitDto

pbn4:
# Pull Request Template

## Issue

Addresses # (https://github.com/bloom-housing/bloom/issues/1557)

- [ ] This change addresses the issue in full
- [ ] This change addresses only certain aspects of the issue
- [ ] This change is a dependency for another issue
- [ ] This change has a dependency from another issue

## Description

This change removes AmiChart object from UnitDto. 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Prototype/POC (not to merge)
- [ ] This change is a refactor/address technical debt
- [ ] This change requires a documentation update
- [ ] This change requires a SQL Script

## How Can This Be Tested/Reviewed?

Please describe the tests that you ran to verify your changes. Provide instructions so we can review. Please also list any relevant details for your test configuration

- [ ] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have reviewed the changes in a desktop view
- [ ] I have reviewed the changes in a mobile view
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have assigned reviewers
- [ ] I have updated the changelog to include a description of my changes

PR #1678, opened 2021-08-17
872/allow for unit sets to have multiple ami charts

emilyjablonski:
# Pull Request Template

## Issue

Addresses #872

## Description

Allows for sets of units to have multiple AMI charts, and in doing so updates the HMI chart visually to display ranges of maximum income values if there are multiple AMI charts in use. Also fixes some existing sorting bugs, fixes the BMR chart seed which I discovered wasn't working, and adds two new seeds to test this behavior that will render each flavor of the HMI table with multiple AMI charts. Screenshots of them both below. I essentially re-wrote the function and added a few comments to try to make it more readable but I don't know that I succeeded there 😨 

Screen Shot 2021-08-16 at 7 08 17 PM
Screen Shot 2021-08-16 at 6 30 42 PM

## Type of change

Check out the new seeds, and the existing seeds against the tables on dev to ensure they didn't break.

## Checklist:

### DIFF
--- a/backend/core/src/seeds/listings/listing-default-bmr-chart-seed.ts
+++ b/backend/core/src/seeds/listings/listing-default-bmr-chart-seed.ts
@@ -1,18 +1,47 @@
 import { ListingDefaultSeed } from "./listing-default-seed"
-import { getDefaultUnits } from "./shared"
+import { getDefaultUnits, getDefaultProperty, getDefaultAmiChart } from "./shared"
+import { UnitCreateDto } from "../../units/dto/unit.dto"
+import { BaseEntity } from "typeorm"
 
 export class ListingDefaultBmrChartSeed extends ListingDefaultSeed {
   async seed() {
     const listing = await super.seed()
     const defaultUnits = getDefaultUnits()
+
+    const unitTypeOneBdrm = await this.unitTypeRepository.findOneOrFail({ name: "oneBdrm" })
+    const unitTypeTwoBdrm = await this.unitTypeRepository.findOneOrFail({ name: "twoBdrm" })
+
+    const amiChart = await this.amiChartRepository.save(getDefaultAmiChart())
+
+    const property = await this.propertyRepository.save({
+      ...getDefaultProperty(),
+    })
+
+    const bmrUnits = [
+      { ...defaultUnits[0], bmrProgramChart: true, monthlyIncomeMin: "700", monthlyRent: "350" },
+      { ...defaultUnits[1], bmrProgramChart: true, monthlyIncomeMin: "800", monthlyRent: "400" },
+    ]
+
+    const unitsToBeCreated: Array<Omit<UnitCreateDto, keyof BaseEntity>> = bmrUnits.map((unit) => {
+      return {
+        ...unit,
+        property: {
+          id: property.id,
+        },
+        amiChart,
+      }
+    })
+
+    unitsToBeCreated[0].unitType = unitTypeOneBdrm
+    unitsToBeCreated[1].unitType = unitTypeTwoBdrm
+
+    await this.unitsRepository.save(unitsToBeCreated)
+
     return await this.listingRepository.save({
       ...listing,
       name: "Test: Default, BMR Chart",
       preferences: [],
-      units: [
-        { ...defaultUnits[0], bmrProgramChart: true, monthlyIncomeMin: "700", monthlyRent: "350" },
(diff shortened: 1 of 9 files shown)


backend/core/src/seeds/listings/listing-default-bmr-chart-seed.ts
```diff
@@ -1,18 +1,47 @@
 import { ListingDefaultSeed } from "./listing-default-seed"
-import { getDefaultUnits } from "./shared"
+import { getDefaultUnits, getDefaultProperty, getDefaultAmiChart } from "./shared"
```
emilyjablonski, line 2: This seed listing just wasn't working anymore and we must have missed it - so now I'm actually adding the appropriate units so we get the BMR chart

emilyjablonski:
I'll make another issue but I also realized in the process of doing this that a few things in this table aren't translated (existing issue)

backend/core/src/shared/units-transformations.ts
```diff
@@ -32,85 +34,131 @@ const minMaxCurrency = (baseValue: MinMaxCurrency, newValue: number): MinMaxCurr
...
+  // All unique combinations of an AMI percentage and an AMI chart across all units
+  const uniquePercentageChartSet: ChartAndPercentage[] = [
+    ...new Set(
+      units
+        .map((unit) => {
+          return { percentage: parseInt(unit.amiPercentage, 10), chart: unit.amiChart }
+        })
+        .map((item) => JSON.stringify(item))
+    ),
+  ].map((uniqueSetString) => JSON.parse(uniqueSetString))
```
seanmalbert, line 66: I think it will be more efficient to only reference the amiChart id here and not the full chart. If I'm reading this hmiData function correctly, the only place we need to know the actual chart item data is in `findAmiValueInChart`, so if you updated that function to take the chart id, and used that to get the chart data, I think some of this could be simplified. This does mean that ami charts would need to be fetched separately, but that could and probably should be cached, since I doubt they change very frequently. While this probably doesn't really help speed up processing time, it should reduce the amount of memory used, especially when you take into account that we could take this approach with ami charts everywhere and only ever request the chart id along with listing units.

seanmalbert (changes requested):
@emilyjablonski , I'm impressed that you were able to get this refactor turned around so quickly, that hmiData method is not easy to follow, at least for me. There are a couple of places that I noticed where I think we can make some improvement. Please see my comments. The main one is having to do with how we're handling and referencing the ami charts. Please let me know if you have any questions.

emilyjablonski:
@slowbot The content wraps where needed already so we should be good to go there, and it will be a little challenging to add the per month and per year since we're just passing translation keys and not pre-translated text and we don't have a way to do that yet for ranges. I would love to do that in a separate PR if that's okay since this one is blocking?

backend/core/src/shared/units-transformations.ts
emilyjablonski: This makes sense to me and I can see the memory gains, but currently I don't think we have access to the amiChart id here without other backend changes. The Unit entity only has the amiChart as `AmiChart`. So I think we need the other PR's [changes](https://github.com/bloom-housing/bloom/pull/1575/files#diff-71332228c9550323008f3cb6962ea93aaa4bf15137a6eeae135ee63dde3ac1a3R56) to move forward with that improvement.

backend/core/src/shared/units-transformations.ts
```diff
@@ -32,85 +34,131 @@ const minMaxCurrency = (baseValue: MinMaxCurrency, newValue: number): MinMaxCurr
...
+
+  type ChartAndPercentage = {
+    percentage: number
+    chart: AmiChart
+  }
+
+  // All unique AMI percentages across all units
+  const allPercentages: number[] = [
+    ...new Set(
+      units.map((unit) => parseInt(unit.amiPercentage, 10)).filter((item) => item != null)
```
seanmalbert, line 51: @emilyjablonski , this looks good, can you just switch the order of the filter and map.

backend/core/src/shared/units-transformations.ts
seanmalbert: I think that's fine, since that PR does fetch the ami charts, we can swap that out there.

seanmalbert (approved):
@emilyjablonski ,
Looks good, just one request before you merge to swap a map and filter placement. See comment in code.

PR #1689, opened 2021-08-18
Add a migration for small changes to `units`, `units_summary`, and `user_roles` tables

anders-schneider:
It looks like there's a small backlog of migrations to be generated. Running `yarn db:migration:generate` resulted in this new migration file. There are a couple separate changes: `units.status` cannot be null and defaults to "unknown", `units_summary.monthly_rent` is an integer, `units_summary` is updated to include a foreign-key reference to the `listings` table, and the foreign-key relationship between `user_roles` and `user_accounts` is updated.


seanmalbert (approved):
Thank you, @anders-schneider!

Issue #1582, opened 2021-07-28
State machine set up for status change


kathyccheng:
@seanmalbert to fill out details

seanmalbert:
We're moving the conversation started in #1541 here to do this work as a blocker to #1541, #1542 and #1580.

dominikx96:
Hey,

I’ve started researching possibilities for how we can implement validators for different states.
For now - I think for state management, [xstate](https://xstate.js.org/docs/) (+ @xstate/react which provides helpful hooks) is the best choice.

To summarize we can assume:
- Validators:
    - draft (probably only one field is required)
    - active (most of the fields are required)
- States:
    - draft | active
- Transitions:
    - draft -> draft (validator: draft)
    - draft -> active (validator: active)
    - active -> draft (validator: draft)
    - active -> active (validator: active)

Each transition should define a proper validation and store information about it.

The main consideration that I have is about what kind of validation we want on the frontend side.
Of course, backend validation still should be implemented.

**1. Frontend and backend validation:**
    - We can create a group of rules which can be shared between frontend and backend, then fetch it from the API during the build and apply proper rules for each field depending on the state and transition after event e.g after clicking on the “publish” button

    - Pros:
        - A better UX, a user don’t need to wait for the API response
        - We still meet with validation patterns like in the other forms in the app (data is validated on both sides)

**2. Backend only validation:**
    - We can validate form data backend side only and return an object with errors from the API, then:
        - map each one backend id with the frontend id
        - map each one validation rule (e.g. required) to a proper frontend value to determine which translation show

I feel like the second option has a similar complexity as the first, because:
- We still have to determine which backend field ID is on the frontend
- We still have to show proper translations

so far - the first option seems to better work for me.

I also think, if we consider implementing a “form configurator” in the future - it’s a good time to think about how we want to build this solution to easily use it in the future without many breaking changes.

**How we can connect backend field names with the frontend?**
Do we plan to create a static object in the frontend app or create a new entity to store frontend field keys?

Do you have any thoughts before we go deeper? :)

jaredcwhite:
I agree with @dominikx96 — either option (1) or (2), we still have a similar amount of work to build a validation rules engine that is compatible with the shape of the JSON.

Thinking of what we can leverage that's already out there…does anyone have experience with Yup? https://github.com/jquense/yup

React Hook Form has an integration point: https://react-hook-form.com/get-started#SchemaValidation

seanmalbert:
[...]

PR #1705, opened 2021-08-23
Initial Listing state machine setup

software-project:
# Pull Request Template

## Issue

Addresses #1582

## Description

Adds state machine with guards and validation.

## Type of change

- [ ] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/backend/core/src/listings/types/listing-status-enum.ts
+++ b/backend/core/src/listings/types/listing-status-enum.ts
@@ -1,5 +1,6 @@
 export enum ListingStatus {
   active = "active",
-  pending = "pending",
   closed = "closed",
+  new = "new",
+  pending = "pending",
 }
--- a/backend/core/package.json
+++ b/backend/core/package.json
@@ -45,6 +45,7 @@
     "@nestjs/throttler": "^1.1.2",
     "@nestjs/typeorm": "^7.1.0",
     "@types/cache-manager": "^3.4.0",
+    "@xstate/fsm": "^1.6.1",
     "async-retry": "^1.3.1",
     "axios": "^0.21.0",
     "cache-manager": "^3.4.0",
@@ -77,7 +78,8 @@
     "ts-node": "^9.0.0",
     "typeorm": "0.2.34",
     "typeorm-naming-strategies": "^1.1.0",
-    "typescript": "^3.9.7"
+    "typescript": "^3.9.7",
+    "yup": "^0.32.9"
   },
   "devDependencies": {
     "@babel/core": "^7.11.6",
--- a/backend/core/src/listings/listing.schema.ts
+++ b/backend/core/src/listings/listing.schema.ts
@@ -1,0 +1,28 @@
+import * as yup from "yup"
+
+const previewSchema = yup.object().shape({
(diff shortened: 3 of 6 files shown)


software-project:
@seanmalbert @pbn4 @emilyjablonski @dominikx96 @jaredcwhite  This is an initial state machine for listings. 
What's working:
- state transitions with guards
- it passes an error when it's one (we need to figure out the best way to handle them and pass them to frontend) @pbn4 do you know if there is a standard way to handle it?

I still need to add some tests and schemas for the validation are not complete. Also there is an caching issue, so even when state is changed, view remains unchanged. @seanmalbert do you know what it might be? Let me know what do you think and I'll make updates this week.

backend/core/src/listings/types/listing-status-enum.ts
```diff
@@ -1,5 +1,6 @@
 export enum ListingStatus {
   active = "active",
-  pending = "pending",
   closed = "closed",
+  new = "new",
```
emilyjablonski, line 4: What does `new` represent as a status?

seanmalbert:
@dominikx96 ,
Can you please review and make sure this makes sense to you. You'll be taking the end result of this to flesh out the rest of the state validation.

backend/core/src/listings/types/listing-status-enum.ts
software-project: It's an initial state for the state machine. Pending state requires already some validations to pass, so we need an initial point.

pbn4 (changes requested):
I like the idea of using xstate for state management and condition guards, it will handling existing and future transition in listing state scalable and easy to maintain. 

Regarding using `yup` I think it represent a different convention than what we already have with `class-validator` and we should stick with the latter to have consistent error messages. `ValidationPipe` can be injected through constructor to `Listings.service.ts` and it exposes `transform` method specifically for inline validation. I think the solution should a mix of `xstate` with transition guards and `ValidationPipe.transform` inline depending (triggered in transition guards) on what Listing state we would like to transition to. It implies that we need separate classes like ListingPublished defined with all the `class-validator` decorators in place with appropriate logic for a given state. For more complex validations (e.g. including DB access) it's possible to do DI in custom validation decorator classes like presented [here](https://github.com/nestjs/nest/issues/528). 

@seanmalbert I can take over this PR.

jaredcwhite:
@pbn4 I thought the point of using Yup is we can validate the JSON payload entirely on the client. Doing it again on the backend is just extra hygiene at that point. Using validation concerns in Nest wouldn't allow us to do client-side validation easily if I'm not mistaken.

software-project:
@dominikx96 @jaredcwhite yes, I agree with you, that returning an error object, so we can assign errors to field would be my prefered way, but I could find a way to do so with yup. One thing to check also, cause I don't remember if I fixed it, was an issue where yup was returning only the first validation error. As there I quite a few things to resolve here I'm happy if @pbn4 could take it over.

seanmalbert:
This work should now be covered with #1796

Issue #1691, opened 2021-08-19
Update HMI Table to include `per year` and `to` in between ranges


emilyjablonski:
Currently ranges are just displayed as `$1000 - $2000` but should display as `$1000 to $2000 per year`

emilyjablonski:
Ensure that HMI table numbers now look like ($1000 to $2000 per year) and not ($1000 - $2000). The BMR rows should also now be translated. We have the property variety in the seed data but we may need to re-seed dev.

slowbot:
@seanmalbert @emilyjablonski Same request here, can we reseed dev so I can finish the QA here? Thanks

PR #1707, opened 2021-08-23
1691/hmi translations

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1691

## Description

When the HMI table was rewritten to include ranges I wrote those ranges as ($1,000 - $2,000) but it should have been ($1000 to $2000 per year). This is that bug fix and also another bug fix where the BMR headers were untranslated. 

## Type of change

Just start up the app and ensure all the seed data HMI tables are correct. The special cases are the BMR listing and both of the multiple AMI listings.

## Checklist:


backend/core/src/shared/units-transformations.ts
```diff
@@ -100,16 +106,20 @@ const hmiData = (units: Units, maxHouseholdSize: number) => {
 
   const getYearlyRangeValue = (incomeRange: MinMaxCurrency) => {
     return incomeRange.min === incomeRange.max
-      ? incomeRange.min
-      : `${incomeRange.min} - ${incomeRange.max}`
+      ? `listings.annualIncome*income:${incomeRange.min}`
+      : `listings.annualIncomeRange*from:${incomeRange.min}*to:${incomeRange.max}`
+  }
+
+  const yearlyCurrencyStringToMonthly = (minimum: string) => {
```
dominikx96, line 113: This argument is used for min and max below, not sure, but maybe we should rename it? :)

dominikx96 (approved):
This looks good to me!

backend/core/src/shared/units-transformations.ts
emilyjablonski: good call!

Issue #1614, opened 2021-08-04
Explore SJ implementation + set up 


kathyccheng:
* As part of this issue, we want to update the use of county_county code to use jurisdiction_id, which means mapping the current CountyCode enum to rows and updating the references.
* This also includes getting to staging with current SJ content - https://housing.sanjoseca.gov/listings
* Let's add some seed data, at least one open and closed

kathyccheng:
Closing, on staging, already demo'd, opened up new issues for remaining jurisdiction updates

PR #1776, opened 2021-08-30
Adds Jurisdiction to listings, users, translations (replaces countyCode)

seanmalbert:
# Pull Request Template

## Issue

Addresses # #1614

## Description

This PR covers adding jurisdiction relantionships to listings, users and translations, which replaces the use of countyCode. You'll note that countyCode is kept as a virtual field, so I wouldn't have to update every instance in the front-end apps. There is now a `JURISDICTION_NAME` environment variable for the public app. The value can be the same as COUNTY_CODE, which can be removed.

## Type of change

You should be able to create a new user from the public app, and they should be assigned to the jurisdiction set in the env var. On the partner's side, you should be able to create a new listing. If the user has access to more than one jurisdiction, they should see a dropdown to select the jurisdiction at the top of the form, otherwise it should set to the user's jurisdiction and not show the dropdown.

This does not address adding different roles for users under different jurisdictions. We can add that when we need it.

- [x] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


sites/public/.env.template
```diff
@@ -6,7 +6,7 @@ NEXTJS_PORT=3000
 MAPBOX_TOKEN=
 LANGUAGES=en,es,zh,vi
 IDLE_TIMEOUT=5
-COUNTY_CODE=Alameda
+JURISDICTION_NAME=Alameda
```
emilyjablonski, line 9: I wonder if our template should be something like 'Development' rather than any reference to Alameda?

sites/partners/src/listings/PaperListingForm/sections/ListingIntro.tsx
```diff
@@ -1,39 +1,90 @@
...
+  AuthContext,
+  Select,
+  ViewItem,
+} from "@bloom-housing/ui-components"
 
 const ListingIntro = () => {
   const formMethods = useFormContext()
 
+  const { profile } = useContext(AuthContext)
+  console.log("profile = ", profile)
```
emilyjablonski, line 17: 🧹

sites/partners/src/listings/PaperListingForm/sections/ListingIntro.tsx
```diff
@@ -1,39 +1,90 @@
...
+
   // eslint-disable-next-line @typescript-eslint/unbound-method
   const { errors, register } = formMethods
 
+  const JuriWrap = useCallback(() => {
```
jaredcwhite, line 22: Not sure about the `JuriWrap` naming… 🤔

seanmalbert: Renamed to `JurisdictionWrapper`

backend/core/src/auth/services/user.service.ts
```diff
@@ -172,6 +173,10 @@ export class UserService {
     user.dob = dto.dob
     user.email = dto.email
     user.language = dto.language
+    // if coming from partners dto.jurisdictions can be set
+    user.jurisdictions = dto.jurisdictions
```
pbn4, line 177: @seanmalbert Could you explain a bit more the reasoning behind this?

backend/core/src/migration/1630250097191-adds-jurisdiction-relations.ts
```diff
@@ -0,0 +1,71 @@
...
+          UPDATE listings
+          SET jurisdiction_id = j.id
+          FROM listings AS l
+            INNER JOIN jurisdictions AS j
+              ON l.county_code = j.name
+        `)
+        // drops county code from listings
+        await queryRunner.query(`ALTER TABLE "listings" DROP COLUMN "county_code"`);
+        await queryRunner.query(`ALTER TABLE "jurisdictions" ADD CONSTRAINT "UQ_60b3294568b273d896687dea59f" UNIQUE ("name")`);
+        await queryRunner.query(`ALTER TABLE "translations" DROP COLUMN "jurisdiction_id"`);
```
pbn4, line 25: `translations.county_code` has been first renamed to `jurisdiction_id` and here it's dropped. Isn't it a data loss?

pbn4: I think the first rename should be county_code to `name` instead of `jurisdiction_id` considering what this UPDATE is doing:

```
        await queryRunner.query(`
          UPDATE listings
          SET jurisdiction_id = j.id
          FROM listings AS l
            INNER JOIN jurisdictions AS j
              ON l.county_code = j.name
        `)
```

backend/core/src/shared/email/email.service.spec.ts
```diff
@@ -47,25 +48,31 @@ describe("EmailService", () => {
...
-              return { "county-code": CountyCode.alameda }
+              return "Alameda"
             },
           },
         },
       ],
     }).compile()
 
+    const jurisdicitonService = await module.resolve<JurisdictionsService>(JurisdictionsService)
+    const jurisdiction = await jurisdicitonService.findOne({ where: { name: "Alameda" } })
```
pbn4, line 65: typo

pbn4 (changes requested):
backend-swagger.ts has a duplicate entry for `Jurisdiction`. It means that DB entity model is exposed directly from some DTO instead of using JurisdictionDto.

backend/core/src/auth/services/user.service.ts
seanmalbert: @pbn4 , As part of the user management flow, a user can belong to more than one jurisdiction and, given they have permissions, when they create a user they will have to select which jurisdiction they belong to. Therefore it would be set on the incoming dto, not from the request header (which partner's doesn't currently use).

backend/core/src/migration/1630250097191-adds-jurisdiction-relations.ts
seanmalbert: @pbn4 ,

> `translations.county_code` has been first renamed to `jurisdiction_id` and here it's dropped. Isn't it a data loss?

No, because before this it fetches the translations so it can set them based off of the county_code before it was dropped.



> I think the first rename should be county_code to `name` instead of `jurisdiction_id` considering what this UPDATE is doing:
> 
> ```
>         await queryRunner.query(`
>           UPDATE listings
>           SET jurisdiction_id = j.id
>           FROM listings AS l
>             INNER JOIN jurisdictions AS j
>               ON l.county_code = j.name
>         `)
> ```

This is for listings though, not translations, which has the rename (listings doesn't), and this occurs before county_code is dropped from listings.

Issue #1687, opened 2021-08-18
Partners: Pagination on Listings Page 


kathyccheng:
As a leasing agent (most likely an admin who can see everything)
I want the list of listings in Partners to be paginated
So that I can see everything and the page will load performantly 

Acceptance Criteria
* When I am on the listings page, I will see pagination 

Notes
* Use our standard pagination

slowbot:
@dominikx96 I did find a bug. If I try and load more than 8 pages in the "show" dropdown on a page other than page 1 I get an error

![image](https://user-images.githubusercontent.com/13915/132532816-7744fa93-4772-4b2e-86a6-2f564f95225d.png)
![image](https://user-images.githubusercontent.com/13915/132532867-c540c8da-1aa4-4403-94c1-4cc3f27622be.png)

PR #1782, opened 2021-09-01
1687/listing pagination issues

pbn4:
# Pull Request Template

## Issue

Addresses # (https://github.com/bloom-housing/bloom/issues/1687)

## Description

GET /listings` filters query param has been changed to support a querystring serialized array of filters, it's a breaking change because comparison property can no longer be an array. Also a property ordering problem has been resolved. Now the strict requirement for every client using the API is to use `qs` serialization format for query params.

## Type of change

- [ ] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


sites/public/pages/index.tsx
```diff
@@ -80,11 +81,18 @@ export default function Home(props: IndexProps) {
...
+        view: "base",
+        limit: "all",
+        $comparison: "<>",
+        status: "pending",
+      },
+      paramsSerializer: (params) => {
+        return qs.stringify(params)
+      },
+    })
+
```
seanmalbert, line 95: And again here, shouldn't there be a filter key that's an array of object? I tested with setting filter: [with different objects here] and it worked.

dominikx96:
@seanmalbert it was because I forgot to update the backend client, can you do a review again? :)

seanmalbert (approved):
@pbn4 and @dominikx96 , this looks good, thanks for all the updates!

avaleske (commented):
Thanks for fixing the filter ordering issue!

@seanmalbert asked me to review this PR, as well as share some other things re filters I've been thinking about.

### Questions about this refactor:
1. Can we update the name of this PR so it matches what's happening?

2. Passing an array of `ListingFilterParams` could allow for some odd edge cases. We're assuming that each ListingFilterParams will only have one comparison field and one filter field, but what if someone passes a `ListingFilterParams` that looks like this:
```
{
  $comparison: Compare["="],
   neighborhood: "Fox Creek",
   status: "pending",
},
```
   which filter field should the filter logic pick? Neighborhood or Status?

   Is there a way in the validation logic to ensure that only one field of `ListingFilterParams` is set? If so, that seems ideal, cause then we can still validate the types of each filter value. Otherwise, an array of objects with an interface like
```
{
  comparison?: Compare,
  fitler: ListingFilterKey,
  filterValue: any,
}
```
   seems less prone to error, although we'd then have to validate filter values some other way.

3. I think it'd be helpful to wrap the querystring logic in a helper function, similar to what we've done in our fork here:
https://github.com/CityOfDetroit/bloom/blob/d7ac378a7839cffd350cf993164fb68f679b8a19/ui-components/src/helpers/filters.ts#L28-L37
That way, next time the URL format changes, there's only one place to change.

### Some additional thoughts on filters that are likely out of scope for this PR
I moved these to this discussion:
https://github.com/bloom-housing/bloom/discussions/1794

avaleske:
@seanmalbert this PR was already merged, but I shared my thoughts anyway since they point toward the future direction for filters.

backend/core/src/listings/dto/listing.dto.ts
```diff
@@ -838,4 +861,5 @@ export const filterTypeToFieldMap: Record<keyof typeof ListingFilterKeys, string
   name: "listings.name",
   neighborhood: "property.neighborhood",
   bedrooms: "unitTypeRef.num_bedrooms",
+  leasingAgents: "leasingAgents.id",
```
avaleske, line 864: Also @seanmalbert this needed to be added to the ListingFilterKeysEnum here:
https://github.com/bloom-housing/bloom/blob/26a54444188e8c3c369a981e3b970ad419dff8f3/backend/core/src/listings/types/listing-filter-keys-enum.ts

avaleske: nvm it was! deleting this comment.

Issue #1653, opened 2021-08-11
Add New Users Invite Flow 


kathyccheng:
As an admin, 
I want to be able to invite new users to the partners portal 
So that I can add new leasing agents to their associated listing 

**Acceptance Criteria**
* When I am on the users page, then I will see a button to add user
* When I click on the button to add user, then I will be take to the add user form which includes Name, Email, and Listings
* When I fill out the information and click Invite, then an email will be sent to the user and I will be taken back to the users page with a success message and the last user at the top of the list with the status Unconfirmed 

**Questions**
* Should email be its own issue?
* Copy for the email 

**Designs**
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4084%3A169393

https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4084%3A169536

https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4084%3A169464

slowbot:
@kathyccheng Adding user flow to this issue

kathyccheng:
@slowbot I drafted some email language above, let me know what you think

slowbot:
@kathyccheng Looks pretty good to me. I would have started with our existing user email too.

pbn4:
>This endpoint should be able to handle not only assigning the user roles but also assigning listings, right?

@seanmalbert Yes thats what I had in mind.

seanmalbert:
@dominikx96 , after you address the issues noted in #1651 , you can start on this. If #1801 hasn't been merged by then, please work off of that branch and with @pbn4 to get started on the front piece of this.

slowbot:
Blocked from QA by https://github.com/bloom-housing/bloom/issues/1980

PR #1801, opened 2021-09-08
Add POST /users/invite and extend /confirm with optional password

pbn4:
# Pull Request Template

## Issue

Addresses # (https://github.com/bloom-housing/bloom/issues/1653)

## Description

This change add `POST /users/invite` endpoint and extends `PUT /users/confirm` with optional password change. 

There is a new required environment variable: `PARTNERS_PORTAL_URL` which has to be added to every heroku app, otherwise an app will crash. 

Invite flow:

1. Frontend posts `UserInviteDto` to `/users/invite`
2. Backend creates a new unconfirmed user account and sends an email to the new user with a  user id bound JWT confirmation token (24h expiration)
3. User clicks the confirmation URL and is redirected to `http://partners_portal_hostname/?token=<token>`
4. Frontend displays a password update form 
5. Frontend submits `ConfirmDto` with new optional password field
6. Backend verifies confirmation token, confirms the user and sets a new password

## Type of change

- [ ] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


backend/core/src/auth/dto/user-invite.dto.ts
```diff
@@ -0,0 +1,28 @@
...
+  "createdAt",
+  "updatedAt",
+  "roles",
+  "leasingAgentInListings",
+] as const) {
+  @Expose()
+  @IsOptional()
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @Type(() => UserRolesCreateDto)
+  roles?: UserRolesCreateDto | null
```
seanmalbert, line 20: Should roles be optional for the invite? If we don't assign them isPartner, then they won't be able to access the portal.

seanmalbert (commented):
Hey @pbn4 ,
For the most part this looks great. I do have a few questions though:
1. should roles be required when inviting a user? If we don't assign them any roles, then they won't be able to login
2. I think jurisdiction should also be required and not empty
3. should the invite endpoint check if the user exists? What happens with what's here and a user with that email already exists?

If we make roles and jurisdiction required, then I think we should have the tests you wrote updated to include cases where those values are not passed in. Likewise with trying to invite a user with an email that already exists.

backend/core/src/auth/dto/user-invite.dto.ts
pbn4: > should roles be required when inviting a user? If we don't assign them any roles, then they won't be able to login

@seanmalbert We don't have any logic in place to prevent partners portal login based on roles, any user can login in (applicants too). 

> I think jurisdiction should also be required and not empty

I have not made it required because in the designs there is no jurisidction input. 

> should the invite endpoint check if the user exists? What happens with what's here and a user with that email already exists?

`invite` reuses `UserService.createUser` which in turn checks that and throws: 
```
  EMAIL_IN_USE: { message: "emailInUse", status: HttpStatus.BAD_REQUEST },
```

pbn4: @seanmalbert Why jurisdictions should be required?

seanmalbert: @pbn4 ,
This doesn't matter so much for user's assigned with isPartner, since they are directly assigned listings and can't create them, but for admin users, they need to be assigned a jurisdiction so we can know which jurisdictions they can create listings in. The mocks don't show it, but if an admin has access to more than one jurisdiction there needs to be an option to select which jurisdictions the new user has. If the admin is only assigned to one, which will be the case for everyone but us (for now anyway), then the input will be hidden with the value set to the jurisdiction they have. So on the backend with these new endpoints, we also need to ensure that the jurisdictions passed into creating a partner user are limited to what the current user has. I think we already updated the user update function to check this.

Issue #1728, opened 2021-08-25
Notifications - Homepage Sign Up Component 


kathyccheng:
As a user
I want to be able to sign up to receive notifications 
So that I know when I should come back to check the site for things to apply to 

**Acceptance Criteria**
* When I'm on the homepage, I'll see a sign up section with a call to action
* When I click Sign Up, I'll be taken to the Granicus sign up page: https://public.govdelivery.com/accounts/CAALAME/signup/29386

**Notes**
* Use our sign up component 
* Sign Up goes to Granicus page to be built by Alameda County + Granicus team 
* This seems like a good first jurisdiction setting so we'll want to add a new table, new relationship for jurisdictions to features (feature flags). Feature flag table and a relationship between features and jurisdiction. 
* We can check for the value from the jurisdiction, if there have 2 columns, if not 1 

**Designs**
![image.png](https://images.zenhubusercontent.com/5f0361f3c34cb3aabaf096f2/406f8e36-1240-41f8-ac95-f0fc3642c1b1)
https://www.figma.com/file/i3IAkbjv4nVPKg1d1E1g1d/bloom-accounts?node-id=333%3A2420

emilyjablonski:
I'm curious how to handle this in core. Will every single jurisdiction have this sign up component and it should always exist? Should it be toggled on via an env variable or a feature flag? Where should we be storing the URL to the sign up if it's on a jurisdiction level?

kathyccheng:
@seanmalbert can you take a look at my notes above and clean that up?

emilyjablonski:
May not be able to QA this on the dev environment since this is a jurisdiction setting, will update back

slowbot:
We'll need to update the icons used here. Will try and locate matching SVGs for each

Font Awesome
https://fontawesome.com/v6.0/icons/house?s=thin
https://fontawesome.com/v6.0/icons/envelope?s=thin

Thin
https://drive.google.com/file/d/0B_JSwOwJ90FTQTBKQlBPdFVhMWc/view?usp=sharing&resourcekey=0-xLSKeGryZThe99KjWr7KSA
https://drive.google.com/file/d/0B_JSwOwJ90FTQ2tpR2ZRTDcwZnc/view?usp=sharing&resourcekey=0-5OacHM5NtxFF2CY5DgIU2Q

emilyjablonski:
Sounds good, we don't currently have these in ui-components I can add them from the Figma link. What is the second "thin" link for?

[...]

PR #1802, opened 2021-09-08
1728/notifications

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1728 

## Description

If a jurisdiction has the notificationsSignUpURL field set, the home page will display a second action block with the sign up link. The migration sets the value for Alameda to be the current sign-up link.

## Type of change

Load up the homepage with the jurisdiction ENV variable set to Alameda and to anything but Alameda to see both states.

## Checklist:

### DIFF
--- a/backend/core/src/jurisdictions/jurisdictions.controller.ts
+++ b/backend/core/src/jurisdictions/jurisdictions.controller.ts
@@ -11,7 +11,7 @@ import {
   ValidationPipe,
 } from "@nestjs/common"
 import { ApiBearerAuth, ApiOperation, ApiTags } from "@nestjs/swagger"
-import { DefaultAuthGuard } from "../auth/guards/default.guard"
+import { OptionalAuthGuard } from "../auth/guards/optional-auth.guard"
 import { AuthzGuard } from "../auth/guards/authz.guard"
 import { ResourceType } from "../auth/decorators/resource-type.decorator"
 import { mapTo } from "../shared/mapTo"
@@ -27,7 +27,7 @@ import {
 @ApiTags("jurisdictions")
 @ApiBearerAuth()
 @ResourceType("jurisdiction")
-@UseGuards(DefaultAuthGuard, AuthzGuard)
+@UseGuards(OptionalAuthGuard, AuthzGuard)
 @UsePipes(new ValidationPipe(defaultValidationPipeOptions))
 export class JurisdictionsController {
   constructor(private readonly jurisdictionsService: JurisdictionsService) {}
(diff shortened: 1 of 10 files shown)


emilyjablonski:
@seanmalbert Here is what I mentioned in standup today. The two issues I have/see now are (1) Getting a 401 on the jurisdictions service and (2) Feels weird to fetch all the jurisdictions, to find our current one, to find the sign-up URL

seanmalbert:
> @seanmalbert Here is what I mentioned in standup today. The two issues I have/see now are (1) Getting a 401 on the jurisdictions service and (2) Feels weird to fetch all the jurisdictions, to find our current one, to find the sign-up URL

1. This shouldn't be the case since anonymous users should be able to read from jurisdictions according to what's in the auth policy in the backend `backend/core/src/auth/authz_policy.csv`, so there's probably a name mismatch
2. I agree. process.env.jurisdictionName should be available in public, assuming you have `JURISDICTION_NAME` set in your .env file. I think we can create a new endpoint for getting a jurisdiction by name. See the jurisdiction controller for how get by id is setup as a reference.

seanmalbert:
@emilyjablonski , I updated jurisdictions to use the optional auth guard so we can now read jurisdictions with an anonymous user. You should now be unblocked on that.

backend/core/src/jurisdictions/jurisdictions.controller.ts
```diff
@@ -50,12 +50,12 @@ export class JurisdictionsController {
     return mapTo(JurisdictionDto, await this.jurisdictionsService.update(jurisdiction))
   }
 
-  @Get(`:jurisdictionId`)
-  @ApiOperation({ summary: "Get jurisdiction by id", operationId: "retrieve" })
-  async retrieve(@Param("jurisdictionId") jurisdictionId: string): Promise<JurisdictionDto> {
+  @Get(`:jurisdictionName`)
```
pbn4, line 53: I would like to keep `GET /jurisidctions/:id` endpoint because it's a REST API convention and most endpoints follow it.  `name` search you need here could be implemented through `list()` endpoint. Please add optional query param to `list and do such filtering there. Frontend then should receive and parse a single element array as a response.

pbn4 (changes requested):

backend/core/src/jurisdictions/jurisdictions.controller.ts
seanmalbert: @pbn4 ,
It makes sense to keep that endpoint, but I don't like using list for this either, since we're only expecting one to return (name column has been set to unique), so having another route like `jurisdictions/byName/:name` makes sense to me.

backend/core/test/jurisdictions/jurisdictions.e2e-spec.ts
```diff
@@ -71,4 +71,31 @@ describe("Jurisdictions", () => {
...
+      .expect(201)
+    expect(res.body).toHaveProperty("id")
+    expect(res.body).toHaveProperty("createdAt")
+    expect(res.body).toHaveProperty("updatedAt")
+    expect(res.body).toHaveProperty("name")
+    expect(res.body.name).toBe("test2")
+
+    const getById = await supertest(app.getHttpServer())
+      .get(`/jurisdictions/byName/${res.body.name}`)
+      .set(...setAuthorization(adminAccesstoken))
```
seanmalbert, line 89: We should be able to get this without the token now. Same with the existing gets above.

seanmalbert (approved):
Looks good! Can you update the e2e tests so that the get requests don't have the admin token, since they should be accessible by an anonymous user.

backend/core/test/jurisdictions/jurisdictions.e2e-spec.ts
emilyjablonski: updated 👍

PR #1808, opened 2021-09-09
Remove listing sorting by

pbn4:
# Pull Request Template

## Issue

Addresses # (https://github.com/bloom-housing/bloom/issues/1165)

- [x] This change addresses the issue in full
- [ ] This change addresses only certain aspects of the issue
- [ ] This change is a dependency for another issue
- [ ] This change has a dependency from another issue

## Description

Removes sorting listing by ID (which is UUID) and adds default sorting by createdAt. 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Prototype/POC (not to merge)
- [ ] This change is a refactor/address technical debt
- [ ] This change requires a documentation update
- [ ] This change requires a SQL Script

## How Can This Be Tested/Reviewed?

Please describe the tests that you ran to verify your changes. Provide instructions so we can review. Please also list any relevant details for your test configuration

- [ ] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have reviewed the changes in a desktop view
- [ ] I have reviewed the changes in a mobile view
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have assigned reviewers
- [x] I have updated the changelog to include a description of my changes

PR #1817, opened 2021-09-10
Improvements to backend listing sorting

anders-schneider:
## Issue

I'm not aware of an issue in https://github.com/bloom-housing/bloom for sorting listings in the backend. On the Detroit side, we've been tracking this in https://github.com/CityOfDetroit/bloom/issues/193.

## Description

This change builds a simple framework for ordering the listings returned by the backend `GET /listings`. The behavior:
- If the `GET /listings` request includes an `orderBy` parameter, the results will be ordered correspondingly. There are only two valid `orderBy` parameters: `mostRecentlyUpdated` and `applicationDates`.
- In the absence of an `orderBy` parameter, the default behavior is to order listings by `applicationDates` (i.e. by soonest applicationDueDate, and then by latest applicationOpenDate).

Note: this change also fixes a bug in the way backend listings ordering worked in conjunction with pagination. Before this change, the `innerFilteredQuery` was ordered by listing.id and the full query was ordered by listing.applicationDueDate and listing.applicationOpenDate. If the request was for e.g. the first page of listings (not all listings), it wasn't guaranteed that the listing with the soonest applicationDueDate would appear, because pagination happens on the `innerFilteredQuery` but ordering by applicationDueDate only happens on the full query. The way this change fixes it: it applies the same `OrderByCondition` on both the `innerFilteredQuery` and the full query. (There are other approaches that would only involve applying the `ORDER BY` clause once, but we opted for the simplicity of this approach; see https://github.com/CityOfDetroit/bloom/issues/193#issuecomment-915539323 for a discussion of this decision.)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

```
$ cd backend/core
$ yarn test
$ yarn test:e2e:local
```

I also verified that the listings at localhost:3000/listings are displayed in order by applicationDueDate.

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I have run `yarn generate:client` if I made backend changes

### DIFF
--- a/backend/core/src/listings/dto/listing.dto.ts
+++ b/backend/core/src/listings/dto/listing.dto.ts
@@ -43,6 +43,7 @@ import {
   UnitsSummaryDto,
   UnitsSummaryUpdateDto,
 } from "../../units-summary/dto/units-summary.dto"
+import { OrderByFieldsEnum } from "../types/listing-orderby-enum"
 import { IdNameDto } from "../../shared/dto/idName.dto"
 
 export class ListingDto extends OmitType(Listing, [
@@ -845,6 +846,18 @@ export class ListingsQueryParams extends PaginationAllowsAllQueryParams {
   @IsOptional({ groups: [ValidationsGroupsEnum.default] })
   @IsString({ groups: [ValidationsGroupsEnum.default] })
   view?: string
+
+  @Expose()
+  @ApiProperty({
+    name: "orderBy",
+    required: false,
+    enum: OrderByFieldsEnum,
+    enumName: "OrderByFieldsEnum",
+    example: "updatedAt",
+  })
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @IsEnum(OrderByFieldsEnum, { groups: [ValidationsGroupsEnum.default] })
+  orderBy?: OrderByFieldsEnum
 
   @Expose()
   @ApiProperty({
--- a/backend/core/src/listings/listings.service.spec.ts
+++ b/backend/core/src/listings/listings.service.spec.ts
@@ -7,6 +7,7 @@ import { ListingsQueryParams, ListingFilterParams } from "./dto/listing.dto"
 import { Compare } from "../shared/dto/filter.dto"
 import { TranslationsService } from "../translations/translations.service"
 import { AmiChart } from "../ami-charts/entities/ami-chart.entity"
+import { OrderByFieldsEnum } from "./types/listing-orderby-enum"
 
 // Cypress brings in Chai types for the global expect, but we want to use jest
 // expect here so we need to re-declare it.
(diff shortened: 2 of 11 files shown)


emilyjablonski:
Just a heads up we currently have another PR open around sorting listings from @pbn4 https://github.com/bloom-housing/bloom/pull/1808

pbn4:
@emilyjablonski This PR takes more comprehensive approach to the problem than what I did in #1808 . Let's continue here.

pbn4 (approved):
No comments on that one. Thank you for contributing.

anders-schneider:
Ah thanks @emilyjablonski and @pbn4 for pointing out that other sorting PR, and thanks for the review!

Our next step on this work will be to figure out the interaction between sorting and filtering. The high-level idea is: if a request indicates that we're filtering on column X and some listings have data in column X while others have `null`, we'd like to display the listings _with_ data in that column above those that have `null`s. If you all have thoughts on that, I'd be happy to discuss. Otherwise I'll send a PR for that change once it's ready.

anders-schneider:
Oh and I believe the Cypress test that's failing in the CircleCI run was already failing before this PR? And it looks like the "run linters" check is failing on some permissions that I don't have.

seanmalbert:
> Ah thanks @emilyjablonski and @pbn4 for pointing out that other sorting PR, and thanks for the review!
> 
> Our next step on this work will be to figure out the interaction between sorting and filtering. The high-level idea is: if a request indicates that we're filtering on column X and some listings have data in column X while others have `null`, we'd like to display the listings _with_ data in that column above those that have `null`s. If you all have thoughts on that, I'd be happy to discuss. Otherwise I'll send a PR for that change once it's ready.

Hey @anders-schneider , I have a question, if you're filtering on column X, you also want to return data where column X is null? It makes sense that if we're sorting on column X that we would want to show the data where column X is not null before the ones with null, but if we're filtering on it, shouldn't we return the ones that only have the requested filter value? I suppose there could be valid use cases where you'd want that, but I think we'd want to make that clear to the end user that that's what's getting returned.

anders-schneider:
> Hey @anders-schneider , I have a question, if you're filtering on column X, you also want to return data where column X is null? It makes sense that if we're sorting on column X that we would want to show the data where column X is not null before the ones with null, but if we're filtering on it, shouldn't we return the ones that only have the requested filter value? I suppose there could be valid use cases where you'd want that, but I think we'd want to make that clear to the end user that that's what's getting returned.

Good question, and yes agreed that for some filters (maybe zip code? neighborhood?) if a listing has a null value, we probably don't want to include that listing in the results. But for others (rent range? unit size?) we may want to include listings that don't have that data at the bottom of the search results. Our plan was to consider each type of filter and make a reasonable decision.

And to your point ("make that clear to the end user that that's what's getting returned"), @avaleske suggested adding a UI component with some explanatory text as a "break" between the matching listings (above the component) and the "we don't know if these match" listings (below the component).

We are still thinking about the right approach here, and would appreciate any thoughts/ideas from your side!

avaleske:
> @avaleske suggested adding a UI component with some explanatory text as a "break" between the matching listings (above the component) and the "we don't know if these match" listings (below the component).

Adding to this, we will be launching with a lot of incomplete listings data, especially around accessibility. To avoid filtering out listings that might match, but for which we don't have data, we plan to add some filtering options such as "include listings with missing accessibility data" with a prompt to contact the property manager.

seanmalbert:
@anders-schneider and @avaleske , continued filter discussion at #1822

PR #1826, opened 2021-09-14
Fix prettier action

emilyjablonski:
Details in a comment below, but we introduced an issue that was causing the prettier action to fail to run

### DIFF
--- a/sites/public/email_templates/email.html
+++ b/sites/public/email_templates/email.html
@@ -419,7 +419,7 @@
                           <tbody>
                             <tr>
                               <td align="center">
-                                <a class="btn btn-primary" href="http://htmlemail.io" target="_blank">See Listing and Apply</a> </td>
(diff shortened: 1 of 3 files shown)


emilyjablonski:
https://github.com/bloom-housing/bloom/pull/1792 may have cause the error that prevented prettier from running @seanmalbert

sites/public/email_templates/email.html
```diff
@@ -419,7 +409,12 @@ <h1>Rental opportunity at <br>/Listing Name/</h1>
                           <tbody>
                             <tr>
                               <td align="center">
-                                <a class="btn btn-primary" href="http://htmlemail.io" target="_blank">See Listing and Apply</a> </td>
```
emilyjablonski, line 422: Here is my only change, everything else is prettier

jaredcwhite:
@emilyjablonski Looks fine to merge in…I think the email template is a draft example anyway so will need further tweaking.

seanmalbert (approved):

PR #1827, opened 2021-09-15
Adds tailwind rtl plugin

seanmalbert:
Per [this discussion](https://github.com/bloom-housing/bloom/discussions/1400#discussioncomment-1213771), we needed to be able to add tailwind plugins and Detroit specifically needs the rtl plugin.

This gets that working, though one I think I don't like about his is that I had to add tailwind to the root package.json. @jaredcwhite , do you have any ideas to get this working without that?

@willrlin, in `sites/public/pages/_app.tsx`, I left some code commented out, so you can test that to see if this works as you expected.


emilyjablonski:
We probably want to rebase off latest dev

willrlin:
I'm getting this error when I try to patch this in and load the page:
```
error - ../../node_modules/next/dist/compiled/css-loader/cjs.js??ruleSet[1].rules[3].oneOf[5].use[1]!../../node_modules/next/dist/compiled/postcss-loader/cjs.js??ruleSet[1].rules[3].oneOf[5].use[2]!../../node_modules/mapbox-gl/dist/mapbox-gl.css
TypeError: getProcessedPlugins is not a function
```

jaredcwhite:
`I had to add tailwind to the root package.json`

That's strange, the tailwind dependency should be a given via the `ui-components` dependency tree. Were you getting a specific error without adding that explicitly?

willrlin:
I pushed up my test patch in this branch: https://github.com/CityOfDetroit/bloom/tree/495/add-rtl-plugin

seanmalbert:
> I pushed up my test patch in this branch: https://github.com/CityOfDetroit/bloom/tree/495/add-rtl-plugin

@willrlin ,
I was able to get it working on your branch by setting tailwindcss version back to 7. One of the packages (mapbox-gl?) seems to be incompatible with version 8. I tested with adding the bit of code to _app.tsx. I hope you don't mind that I pushed a commit with the package changes to your branch, so you can review.

seanmalbert:
> @willrlin ,
> I was able to get it working on your branch by setting tailwindcss version back to 7. One of the packages (mapbox-gl?) seems to be incompatible with version 8. I tested with adding the bit of code to _app.tsx. I hope you don't mind that I pushed a commit with the package changes to your branch, so you can review.

In all the places I say that tailwindcss got set back to 7, I meant postcss' version got set back to 7.

willrlin:
Great, thanks! I was able to get this working. Weirdly in one case that I can find, adding a className of `float-end` doesn't work correctly, but `float-right` does (see our `sites/public/pages/listings.tsx` for a `float-right` on the `ListingCard`). Other `tailwindcss-rtl` attributes like `ms-10` work correctly. Do you have any leads for figuring out why in that one case it doesn't work?

seanmalbert:
> > I had to add tailwind to the root package.json
> 
> @seanmalbert That's strange, the tailwind dependency should be a given via the `ui-components` dependency tree. Were you getting a specific error without adding that explicitly?

@jaredcwhite ,
Yea, if I pull the dependency from the root package, it'll throw `Error: Cannot find module 'tailwindcss/plugin'`.

seanmalbert:
> Great, thanks! I was able to get this working. Weirdly in one case that I can find, adding a className of `float-end` doesn't work correctly, but `float-right` does (see our `sites/public/pages/listings.tsx` for a `float-right` on the `ListingCard`). Other `tailwindcss-rtl` attributes like `ms-10` work correctly. Do you have any leads for figuring out why in that one case it doesn't work?

Not that I've found, especially since those just map to float-dir properties (https://github.com/20lives/tailwindcss-rtl/blob/master/src/floatUtilities.js). It did pop into my head that there might be issues with using that inside a `flex` element like the ListingCard, but I would think the regular floats would have issues there too, if that was the case.

PR #1848, opened 2021-09-16
dependency issues

emilyjablonski:
Our dependency warnings on fresh install went from 3 --> 23 after the tailwind/node/etc upgrade. I think we just missed a number of packages that also needed to be upgraded as we moved to React 17. Upgraded the packages I could which involved some tiny code updates in some places, I commented on all of them.

After this PR we have 5 warnings remaining, 4 of which we cannot fix ourselves. They are:

-------------
- `warning "workspace-aggregator-e0d01dfa-ee0d-47ef-8d50-0f1c5d0df64e > @bloom-housing/ui-components > react-text-mask@5.4.3" has incorrect peer dependency "react@^0.14.0 || ^15.0.0 || ^16.0.0".`
- https://github.com/text-mask/text-mask/pull/1038 This will be fixed when this PR goes in but it's unclear if it's being actively maintained so maybe not soon, but doesn't seem to cause any issues
--------------
- `warning "workspace-aggregator-e0d01dfa-ee0d-47ef-8d50-0f1c5d0df64e > @bloom-housing/ui-components > @storybook/api > @reach/router@1.3.4" has incorrect peer dependency "react@15.x || 16.x || 16.4.0-alpha.0911da3".`
- `warning "workspace-aggregator-e0d01dfa-ee0d-47ef-8d50-0f1c5d0df64e > @bloom-housing/ui-components > @storybook/api > @reach/router@1.3.4" has incorrect peer dependency "react-dom@15.x || 16.x || 16.4.0-alpha.0911da3".`
- `warning "workspace-aggregator-e0d01dfa-ee0d-47ef-8d50-0f1c5d0df64e > @bloom-housing/ui-components > @storybook/api > @reach/router > create-react-context@0.3.0" has incorrect peer dependency "react@^0.14.0 || ^15.0.0 || ^16.0.0".`
- There's an open issue for this here: https://github.com/reach/router/issues/429 Seems like Storybook is on it in terms of moving off of reach/router if they don't update to React 17
------------------
- `warning "workspace-aggregator-e0d01dfa-ee0d-47ef-8d50-0f1c5d0df64e > @bloom-housing/ui-components > @storybook/react > @storybook/react-docgen-typescript-plugin > react-docgen-typescript@2.1.0" has incorrect peer dependency "typescript@>= 4.3.x".`
- Upgrading typescript just causes a different dependency warning from nest for typescript then being too high of a version for the nest version we have, so upgrading typescript seems like a larger issue outside of this PR
---------------

To test you can fresh `yarn install` on dev and then on this branch!


sites/partners/src/applications/ApplicationsColDefs.ts
```diff
@@ -66,7 +66,6 @@ export function getColDefs(maxHouseholdSize: number, countyCode: string) {
       width: 150,
       minWidth: 120,
       pinned: "left",
-      type: "leftAligned",
```
emilyjablonski, line 69: No longer exists and is default behavior anyway

ui-components/.jest/setup-tests.js
```diff
@@ -1,12 +1,8 @@
 import "@testing-library/jest-dom/extend-expect"
 
-import { configure } from "enzyme"
```
emilyjablonski, line 3: We weren't actually using enzyme anymore

ui-components/src/global/mixins.scss
```diff
@@ -1,3 +1,5 @@
+@use "sass:math";
```
emilyjablonski, line 1: Changes here needed from the ag-grid update (the old division way is deprecated now)

yarn.lock
```diff
@@ -20674,20 +20443,20 @@ swagger-ui-express@^4.1.4:
...
-    dequal "2.0.2"
-
 swr@^0.5.5:
   version "0.5.5"
   resolved "https://registry.yarnpkg.com/swr/-/swr-0.5.5.tgz#c72c1615765f33570a16bbb13699e3ac87eaaa3a"
   integrity sha512-u4mUorK9Ipt+6LEITvWRWiRWAQjAysI6cHxbMmMV1dIdDzxMnswWo1CyGoyBHXX91CchxcuoqgFZ/ycx+YfhCA==
   dependencies:
     dequal "2.0.2"
 
+swr@^1.0.1:
```
jaredcwhite, line 20453: I wonder how we still have two different versions of SWR? Should be the same one used for partners and public, right?

jaredcwhite (approved):
Thanks for digging into this. Changes look good to me! (just one question)

yarn.lock
emilyjablonski: Hmmmmmm that is weird! I'll take a look, thanks for noticing :)

seanmalbert (approved):
Looks good! For the UnitForm updates, I made a minor change to use the existing check for defaultUnit.

Issue #1811, opened 2021-09-09
Listings Mgmt Validations: Preview and Save as Draft - FrontEnd


seanmalbert:
Continuing from #1541 , this is the frontend implementation of that work. This work entails getting the response from the backend and parsing and setting errors for the fields.

emilyjablonski:
@kathyccheng @slowbot The backend is validating the whole form bc of existing required fields, so I will need some more information about what within the unit form should be considered required.

kathyccheng:
@slowbot can you chime in on what fields in the unit form you think should be required?

slowbot:
@emilyjablonski Can we add a label to the default state of photos and units similar to the mockups 

![image](https://user-images.githubusercontent.com/13915/136435562-9ea3e78b-6154-440c-ab83-766e32b76561.png)

Also open to creating a new issue for this

emilyjablonski:
Fixed!

Issue #1542, opened 2021-07-20
Listings Mgmt Validations: Required to Publish 


kathyccheng:
As a leasing agent, 
I want to know what's required to publish a listing
So that I know I have inputted all the right information

**Acceptance Criteria** 
* When I am filling out a listing, and I click publish, if I don't have all of the required fields filled out, then I will see an error message message and all of the fields I'm missing highlighted in red 
* When I have filled out all of the required fields, and I click publish, then I will be taken to the summary view where I can preview the listing and I will see the listing live on the public site 

**Notes**
* Update error message from "There was an issue submitting the form" to "Please resolve any errors before publishing your listing." 
* Validating the listing to transition to Open state 

**Required Fields** 
* Name
* Developer
* Asset
* Is there a digital application? - boolean
* Is there a paper application? - boolean
* Is this a referral opportunity? - boolean
* Building Address
* Units - At least one unit
* Deposit Min
* Deposit Max
* Rental Assistance
* How is the application review order determined?
* Is the waitlist open?
* Leasing agent email
* Leasing agent name
* Leasing agent phone 

**Notes**
* This issue DOES NOT consider dependency or sub-object required dependencies

**Reference Spreadsheet**
https://docs.google.com/spreadsheets/d/1yzy7t1n0lNuhjmElnBH7rFOk2qZYdK7iHwhnyY3QrmM/edit#gid=515768701

kathyccheng:
@slowbot please update error copy

kathyccheng:
Need to figure out how we move between states

seanmalbert:
From conversation the other day, we're going to create two separate "dtos" to validate the fields against. These could be usable on the frontend and backend. @pbn4 , can you add more info based off of what you're doing with state management.

seanmalbert:
On the boolean types, if the user says "yes", then they need to add at least one of those items.

seanmalbert:
For these types:
Is there a digital application? - boolean
Is there a paper application? - boolean
Is this a referral opportunity? - boolean

We want to add actual boolean fields on the model to make this easier.

seanmalbert:
@pbn4 ^

Issue #1541, opened 2021-07-20
Listings Mgmt Validations: Preview and Save as Draft - Backend


kathyccheng:
As a leasing agent, 
I want to it to be easy to save a draft and preview,
So that I can easily save work as I go along 

**Acceptance Criteria** 
* When I am on a new listing form, and I click Save without having filled anything in, then I expect to see an error letting me know I need to add a listing name
* When I fill in a listing name, and I click Save, then I expect it to be a success and I will see the listing summary page
* When I am on a new listing form, I expect to see the defaulted values in the form 

**Notes** 
* Move validation to back end

**Required Fields** 
* The only field that is required for Save as a Draft or Preview is Listing Name 
* I believe this is the current behavior already 

**Updates to the Listing Form** 
* Remove from the form defaults - 
  * Year Built (2021), 
  * Leasing Agent Email (test@email.com)
  * Application Fee $0
* Add form default - Is waitlist open --> NO 
* Add default rental assistance text to form: The property is subsidized by the Section 8 Project-Based Voucher Program. As a result, Housing Choice Vouchers, Section 8 and other valid rental assistance programs are not accepted by this property.
* Add form default Deposit Min and Deposit Max to $0

**Updates to the Public Site Defaults** 
* Remove from public form defaults - 
  * Neighborhood if empty, 
  * Application Fee section if empty, 
  * Rental Assistance text (because this should be on the form instead) 
  * It should not default to the Referral type with the For More Information, that block should only be there if they choose referral in application types 

**Spreadsheet** 
This summarizes all of the updates above 
https://docs.google.com/spreadsheets/d/1yzy7t1n0lNuhjmElnBH7rFOk2qZYdK7iHwhnyY3QrmM/edit#gid=515768701

Public Listing Screenshot for Reference 
https://share.getcloudapp.com/DOuqOlXn

seanmalbert:
@pbn4 , @software-project and @jaredcwhite ,

So I can make sure I understand what was proposed from the planning meeting regarding state and validation management, here is how I'm understanding it from a basic high level.

- On the backend we want to define some structure or structures that can be used to determine what fields on the front end should be required. This structure could also be used to maintain the error states based off of the state change action.
- in the backend service, we can use this structure along with the data submitted to determine if the listing can be updated to the intended state.
  - there seems to be some discussion needed around whether we want to save any part of that as unpublished first and then return the errors (if action is publish)
[...]

PR #1850, opened 2021-09-17
feat: listings management draft and publish validation backend & frontend

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1543 #1541 #1811 #1542

## Description

Adds the backend and frontend pieces for validations on listings management on both draft and publish. I also refactored a little how we are submitting data so we don't need to keep track of status or submitData in state.

For draft, required fields are:
* Listing name
* Jurisdiction (only shows if user has more than one, otherwise defaults)

For publish, required fields are:
* Listing name
* Jurisdiction
* Developer
* Image
* Is there a digital application?
* Is there a paper application? 
* Is this a referral opportunity? 
* Building Address (specifically street, city, state, zipCode)
* Units - At least one unit
* Deposit Min (has a default)
* Deposit Max (has a default)
* Rental Assistance (has a default)
* How is the application review order determined? (has a default)
* Is the waitlist open?
* Leasing agent email
* Leasing agent name
* Leasing agent phone

## Type of change

Create a new listing from scratch and without filling anything out and try to save as draft and publish and the frontend validations should kick in. Also try to edit an existing listing.

The one kinda wonky frontend validation is buildingAddress, because if you send nothing we get a generic "buildingAddress" error but if at least one field is filled out we get errors specific to the four required fields. Should show as individual errors tho.

## Checklist:


backend/core/src/listings/dto/listing-published-create.dto.ts
```diff
@@ -38,6 +39,12 @@ export class ListingPublishedCreateDto extends OmitType(ListingCreateDto, [
   "reviewOrderType",
   "units",
 ] as const) {
+  @Expose()
+  @ValidateNested({ groups: [ValidationsGroupsEnum.default], each: true })
+  @Type(() => AssetCreateDto)
```
pbn4, line 44: I know it's a WIP but i looked at it out of curiosity. Could you put class-validator decorators together and @Type() next to the property definition?

backend/core/src/shared/entities/address.entity.ts
```diff
@@ -32,10 +32,9 @@ export class Address {
 
   @Column({ type: "text", nullable: true })
   @Expose()
-  @IsOptional({ groups: [ValidationsGroupsEnum.partners] })
   @IsString({ groups: [ValidationsGroupsEnum.default] })
   @MaxLength(64, { groups: [ValidationsGroupsEnum.default] })
-  city?: string | null
+  city: string | null
```
pbn4, line 37: I think `| null` can be removed too if they are not optional because frontend side we do not distinguish between undefined and null, right?

sites/partners/src/listings/PaperListingForm/index.tsx
```diff
@@ -456,54 +442,87 @@ const ListingForm = ({ listing, editMode }: ListingFormProps) => {
...
+              if (
+                typeof obj[key] === "object" &&
+                !Array.isArray(obj[key]) &&
+                Object.keys(obj[key]).length === 0
+              ) {
+                delete obj[key]
+              }
+            })
+          }
+          removeEmptyFields(formattedData)
```
dominikx96, line 493: Looks like a good candidate to be a helper function in a separate file 👀 Especially that this component is veeeeery long.

sites/partners/src/listings/PaperListingForm/index.tsx
```diff
@@ -456,54 +442,87 @@ const ListingForm = ({ listing, editMode }: ListingFormProps) => {
...
+          setAlert("api")
+          const { data } = err.response || {}
+          if (data.statusCode === 400) {
+            data?.message?.forEach((errorMessage: string) => {
+              const fieldName = errorMessage.split(" ")[0]
+              const readableError = getReadableErrorMessage(errorMessage)
+              if (readableError) {
+                setError(fieldName, { message: readableError })
+              }
+            })
```
dominikx96, line 523: Shouldn't we catch other errors (different than 400) as well?

dominikx96 (changes requested):

sites/partners/src/listings/PaperListingForm/index.tsx
emilyjablonski: We are catching all errors but showing field-specific errors on 400

sites/partners/src/listings/PaperListingForm/index.tsx
emilyjablonski: Good idea :)

dominikx96:
@emilyjablonski I think the error message should be different than generic:
https://github.com/bloom-housing/bloom/issues/1542

"Please resolve any errors before publishing your listing."

dominikx96 (changes requested):
@emilyjablonski

**1. I think the error message should be different than generic:**
https://github.com/bloom-housing/bloom/issues/1542
"Please resolve any errors before publishing your listing."

**2. Noticed a problem when I trigger errors and then fill the form with correct values:**
See video: https://share.getcloudapp.com/NQuYXn1R
So the correct jurisdiction value has been removed and I still see an error.

**3. When I put a user agent phone, then the phone should be validated using phone regex**
Currently, I am able to put a phone with the incorrect format: https://share.getcloudapp.com/o0uP01R0

**4. "Is the waitlist open?" field is not validated**
https://share.getcloudapp.com/d5uR5QgB

**5. Lottery fields**
When I choose the lottery option, I suppose date fields also should be required (?)
https://share.getcloudapp.com/nOuvO476

**6. Add unit**
When I tried to add a new unit type (I've just selected unit type "studio" from the select field, other fields were empty), I've noticed the error: https://share.getcloudapp.com/wbubgOgw

**7. Edit and delete unit don't work**
Nothing happened after click
https://share.getcloudapp.com/E0ujvN1B

**8. Building address validation**
https://share.getcloudapp.com/NQuYXn84

When I filled only street address, other errors have gone:
https://share.getcloudapp.com/9ZuQp65e

**9. Application types validation doesn't work:**
https://share.getcloudapp.com/bLuqL21x

emilyjablonski:
@dominikx96 

Everything is either fixed or waiting on a product update (1, 6) or a followup ticket (6) :)

1. I’ll message Kathy/Jesse for a copy update!
2. This was an existing problem with the Jurisdiction select before this PR. It would occasionally just….not select? But I fixed it by removing the Controller.
3. We actually don’t have any existing backend validations for phone number, elsewhere in our apps where we validate phone number fields we must be validating on the frontend. I added a backend phone number validator for this field though so it works now!
4. Oops! Missed this because we assign it “false” on submit once all other fields are validated. I removed this default value so we can tell if they actually clicked the field yes/no.
5. They shouldn’t be, we actually made a fix a few weeks ago so that they wouldn’t be required because apparently often leasing agents won’t know the lottery date yet on listing publish, but thanks for asking!
6. My bad, didn’t expect anything in units to be validating yet, and I always submitted units with an AMI chart! You can now submit a unit without an AMI chart. Unclear if AMI on every unit should be required on publish of a listing. I’m still waiting to hear back from Kathy/Jesse on any info concerning what should be required in the unit form, and I think ideally that’s handled in a separate ticket since we’ll be doing a sort of pre-publish backend call to validate.
7. Similar to 6. Edit and delete now work.
8. The screenshot was expected behavior and I mentioned this a little in the PR description, but I agree w your feedback that it’s weird and I refactored it a little so now I think it probably meets what you would expect.
9. Same as 4. Fixed!

Ty again for this ✨  amazing ✨ review! :D

emilyjablonski:
Everything is broken after the rebase 😭 Will be a bit, need to work in the new publish modal

dominikx96:
@emilyjablonski 

1. I still have a problem with the phone number, it's validated, but then I trigger an error (e.g. phone field is empty) - then fill it correctly, I'm still facing the error:
https://share.getcloudapp.com/v1u07GWO

2. When I'm trying to save as a draft, I see building address errors (I think they shouldn't be required in the draft state):
https://share.getcloudapp.com/L1urAldZ

3. When I'm editing a listing and triggering publish action (with some issues, empty fields), then listing unit actions (edit, delete) don't work.

Otherwise looks good to me, it's a large amount of great work, @emilyjablonski  ! 💪

dominikx96 (changes requested):
@emilyjablonski 

1. I still have a problem with the phone number, it's validated, but then I trigger an error (e.g. phone field is empty) - then fill it correctly, I'm still facing the error:
https://share.getcloudapp.com/v1u07GWO

2. When I'm trying to save as a draft, I see building address errors (I think they shouldn't be required in the draft state):
https://share.getcloudapp.com/L1urAldZ

3. When I'm editing a listing and triggering publish action (with some issues, empty fields), then listing unit actions (edit, delete) don't work.

Otherwise looks good to me, it's a large amount of great work, @emilyjablonski  ! 💪

emilyjablonski:
@dominikx96 Thank you again 🙏 
1) I can't repro this :( Will slack you
2) Fixed! Introduced by latest rebase, hadn't seen that
3) Discovered there's even more issues beyond that with units while testing this. The units will straight up just disappear sometimes while editing. Testing / fixing this now.

emilyjablonski:
Okay the unit issues is fixed! And as far as the phone number issue, the number 444-444-4444 fails (and it does too in the validator's official demo so it is supposed to http://libphonenumber.appspot.com/phonenumberparser?number=4444444444&country=US) and I believe this is because it is validating that the phone number is a valid US phone number down the area code, and 444 isn't a valid US area code. If I put in for example my phone number, it works. I assume we want to restrict to US phone numbers?

dominikx96:
@emilyjablonski you're right, I tried with the US number and it works better than I expected! :D 

I noticed one thing, building address probably should be optional for draft, but I still see errors:
https://share.getcloudapp.com/6quYgoZ7

emilyjablonski:
@dominikx96 I can't repro that, will slack you

dominikx96:
> @dominikx96 I can't repro that will slack you

As we discussed at the meeting, it was my fault, I forgot to fetch changes, LGTM! 👍

backend/core/src/shared/url-helper.ts
```diff
@@ -23,6 +23,10 @@ export const formatUrlSlug = (input: string): string => {
...
   const { name } = listing
 
-  const { city, street, state } = listing?.property?.buildingAddress
-  return formatUrlSlug([name, street, city, state].join(" "))
+  if (listing?.property?.buildingAddress) {
+    const { city, street, state } = listing?.property?.buildingAddress
+    return formatUrlSlug([name, street, city, state].join(" "))
+  } else {
+    return ""
+  }
```
seanmalbert, line 31: Should we at least return the name part of the slug?

sites/partners/lib/helpers.ts
```diff
@@ -236,3 +236,21 @@ export function formatIncome(value: number, currentType: IncomePeriod, returnTyp
...
+    }
+    if (
+      typeof obj[key] === "object" &&
+      !Array.isArray(obj[key]) &&
+      Object.keys(obj[key]).length === 0
+    ) {
+      delete obj[key]
+    }
+  })
+}
```
seanmalbert, line 256: Can you create unit tests for this, or at least spin up an issue in next sprint to create unit tests for this function.

sites/partners/src/listings/PaperListingDetails/sections/DetailListingPhoto.tsx
```diff
@@ -18,9 +18,12 @@ const DetailListingPhoto = () => {
...
     listingFormPhoto = listing.assets.find((asset) => asset.label == "building")
   }
-  const listingPhotoUrl = /https?:\/\//.exec(listingFormPhoto.fileId)
-    ? listingFormPhoto.fileId
-    : cloudinaryUrlFromId(listingFormPhoto.fileId)
+
+  const listingPhotoUrl = listingFormPhoto
+    ? /https?:\/\//.exec(listingFormPhoto.fileId)
+      ? listingFormPhoto.fileId
+      : cloudinaryUrlFromId(listingFormPhoto.fileId)
```
seanmalbert, line 25: I know this was already here, but regex.test might be better used here instead of exec.

sites/partners/src/listings/PaperListingDetails/sections/helpers.ts
```diff
@@ -16,3 +16,23 @@ export const getDetailFieldDate = (listingDate: Date) => {
 export const getDetailFieldTime = (listingTime: Date) => {
   return listingTime ? moment(new Date(listingTime)).format("hh:mm:ss A") : t("t.none")
 }
+
+export const getReadableErrorMessage = (errorMessage: string | undefined) => {
```
seanmalbert, line 20: can you also create unit tests, or an issue to create them for this function too.

sites/partners/src/listings/PaperListingForm/sections/BuildingDetails.tsx
```diff
@@ -147,7 +152,12 @@ const BuildingDetails = ({
             label={t("application.contact.streetAddress")}
             name={"buildingAddress.street"}
             id={"buildingAddress.street"}
+            error={errors?.buildingAddress?.street !== undefined}
+            errorMessage={errors?.buildingAddress?.street?.message}
```
seanmalbert, line 156: Wondering if it would be useful to have a function to handle these that would accept the path like "buildingAddress.street", so we don't have to do the same checks everywhere. Something like `hasError: (path: string) => boolean` and `errorMessage: (path: string) => string`

seanmalbert (commented):
@emilyjablonski ,
Took an initial pass and added some comments. Going to finish review in the morning.

sites/partners/src/listings/PaperListingDetails/sections/helpers.ts
emilyjablonski: We don't have unit tests in partners rn but I will create an issue

sites/partners/lib/helpers.ts
emilyjablonski: We don't have unit tests in partners rn but I will create an issue

jaredcwhite (changes requested):
Nice work @emilyjablonski! I think my main comment would be that if `src/listings/PaperListingForm/index.tsx` is getting changed to this degree, it'd be a great opportunity to refactor. The file's getting pretty large (~800 lines) and feels like it needs more of an architecture. What do you think?

emilyjablonski:
@jaredcwhite I think that's an excellent idea that given the already large size and complexity of this PR should be handled in a separate issue

seanmalbert (approved):
Looks good @emilyjablonski . I created a new issue off of @jaredcwhite's comment on refactoring the main listing form.

Issue #1780, opened 2021-08-31
Cache Revalidation fallback and Open Application Date for Publishing


seanmalbert:
We want to make sure cache revalidation is working and that it cannot fallback to an error state, so we can re-enable on alameda production. In addition to this, we want to Add application open date/time to the form, so that if that datetime exists, and the user publishes, that the cache is cleared, but on the frontend the listing doesn't show until the the open datetime is >= to now.

seanmalbert:
This was never disabled from dev, so on https://ala.bloom.exygy.dev/ and https://partners.housingbayarea.bloom.exygy.dev/, you should be able to edit listings and see the changes, including status updates reflected on the public side within 60 seconds.

kathyccheng:
Pushed live, closing

Issue #1618, opened 2021-08-05
Public Site - Fail Deployments if Error on Builds


seanmalbert:
We have NextJS specific logic with getStaticProps and getStaticPaths which call the backend. These are implementing try catches which are preventing errors from stopping the build. We want to keep the try catches in place, so we can log the errors and we want to throw the error. This way the deployment fails and we reduce the risk of publishing a bad deployment.

PR #1861, opened 2021-09-21
1780/Cleanup getStaticProps and getStaticPaths

seanmalbert:
## Issue

Addresses #1780 

## Description

The cache revalidation piece doesn't actually change here, the main changes are is so that errors aren't swallowed on builds and fails them (should cover #1618) and I removed the export script, since it isn't valid with [fallback: true](https://nextjs.org/docs/advanced-features/static-html-export#caveats). So we'll have to change the build command to replace `export` with `start`.

This doesn't address the second part of #1780, which is to add applicationOpenDate to the form and check against that before displaying, since that feels separate enough to warrant another PR.

## Type of change

I tested locally by starting the back end, then building and starting public. If you have partners started, you should be able to add/edit a listing and after `CACHE_REVALIDATE` (set in your .env) expires, you should see the update after refreshing. You should also be able to stop the backend and continue to see listings and go to detail pages.

- [x] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


sites/public/pages/index.tsx
```diff
@@ -63,14 +56,7 @@ export default function Home(props: IndexProps) {
...
         </AlertBox>
       )}
-      <Hero
-        title={heroTitle}
-        buttonTitle={t("welcome.seeRentalListings")}
-        buttonLink="/listings"
-        allApplicationsClosed={
-          !props.listings.some(listingOpen) && !props.listings.some(openDateState)
-        }
-      />
```
seanmalbert, line 73: This piece has been updated on Alameda for some time now, since the current logic doesn't account for listings with no application due date. I'm not sure we really need this check here, and it gives us one less route where we're querying listings.

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ All notable changes to this project will be documented in this file. The format
 
   - Upgrade the public and partners sites to Next v11 and React v17 ([#1793](https://github.com/bloom-housing/bloom/pull/1793)) (Jared White)
     - **Breaking Change**
+  - The cache revalidation piece doesn't actually change here, the main changes are is so that errors aren't swallowed on builds and fails them (should cover #1618) and I removed the export script, since it isn't valid with [fallback: true](https://nextjs.org/docs/advanced-features/static-html-export#caveats). So we'll have to change the build command to replace `export` with `start`. ([#1861](https://github.com/bloom-housing/bloom/pull/1861))
```
emilyjablonski, line 49: `are is so`

emilyjablonski:
I'm a little confused as to why this one is titled cache revalidation / what is changing there? Maybe just the title needs to be changed?

seanmalbert:
> I'm a little confused as to why this one is titled cache revalidation / what is changing there? Maybe just the title needs to be changed?

Updated title to be more appropriate for what changed.

seanmalbert:
@emilyjablonski , updated change log and changed PR title.

emilyjablonski:
LGTM apart from the build issues, not sure if they were going to be resolved pre- or post-merge

seanmalbert:
> LGTM apart from the build issues, not sure if they were going to be resolved pre- or post-merge

So the reason why the public build fails on this branch is because the build actually depends on the backend server. If you look at passing public builds, you'll see errors that are swallowed by the try catches. So I'll need to think about the best way to handle this that doesn't require adding the catch back in.

PR #1874, opened 2021-09-23
fix: update alameda's notification sign up URL

emilyjablonski:
Alameda has a new notification sign-up URL. No issue, came out of a meeting with their team.


jaredcwhite:
Wouldn't we want to put any jurisdiction-specific data in `housingbayarea` repo?

emilyjablonski:
@jaredcwhite We're storing jurisdiction-specific data in the DB, so with our current dev flow, I don't think so

Issue #1467, opened 2021-07-02
Mobile menu coming from the top instead of the side


akegan:
We just did a visual design review of our react homepage ahead of launch, and the design team found some design issues that they marked as blockers for deploy.

They had a couple of different pieces of feedback about the menu:

## 1. Different menus across different pages could be confusing
They noted that since we're going to be in a mixed state of angular and react pages for some time, that it might be confusing for users to have two different mobile menu experiences on different pages.

Did you consider keeping the same style of menu we had on DAHLIA where it came out from the side as a drawer? Were there technical limitations that prevented you from implementing it? Would you be open to switching to a drawer component that's more similar to what we have for DAHLIA if we contribute it?
![Mobile menu](https://user-images.githubusercontent.com/11825994/124334564-4fef1c80-db4c-11eb-9659-fe487470f85a.png)

## 2. Mobile menu pushes down content instead of sitting on top of it

(see image above). One designer noticed that the menu pushes down the content instead of unfurling on top of it and thought it was a bit strange. Was that an intentional choice? or a technical limitation?

cc @emilyshore

akegan:
Could add an option to menubar to make it a drawer instead. Could be related to https://github.com/bloom-housing/bloom/issues/1345. 

We might also need to look into a11y on the mobile navbar to make sure that the focus works as expected.

akegan:
Next step is to write a unified story about desktop/mobile navbar and layout options.

slowbot:
@akegan Starting to sketch the component and options for nav https://www.figma.com/file/tUtgsJO8XnF9Ij2dwu22EF/bloom-general?node-id=387%3A768

Issue #1345, opened 2021-06-09
Consider removing Bulma for navbar, writing our own


emilyjablonski:
Due to some new feature requests in the navbar, maintaining our use of Bulma has become more challenging.

There are changes we want to make that Bulma does not easily support (like custom logo widths and image-only logos) that are possible for us now but require quite a bit of CSS overriding to remove their default styles.

I think it would be cleaner and easier in the long run to remove Bulma and write our own nav bar so we have full control and customization.

jaredcwhite:
Sounds good to me. We added Bulma originally for a quick leg up, but it's clearer by now what sorts of features we'll want and the best way to achieve them.

emilyjablonski:
Can you re-assign me issues when they're kicked back to QA? Otherwise they really easily get lost :) Thanks!

slowbot:
@seanmalbert @emilyjablonski Lets talk about the process. Not sure if its best to reassign or put it at the top of the stack for whomever to pick up. I try to at least comment the original engineer but see I forgot to do that here.

emilyjablonski:
QA: All three of the above visual issues

PR #1885, opened 2021-09-27
refactor: re-write header menu bar, remove bulma

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1345 #1467

## Description

We had been using Bulma as a dependency for our Site Header. We had four main issues with this, which were that customizing the header bar was becoming increasingly difficult as we had to override their styles and behavior, the Menu text was untranslated, some of the styles were just a little visually off, and the dropdown menu was not accessible on mobile.

This PR re-writes our SiteHeader from mostly scratch. There are new props to toggle certain visual choices on/off. I also refactored how we are sending in languages and the home page link to remove business logic. Also fixed a bug where header links were still showing in partners when signed-out.

## Type of change

Check out the new stories, however please note that Storybook is not accurately displaying the mobile view in its default viewing window. Would recommend starting locally or the deploy preview to view mobile.

In Bloom we are not flattening mobile menus, using the mobile menu button, using the mobile dropdown, and not showing the feedback bar on mobile.
In DAHLIA, they will be flattening mobile menus, using mobile text as a button, using the mobile drawer, and showing the feedback bar on mobile.

These are all toggleable features through props.

I would test:
* Accessibility in desktop and mobile (can you navigate through all things in the header and dropdowns using your keyboard, are they all focusable and actionable on enter)
* Changing the window sizes to view the transition between desktop and mobile
* Comparing visually this PR to https://dev-bloom.netlify.app/  and https://partners.bloom.exygy.dev/ - there are some styling differences which are purposeful based off of the designs here https://www.figma.com/file/tUtgsJO8XnF9Ij2dwu22EF/bloom-general?node-id=387%3A768

## Checklist:


akegan:
@emilyjablonski do you have a sense of why the menus aren't working as expected on storybook?

ui-components/src/headers/SiteHeader.stories.tsx
```diff
@@ -1,171 +1,339 @@
...
-    <a href="#" className="navbar-item">
-      Hello World
-    </a>
-  </SiteHeader>
+    menuLinks={[]}
+  />
 )
 
-export const standardWide = () => (
+export const withLanguageAndNotice = () => (
```
akegan, line 44: would it be possible to test withLanguageAndNotice with the notice on mobile view too? I realize that makes a lot of variations though. Would it make sense to set up knobs for this component because there are so many props?

emilyjablonski:
@akegan Not 100% sure, but Storybook adds extra padding around the rendered components that looks like it might be the issue. If you test it in the mobile flavor by just making the window smaller, then it behaves as it does locally, but opening the console and setting the view to mobile gives me different results. I don't know of a great way to pre-release it with the DAHLIA set of props turned on but we do have the [Netlify deployment of the PR](https://deploy-preview-1885--dev-bloom.netlify.app/) of Bloom. I could start it for you locally if the smaller Storybook windows isn't enough QA.

akegan:
LGTM, just with one note. The close button for the mobile menu with our style needs an aria label 
image

emilyjablonski:
@akegan We have a ticket to set up the same a11y linting so hopefully that will help!

akegan (approved):
LGTM for DAHLIA!

slowbot (changes requested):
1. It seems we could use some padding around the text based logo on mobile
2. Can we add full height for mobile drawer menu
3. Can we use rems instead of px whenever possible?

emilyjablonski:
@slowbot we can add padding for sure! and the mobile drawer is full height there's just no left border and nothing for it to cover up (check dahlia prod for an example)

emilyjablonski:
@slowbot For sure but I just did an overview of every case where I'm using px in the scss file and I can't change any of them to rem bc they need to be exact numbers for positioning

slowbot (approved):
LGTM

jaredcwhite (changes requested):
I'd like to discuss how to adjust all the navbar links so they're `a` tags and remove most of the click handlers and Next router statements. Unless I'm missing something, most of the links are indeed hrefs to other pages, so it's not semantic to use `button` and click handlers to navigate to a new page. We can also make use of `NavigationContext`/`LinkComponent` to ensure proper routing using the current Next locale.

jaredcwhite:
Looking at the public website:
Screen Shot 2021-10-05 at 3 34 38 PM

Semantically I would expect these all to be `a` tags, unless it's something that only performs an action in-page (aka no `href`), in which case it should be a `button` tag just styled to look like a link.

The generated HTML seems to be just `div`/`span` tags which wouldn't behave like a link or a button:

Screen Shot 2021-10-05 at 3 34 27 PM

emilyjablonski:
@jaredcwhite I used buttons purposefully because I think they're more accessible but not 100% sure there, and I removed the link component / next stuff purposefully because we needed to remove Bloom implementation-specific bits from this component but down for other ways to accomplish that

emilyjablonski:
@jaredcwhite Okidokie I made some updates. I'm not able to blanket use LinkComponent because some of the "links" in the SiteHeader actually just have onClicks and no href, so I'm toggling LinkComponent or button on and off based on what was passed in, and there are no more vanilla a tags. I had to modify NavigationContext to get it to fully work which I think was the whole initial issue DAHLIA was having with localization. We need to spread on all props in order to pass in some of what we need for the SiteHeader and for consumers to localize. 

The LinkComponent is my biggest remaining area of concern still when it comes to making consumers use the NavigationContext and LinkComponent, because if DAHLIA (or any other consumer) sets up their LinkComponent and its props incorrectly, any component we have that uses LinkComponent won't even build for them, so we're kind of relying on them to set theirs up and I still think we should have them pass in a pre-localized href. But I think if we want to do that at all that should be a separate issue considering the priority of this PR so it doesn't matter atm anyway :)

jaredcwhite (approved):
@emilyjablonski Approach looks good. Let's continue to discuss ways we can make the link component injection easier to understand for consumers after this PR is merged.

One thing I noticed (shouldn't hold up the merge): the classes in the HTML have extra `undefined`…probably just missing array values somewhere in the component chain that are getting mapped to the string as undefined.

image

PR #1888, opened 2021-09-28
test: fix broken a11y test, stories

emilyjablonski:
No issue, I accidentally merged a broken a11y test into dev. The actual a11y fix is in ListingCard (the issue was headers increasing by more than one numeric value). The other changes are stories that were / have been broken by the register prop becoming typed.


dominikx96 (approved):
@emilyjablonski LGTM! 👍

PR #1920, opened 2021-10-04
fix: fix dev seeds with new priority types

emilyjablonski:
No issues, names of priority types changed so updated the seeds so we can re-seed again!


seanmalbert (approved):
Looks good!

Issue #1401, opened 2021-06-23
New or updated ui-component for DAHLIA listing dir pricing tables


akegan:
[Our Jira ticket](https://sfgovdt.jira.com/browse/DAH-829)

On DAHLIA, we have an updated table style 
In a meeting a few weeks ago, Jesse suggested maybe updating the  “minimal” table design to accommodate this rather than adding a new component.

Here's what it looks like on the [angular version of DAHLIA](https://housing.sfgov.org/listings/for-rent)
Mobile view:
![image](https://user-images.githubusercontent.com/11825994/123144091-275e7880-d410-11eb-9716-65df04c79f28.png)
Desktop view:
![image](https://user-images.githubusercontent.com/11825994/123144111-2f1e1d00-d410-11eb-9249-95a437f51e78.png)

Differences to highlight:

- White with lines instead of blue and alternating
- Subtitles for per month, etc
- Different component/format for mobile (in angular we actually have totally different markup for mobile to reduce css magic)

Questions for Bloom team:

1. Should we create this component by extending the "minimal" table component?
2. Are folks up for having a component that has separate markup for mobile and desktop?

slowbot:
@akegan 
1. I think @jaredcwhite may have suggested that, but after looking at the code I think this table style deserves its own component
2. Our preference would be to try and have one component that switches see https://designsystem.digital.gov/components/table/

emilyjablonski:
This isn't used anywhere in Bloom so Storybook under StackedTable is what I would check for QA

PR #1937, opened 2021-10-05
feat: responsive table refactor

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1926 #1401

## Description

The responsive table was not exported, wasn't meeting the AC, and was going to be very fragile and difficult to uptake, so I refactored it here and renamed it StackedTable. This flavor will be used on DAHLIA (https://github.com/bloom-housing/bloom/issues/1401) and is required for their listing directory React page.

## Type of change

Ensure public site still looks as it should / no regressions.

Check out the new stories under StackedTable and ListingCard checking the mobile view as well.

## Checklist:

### DIFF
--- a/ui-components/__tests__/page_components/ListingCard.test.tsx
+++ b/ui-components/__tests__/page_components/ListingCard.test.tsx
@@ -18,11 +18,11 @@ describe("<ListingCard>", () => {
         }}
         tableProps={{
           headers: {
-            unitType: "unit type",
-            minimumIncome: "minimum income",
-            rent: "rent",
+            unitType: "t.unitType",
+            minimumIncome: "t.incomeRange",
+            rent: "t.rent",
           },
-          data: [{ data: [{ unitType: "cellA", minimumIncome: "cellB", rent: "cellC" }] }],
+          data: [{ unitType: "cellA", minimumIncome: "cellB", rent: "cellC" }],
           responsiveCollapse: true,
           cellClassName: "px-5 py-3",
         }}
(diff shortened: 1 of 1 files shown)


ui-components/src/page_components/listing/ListingCard.tsx
```diff
@@ -47,7 +52,15 @@ const ListingCard = (props: ListingCardProps) => {
           </h4>
         )}
         <div className="listings-row_table">
-          {tableProps.data && <GroupedTable {...tableProps} />}
+          {(tableProps.data || tableProps.stackedData) && (
+            <>
+              {tableHeaderProps?.stackedTable ? (
+                <StackedTable {...(tableProps as StackedTableProps)} />
+              ) : (
+                <StandardTable {...(tableProps as StandardTableProps)} />
```
jaredcwhite, line 60: I think we'll still want to use `GroupedTable` unless the stacked mode is requested, otherwise we'll lose the special display of unit categories (reserved for seniors, etc.) like so:

image

jaredcwhite (changes requested):
Looks good to me! Just saw the one issue about moving away from the grouped table on listing cards.

ui-components/src/page_components/listing/ListingCard.tsx
emilyjablonski: I believe we purposefully removed this from the UI a while ago @slowbot is this the case?

akegan (approved):
Looks great to me! Much simpler to use than the previous version too.

ui-components/src/page_components/listing/ListingCard.tsx
emilyjablonski: @jaredcwhite Jesse and I connected offline, the orange bar for reserved is indeed deprecated :) We were still using the GroupedTable but we had stopped passing in group units a while ago. The reasoning is that we will only ever use the Reserved indication as a tag above the table and not within the table.

jaredcwhite (approved):
LGTM!

Issue #1837, opened 2021-09-16
Preferences - Updates for Jurisdictions (and general cleanup)


seanmalbert:
Preferences are a little bit of a mess right now. Each preference is tied to a listing, so to create the preference list for the listing management form, we get all of the preferences and then make a unique list.

To do this properly we need to change up the relationships between listings and preferences and add in jurisdictions. We should have a preference list that is independent of listings. These preferences should be tied to jurisdictions by many-to-many. This way preferences can be shared across jurisdictions. This allows us to maintain a core list of preferences and enable them for different jurisdictions.

Listings has a one-to-many relationship with preferences. This will also need to change to a many-to-many.

The end result should be that I can query for a list of preferences, filtering by jurisdiction. I should then be able to take that list and assign any number of those preferences to a listing.

### QA Review Instructions
1. Go to https://dev-partners-bloom.netlify.app/
2. create or edit a listing
3. with no jurisdiction selected, go to add a preference and you should see a message "You must first select a jurisdiction"
4. select a jurisdiction (dev has been reseeded so all jurisdictions should have the same preferences to select from)
5. scroll down to programs and you should be able to select, order and delete preferences and save at any iteration
6. edit form and detail should display preferences as previously selected and saved
7. Go to https://dev-bloom.netlify.app/ and a listing for which you were modifying preferences
8. Go through the application process and at https://dev-bloom.netlify.app/applications/preferences/all, you should see the preferences on each page in the order you specified on the listing form in the partner portal
9. fill out the preferences and complete the application
10. Back in https://dev-partners-bloom.netlify.app/, go to the applications for the listing you applied to and find the application(s) you submitted and ensure that the preferences you selected in the application are showing correctly on the app detail and form

pbn4:
@seanmalbert 

> Each preference is tied to a listing, so to create the preference list for the listing management form, we get all of the preferences and then make a unique list.

Does that mean:
1. listing create and update actions should now take just the ID of the preference?
2. preference create and update actions should be performed through /preferences endpoint?
3. How do we want to handle existing "duplicate" preferences? Should we somehow collapse them into one? 
4. Should I assign them all by default to Alameda in a migration (how does that affect Detroit running this migration)?

seanmalbert:
1. yes
2. yes
[...]

PR #1947, opened 2021-10-06
Preferences cleanup

pbn4:
# Pull Request Template

## Issue

Addresses # (#1837 )

## Description

Introduce N-M Listing-Preference relation through a self managed (not TypeORM managed) intermediate entity ListingPreference, which now holds ordinal and page. Remove Preference entity entirely with an appropriate DB migration. 

@emilyjablonski @dominikx96 @jaredcwhite This is a breaking change for everything related with retrieving preferences through Listing model. 

Listing.preference data has been moved to Listing.listingPreferences[].preference. `Page` and `Ordinal` are moved from Preference to ListingPreference. Example:

```
{
  ...listing,
  listingPreferences: [
    {
      preference: {id: preferenceId, title: ...}
      ordinal: 1,
      page: 1
    },
    {
      preference: {id: preferenceId2, title: ...}
      ordinal: 2,
      page: 2
    }
  ]
}
```

## Type of change

Places preferences are used:
* Displayed on the individual listing view (public)
* Submitting an application (public)
* Viewing an application (partners)
* Editing an application (partners)
* Adding and editing preferences on a listing (partners)

## Checklist:


backend/core/src/preferences/preferences.service.ts
```diff
@@ -1,9 +1,62 @@
...
+import { assignDefined } from "../shared/assign-defined"
+import { addFilters } from "../shared/filter"
+import { PreferencesListQueryParams } from "./dto/preferences-list-query-params"
+import { PreferencesFilterParams } from "./dto/preferences-filter-params"
+import { jurisdictionFilterTypeToFieldMap } from "./dto/jurisdictionFilterTypeToFieldMap"
+
+export class PreferencesService {
+  constructor(@InjectRepository(Preference) private readonly repository: Repository<Preference>) {}
+
+  list(params?: PreferencesListQueryParams): Promise<Preference[]> {
```
emilyjablonski, line 17: How do I fetch preferences if a user has more than one jurisdiction?

seanmalbert (changes requested):
@pbn4 and @emilyjablonski ,
There are two pieces that seem off:

1. In partners, if I remove preferences and add them to a listing `listingPreference.page` becomes null, which means they won't show up on the application. We don't have a UI for setting the page of a selected preference, and until we do, we may need to keep a page column on `preferences` too, so we can set that on `listings_preferences` when adding to a listing. We should bring this up in stand-up, but it may be the best solution until we have an improved UI for adding preferences on a listing
2. This next one isn't a bug, but I want to call it out here along with item 1. When adding a preference to a listing and once it's added, there isn't any indication in partners that the preference will be hidden on the listing detail. Even once we have a UI for adding preferences, it would be nice to have an indicator on the LM form/detail that the preference will be hidden.

seanmalbert:
As discussed, we're going to keep pages for now and we need to update the LM form, so that the listingPreferences getting set are using the page value from the preference, so that preferences show on the application. In a separate issue, we are going to remove pages and update the application to split preferences across pages based off their ordinal.

emilyjablonski:
@pbn4 had some significant backend merge conflicts w the new jurisdiction language piece but i think i resolved them ok

seanmalbert:
removed the no longer needed test and updated others, since preferences are no longer on the same page.

Issue #1746, opened 2021-08-25
Figure out how to make CSV Export for applications more reusable 


kathyccheng:
* Most columns are the same listing by listing
* But the columns that are dynamic relate to the preferences 
* It's probably most consistent to have a column for each sub-option tied to a preference and the row values are yes or blank

@seanmalbert to think this through

PR #1955, opened 2021-10-07
1746/Update csv export

seanmalbert:
# Pull Request Template

## Issue

Addresses #1746

## Description

This updates the CSV to not rely on mappings or configurations and relies on the data passed in to create the CSV. For simplicity sake, I removed all of the arguments like the header and demographic settings. If we do indeed need those, I'll add them back in.

The goal here was to not have to maintain different csv format types so we could add any type of data and have it work. Because this removes all of the CSV formatting type pieces there was a fair amount of code removal.

## Type of change

Export a CSV on a listing that has applications and review.

- [x] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:


backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+  data: { header: string; mappedField: string; val: any }[]
+  inProgress: boolean
+  excludeKeys: string[]
+  mappedFields: { [key: string]: string }
+  constructor(private readonly csvEncoder: CsvEncoder) {
+    this.headers = {}
+    this.data = []
+    this.inProgress = false
+    this.excludeKeys = [
+      " id",
```
pbn4, line 22: Why there are whitespaces in excludedKeys?

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+      "appUrl",
```
pbn4, line 23: BTW I recently found out this nice method of typing strings like those using: https://stackoverflow.com/a/42516869/2429333

`propertyOf` method, so we make sure that if any of those strings is removed from the original class this part of the code will throw a type error too. Could you update this code to take advantage of that?

pbn4: Same comment for keys of this.mappedFields, fields referenced in `mapAddressFields`, `mapPrimaryApplicantFields` and other methods.

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+
+  private getExcludeKeysRegex() {
+    return new RegExp(this.excludeKeys.join("|"))
+  }
+
+  private addToExlucdedKeys(key) {
+    this.excludeKeys.push(key)
+  }
+
+  private parseApplication(key: string, val: any, header: string, index = 0): void {
```
pbn4, line 174: Method is called `parseApplication`, but the `val` is actually `any` so I think we should consider a better method name.

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+import { ApplicationDto } from "../applications/dto/application.dto"
+import { capitalizeFirstLetter } from "../libs/stringLib"
+
+const HEADER_COUNT_REG = new RegExp(/\s\(\d\)/)
 
 @Injectable()
 export class CsvBuilder {
-  constructor(private readonly csvEncoder: CsvEncoder) {}
+  headers: { [key: string]: number }
+  data: { header: string; mappedField: string; val: any }[]
```
pbn4, line 13: Is `data` keeping output content metadata? Maybe consider renaming it to `outputData` for readability

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+  private capAndSplit(str: string): string {
+    let newStr = capitalizeFirstLetter(str)
+    newStr = newStr.split(/(?=[A-Z])/).join(" ")
+    return newStr
+  }
+
+  private setHeader(header: string, index: number): [string, string?] {
+    let newKey = header
+    // we pass along the mapped key for reference, so we can more easily set data to the correct column
+    const test1 = this.mappedFields[header]
```
pbn4, line 133: Could you write an example in a comment here? It's hard to determine what is going to happen here and what is expected

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
-      } else {
-        const metadata: FormattingMetadata = metadataObj as FormattingMetadata
-        try {
-          outputRows.push(metadata.formatter(value))
-        } catch (e) {
-          outputRows.push("")
-        }
+    }
+
+    if (index !== 0) {
```
pbn4, line 145: What's the purpose of index in newKey and mappedStr?

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
-        if (md.size !== null) {
-          return
-        }
-        md.size = Math.max(
-          ...arr.map((item) => {
-            const value = this.retrieveValueByDiscriminator(item, md.discriminator)
-            if (!value || !Array.isArray(value)) {
-              return 0
+  public build(arr: any[], includeDemographics?: boolean): string {
+    if (this.inProgress === true) {
```
pbn4, line 225: I think this class should be stateless and request scoped so that we do not have to store local state in singleton instance and throw 429 if some other requests already uses that code. At least I don't see a strong reason for doing so.

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+          error: "Download Already in Progress",
+        },
+        HttpStatus.TOO_MANY_REQUESTS
+      )
+    }
+    this.setProgress(true)
+    let dataString = ""
+    try {
+      if (!includeDemographics) {
+        this.addToExlucdedKeys("demographics")
```
pbn4, line 238: typo in method name

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+      }
+      arr.forEach((row: ApplicationDto) => {
+        this.parseApplication("", row, "")
+        this.data.push({ header: "_endRow", mappedField: undefined, val: "_endRow" })
```
pbn4, line 242: What the purpose of _endRow?

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+        return !this.getExcludeKeysRegex().test(header) && mappedField
+      })
+
+      // initiate arrays to insert data
+      const rows = Array.from({ length: arr.length }, () => Array(headerArray.length))
+
+      headerArray = headerArray
+        .map((column) => {
+          let columnString = ""
+          const test1 = this.mappedFields[column.replace(HEADER_COUNT_REG, "")]
```
pbn4, line 259: This seems to be a duplicate logic of whats in line 133

backend/core/src/csv/csv-builder.service.ts
```diff
@@ -1,130 +1,330 @@
...
+          const test2 = this.mappedFields[column.split(" ")[0]]
+          let mappedField = test1 ?? test2
+          if (mappedField) {
+            if (!test1 && test2) {
+              const [, ...fields] = column.split(" ")
+              fields.forEach((field) => {
```
pbn4, line 265: Are fields actually tokens of a header? Maybe a better name would be `headerTokens`

backend/core/src/csv/csv-builder.service.ts
seanmalbert: because the keys consist of things like `householdMembers id`, but we do want `id`.

backend/core/src/csv/csv-builder.service.ts
seanmalbert: Now that you say this, CsvBuilder shouldn't care about the properties on these keys, and in this case should probably be defined in ApplicationCsvExporter

backend/core/src/csv/csv-builder.service.ts
seanmalbert: How about parseData?

backend/core/src/csv/csv-builder.service.ts
seanmalbert: It's technically not output data either, but it is the parsed data, so out how about `parsedData` to go along with `parseData`

backend/core/src/csv/csv-builder.service.ts
seanmalbert: It was in both spots, so still worked ;-)

backend/core/src/csv/csv-builder.service.ts
seanmalbert: _endRow is used below to determine when we're at the end of a row and to go to the next one. Farther down `rows` is defined as an array of arrays with all of the necessary slots created, according to the header columns and the length of `arr`, so then we can fill data rows according to their matching header.

Issue #1851, opened 2021-09-20
Automatically version and release packages on commit to master


emilyjablonski:
Blocked by adding in conventional commits

PR #1959, opened 2021-10-07
chore: add commands to allow for automatic versioning and release

emilyjablonski:
# Pull Request Template

## Issue

Addresses #1851

## Description

This change creates the scripts needed to automatically version and publish, but actually setting that up will be done through GitHub Actions. Lerna can only understand commits with the conventional standard, so only those will be used to version and be added to the changelog. It applies the version implied by the commit message to packages changed in that commit. It takes all commits since the last version as whole so if there are multiple breaking changes it will only increment by one major version.

From root, `version:all` will version all four packages based on the commits since the last version and then commit to dev with the version changes and changelog updates. It understand our packages relationships to each other, so if we patch bump ui-components it will also bring public and partners to that latest version and increment their version by a patch.

From root, `publish:all` will publish backend/core and ui-components, as we don't want to publish public and partners right now. It will check to see if any local versions are not on npm and publish if they are not.

From root, `version:prerelease:ui-components` will version ui-components as a pre-release version.

On every push to master, we will version and publish all.

On every push to dev, we will pre-release version ui-components and publish all.

## Type of change

- [x] This change is a refactor/addresses technical debt

I'd honestly recommend just seeing how this pans out or pairing with me to test.

It's a little tricky. **If you want to test this setup is important so you don't accidentally publish to npm!!** (Although it _probably_ wouldn't let you anyway for auth reasons unless you already have npm logged in through the terminal). 

If you want to test it locally you can with Verdaccio, a local package registry you spin up. You would need to create a branch off of this branch.

npm install -g verdaccio
npm set registry http://localhost:4873
verdaccio

And within lerna.json, change the public command to:
"publish": {
      "conventionalCommits": true,
      "registry": "http://localhost:4873",
      "message": "chore(release): publish"
    },

Then you can commit on your branch, run the version and publish commands, and see the packages appear in Verdaccio.

## Checklist:


seanmalbert (approved):
@emilyjablonski , I think this looks good to merge and see how it works out in the wild.

Issue #1912, opened 2021-10-01
Lowercase user and applicant email addresses


seanmalbert:
For users and applicants we want to lowercase the email addresses. This means two things:

1. We need migration script that lowercases all existing email address
2. When an email address is supplied by a user either for entry or for a an action like logging in, we want to lower case input (we can do this server side)

PR #1972, opened 2021-10-12
fix(backend): enforces lower casing of emails

YazeedLoonat:
## Issue
https://github.com/bloom-housing/bloom/issues/1912

Addresses #issue

## Description
Addresses Issue #1912  enforcing lower casing of incoming emails when:
new users are created, a look up of a user by email address happens, during user confirmation, and user log in

## Type of change
- [x] This change requires a SQL Script

Create a new user with some combination of upper and lower case letters in the email address
attempt to log in as that user but only input the email address in lower case (no upper case characters) you should be able to log in as if the cases matched

if you have DB access, peep the new record and verify the email stored in the db is lower cased

## Checklist:


backend/core/test/user/user.e2e-spec.ts
```diff
@@ -570,4 +570,27 @@ describe("Applications", () => {
...
+      middleName: "Mid",
+      lastName: "Last",
+      dob: new Date(),
+    }
+    const mockWelcome = jest.spyOn(testEmailService, "welcome")
+    const res = await supertest(app.getHttpServer())
+      .post(`/user`)
+      .set("jurisdictionName", "Alameda")
+      .send(userCreateDto)
+    expect(mockWelcome.mock.calls.length).toBe(1)
```
pbn4, line 590: Checking whether mockWelcome has been called is not needed here.

backend/core/src/migration/1633557587028-user-email-lower-case.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class userEmailLowerCase1633557587028 implements MigrationInterface {
+  name = "userEmailLowerCase1633557587028"
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
```
pbn4, line 6: @seanmalbert @YazeedLoonat  Do we only want to change login related email handling here?

We have more columns like %email% in our DB schema:
```
bloom=# SELECT t.table_schema, t.table_name, c.column_name fROM information_schema.tables t INNER JOIN information_schema.columns c on c.table_name = t.table_name and c.table_schema = t.table_schema WHERE c.column_name LIKE '%email%' and t.table_schema not in ('information_schema', 'pg_catalog') and t.table_type = 'BASE TABLE' order by t.table_schema;
 table_schema |    table_name     |     column_name
--------------+-------------------+---------------------
 public       | user_accounts     | email
 public       | listings          | leasing_agent_email
 public       | household_member  | no_email
 public       | applicant         | email_address
 public       | applicant         | no_email
 public       | alternate_contact | email_address
 public       | household_member  | email_address
```

Shouldn't we update those too? E.g. when detecting duplicates in applications (email rule set) by lower casing them we also prevent some problems with missing duplicates.

backend/core/test/user/user.e2e-spec.ts
```diff
@@ -570,4 +570,27 @@ describe("Applications", () => {
...
+    expect(res.body).toHaveProperty("id")
+    expect(res.body).not.toHaveProperty("passwordHash")
+    expect(res.body).toHaveProperty("email")
+    expect(res.body.email).toBe("testinglowercasing@lowercasing.com")
```
pbn4, line 594: I would be nice to add a login check here, to see if login in with lower cased email works, but that requires you do use admin credentials to confirm a user (set confirmed_at through `PUT /user/:id` endpoint or by setting confirmed_at with UserRepository.

YazeedLoonat:
Hey @pbn4 got this update per your comments, lemme know what you think

pbn4 (approved):

Issue #1949, opened 2021-10-06
Incorrect listing status


dominikx96:
I noticed a problem with the listing status if someone provide a past `application due date`. It occurs status change to `closed`, but it still should be draft.

PR #2015, opened 2021-10-18
fix: Incorrect listing status

dominikx96:
# Pull Request Template

## Issue

Addresses #1949

## Description

Fixes incorrect status (closed) if the application due date is before the current time even if the application is a draft.

## Type of change

- [ ] This change is a refactor/addresses technical debt

Try to create a new listing with a past due date ad try to save it as a draft. After saving, a status still should be `draft`;

## Checklist:

### DIFF
--- a/sites/partners/src/listings/Aside.tsx
+++ b/sites/partners/src/listings/Aside.tsx
@@ -63,7 +63,7 @@ const Aside = ({
     })()
 
     submitFormWithStatus(false, newStatus)
-  }, [getValues, listing.status, submitFormWithStatus])
+  }, [getValues, listing, submitFormWithStatus])
 
   const actions = useMemo(() => {
     const elements = []
@@ -252,8 +252,7 @@ const Aside = ({
 
     return elements
   }, [
-    listing.events,
-    listing.status,
+    listing,
     listingId,
     saveAndExit,
     showCloseListingModal,


seanmalbert (changes requested):
@dominikx96 , please see @emilyjablonski's comment

dominikx96:
> From what's in the issue this looks good but I cannot get back from the Closed state. So if I make a draft listing with a due date in the past this branch still keeps it as a draft which is good. Then I publish it and its closed which makes sense. Then I change the due date to the future and save but it is still closed. It shows up in the closed section in the website but its due date is in the future.

Hm when I click the "Publish" button, a listing is active again. This behavior is only when you click on the "Save and exit" button.

How do we want to resolve it? I think when a listing is closed, you should use `Publish` to change the status to `active` even if a `dueDate` is future. Maybe we should consider changing status to `pending` when:

- listing is closed
- user tries to use the `Save and exit` button

@emilyjablonski @seanmalbert

emilyjablonski:
@dominikx96 Just putting a note here from what we talked about in standup - when a listing is Closed and we change the due date to the future and hit save as draft, the status should be published

emilyjablonski:
Getting this on the listing/add page I think you just need an optional
Screen Shot 2021-10-21 at 6 39 45 PM

Issue #1951, opened 2021-10-06
Filter front end by jurisdiction 


kathyccheng:
As a visitor
I only want to see listings for the jurisdiction site that I am on
So that I don't get confused 

**Acceptance Criteria**
* When I'm on the listings page, then I will only see listings with the jurisdiction that I'm on 

**Notes**
* Needs to be added to filter keys type
* Jurisdiction is on the request
* Jurisdiction is also an env variable
* Query param filter

kathyccheng:
@seanmalbert this looks good on San Jose, let's add to Alameda + SMC and then push live when we're ready

PR #2027, opened 2021-10-19
feat: adds jurisdiction filtering to listings

seanmalbert:
# Pull Request Template

## Issue

Addresses #1951 

## Description

This adds the ability to filter listings by jurisdiction by way of a query param. The public site has been updated to filter by jurisdiction and e2e tests were added. I updated the seeds to specific the jurisdiction they belong to with all but one getting assigned to Alameda. This also adds a jurisdiction index to listings.

## Type of change

- [ ] This change is a refactor/addresses technical debt

After you have booted this locally, you can go into partners, edit or create a listing, and assign it to a jurisdiction other than the one you have specified in your public .env file (probably Alameda). You shouldn't see this listing show on public. Now change it to the one you have specified, and you should see it show (assuming it's published).

## Checklist:

### DIFF
--- a/backend/core/test/listings/listings.e2e-spec.ts
+++ b/backend/core/test/listings/listings.e2e-spec.ts
@@ -124,7 +124,6 @@ describe("Listings", () => {
 
   it("should return listings with matching Alameda jurisdiction", async () => {
     const jurisdictions = await jurisdictionsRepository.find()
-    expect(jurisdictions.length).toBe(4)
     const alameda = jurisdictions.find((jurisdiction) => jurisdiction.name === "Alameda")
     const queryParams = {
       limit: "all",
--- a/sites/public/lib/hooks.ts
+++ b/sites/public/lib/hooks.ts
@@ -115,7 +115,7 @@ export async function fetchBaseListingData() {
   return listingData
 }
 
-let jurisdiction: Jurisdiction
+let jurisdiction: Jurisdiction | null = null
 
 export async function fetchJurisdictionByName() {
   try {
@@ -128,9 +128,9 @@ export async function fetchJurisdictionByName() {
       `${process.env.backendApiBase}/jurisdictions/byName/${jurisdictionName}`
     )
     jurisdiction = jurisdictionRes?.data
-
-    return jurisdiction
   } catch (error) {
     console.log("error = ", error)
   }
+
+  return jurisdiction
 }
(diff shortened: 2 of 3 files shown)


backend/core/src/seeds/listings/listing-default-sanjose-seed.ts
```diff
@@ -0,0 +1,111 @@
...
+          items: [
+            {
+              percentOfAmi: 80,
+              householdSize: 1,
+              income: 777777,
+            },
+          ],
+        },
+      ],
+      name: "Test: Default, Two Preferences",
```
emilyjablonski, line 101: We should change the name of the seed

emilyjablonski (approved):
LGTM apart from seed name change which I also think could stay the way it is but I think is just more clear if it's "San Jose..."

YazeedLoonat (approved):
looks good to me

backend/core/test/listings/listings.e2e-spec.ts
```diff
@@ -107,12 +115,70 @@ describe("Listings", () => {
...
+      view: "base",
     }
     const query = qs.stringify(queryParams)
     const res = await supertest(app.getHttpServer()).get(`/listings?${query}`).expect(200)
     expect(res.body.items.length).toBeGreaterThanOrEqual(2)
   })
 
+  it("should return listings with matching Alameda jurisdiction", async () => {
+    const jurisdictions = await jurisdictionsRepository.find()
+    expect(jurisdictions.length).toBe(4)
```
pbn4, line 127: I don't think this check brings anything to the test other than breaking a test suite when we add a new jurisidiction to the seeds.

pbn4: Same comment for other tests that check the length.

sites/public/lib/hooks.ts
```diff
@@ -109,3 +114,23 @@ export async function fetchBaseListingData() {
...
+let jurisdiction: Jurisdiction
+
+export async function fetchJurisdictionByName() {
+  try {
+    if (jurisdiction) {
+      return jurisdiction
+    }
+
+    const jurisdictionName = process.env.jurisdictionName
+    const jurisdictionRes = await axios.get(
```
pbn4, line 127: Why not use an autogenerated client service here?

pbn4 (changes requested):

sites/public/lib/hooks.ts
seanmalbert: Because it actually feels like more work to do that, since we have to set an axios instance to use it anyway. We have this handled in AuthContext, but that's used for React hooks, so we need to setup a different more reusable function to make it easier. I'll create task to do this. Since we have shared-helpers, that would be a good place to add it.

Issue #1869, opened 2021-09-22
Application Program Display


kathyccheng:
As an applicant 
I want to be able to see the housing programs offered by this application 
So that I can answer the questions that I need to 

**Acceptance Criteria:** 
* If program questions exist, they should show up in the application in the order specified at the end of the household member section, after the ADA question 
* If a program question exists, it should show up in the application summary screen
https://www.figma.com/file/oU5mI1afPd9zjt9vNxcLFa/bloom-short-form?node-id=709%3A5171

**Flow:** 
https://www.figma.com/file/ERmjqfhokoKQ1Lh8FDzT8Q/Application?node-id=0%3A1

### QA Notes
1. Go to https://dev-partners-bloom.netlify.app/
2. Edit an existing listing or create a new one and select programs. Make sure the listing has a digital application.
3. Go to https://dev-bloom.netlify.app/ and find the listing you were working with in the partner portal
4. Apply to the listing and in the `/applications/household/programs` section, you should see the programs you selected for the listing appear each on their own page
5. complete the application and in the review portion, you should see the answers you selected for each program
6. Go back to https://dev-partners-bloom.netlify.app/ and find the application you submitted
7. In the detail and when you edit the form, you should see the programs you selected in the public application

kathyccheng:
@slowbot can you update the application flow diagram to include where this set goes?

emilyjablonski:
@jaredcwhite 
` title: "Veteran",
  subtitle: "Should your application be chosen, be prepared to provide supporting documentation.",
  description: "Have you or anyone in your household served in the US military?"`

` title: "Transition Age Youth",
  subtitle: "Should your application be chosen, be prepared to provide supporting documentation.",
  description:
    "Are you or anyone in your household a transition age youth (TAY) aging out of foster care?",`

`title: "Developmental Disability",
  subtitle: "Should your application be chosen, be prepared to provide supporting documentation.",
  description:
    "Do you or anyone in your household have a developmental disability or mental illness?",`

`title: "Housing Situation",
  subtitle: "Should your application be chosen, be prepared to provide supporting documentation.",
  description:
    "Thinking about the past 30 days, do either of these describe your housing situation?",`

kathyccheng:
@jaredcwhite I think we'd like the page to look like this with the question as the title. this might be a little different than how the preference pages are set up 

https://www.figma.com/file/oU5mI1afPd9zjt9vNxcLFa/bloom-short-form?node-id=593%3A4245

[...]

PR #2070, opened 2021-10-26
feat: two new common app questions - Household Changes and Household Student

emilyjablonski:
# Pull Request Template

## Issue Overview

This PR addresses #1735 

- [x] This change addresses the issue in full
- [ ] This change addresses only certain aspects of the issue
- [ ] This change is a dependency for another issue
- [ ] This change has a dependency from another issue

## Description

Two new common app questions, detailed in the issue. There are two new screens after ADA in the Household section - household changes and household student.

Overall adding a question took a lot of setup and we are storing the step order in a lot of places - we should refactor this in the future.

## How Can This Be Tested/Reviewed?

Submit an application. In the household section ensure the new sections are in the changes-->student order and that you can navigate forward and backward. Ensure they are on the summary screen before submission. Then in partners open the application and ensure it looks good on the detail and edit pages. Then export as CSV and ensure the column appears.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have reviewed the changes in a desktop view
- [x] I have reviewed the changes in a mobile view
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have assigned reviewers
- [x] I have run `yarn generate:client` and/or created a migration if I made backend changes that require them
- [ ] I have exported any new pieces added to ui-components
- [x] My commit message(s) is/are polished, and any breaking changes are indicated in the message and are well-described
- [x] Commits made across packages purposefully have the same commit message/version change, else are separated into different commits

## Reviewer Notes:

Steps to review a PR:

- Read and understand the issue
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Also review the acceptance criteria on the Netlify deploy preview (noting that these do not yet include any backend changes made in the PR)
- Either explicitly ask a clarifying question, request changes, or approve the PR if there are small remaining changes but the PR is otherwise good to go

## On Merge:

If you have one commit and message, squash. If you need each message to be applied, rebase and merge.

PR #2114, opened 2021-10-31
1869/application program display

jaredcwhite:
## Issue Overview

This PR addresses #1869

## Description

This adds application form steps to the public site which allow applicants to choose one or more programs with a yes/no answer, and shows the answers in the application summary page.

There are two available programs in the Coliseum listings, so you can try applying.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


jaredcwhite:
FYI, the failing Cypress tests were a problem on dev (same failing tests). The new tests I added for this PR are working.

jaredcwhite:
This PR touches some of the same areas in the code as https://github.com/bloom-housing/bloom/pull/2070 so depending on which merges in first, the other will likely need to merge conflicts resolved.

jaredcwhite:
@dominikx96 OK I updated the seeds based on updates in your PR, let me know if you have any questions.

ui-components/src/locales/general.json
```diff
@@ -378,6 +378,49 @@
...
+          "label": "Yes"
+        },
+        "doNotConsider": {
+          "label": "No"
+        }
+      },
+      "tay": {
+        "summary": "Transition Age Youth (TAY)",
+        "tay": {
+          "label": "Yes"
```
emilyjablonski, line 394: Ideally we don't have duplicate translation strings, is there a way to prevent that here?

jaredcwhite: Probably not for the "Yes" options since they could be keyed anything, but the No options are always keyed "doNotConsider" so I could DRY that up.

emilyjablonski: If for programs there are only ever two answers, couldn't on the frontend we always use the "yes" label for the first key?

jaredcwhite: I was hoping that would be the case, but the homeless one is an oddball, it's not Yes/No answers. 🤷‍♂️

emilyjablonski: Ahhh okay no worries

emilyjablonski:
Screen Shot 2021-11-15 at 11 46 51 AM
Screen Shot 2021-11-15 at 11 46 03 AM
Screen Shot 2021-11-15 at 11 45 10 AM

jaredcwhite:
@emilyjablonski Hey, so those screenshots look like an older version of the PR. (Shouldn't ever see "PROGRAM TITLE", checkboxes should be radio buttons, etc.) Can you double-check you've pulled down the latest commits?

emilyjablonski:
@jaredcwhite For sure that was a problem! Scratch the first two screenshots 😓  Still seeing the third issue which may or may not be unrelated?

jaredcwhite:
@emilyjablonski OK, good to hear about the latest fixes. On that third screenshot, it looks like the preferences page? Not the programs page? Though that preferences page should definitely be skipped there if there are no preferences in the data. I don't think I touched that though in this PR. Hmm…

emilyjablonski:
Cypress tests are passing now

sites/public/src/forms/applications/FormSummaryDetails.tsx
```diff
@@ -327,6 +327,19 @@ const FormSummaryDetails = ({
               </Fragment>
             ))}
           </ViewItem>
+          {application.programs
```
YazeedLoonat, line 330: I think it'd be better to switch this to using `.reduce` instead of `.filter` and then `.map`

YazeedLoonat (changes requested):
looks good, small code change request

dominikx96 (changes requested):
@jaredcwhite in the summary, one of the transitions seems broken

Screenshot 2021-11-17 at 17 56 56
:

shared-helpers/src/programHelpers.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (value == "none") {
+      claimed = false
+      program?.formMetadata?.options.forEach((option) => {
+        options.push({
+          key: option.key,
+          checked: false,
+          extraData: [],
+        })
+      })
+    } else {
```
seanmalbert, line 24: under what circumstances does value === "none"?

shared-helpers/src/programHelpers.ts
```diff
@@ -0,0 +1,58 @@
+import { ApplicationProgram, Program } from "@bloom-housing/backend-core/types"
+
+export const PROGRAMS_FORM_PATH = "application.programs"
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const mapProgramsToApi = (programs: Program[], data: Record<string, any>) => {
```
seanmalbert, line 6: In testing this with the seeded data, it appears that this function in practice always takes data with one key like `{tay: 'preferNotToSay'}` and returns an array with one element like `[{key: 'tay', claimed: true, options: Array(3)}]`.  This may be operating off an old model of having multiple programs/preferences per page, so that would make sense. Given that there is only one coming in and out of this function in reality, we can probably simplify this and perhaps code further upstream and downstream as well.

seanmalbert (changes requested):

ui-components/src/locales/general.json
```diff
@@ -378,6 +378,49 @@
...
+      },
+      "disabilityOrMentalIllness": {
+        "summary": "People with Developmental Disabilities or Mental Illness",
+        "disabilityOrMentalIllness": {
+          "label": "Yes"
+        },
+        "doNotConsider": {
+          "label": "No"
+        }
+      },
```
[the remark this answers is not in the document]
jaredcwhite: Hmm, where is this happening? I have some code that always pulls from `t.prefersNotToSay` for that particular key on any program.

sites/public/pages/applications/household/programs.tsx
```diff
@@ -0,0 +1,203 @@
...
+      const programsDataKeys = programsData.map((program) => program.key)
+      const currentPrograms = application.programs.filter((program) => {
+        return !programsDataKeys.includes(program.key)
+      })
+      programsData = [...currentPrograms, ...programsData]
+      conductor.currentStep.save(programsData)
+    } else {
+      // Otherwise, submit all at once
+      conductor.currentStep.save(programsData)
+    }
```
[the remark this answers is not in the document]
jaredcwhite: I think the comment on line 67 doesn't really communicate what this is. We always have to divide the total number of programs into different "pages" here since it's simulating Next's page-based navigation but within a single file. Each program ordinal (1, 2, 3) will then place that program on each of the "pages".

shared-helpers/src/programHelpers.ts
jaredcwhite: Whoops, I think this was left here by mistake after my latest refactor. I'll clean it up.

shared-helpers/src/programHelpers.ts
jaredcwhite: Good point, I can simplify that.

ui-components/src/locales/general.json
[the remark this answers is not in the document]
jaredcwhite: Oh this must be the issue that Dominik found. Will fix.

seanmalbert (approved):
Thanks for the updates, @jaredcwhite. This is looking good to me.

sites/public/src/forms/applications/FormSummaryDetails.tsx
YazeedLoonat: the end result of the maps is an array of arrays containing jsx which is the accumulated value

Ultimately this is an efficiency change though, so I'm okay with leaving this in place for now and if we run into efficiency concerns later we can make this change then

YazeedLoonat (approved):
LGTM

Issue #2116, opened 2021-11-01
Public Cypress Tests - Application Flow


emilyjablonski:
- Does every step render, and does the application order follow what we expect (somewhat existing)
- *** Can I submit an application from start to finish on a few different listing types, and then does the application data match on both the ending summary page and within the My Applications tab of user accounts (we can't check this from partners in Cypress) - we currently do not submit an application start to finish following all the steps sequentially, are using untyped mock data, and do not check that the data successfully was saved, so this is the biggest addition
- Do errors appear at every step if I try to continue without editing (somewhat existing)
- Does autofill work
- If I am signed in, does it tie the application to my account

I'm hoping to remove dependencies on brittle selectors like classes and text, and add data test ids where we need them.

PR #2118, opened 2021-11-02
test: refactor and add public site application flow cypress tests

emilyjablonski:
# Pull Request Template

## Issue Overview

This PR addresses #2116

## Description

Split up the issue into this PR (a refactor) and another (additional coverage) since this one is large. This PR does include two new tests though that go through the application end to end and submit it with one filling out every field and one filling out minimal fields.

The refactors here:
* Consistently named all our test files and test names
* Removed duplicate tests
* Removed tests testing legacy behavior
* Removed all untyped json data/fixtures, in favor of typed objects
* Removed any selectors that were based on css ids, classnames, dom elements, or english strings, as these are brittle and also implementation details which can cause false negatives - added data-test-ids for these (this test-id work can also enable us to test in multiple languages in the future if we want to)
* In each individual application step test, I moved out the pieces where we were checking what data each step is saving, since in order to do that we were mocking the current application data using session storage, and that approach to testing is deeply tied to implementation details. Now each individual step just tests ui pieces within that form step only.
* Added two longer tests that fills out an entire application and submits it, with all the methods reusable

CircleCI runtime is the same as before (between 8-25 minutes, usually around 15, not sure why it is so variable)

Remaining application tests for my follow-up PR:
* In the submission tests, confirm all the information appears correctly on the summary page pre-submit, and use the edit links to go back to form steps and edit the data
* In the submission tests, confirm all the information was submitted correctly by logging in before submitting, and viewing the application in the account tab
* Autofill

The diff is suuuuper wonky because of how much code was either removed by de-duplicating or just moved to a different file. I'd suggest pulling it down and checking out the cypress directory. The majority of the submission logic is in `commands`.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


emilyjablonski:
TODO for myself: double-check that there are no new console errors created by spreading the test-ids onto Buttons

jaredcwhite:
I'm not sure about the idea of sprinkling test data attributes across all of our HTML. It's additional visual noise both in the source as well as output. For elements which aren't form fields and don't yet have a DOM ID, can't we just use `id`?

`<a class="button" id="app-choose-language-sign-in-button" href="...">Sign In</a>`

vs.

`<a class="button" data-test-id="app-choose-language-sign-in-button" href="...">Sign In</a>`

And for form fields, they already have a unique name that's a product of the data schema. I don't think they're likely to change…and if they do, it means the underlying data schema has changed as well so I would expect tests to require updating. For example:

```
<PhoneField
    label={t("application.contact.yourPhoneNumber")}
    caps={true}
    required={true}
    id="applicant.phoneNumber"
    name="applicant.phoneNumber"
    placeholder={clientLoaded && noPhone ? t("t.none") : null}
    error={!noPhone ? errors.applicant?.phoneNumber : false}
    errorMessage={t("errors.phoneNumberError")}
    controlClassName="control"
     control={control}
     defaultValue={application.applicant.phoneNumber}
     disabled={clientLoaded && noPhone}
     dataTestId={"app-primary-phone-number"}
   />
```

The `app-primary-phone-number` test ID seems redundant because `applicant.phoneNumber` would uniquely identify the field and won't change unless the schema changes. Like if we changed this for some reason to "applicant.contactMeUsingThisPhoneNumber`, I wouldn't be surprised if that breaks the integration test and the test needs to be updated.

jaredcwhite (changes requested):
oops, forgot to go through the review button

jaredcwhite:
@emilyjablonski Thanks for the link, I read through it for the added context. I don't entirely agree with their reasoning. I can see the value of it for arbitrary UI elements perhaps, but in the case of form fields￼  we shouldn't think of them as brittle…the name attribute is very unlikely to change once its initially conceived. Let's open this up for a wider discussion and see what others think.￼

seanmalbert:
@emilyjablonski and @jaredcwhite , I agree that the names shouldn't be brittle and I also like the use of the test IDs. My preference would be to use the test Ids for consistency. It may also help us identify what isn't covered, by only adding the test ids to fields we're currently testing.

seanmalbert (changes requested):
@emilyjablonski ,
I merged the latest dev locally and resolved the conflicts. Everything work well up until the submission tests at the end. Can you please check on that? Other than that, I think this looks great!

YazeedLoonat:
@jaredcwhite @seanmalbert @emilyjablonski 
I'm in favor of switching to the test IDs as well. I would rather stick to cypress's recommendations where possibly unless we had a very strong reason for not doing so

If we took the case of renaming a column in the db though
with test-ids if we named them well we wouldn't have to go through and update every single integration test

without test-ids we would have to update them

I think those two reasons are enough to warrant a little bit more code written

Issue #2085, opened 2021-10-27
Back end error handling 


kathyccheng:
For the first pass on this we want the following to happen:
1) Similar to how `EntityNotFoundExceptionFilter` is attached we would like to attach a 2nd generic error filter. This filter can function simply as a `console.error` for now on the error to simplify tracking down any errors on the backend. 
Please try to keep this similar to best practices/standard usage of nestjs

2) Look into and add into this nestjs best practice/standard error handling strategies

Nestjs documentation on exception filtering: https://docs.nestjs.com/exception-filters 

https://github.com/bloom-housing/bloom/pull/2090 is a rough-hewn example of a generic error filter that we can use to build up

Next steps:
1) contemplate removing the current error filter and either rolling into the new generic one or getting rid of it entirely 

2) remove try-catch clauses in the backend so that errors can bubble into the the top level error filter (this is more similar to standard practice of nestjs)

3) If standard/best practice dictate how the error logger should be attached we should change our logger attachment to meet that

PR #2132, opened 2021-11-03
2085/adds top level catchAll exception filter

YazeedLoonat:
## Issue Overview

This PR addresses #issue
https://github.com/bloom-housing/bloom/issues/2085

## Description
This adds a top level exception filter where exceptions will bubble up to/through

This is to give a console.error when errors do happen so backend errors that happen throughout the various nextjs built in logic and decorators can be easier to track down

we can add more onto here later as this is a starting point. Next steps from here would be to go through and remove try-catch blocks throughout the backend so that this top level error handler can be fleshed out/function more in-line with nextjs best practice.

I rolled logic into here from the entity not found exception filter, but in a second pass we should remove the exception filter specific to it and possibly remove this custom logic all together if its not needed. 

Also the way we have our logger attached is in line with best practice so no changes will be needed there

In order to review this you can run the tests to verify that all tests passed locally

then to see it in action modify one of the backend e2e tests in a way that will error (such as modifying what should be a string into a number or a similar change)

run the test and see that the console.error fires letting you know what the error is and the stack trace for it 

outside of that this shouldn't affect any current logic/behavior so without triggering something we shouldn't see any real difference

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend/core/src/filters/catch-all.filter.ts
```diff
@@ -0,0 +1,15 @@
+import { Catch, ArgumentsHost } from "@nestjs/common"
+import { BaseExceptionFilter } from "@nestjs/core"
+
+@Catch()
+export class CatchAllFilter extends BaseExceptionFilter {
+  catch(exception: any, host: ArgumentsHost) {
+    console.error({ message: exception?.response?.message, stack: exception.stack, exception })
+    if (exception.name === "EntityNotFound") {
```
pbn4, line 8: I think it will never be triggered because as per NestJS docs:

> The @Catch(HttpException) decorator binds the required metadata to the exception filter, telling Nest that this particular filter is looking for exceptions of type HttpException and nothing else. The @Catch() decorator may take a single parameter, or a comma-separated list. This lets you set up the filter for several types of exceptions at once.

and 

```
@Catch(EntityNotFoundError)
export class EntityNotFoundExceptionFilter implements ExceptionFilter {
  catch(exception: EntityNotFoundError, host: ArgumentsHost) {
    const response = host.switchToHttp().getResponse()
    response.status(404).json({ message: exception.message })
  }
}
```

returns a response and by that short circuits filters chain, so CatchAllFilter will never be called when EntityNotFoundError is thrown. 

Could you rephrase that part: 

>I rolled logic into here from the entity not found exception filter, but in a second pass we should remove the exception filter specific to it and possibly remove this custom logic all together if its not needed.

What do you want to do with `EntityNotFoundExceptionFilter`?

YazeedLoonat: Hey @pbn4 

for `I think it will never be triggered because as per NestJS docs:` peep https://docs.nestjs.com/exception-filters#catch-everything

when you don't pass an exception list in it catches everything

for `so CatchAllFilter will never be called when EntityNotFoundError is thrown`
I thought so to, but weirdly there was an e2e test failing where the EntityNotFoundError was only getting caught in the new catch all filter 

also for `Could you rephrase that part:` sorry, I meant that with the catch all filter in place we no longer need that specific filter anymore so we can get rid of the specific filter for it code wise. I thought maybe in the future we could get rid of that entirely and let nest's default exception list cover it for us. 

Lemme know if you have any other questions or anything like that

pbn4: > for so CatchAllFilter will never be called when EntityNotFoundError is thrown
I thought so to, but weirdly there was an e2e test failing where the EntityNotFoundError was only getting caught in the new catch all filter

OK I see what the problem was, the `EntityNotFoundExceptionFilter` is never called because TypeORM does not throw a proper EntityNotFoundError class instance, it throws a JS object. A proper implementation of this filter should be 
```
@Catch(NotFoundException)
export class NotFoundExceptionFilter implements ExceptionFilter {
  catch(exception: NotFoundException, host: ArgumentsHost) {
    const response = host.switchToHttp().getResponse()
    response.status(404).json({ message: exception.message })
  }
}
```

because most of the services use findOne and check whether returned object is undefined and then throw NotFoundException. 

Anyway your approach is better so let's keep that, you can remove `EntityNotFoundExceptionFilter` completely in this PR.

pbn4 (approved):

seanmalbert (approved):

PR #2146, opened 2021-11-06
refactor: make backend a peer dependency in ui-components

emilyjablonski:
It has been very challenging to uptake the components on DAHLIA while they are still on node 12, and the ui-components work just fine on node 12 so I made the backend-core package a peer dependency of ui-components so that it can be uptaken on node 12.

Also fixed a broken cypress test.


jaredcwhite (approved):
Makes sense. :+1:

Issue #2080, opened 2021-10-27
Waitlist Fixes


kathyccheng:
Updated Acceptance Criteria: https://github.com/bloom-housing/bloom/issues/2080#issuecomment-959914977
QA instructions: https://github.com/bloom-housing/bloom/issues/2080#issuecomment-982165407

----

As a leasing agent, 
I might want to show information about the waitlist
So that I can give the full picture to the applicant for an available unit 
I do not want to tell applicants they can apply for spots on the waitlist yet
Since waitlist availability type is not currently supported 

Acceptance Criteria 
* When I am on the application section of Listing Mgmt, then I should see this waitlist question "Do you want to show a waitlist" instead of "Is the Waitlist Open" 
* When I choose Yes to "Do you want to show a waitlist" but No to "Do you want to show waitlist sizes", then the public listing page should show Waitlist is Open header with subtext "**Available Units and Waitlist** (header)
Once ranked applicants fill all available units, the remaining ranked applicants will be placed on a waitlist for those same units." 
* When I choose Yes to Do you want to show waitlist size, then I have to fill out Maximum Waitlist size instead of Current Waitlist Size 
* When I add a Current Waitlist Size, then I expect that to show up on the listing, but only if I added it 
* When I add a Spots open on the listing, then I expect that to show up on the listing but only if I added it 
* If yes, but no size, Show Waitlist is Open title with new text: After available units... 

Notes 
* Change question for first waitlist question 
* Make Maximum vs Current waitlist size required 
* User can input Max waitlist size + Current, or Max waitlist size + Spots Open or Max + Current + Spots Open 
* Make sure listing is only showing waitlist sizes based on what user is inputting, not any calculation

jaredcwhite:
@kathyccheng Are these just copy/display changes based on an unchanged existing data schema, or does this story require new/changed schema? I'm not entirely clear on that just reading the description. Thanks!

jaredcwhite:
> If yes, but no size, show Waitlist is Open title with new text: After available units...

I don't know what this means. Also it seems to contradict:

> When I choose Yes to "Do you want to show a waitlist" but No to "Do you want to show waitlist sizes", then the public listing page should show Waitlist is Open header with subtext "TBD"

kathyccheng:
![image](https://user-images.githubusercontent.com/3453407/140102840-06950a62-214e-4416-b42c-1bf5678dc2d6.png)

kathyccheng:
[...]

PR #2168, opened 2021-11-11
feat: simplify Waitlist component and use more flexible schema

jaredcwhite:
## Issue Overview

This PR addresses #2080 

## Description

The Waitlist component on the Listing sidebar has been simplified so it either displays or doesn't depending on the `isWaitlistOpen` boolean, and if it does display, it shows one, or, or three of the different waitlist values depending on how they were set.

In addition, the show waitlist sizes boolean in the partners LM form has been removed.

Examples of the different states are in Storybook under Listing Sidebar > Waitlist, and you can view different listings on the public site for concrete examples.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


jaredcwhite:
Can I get an early look at this @kathyccheng, as I'm not entirely sure I covered all the bases and want to double-check the new logic makes sense.

ui-components/src/locales/general.json
```diff
@@ -1063,7 +1064,7 @@
       "maxSize": "Maximum Waitlist Size",
       "maxSizeQuestion": "What is the maximum size of the waitlist?",
       "open": "Open Waitlist",
-      "openQuestion": "Is the waitlist open?",
```
emilyjablonski, line 1066: Just a reminder that if we are changing a string we need to remove the keys in the translations files so that we know the string needs to be updated / is now wrong (unless maybe this is just a partners string and then NBD!)

jaredcwhite: Good point…I searched around and this is only for partners, so we I think we're in the clear.

emilyjablonski (changes requested):
A couple issues I've noticed from the AC in the issue:
- If select yes to waitlist, leave all the next inputs blank, and have either no application methods or the referral method, the waitlist section does not show
- If there is a waitlist and available units we should be showing the number of available units and not the current waitlist size
- How did you test the state with a waitlist and no available units right now since we can't yet add units that aren't available but we will be able to do that soon?
- Current waitlist size when shown should not be bold
- The three extra inputs are not being removed from the data model if I add them and save and then select no to the waitlist question and save

jaredcwhite:
@emilyjablonski Thanks for looking through this in detail. So there was a progression of the AC over the life of the issue and I think the initial issue description never got updated. :( Can you take a look at https://github.com/bloom-housing/bloom/issues/2080#issuecomment-959914977 and let me know if that addresses some of the outstanding questions?

> * The three extra inputs are not being removed from the data model if I add them and save and then select no to the waitlist question and save

Def seems like a bug to me, I'll go investigate.

emilyjablonski:
@jaredcwhite Sounds good! And that is the comment I was using

jaredcwhite:
Hmm, I think you're right. I must admit I'm throughly confused about the AC at this point. I'll reach out to Kathy and see if we can draw up a new, canonical AC that makes sense to me. 😅

kathyccheng:
@emilyjablonski -- @jaredcwhite and I sync'ed so some responses to your notes

>> If there is a waitlist and available units we should be showing the number of available units and not the current waitlist size
* We're going to ignore this case for now. Based on current logic in this issue, I think Available Units will never show. And that's okay for now, will sync with Jesse on a path forward for available units. 

How did you test the state with a waitlist and no available units right now since we can't yet add units that aren't available but we will be able to do that soon?
* Units are right now either available or non-existent. We don't have to worry about units with any other "availability" type for now. But also based on the comment above, I think this is the only state below and the variance is just in showing 1, 2, 3 or none of the waitlist numbers. 
image

emilyjablonski:
@kathyccheng Sounds great! So I think we only need to worry about bullets 1 and 5 :)

YazeedLoonat (changes requested):
Looks good, just a few small code suggestions

YazeedLoonat (approved):
LGTM

emilyjablonski:
LGTM other than one remaining issue which is that nothing shows if the number is zero, ie max is 100, current is 0, open is 100 - the current line does not show

jaredcwhite:
So current would be the only item that should show if the data is 0? Or is that true for all of them? cc @slowbot @kathyccheng

kathyccheng:
Do we have a way to distinguish 0's from null? Ideally, if they enter 0 it should show. But if there's no entry, it would not show for all of the 3 values.

jaredcwhite:
@kathyccheng @emilyjablonski Think I got that working now, see the various examples here: https://deploy-preview-2168--dev-storybook-bloom.netlify.app/?path=/story/listing-sidebar-waitlist--open-with-some-zeroes

emilyjablonski:
Screen Shot 2021-11-23 at 1 37 14 PM
This is still not distinguishing between null and 0. Believe if the first two values are null we don't want to show the row at all.

jaredcwhite:
Back to the drawing board! 🤪

jaredcwhite:
OK, this should be good to go. I added more examples to the storybook.

ui-components/src/page_components/listing/listing_sidebar/Waitlist.tsx
```diff
@@ -1,77 +1,46 @@
...
   return (
     <section className="aside-block is-tinted">
-      <h4 className="text-caps-tiny">{header}</h4>
+      <h4 className="text-caps-tiny">{t("listings.waitlist.unitsAndWaitlist")}</h4>
       <div>
-        {subheader && <p className="text-tiny text-gray-800 pb-3">{subheader}</p>}
-        {<ul>{waitlistItems}</ul>}
+        <p className="text-tiny text-gray-800 pb-3">{t("listings.waitlist.submitAnApplication")}</p>
+        <ul>
+          {props.waitlistCurrentSize != null && (
```
emilyjablonski, line 27: !==

ui-components/src/page_components/listing/listing_sidebar/Waitlist.tsx
```diff
@@ -1,77 +1,46 @@
...
+            <WaitlistItem
+              value={props.waitlistCurrentSize}
+              text={t("listings.waitlist.currentSize")}
+            />
+          )}
+          {props.waitlistOpenSpots != null && (
```
emilyjablonski, line 33: !==

emilyjablonski (commented):
Could we add a test for this case?

ui-components/src/page_components/listing/listing_sidebar/Waitlist.tsx
jaredcwhite: Ah, so this is what got me off on the wrong tangent before. 😅 When checking for something that's either null or undefined, you can't use the double-equals. Some more context here: https://stackoverflow.com/a/15992131

jaredcwhite:
Yep, just added.

ui-components/src/page_components/listing/listing_sidebar/Waitlist.tsx
seanmalbert: In cases like this, I'd rather us explicitly check for both, so `props.waitlistCurrentSize !== null && props.waitlistCurrentSize !== undefined`

seanmalbert: Something that we could do (not here and for a larger point of discussion), and I've done in past projects, is to avoid using null altogether. Right now we have this mix of values being null or undefined so we always have to check for both, but we could make a system wide decision to clean up all the nulls and use undefined instead. There's likely a lot of work to do there, but it would be nice to have some consistency.

Issue #2098, opened 2021-10-28
Applications only showing 3


kathyccheng:
**Acceptance Criteria**
* When I'm looking at applications for a listing in the partners portal, and the default is set to 8, I only see 4 applications. I expect to see 8. 

**Notes**
* Sometimes it's 3 or 4 or 5

**QA NOTES**
Create or edit an application on a listing
on that application, add several preferred units
now go to the partner's portal, and go the table of applications for the listing you created (something like https://dev-partners-bloom.netlify.app/listings/5fc338b3-45bd-4054-a383-869f2fe02ec4/applications )

there should now be the correct number displayed on that table, depending on the number that should show

### QA Review Instructions
1. go to https://dev-partners-bloom.netlify.app/
2. View the applications for a listing by clicking on the Listing Status link (https://dev-partners-bloom.netlify.app/listings/0967a60b-2de4-45e5-aacc-c905ec228547/applications)
3. the per page selection should be accurately reflected

seanmalbert:
@YazeedLoonat , your PR was merged. Once it's up on dev, please review and move over to QA.

seanmalbert:
Hey @YazeedLoonat , due to the failing tests on the backend, I reverted your PR for this. Can you re-open it with the latest PR you put out to fix it in there. That still had a test failing, which may be in part to the package version change bump.

slowbot:
LGTM @seanmalbert

PR #2188, opened 2021-11-16
2098/fix partner applications grid display

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/2098)

## Description
When an application had several preferred units selected it was messing up the partner's listing -> applications table

Create or edit an application on a listing
on that application, add several preferred units 
now go to the partner's portal, and go the table of applications for the listing you created (something like https://dev-partners-bloom.netlify.app/listings/5fc338b3-45bd-4054-a383-869f2fe02ec4/applications )

there should now be the correct number displayed on that table, depending on the number that should show

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


emilyjablonski:
Is there an easy way we could add a test for this on the backend?

YazeedLoonat:
Hey @emilyjablonski I'm not sure if there is
the issue was that when there was multiple preferred unit types selected, the query was getting each of those which was causing duplicate data to be returned

with this fix it should no longer be querying for preferred units when we didn't need preferred units. So the fix here would be that preferred units info was no longer a part of the query/result set which I'm sure if that would be a valuable test or not



if you feel like it should be tested, I can add something like that in though

emilyjablonski:
@YazeedLoonat Makes sense! From an AC perspective I could see ensuring we get the amount we expect on a query but NBD :)

seanmalbert (changes requested):
Hey @YazeedLoonat ,
I think we should address this issue with something along the lines of what we did with the listing "views". See `backend/core/src/listings/views/view.ts`. The idea there is that we define different views for different use cases, so we can more easily control what fields and joins we want.

I think this works for applications as well, as it looks like the list doesn't need a lot of those other joins either. Let me know if you have any questions and feel free to iterate off of what's there for listings in applying it to applications with improvements as you see them.

YazeedLoonat:
Hey @seanmalbert I made changes so it uses views like you requested, one thing I didn't have a chance to do was actually fill out the left join and select portions completely, I basically just used leftJoinAndSelect for everything but the join that was causing the issue in the first place.

So I'd like to come back and resolve that and add a cypress test, but if you can peep and let me know what you think I would appreciate

emilyjablonski:
@seanmalbert Do you mind re-checking your requested changes?

backend/core/src/applications/views/config.ts
```diff
@@ -0,0 +1,57 @@
...
+    ["applicant.workAddress", "applicant_workAddress"],
+    ["application.alternateAddress", "alternateAddress"],
+    ["application.mailingAddress", "mailingAddress"],
+    ["application.alternateContact", "alternateContact"],
+    ["alternateContact.mailingAddress", "alternateContact_mailingAddress"],
+    ["application.accessibility", "accessibility"],
+    ["application.demographics", "demographics"],
+    ["application.householdMembers", "householdMembers"],
+    ["householdMembers.address", "householdMembers_address"],
+    ["householdMembers.workAddress", "householdMembers_workAddress"],
```
seanmalbert, line 53: I think this can be reduced to:
```
leftJoinAndSelect: [
  ["application.applicant", "applicant"],
  ["application.householdMembers", "householdMembers"],
],
```

backend/core/src/applications/views/config.ts
```diff
@@ -0,0 +1,57 @@
...
+    { join: "application.alternateAddress", alias: "alternateAddress" },
+    { join: "application.mailingAddress", alias: "mailingAddress" },
+    { join: "application.alternateContact", alias: "alternateContact" },
+    { join: "alternateContact.mailingAddress", alias: "alternateContact_mailingAddress" },
+    { join: "application.accessibility", alias: "accessibility" },
+    { join: "application.demographics", alias: "demographics" },
+    { join: "application.householdMembers", alias: "householdMembers" },
+    { join: "householdMembers.address", alias: "householdMembers_address" },
+    { join: "householdMembers.workAddress", alias: "householdMembers_workAddress" },
+  ],
```
seanmalbert, line 40: So there isn't any confusion about what's being used here, can you remove select and leftJoins.

seanmalbert (changes requested):
@YazeedLoonat ,
This looks good, see my comments for a little cleanup.

backend/core/src/applications/views/config.ts
YazeedLoonat: so I played around with this and reduced the number of joins down to what I think is the minimum, the other joins are needed for the rest of the application grid to display properly

backend/core/src/applications/views/config.ts
YazeedLoonat: done

backend/core/src/applications/views/config.ts
```diff
@@ -0,0 +1,57 @@
...
+      ["application.alternateAddress", "alternateAddress"],
+      ["application.mailingAddress", "mailingAddress"],
+      ["application.alternateContact", "alternateContact"],
+      ["alternateContact.mailingAddress", "alternateContact_mailingAddress"],
+      ["application.accessibility", "accessibility"],
+      ["application.demographics", "demographics"],
+      ["application.householdMembers", "householdMembers"],
+      ["householdMembers.address", "householdMembers_address"],
+      ["householdMembers.workAddress", "householdMembers_workAddress"],
+      ["application.preferredUnit", "preferredUnit"],
```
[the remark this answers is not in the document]
YazeedLoonat: with the way this is set up the base set is used throughout by the 
`_getQb` function in applications/services/applications.service.ts file 

so I think we do need this

YazeedLoonat:
@seanmalbert I made some updates per your requests and left comments against your comments with some things I didn't change lemme know what you think

seanmalbert (approved):
Thanks for the updates, @YazeedLoonat!

Issue #2225, opened 2021-11-24
Password Reset Back End Work 


kathyccheng:
As part of our security protocol updates, we want to require a partner portal user reset their password every n days. The default should be 180. We can hold off on creating a jurisdictional setting for this until different durations are requested.

A few things need to happen:
1. We need to keep track of when the user last updated their password
2. Using the password_updated_at, when a user logs in, we need to let the user know that they must reset their password before doing anything else (right now this can be a message returned from the login attempt)
3. ensure that the old password does not match the new one 
4. ensure that the update password endpoint works properly

Considerations:
If the user's password has expired, we probably don't want to give them access to anything before they update their password, so instead of logging them in, we want to prompt the user to update their password. Since update password currently requires a token, we may want to create a new endpoint that takes the current password, the new password and the new password confirmation. In either case, a new confirmation token should be generated, and the user wouldn't be able to login until they confirm the password update.

seanmalbert:
@pbn4 , here is the figma files for MFA, which contains all of the copy https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4649%3A181148

Issue #2017, opened 2021-10-18
Security - Password Reset


kathyccheng:
As an admin, I want to ensure all account are up to date and that passwords are reset on a regular basis to assure this.

**Acceptance Criteria**
- After a set amount if time (180 days in SJ) the user is prompted to reset their password
- User must enter new password and confirm of new password
- Once password is updated, they will need to confirm via email to continue

**Flow**
https://www.figma.com/file/kKCZvQHjPt3hHDwJTyk9sK/bloom-security-flow?node-id=0%3A1

**Screen**
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4773%3A182762
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4810%3A184365

dominikx96:
@kathyccheng I don't have access to this file: https://www.figma.com/file/kKCZvQHjPt3hHDwJTyk9sK/bloom-security-flow?node-id=0%3A1

kathyccheng:
@slowbot ^^

dominikx96:
@kathyccheng @slowbot @seanmalbert 

During development, I realized that those views don't fit the current backend implementation. Currently, we use forgot password flow. When a user login using an account with an expired password, then it sees a red alert:

![Uploading Screenshot 2022-01-12 at 10.48.52.png…]()

1. We need a proper translation with information that the password is already expired
2. A user should be able to click a button to resend the reset password email with the reset link, so I think we should design a modal with a description + button "send email" - it can fire after a login failure.

then, a change password flow is the same as in the forgot-password, even the same form.

@seanmalbert can you confirm?

seanmalbert:
@dominikx96 ,
1. We shouldn't see the invalid password message if the user's password is correct and has expired, so yes, that needs to change.
2. Per https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4773%3A182762, upon logging in with the expired password, they should be taken to a reset password screen

slowbot:
@dominikx96 

I see your point. I had the modals backwards. The sign in for should generate a modal first depending on whether password is stale and send a reset link which directs to a page which updates the password.

https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=5015%3A168693

I'm open to using the same form or the new form with use case specific text.
[...]

PR #2244, opened 2021-12-01
feat(backend): add user password expiration

pbn4:
# Pull Request Template

## Issue Overview

This PR addresses #2225 

## Description

This change adds user password expiration after 180 days and additional fields to UserDto (fetchable via UserProfile) about when will the password expire. 

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend/core/src/auth/passport-strategies/jwt.strategy.ts
```diff
@@ -35,9 +36,15 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
...
     }
     const userId = payload.sub
-    return this.userRepository.findOne({
+    const user = await this.userRepository.findOne({
       where: { id: userId },
       relations: ["leasingAgentInListings"],
     })
+
+    if (UserService.isPasswordOutdated(user)) {
+      throw new UnauthorizedException()
```
seanmalbert, line 45: We want to return a message to the user that lets them know that their password has expired. Can you update this to use `HttpException` and add an entry with translation string for `USER_ERRORS`. The frontend swallows this right now, but we can address that in the frontend issue for this (which will be #2017).

seanmalbert (changes requested):
@pbn4 ,
This is off to a good start. A couple of things that are still needed:
1. since a user cannot login if their password is expired, `update-password` won't work, because it currently requires a token. The current update password endpoint is still valid, because we want users to be able to reset their password while logged in. What do you think about creating a new `reset-password` end point, which takes a user's email, old password, new password and new password confirmation. This should update their password, create a new confirmation token and send an email to the user for their confirmation. You should be able to write e2e tests for this flow, and verify that the user cannot login with the new password until they have confirmed it.
2. in the password-reset endpoint, check that the old password does not match the new one
3. see my comment on sending a friendly error back to the user

backend/core/src/auth/passport-strategies/jwt.strategy.ts
pbn4: Currently we have two ways to update the password:

1. PUT /user/:id (requires access token)
2. PUT /user/update-password (part of forgot password flow)

> since a user cannot login if their password is expired, update-password won't work, because it currently requires a token. The current update password endpoint is still valid, because we want users to be able to reset their password while logged in.

I was under an impression that we want to use forgot password flow when user account becomes locked because password is outdated. User requests forgot pasword email, receives reset URL, clicks the URL and sets up new password. 

I'm OK with implementing another endpoint only for resettting the password, if this is not sufficient. Please let me know how to proceed with this.

seanmalbert: Ah, yes, forgot password flow will work fine.

seanmalbert (approved):
@pbn4 ,
This looks good. Can you please address the failing backend tests and then we should be good to merge.

Issue #2235, opened 2021-11-30
Cypress Coverage Report


seanmalbert:
We want to be able to generate cypress coverage reports. The documentation looks straight forward https://docs.cypress.io/guides/tooling/code-coverage#Instrumenting-code, where we can add in another library and update some settings and add a new script.

seanmalbert:
### UI-Component Unit Test Coverage

File                                            | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                                                                                    
------------------------------------------------|---------|----------|---------|---------|------------------------------------------------------------------------------------------------------
All files                                       |   76.02 |    53.59 |      55 |   76.98 |                                                                                                      
 actions                                        |   96.67 |    92.31 |     100 |   98.63 |                                                                                                      
  Button.tsx                                    |   93.94 |    90.24 |     100 |   96.15 | 54                                                                                                   
  ExpandableContent.tsx                         |     100 |      100 |     100 |     100 |                                                                                                      
  ExpandableText.tsx                            |   96.15 |    94.44 |     100 |     100 | 38                                                                                                   
  LinkButton.tsx                                |     100 |      100 |     100 |     100 |                                                                                                      
  LocalizedLink.tsx                             |     100 |      100 |     100 |     100 |                                                                                                      
 authentication                                 |   77.78 |    65.71 |   64.71 |   77.63 |                                                                                                      
  RequireLogin.tsx                              |     100 |    94.12 |     100 |     100 | 41                                                                                                   
  timeout.tsx                                   |   68.25 |    38.89 |   53.85 |   67.31 | 18-21,32,64-77,86-87,111-112                                                                         
 blocks                                         |     100 |    91.14 |     100 |     100 |                                                                                                      
  ActionBlock.tsx                               |     100 |    84.21 |     100 |     100 | 25,32-33                                                                                             
  DashBlock.tsx                                 |     100 |      100 |     100 |     100 |                                                                                                      
  DashBlocks.tsx                                |     100 |      100 |     100 |     100 |                                                                                                      
  FormCard.tsx                                  |     100 |      100 |     100 |     100 |                                                                                                      
  HousingCounselor.tsx                          |     100 |      100 |     100 |     100 |                                                                                                      
  ImageCard.tsx                                 |     100 |    92.86 |     100 |     100 | 53                                                                                                   
  InfoCard.tsx                                  |     100 |     87.5 |     100 |     100 | 31                                                                                                   
[...]

PR #2258, opened 2021-12-02
test: adds cypress test coverage

seanmalbert:
## Issue Overview

This PR addresses #2235 

## Description

This adds cypress test coverage. I followed https://docs.cypress.io/guides/tooling/code-coverage#Install-the-plugin and https://github.com/bahmutov/next-and-cypress-example. Reading the latter will be more useful to you in checking out the coverage reports.


emilyjablonski:
Deploy failure looks like a type issue in coverage.ts

PR #2290, opened 2021-12-08
fix: units with invalid ami chart

seanmalbert:
# Pull Request Template

## Issue Overview

This PR addresses #issue

## Description

This issue came up because there are jurisdictions like San Mateo and San José on production, that don't have ami charts, but units were added to listings with those jurisdictions before we tied ami charts to jurisdictions. This is marked as only a partial fix, because while it does fix the issue, we also need to add the ami charts, which are in progress at least for SJ.

You can test this by:
1. create a new draft listing with
2. create a unit and select an ami chart (there should be one per jurisdiction from the seed)
3. save the listing
4. in your local db, reassign the ami chart you selected to a different  jurisdiction
5. edit the listing and one of the overrides on the listing (you shouldn't be able to select an ami chart now)
6. save the listing and you shouldn't see an error, but if you comment out the one line that fixes this, you should see the error. 

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/backend/core/src/shared/units-transformations.ts
+++ b/backend/core/src/shared/units-transformations.ts
@@ -97,12 +97,11 @@ const hmiData = (units: Units, maxHouseholdSize: number, amiCharts: AmiChart[]) 
           if (unit.amiChartOverride) {
             amiChart = mergeAmiChartWithOverrides(amiChart, unit.amiChartOverride)
           }
-          return {
+          return JSON.stringify({
             percentage: parseInt(unit.amiPercentage, 10),
             chart: amiChart,
-          }
+          })
         })
-        .map((item) => JSON.stringify(item))
     ),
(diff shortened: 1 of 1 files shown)


YazeedLoonat (approved):
LGTM only thing I would suggest is maybe switching 
`.filter` then `.map`
then `.reduce` so we only need to loop over once

emilyjablonski:
Makes sense to me 👍

seanmalbert:
> LGTM only thing I would suggest is maybe switching `.filter` then `.map` then `.reduce` so we only need to loop over once

@YazeedLoonat ,
I removed the unnecessary additional map, but opted to keep the Set and final map, since I think that involves less logic than adding a reduce and creating the unique list with that.

Issue #2337, opened 2021-12-21
Investigate adding `max-lines` & `max-lines-per-function` rules for ESLint


jaredcwhite:
**Update:** looking into a `max-lines` rule as well, so we don't end up overall with really long files which might also be considered a code smell.

----

As part of tech debt cleanup, I wanted to get this on the radar. I did look into it a while back.

Basically the sales pitch here is that we have a number of files with really long functions, as in hundreds of lines, and while sometimes that's necessary and unavoidable, other times I would argue it's a code smell (needs to be refactored, various things broken out, maybe a different pattern would work better, etc.). Adding the linting rule is easy enough, but figuring out what the ideal cap should be might take some doing. The added complication is there can be a lot of lines in a React JSX template function which isn't necessarily that complex because it's largely HTML markup, so that can otherwise throw off a reasonable linting rule.

I'll play around with some different values to try to get a sense of what's passing vs. what's failing, and then maybe I can provide a more informed recommendation for the team to review.

FYI: https://eslint.org/docs/rules/max-lines-per-function

jaredcwhite:
P. S. I need to double-check this, but I'm pretty sure the rule equally applies to methods in a class definition.

kathyccheng:
Closing, associated PR still relevant

PR #2360, opened 2022-01-04
chore: add max lines linting rules and refactor

jaredcwhite:
## Issue Overview

This PR addresses #2337 

## Description

Configures eslint with two new rules: max-lines-per-function and max-lines (per file). The intention is to encourage breaking up large single functions (sometimes with nested functions) into smaller discrete functions and perhaps additional files, so code blocks are simpler to maintain and easier to unit test—and possibly improve performance or memory usage as well.

FYI, the numbers I chose for the rules was after tweaking the values a few different times to find a sweet spot between not enough stuff/too much stuff getting caught in our existing codebase. Also the way React Hook Form-related components get written ends up with really long functions much of the time, so I'm not sure how much we can mitigate against that, but it does seems like there's some room for improvement in a few places.

Obviously we can tell if the linter & tests all run as expected, but in the case of refactored files we should try to QA the pages or API endpoints which were modified.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/backend/core/src/shared/hmi-data.ts
+++ b/backend/core/src/shared/hmi-data.ts
@@ -1,0 +1,220 @@
+import { Unit } from "../units/entities/unit.entity"
+import { AmiChart } from "../ami-charts/entities/ami-chart.entity"
+import { AmiChartItem } from "../ami-charts/entities/ami-chart-item.entity"
+import { UnitAmiChartOverride } from "../units/entities/unit-ami-chart-override.entity"
+import { MinMaxCurrency } from "../units/types/min-max-currency"
+import { minMaxCurrency, usd, yearlyCurrencyStringToMonthly } from "./utils/currency"
+
+type ChartAndPercentage = {
+  percentage: number
+  chart: AmiChart
+}
+
+const findAmiValueInChart = (
+  amiChart: AmiChartItem[],
+  householdSize: number,
+  percentOfAmi: number
+) => {
+  return amiChart.find((item) => {
+    return item.householdSize === householdSize && item.percentOfAmi === percentOfAmi
+  })?.income
+}
+
+const getAmiChartItemUniqueKey = (amiChartItem: AmiChartItem) => {
+  return amiChartItem.householdSize.toString() + "-" + amiChartItem.percentOfAmi.toString()
+}
+
+// All unique AMI percentages across all units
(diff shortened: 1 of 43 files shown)


emilyjablonski:
Is this ready for review?

backend/core/src/shared/hmi-data.ts
```diff
@@ -0,0 +1,220 @@
+import { Unit } from "../units/entities/unit.entity"
```
seanmalbert, line 1: @jaredcwhite , just want to confirm, that nothing was changed here, except for pulling these pieces out of unit-transformations to get the file size down under 400 lines, correct?

seanmalbert: Commented too soon, looks like `amiPercentagesAcrossUnits`, `findAmiValueInChart`, and `setupHmiHeaders` was pulled out

jaredcwhite:
@emilyjablonski yes, sorry I forgot to add the needs review label.

YazeedLoonat:
@jaredcwhite could you explain how you got to your cut offs for these?

YazeedLoonat:
Hey @jaredcwhite @seanmalbert @emilyjablonski 
I feel like that limits for max lines per function and max lines per file are somewhat arbitrary in this pr. I tried to look into if there are standards set out by some of the larger open source projects I'm familiar with and none of them seem to use these rules at all. 

https://github.com/mui-org/material-ui/blob/master/.eslintrc.js
https://github.com/microsoft/TypeScript/blob/main/.eslintrc.json
https://github.com/facebook/react/blob/main/.eslintrc.js


I'm not entirely sure I agree with setting these rules in the first place since it seems like in a lot of places already we had to set up files and functions to disable/ignore the rules anyway 

I do think it is a good idea to have separated out the functions and files that you already did in this pr. I feel like code reviews should catch if a function can be broken up, or a file can be split up and we won't have to worry about a seemingly arbitrary rule guiding that

emilyjablonski:
I'm in agreement with Yazeed here - I think it's an awesome best practice that we can and should incorporate in PR reviews but I don't see a ton of value out of enforcing it this way, considering there are places we want to disable it and it kind of depends on the context.

emilyjablonski:
Seems like the PR review resulted in passing on this, and given it's been almost a year going to close, lmk if you'd prefer a different resolution!

Issue #2359, opened 2022-01-04
Public Listing Detail - Submit a Paper Application


seanmalbert:
The text instruction for sending an application by US mail depend on both the application due date and the postmark by date. We have cases where we may have the application due date, but not the post mark date filled out and others where neither are present, but we still want to display the text.

Please also make sure that the dates appear as they should, since application due date isn't showing the proper time.

If there is an application due date, but no postmark by date, it should read:
```
Applications must be received by the deadline. If sending by U.S. Mail, the application must be postmarked by %{applicationDueDate}. %{developer} is not responsible for lost or delayed mail.
```

If there is a postmark by date, but no application due date, it should read:
```
Applications must be received by the deadline. If sending by U.S. Mail, the application must be received by mail no later than %{postmarkReceivedByDate}. Applications received after %{postmarkReceivedByDate} via mail will not be accepted. %{developer} is not responsible for lost or delayed mail.
```

If there is neither, it should read:
```
%{developer} is not responsible for lost or delayed mail.
```

If it has both, it should read as is:
```
Applications must be received by the deadline. If sending by U.S. Mail, the application must be postmarked by %{applicationDueDate} and received by mail no later than %{postmarkReceivedByDate}. Applications received after %{postmarkReceivedByDate} via mail will not be accepted even if they are postmarked by %{applicationDueDate}. %{developer} is not responsible for lost or delayed mail.
```

slowbot:
Mockups

Leasing Agent
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4982%3A168895

Address 1
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4982%3A168486

Address 2
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=4982%3A168775

emilyjablonski:
Solves #2384

emilyjablonski:
QA:
1) Fix for new submit by mail strings: Make four iterations of a listing using the above combinations of postmark date and app due date and ensure the right strings are used
2) Fix for app due date ticking forward: Save a listing with an application due date, re-open the listing to edit it, and ensure the date pre-filled in the app due date form field is correct
[...]

PR #2362, opened 2022-01-06
fix: dates showing as invalid in send by mail section

emilyjablonski:
# Pull Request Template

## Issue Overview

This PR addresses #2359

## Description

In the send application by mail section on the public site, some of the dates were showing as invalid and not respecting the times.

We also have some new strings for different scenarios of date combinations (in the ticket).

This PR also fixes a bug where you are currently unable to save a mailing address.

Test the four scenarios in the ticket and ensure the strings are as expected and that it respects the time.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


seanmalbert (changes requested):
@emilyjablonski, your updates look good. In testing this, I noticed a number of issues that aren't called out in the issue, but I think we need to fix them now to make the submit application section usable:

- the application due date ticks forward a day with each save and it displays the previous day on the public side. This was probably caused by the move to day.js
- the numbering is off, but maybe we can get rid of it altogether (see where it has the number two next to Submit a Paper Application)
Screen Shot 2022-01-07 at 4 35 16 PM
- the SEND APPLICATION BY US MAIL section only seems to show if there's an applicationMailingAddress (`Paper applications are mailed to another address` is checked and the address filled out), but the language on the listing management form implies if that's not checked, that the leasing agent address should be used as the mailing address
- I cannot clear out the Leasing Agent Address, but I think clearing those fields from the form should remove the leasingAgentAddressId

Let me know if you think some or all of these should be their own issues.

emilyjablonski:
@seanmalbert Thanks for the review! I can fix 1 2 and 4 in this PR but for 3 can you check out the message I sent you/Kathy/Jesse a few days ago? I had concerns about the assumption that there is always a mailing address

YazeedLoonat (changes requested):
I think this looks good, small code change, and a request for a logic tweak: 

I think if no mailing address is on the listing, but there is a leasing agent address, I think we we should be falling back on the leasing agent address for this otherwise the section displaying this info doesn't show

sites/public/src/ListingView.tsx
```diff
@@ -296,20 +296,22 @@ export const ListingView = (props: ListingProps) => {
 
   // Move the above methods into our shared helper library when setup
 
+  const getDateString = (date: Date, format: string) => {
+    return date ? moment(date).format(format) : null
```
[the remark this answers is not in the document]
emilyjablonski: It was coming in as undefined which is why the dated look like "Invalid" - and for sure I should rebase and pull in dayjs!

emilyjablonski:
I let the scope creep a little / probably should have separate out the feedback, updates:


* New update: the application due time should always default to 5pm if not entered and due date is entered
    * Flows
        * Save a listing without either and ensure no time is set
        * Save a listing with just a date and ensure 5pm is set
        * Save a listing with a date and a time and ensure custom time is set
    * As part of this removed the applicationDueTime field since it is redundant with the applicationDueDate field which can just hold the time
* New update: there is no longer an assumption that there is always a mailing address, so now mailing address looks the same as pick up and drop off address where you have to select if there is one, then select if it is the leasing agent address or a custom address
    * This means the removal of mailing address on the address type enum bc it is no longer a radio button
    * Should be able to add and  remove mailing/dropoff/pickup addresses
* In the Apply sidebar
    * Removed section numbers
    * The mailing, drop-off, and pickup address sections should only show if that specific address exists or if that address type is leasing agent, otherwise the section does not show at all
* The four cases in the ticket for combinations of due date and postmark date produce the right strings and valid dates in the mailing section
* App due date does not reload the edit page with an incremented due date (dayjs issue)

backend/core/src/seeder/seeds/listings/listing-coliseum-seed.ts
```diff
@@ -78,8 +77,8 @@ const coliseumListing: ListingSeedType = {
...
   criminalBackground: null,
-  depositMax: "1,781",
-  depositMin: "1,284",
+  depositMax: "200",
```
emilyjablonski, line 80: Coliseum will now save as draft after reseed 😄

sites/partners/cypress/integration/listing.spec.ts
```diff
@@ -94,6 +94,13 @@ describe("Listing Management Tests", () => {
       cy.getByID("leasingAgentAddress.zipCode").type(listing["leasingAgentAddress.zipCode"])
       cy.getByID("leasingAgentAddress.state").select(listing["leasingAgentAddress.state"])
 
+      cy.getByTestId("mailing-address-checkbox").check()
```
emilyjablonski, line 97: We have a bug right now on dev where you can't save a mailing address, fixed and added to Cypress

ui-components/src/page_components/listing/listing_sidebar/events/EventDateSection.tsx
```diff
@@ -3,18 +3,19 @@ import { ListingEvent } from "@bloom-housing/backend-core/types"
 import dayjs from "dayjs"
 
 const EventDateSection = (props: { event: ListingEvent }) => {
+  const getRangeString = () => {
```
emilyjablonski, line 6: #2384

backend/core/src/migration/1641855882656-remove-application-due-time.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class removeApplicationDueTime1641855882656 implements MigrationInterface {
+  name = "removeApplicationDueTime1641855882656"
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "listings" DROP COLUMN "application_due_time"`)
+  }
```
seanmalbert, line 8: For existing listings, we'll need to make sure the time set on application_due_date before dropping the column.

sites/partners/src/listings/PaperListingForm/formatters/DatesFormatter.ts
```diff
@@ -5,22 +5,22 @@ import { createDate, createTime } from "../../../../lib/helpers"
...
-    )
+    const dueDate = createDate(this.data.applicationDueDateField)
+    this.data.applicationDueDate =
+      this.data.applicationDueTimeField?.hours && this.data.applicationDueTimeField?.minutes
+        ? createTime(dueDate, this.data.applicationDueTimeField)
+        : createTime(dueDate, {
+            hours: "05",
+            minutes: "00",
+            period: "pm",
+          })
```
seanmalbert, line 16: Since we're defaulting this here to 5pm, should we default it on the form too?

sites/public/src/ListingView.tsx
```diff
@@ -296,37 +295,41 @@ export const ListingView = (props: ListingProps) => {
...
+          postmarkedApplicationsReceivedByDate: getDateString(
+            listing.postmarkedApplicationsReceivedByDate,
+            `MMM. DD, YYYY [${t("t.at")}] hh:mm A`
+          ),
           developer: listing.developer,
-          applicationsDueDate: dayjs(listing.applicationDueDate).format(
-            `MMM. DD, YYYY [${t("t.at")}] h A`
+          applicationsDueDate: getDateString(
+            listing.applicationDueDate,
+            `MMM. DD, YYYY [${t("t.at")}] hh:mm A`
```
seanmalbert, line 332: The formatting of the month doesn't quite work with May, since it's the only one that isn't abbreviated. I'm not sure if this is a problem since it was already like this, but can we either change the formatting so the period isn't necessary or add logic so that it doesn't appear for May?

seanmalbert (changes requested):
@emilyjablonski ,
This is looking really good and I really like that you combined the application due date and time, not to mention updating the coliseum seed. Please see the few comments I added.

backend/core/src/migration/1641855882656-remove-application-due-time.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export class removeApplicationDueTime1641855882656 implements MigrationInterface {
+  name = "removeApplicationDueTime1641855882656"
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const listings = await queryRunner.query(
+      `SELECT id, application_due_time, application_due_date from listings`
+    )
+
+    for (const listing of listings) {
```
emilyjablonski, line 11: @seanmalbert @YazeedLoonat Made the updates here to put existing time fields onto the date fields as we remove the time, but I struggled a little here so would love some eyes on this new piece before it goes in

seanmalbert: @emilyjablonski ,
I think this can be simplified by splitting the date and time on `T`, and combining them.

seanmalbert (approved):
@emilyjablonski ,
This looks good to me. See my comment on the migration script, since you can probably make that simpler before merging.

Issue #2403, opened 2022-01-13
Paper Applications - Saves Partner User as Application User


seanmalbert:
**Describe the bug**
When submitting a paper application from the partner portal, it saves the user for the application as the partner portal user.

**To Reproduce**
1. Go to the partner portal and fill out a paper application
2. find that application in the database and you'll see that the associated user is the user you were signed in as

**Expected behavior**
There should be no user tied to the application.

**Notes**
* Clean up existing ones too (to do when pushed live)

### QA Review Notes
1. Go to https://dev-partners-bloom.netlify.app and submit a new application for any listing
2. Go to public portal and log in with the same credentials. You should not see any applications in my applications tab
3. fill out an application on the public portal signed in as a user and make sure that you do see the application in the my applications tab

slowbot:
@seanmalbert Tested with these AC and looks fine but not sure I get the first part. Let me know if you have time to sync on it.

kathyccheng:
LGTM

Issue #2335, opened 2021-12-20
Adding admin error


kathyccheng:
When I try to add an admin account on production, I enter a name, email, admin type and access to all listings option, I get an error Something Went Wrong.

Notes
* When someone in partners portal adds a user and they're already in there with a public account, assign role and listing 

**QA NOTES:**
Go into public site and add a new user
go into the partner site and attempt to add that exact same user as a partner (there is currently discussion going on around if admin users should show in the user table)
the user should now be added correctly

slowbot:
@kathyccheng LGTM

Issue #1873, opened 2021-09-23
Check User Roles on Partner Portal Login


seanmalbert:
Update the login to check the user's roles when logging into the portal. If they don't have a role, they should get an unauthorized message.

seanmalbert:
I was originally thinking that in either an auth guard or in `generateAccessToken`, preferably a guard, we would check a user's roles for login and token, but we should be able to update the auth policy to require a user to be an admin or partner for those endpoints.

seanmalbert:
Adding this to the next sprint, since it came up in #2415. I think it makes sense to create a separate login for partners that uses the `AuthzGuard`, so that only those with an admin or partner role can login. This necessitates updating the auth policy as well.

PR #2415, opened 2022-01-18
2335/fixes adding user that's already created

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/2335)

## Description
This fixes an error where if an email was already tied to a public user, then trying to add them as a partner was causing an error

Go into public site and add a new user 
go into the partner site and attempt to add that exact same user as a partner (there is currently discussion going on around if admin users should show in the user table)
the user should now be added correctly 

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


seanmalbert (changes requested):
@YazeedLoonat ,

I think we need a little more work to get this working as expected. You updated `_createUser` and it returns the user to `invitePartnersPortalUser`, and then the invite flow continues. We probably want them to still receive the confirmation, but since `_createUser` is now returning before the confirmation token can be generated, the confirmation link in the email isn't valid.

Another problem is that `_createUser` is also used by `createPublicUser` and in that case we still want `_createUser` to return the email in use error, so we either need to add an argument to `_createUser` to add some logic or split some things out so they're not used by both functions.

YazeedLoonat:
Hey @seanmalbert I got the issues you pointed out addressed, lemme know what you think

emilyjablonski:
@YazeedLoonat Heads up on some failing backend tests to resolve

emilyjablonski (changes requested):
If I create a user in public I can log-in as that user to partners even though they don't have access, though it doesn't show me any listings. The same occurs if I create a user in partners, I can log-in to public without having created a public account (it seems to default the birthday).

seanmalbert:
> If I create a user in public I can log-in as that user to partners even though they don't have access, though it doesn't show me any listings. The same occurs if I create a user in partners, I can log-in to public without having created a public account (it seems to default the birthday).

@YazeedLoonat ,

There is an existing [issue](https://github.com/bloom-housing/bloom/issues/1873) to cover the case where a public user can login to the partner portal, but cannot access anything. I think it's acceptable that a partner user can login to the public site as long as we fix #2403. I moved both issues into next sprint.

seanmalbert (changes requested):
@YazeedLoonat ,

This may have been a problem with how you had it before, so I apologize for not calling this out earlier if that's the case, but this does not correctly handle a public account getting created with the same email address multiple times. It doesn't create duplicates, but starts the confirmation process over again for the user.

Rather than try and get `_createUser` working for both public and partner portal cases, I think we should have two separate functions. Looking at the user service as a whole now, it could use a refactor to split out into two services, but that can be for another issue.

Please make sure you create tests for the new function(s).

YazeedLoonat:
This should be good for re-review now

emilyjablonski:
Issue AC LGTM, I could have lost some of the desired behavior but currently I if I already have a partners account I cannot create a public account with that e-mail, is that expected?

seanmalbert:
> Issue AC LGTM, I could have lost some of the desired behavior but currently if I already have a partners account I cannot create a public account with that e-mail, is that expected?

Yes, since there are no roles associated with a public account, all partner portal users have the ability to login to the public site.

seanmalbert (approved):

Issue #2391, opened 2022-01-12
Account lockout shouldn't be on public 


kathyccheng:
**Acceptance Criteria**
* When I'm trying to sign in as a public user and I enter the wrong username and password, then I should not see the account lockout warning
* If I enter my username and password incorrectly after 5 tries, I should not be locked out of my account 

**Notes**
* Remove account lockout for public accounts on the /sign-in page

seanmalbert:
Closing issue, since the lockout on public also provides benefit.

PR #2433, opened 2022-01-25
2391/public account lockout

pbn4:
# Pull Request Template

## Issue Overview

This PR addresses #2391 

## Description

This change makes authentication logic handle account lock out only for admins/partners accounts (those that have .roles defined). 

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend/core/src/auth/entities/user.entity.ts
```diff
@@ -165,4 +165,17 @@ export class User {
...
+  }
+
+  hasLoggedInAtLeastOnce(): boolean {
+    return !!this.lastLoginAt
+  }
+
+  hasOutdatedPassword(): boolean {
+    return new Date(this.passwordUpdatedAt.getTime() + this.passwordValidForDays * 24 * 60 * 60 * 1000) <
+    new Date() && this.isPartnerOrAdmin()
+  }
```
seanmalbert, line 180: It doesn't matter too much, but can you update this to use day.js for the date comparison.

backend/core/src/auth/passport-strategies/local-mfa.strategy.ts
```diff
@@ -31,31 +30,31 @@ export class LocalMfaStrategy extends PassportStrategy(Strategy, "localMfa") {
...
 
     const user = await this.userRepository.findOne({
       where: { email: loginDto.email.toLowerCase() },
-      relations: ["leasingAgentInListings"],
+      relations: ["leasingAgentInListings"]
     })
 
     if (user) {
-      if (user.lastLoginAt) {
+      if (user.hasLoggedInAtLeastOnce() && user.isPartnerOrAdmin()) {
```
seanmalbert, line 42: Why does it matter if the user has logged in at least once before? Would it be more appropriate to check if the user is confirmed?

seanmalbert (changes requested):
@pbn4 ,

A few things:
1. can you update the PR to merge into dev without any MFA code
2. See my comment on checking if the user has logged in at least once, I don't think I'm clear on why that's the case anymore
3. This is more of a question for @kathyccheng . After thinking about it more, it makes sense to me that we don't want to have password expiration for public users, but I think it would be beneficial to keep password lockout, since that login is more exposed to the public and may be a larger target for brute forcing into people's accounts.

seanmalbert:
closing, since we're ok with this being on public

Issue #2372, opened 2022-01-06
Unit summary should hide unit detail information on listing page 


seanmalbert:
**Describe the bug**
On the listing management form in the units section, the radios for unit types or individual units doesn't do anything.

**To Reproduce**
1. Go to https://dev-partners-bloom.netlify.app
2. Edit or create a new listing and go to the unit section
3. select one of the options and save
4. edit the listing (make sure you refresh or the radio may still be selected)
5. you should see nothing selected (nothing gets updated in the db either)

**Expected behavior**
Unit summary should hide expandable section on listing

kathyccheng:
@slowbot ^^

slowbot:
@seanmalbert 

1. individual units: should enable individual unit accordion to be displayed on the single listing page
![image](https://user-images.githubusercontent.com/13915/149243135-591a3a62-861a-4499-895c-38ebcc930cf3.png)


2. Unit types: should only display a summary of unit group and not have accordion for individual units
![image](https://user-images.githubusercontent.com/13915/149243222-a3d9b431-9d9a-4681-a405-e628fd3c222a.png)

emilyjablonski:
QA: Changing the radio button about units should enable and disable the accordion, and the value should save appropriately as well as pre-fill with the correct value when re-editing the listing

slowbot:
@emilyjablonski Can we add an icon to the enabled state of the accordion?

![image](https://user-images.githubusercontent.com/13915/153260761-2f08bfa5-66b4-4e71-be55-e2748d032637.png)

We should have a carrot in the set of svg

emilyjablonski:
QA: Expandable unit table sections will have the carat/x on open/close

slowbot:
@emilyjablonski LGTM

PR #2451, opened 2022-01-28
fix: unit accordion radio button not showing default value

emilyjablonski:
# Pull Request Template

## Issue Overview

This PR addresses #2372 

## Description

When testing this I did see that changing the `Do you want to show unit types or individual units?` radio was already working to enable/disable the units accordion. What is fixed here is that the radio button was not loading the set value either on first open or after a refresh in the listing form.

Also fixed a UI issue in the accordion with the square feet value.

Ensure a listing loads the correct value on first open and after a refresh, and defaults to individual units.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


sites/partners/src/listings/PaperListingForm/sections/Units.tsx
```diff
@@ -48,8 +48,9 @@ const FormUnits = ({ units, setUnits, disableUnitsAccordion }: UnitProps) => {
   }
 
   useEffect(() => {
-    setValue("disableUnitsAccordion", disableUnitsAccordion ? "true" : "false")
-  }, [disableUnitsAccordion, setValue])
+    reset({ ...getValues(), disableUnitsAccordion: disableUnitsAccordion ? "true" : "false" })
```
emilyjablonski, line 51: IMO this is a weird solution but makes the default value work and does not appear to cause any extra renders

ui-components/src/page_components/listing/UnitTables.tsx
```diff
@@ -66,7 +66,13 @@ const UnitTables = (props: UnitTablesProps) => {
             number: unit.number,
             sqFeet: (
               <>
-                <strong>{unit.sqFeet}</strong> {t("t.sqFeet")}
+                {unit.sqFeet ? (
```
emilyjablonski, line 69: Screen Shot 2022-01-28 at 3 16 42 PM
Fixes these two issues

dominikx96 (changes requested):
@emilyjablonski changes from the PR scope look good to me. I've just noticed the same problem with the radio button in the "Is there an application due date?" section, when I select "yes" and refresh - an old value is gone

https://share.getcloudapp.com/jkumWxdl

emilyjablonski:
@dominikx96 Potentially this is expected, what I'm seeing is if I save a listing with app due date as yes and then don't add an app due date it does not save as yes bc there is no due date added. Let me check in w Kathy!

emilyjablonski:
Checked in w Kathy and the default to no if no date is ok 👍

sites/partners/src/listings/PaperListingForm/sections/Units.tsx
seanmalbert: I agree and I think a longer term solution (for another issue) should be to refactor FieldGroup, because this isn't the only place we're having to do some weird things because of that field.

seanmalbert (approved):

dominikx96 (approved):

Issue #2413, opened 2022-01-18
Render map when visible in viewport


seanmalbert:
Currently the map sits at the bottom of the page and we also have a `View on Map` link at the top making it so that we should be able to drastically cut our mapbox rendering by only rendering once (it looks like we're rendering it several times now on page load) when the map is in view.

Checkout [this](https://stackoverflow.com/questions/123999/how-can-i-tell-if-a-dom-element-is-visible-in-the-current-viewport) post for reference. 

**QA Review Instructions**
This is to be filled out by the developer who completes this issue, before passing to QA.

slowbot:
@jaredcwhite Seems to load well. Not sure if there are specific AC I should follow.

jaredcwhite:
@slowbot Thanks, I think we should be fine. It was just a low-level rendering optimization…just want to make sure it didn't break somehow.

PR #2470, opened 2022-02-03
feat: only show MapBox map when it's in the viewport

jaredcwhite:
## Issue Overview

This PR addresses #2413 

## Description

This PR adds a new `useIntersect` React hook which will supply state to control the visibility of an element relative to
an ancestor element (or by default the browser viewport). The map widget on listings then uses this so the map is only initialized once the browser scrolls down close to the map region (there's an extra buffer of 500 pixels).

Wait a beat after the listing has loaded, then scroll as fast as possible to the bottom of the listing, and you'll notice the map loads once it has become visible. Also you can reload & start at the top of the listing, and scroll down slowly, and likely by the time you get to the map it will have already loaded.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/ui-components/src/page_components/listing/ListingMap.tsx
+++ b/ui-components/src/page_components/listing/ListingMap.tsx
@@ -36,8 +36,12 @@ const isValidLongitude = (longitude: number) => {
 }
 
 const ListingMap = (props: ListingMapProps) => {
-  // Lazy load the map component only when it's visible on screen
-  const { setIntersectingElement, intersecting } = useIntersect({ rootMargin: "500px" })
+  // Lazy load the map component only when it will become visible on screen
+  const { setIntersectingElement, intersecting } = useIntersect({
+    // `window` isn't set for SSR, so we'll use `global` instead—doesn't really
+    // matter because the map won't ever get rendered in SSR anyway
+    rootMargin: `${global.innerHeight || 0}px`,
+  })
   const [hasIntersected, setHasIntersected] = useState(false)
   if (intersecting && !hasIntersected) setHasIntersected(true)
 
--- a/ui-components/src/page_components/listing/ListingMap.tsx
+++ b/ui-components/src/page_components/listing/ListingMap.tsx
@@ -1,9 +1,10 @@
-import React, { useState, useCallback, useEffect } from "react"
+import React, { useState, useCallback, useEffect, useMemo } from "react"
 import "mapbox-gl/dist/mapbox-gl.css"
 import MapGL, { Marker } from "react-map-gl"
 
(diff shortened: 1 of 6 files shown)


emilyjablonski:
This could be unrelated to this specific ticket or my misunderstanding, but if I put a console log right next to where <MapGL> is being rendered, I'm seeing two renders on page load and a re-render every time the screen window changes size.

https://user-images.githubusercontent.com/65367387/152831049-e38bdecd-d40b-4a91-ab8c-ee957f9e0f51.mov

ui-components/src/page_components/listing/ListingMap.tsx
```diff
@@ -35,6 +36,11 @@ const isValidLongitude = (longitude: number) => {
 }
 
 const ListingMap = (props: ListingMapProps) => {
+  // Lazy load the map component only when it's visible on screen
+  const { setIntersectingElement, intersecting } = useIntersect({ rootMargin: "500px" })
```
dominikx96, line 40: Can you explain why it's exactly `500px`? Is there any option (and sense) to use e.g ` window.innerHeight` value?

dominikx96 (approved):
Looks good! I've just put one comment/question

ui-components/src/page_components/listing/ListingMap.tsx
jaredcwhite: Oh I like that idea…500px was just an arbitrary value I chose so when someone's leisurely scrolling along the page they probably wouldn't see MapBox start up right in view.

jaredcwhite:
@emilyjablonski yeah, I noticed that too but I think you'll probably see that behavior on any number of the React components on the page there. My assumption is React runs the functions multiple times but only syncs the VDOM with the actual DOM if there's a state change, so we should be OK regarding the MapBox plugin.

seanmalbert (approved):
LGTM, @jaredcwhite. I saw @dominikx96 comment on using window.innerHeight to calculate the margin. I'm not sure we need to do that, since in most cases it will probably be greater than 500px, and I'd argue that's probably greater than we need it to be, but if we wanted to use it maybe we could use it minus some percentage of innerHeight.

jaredcwhite:
@seanmalbert haha, I saw your comment right after committing a change to use innerHeight. I think it will be OK…the point is basically so someone casually scrolling along probably won't see the map load in, it will have already loaded in the background. But it still wouldn't have interfered with the initial page load. If that seems good to you, I'll merge in.

seanmalbert:
@jaredcwhite ,
I think we're good to merge in. What's it's up on dev, please move over to QA.

Issue #2514, opened 2022-02-16
Translations - Audit of Partner Strings


kathyccheng:
Notes 
* Move out all things specific to Partners to override files 
* Move everything to keys
* Not in this issue, but for future, will re-format script 
* Additionally flatten the strings

emilyjablonski:
QA Notes: This will be challenging to QA. We have moved a lot of our strings around that could cause some strings to not render / show just a translation key. You could do a general walkthrough of the app and ensure the strings are still in place. (We did this manual QA in code review as well).

kathyccheng:
Closing this, will file new issues if I find them

PR #2582, opened 2022-03-23
refactor: flatten translations strings, separate partners strings

emilyjablonski:
# Pull Request Template

## Issue Overview

This PR addresses #2514

Remaining work: jurisdiction uptake, removing all plaintext strings from partners

## Description

* Flattened all of our translations files from their object format for usability - added a script to do this so we can use it in jurisdictions
* Separated out strings that are only used in partners to the partners english strings override file (did this manually 😭)
* Removed unused keys that do not have translations, kept in unused keys that do have translations
* Removed duplicate translations where possible (some are duplicated across preferences, where we are dynamically getting the key based on the preference prefix, so removing these duplicates would be a bigger change) - added a script to find duplicate values
* Updated the missing translations script to account for the new flattened format (here are the [missing public strings](https://docs.google.com/spreadsheets/d/1B063YeMOt2vakl1uZxsLvbAI8EwfbdV8B_z_uNKtiJw/edit#gid=0) off of this PR)
* Removed a few plaintext strings from partners, but my recommendation is to do this as we go from now on as we see them
* Added a script to check if there are keys in foreign translation files that do not exist in the english file

Uptaking this in jurisdictions might be a little tricky since there are some jurisdictionally specific strings that may have been added to the general file. My recommendation would be to run the flatten script over there, pull out the same strings to the partners file based on our core partners overrides, and check for differences to see if there are any strings added to the english file. Could be a little tedious, but we shouldn't just keep the same core files without checking.

I think this might be pretty challenging to review. Since pulling out the partners strings required lots of manual checking, it's possible I moved a string into the partners overrides file that is still used in public. I did run through all the strings twice, but that seems challenging to test. You could review the flatten script and ensure it's pulling all the strings through.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


YazeedLoonat (changes requested):
Hey @emilyjablonski I really like these changes
only thing I found on an application's detail page on the partner's side we have a few "missing translations" pop up in the console

emilyjablonski:
@YazeedLoonat That's an existing issue not to do with translations where the minimal table accepts translations keys and not pre-translated values - I think we're actually doing that in a lot of places, so I would love to handle that in a separate ticket, but let me know what you think!

YazeedLoonat (approved):
LGTM,
@emilyjablonski yeah I think we should spin up that separate ticket for it as that seems like we want to be able to test that separately

seanmalbert (approved):
@emilyjablonski ,

LGTM, I wasn't sure if it makes sense to keep `reformat-strings` now, so you may want to remove that.

emilyjablonski:
@seanmalbert I think we will want it for when we uptake into jurisdictions, since we will need to do some massaging of the general files there as well for strings we added just in that jurisdiction

Issue #2576, opened 2022-03-22
Base Email Layout Template


seanmalbert:
**What is this feature or what feature is this part of?**
Emails with an HTML layout and images.

**Describe the solution you'd like**
We currently have very basic email templates found in `backend/core/src/shared/views` and it makes sense to have a more robust wrapper that includes HTML. This is accomplished in Detroit with templates like [`backend/core/src/views/new-listing.hbs`](https://github.com/CityOfDetroit/bloom/blob/dev/backend/core/src/views/new-listing.hbs).

Using the email templates created by @slowbot  in https://github.com/bloom-housing/bloom/pull/2556, we can create a base layout.

If we have a need for images in the email that are for the header or footer or something that's static for a jurisdiction, I think we can accomplish that in a number of ways. Two options that I'm thinking of now are:
1. store and use assets from `sites/public/public/images` or `sites/partners/public/images`. So if the layout requires a header and footer image, we could add default images to core like `email-header` and `email-footer`. On each public site, we would then need to replace the defaults with what the jurisdiction provided. This approach also means that if we're using this template for partners, then emails would use the same image for all jurisdictions, which may be ok.
2. we store the images in Cloudinary and have the URLs as jurisdictional settings. This approach has a couple of benefits over option one: it allows for emails sent from the partners portal to have different images, depending on the user's jurisdiction and it allows us to update the email images without having to re-deploy a site.

A mix of the two may be the answer depending on what we need to accomplish.

jaredcwhite:
Note: working on this in conjunction with #2576 for a single PR.

seanmalbert:
This is merged and QA can be done with #2513 .

Issue #2513, opened 2022-02-16
Update Confirmation Email - Template


kathyccheng:

kathyccheng:
@slowbot can we have a version of the confirmation that's for a Lottery? Also should the contact information be just the phone number or phone number and email if they provided that?

kathyccheng:
I think that should be fine. Can we still adjust content jurisdiction by jurisdiction?

jaredcwhite:
I think that can be done via the translations system, aka there'd be the "general" translation and then a per-jurisdiction translation(s)

jaredcwhite:
@kathyccheng @slowbot So this is getting pretty close now — a few things I still need to pin down:

- I was able to snip out an image of the Alameda logo from the Figma mockup (attached for reference), but I'll need the SJ and SMC logos as well. (FYI they should all be the same pixel dimensions, otherwise we'd have to store width/height pixel values for each jurisdiction.)
- What should the link be for **Read about how you can prepare for next steps** in each jurisdiction?
- As mentioned above, need alternate copy for Lottery (only have FCFS).


<img width="300" alt="alameda-email-logo" src="https://user-images.githubusercontent.com/658496/161353331-440042b8-ad83-4ccc-b844-5bf8d4d0bdec.png">

slowbot:
@jaredcwhite 
Lottery text https://www.figma.com/file/tUtgsJO8XnF9Ij2dwu22EF/bloom-public-general?node-id=1799%3A7983

Logos
![sanmateo](https://user-images.githubusercontent.com/13915/163838485-87dd8976-b816-424c-a303-93e5473e931e.png)
![sanjose](https://user-images.githubusercontent.com/13915/163838487-7279a41f-4ec5-4da0-818b-4df3351a7971.png)
![alameda](https://user-images.githubusercontent.com/13915/163838489-d5a1bc08-7d69-4c4a-8a94-8485617d8b46.png)

jaredcwhite:
Thanks @slowbot!

jaredcwhite:
Finally working again on this!

seanmalbert:
For QA, see @slowbot's comment, https://github.com/bloom-housing/bloom/pull/2594#issuecomment-1149128474, to create a user with the email address [exygy.33c7-42mwk@litmusemail.com](mailto:exygy.33c7-42mwk@litmusemail.com), for litmus testing.

slowbot:
[...]

PR #2594, opened 2022-03-29
2576/new email templates

jaredcwhite:
## Issue Overview

This PR addresses #2576 & #2513 

## Description

This adds a base HTML layout which can be used for various types of designed emails going forward. The first type of email is the confirmation email sent out when an application is submitted.

TBD

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


jaredcwhite:
@seanmalbert I'm not sure what the procedure is for testing this. I haven't actually gotten emails sent locally yet… do I need to get a developer Sendgrid key?

backend/core/src/email/email.service.spec.ts
```diff
@@ -94,27 +101,49 @@ describe("EmailService", () => {
       language: Language.en,
       translations: {
         confirmation: {
-          yourConfirmationNumber: "Here is your confirmation number:",
+          gotYourConfirmationNumber: "We got your application for",
+          yourConfirmationNumber: "Your Confirmation Number", // UPDATED
```
emilyjablonski, line 105: Do we need the comment here?

backend/core/src/email/email.service.spec.ts
```diff
@@ -143,6 +173,22 @@ describe("EmailService", () => {
...
+    })
+
+    await translationsService.create({
+      jurisdiction: {
+        id: jurisdiction.id,
+      },
+      language: Language.es,
+      translations: {
+        confirmation: {
+          yourConfirmationNumber: "SPANISH NUMBER", // UPDATED
```
emilyjablonski, line 188: It seems like we likely don't want this string and the one on 191 to include `SPANISH` ?

backend/core/src/email/email.service.ts
```diff
@@ -186,19 +199,41 @@ export class EmailService {
...
-    )
-    const genericTranslations = await this.translationService.getTranslationByLanguageAndJurisdictionOrDefaultEn(
-      language,
+    let jurisdictionalTranslations: Translation | null,
+      genericTranslations: Translation | null,
+      jurisdictionalDefaultTranslations: Translation | null
+
+    if (language != Language.en) {
+      if (jurisdiction) {
+        jurisdictionalTranslations = await this.translationService.getTranslationByLanguageAndJurisdictionOrDefaultEn(
```
emilyjablonski, line 208: I think we can dry this up with the block on 220 so we don't have to call it twice

emilyjablonski (changes requested):
A couple visual notes:
* The line that connects the three progress circles breaks in between them
Screen Shot 2022-05-26 at 1 28 16 PM
* The button text should be white here 
Screen Shot 2022-05-26 at 1 28 26 PM

backend/core/src/email/email.service.ts
jaredcwhite: Hmm, maybe we should try to clean this up on the translationService level, because the issue I was seeing before was a lack of proper fallback — aka it wasn't going through the flow of 

1. Look up specific jurisdiction and specific language
2. Not found? look up specific jurisdiction and English language
3. Not found? look up non-jurisdiction and specific language
4. Not found? finally look up non-jurisdiction and English language

If we try to DRY up this section we lose some of those fallbacks. Thoughts?

emilyjablonski: Good to manage some cleanup within the service later 👍

jaredcwhite:
I think the visual issues are resolved now. Here are a couple of screenshots from my phone after receiving the email:

IMG_4813
IMG_4814

emilyjablonski (changes requested):
Screen Shot 2022-06-06 at 11 08 23 AM
Seeing the translations come through which is great! But I'm also seeing the `sr-only` strings come through - and it looks like the lines are still not quite reaching across. I'm seeing that you didn't get the `sr-only` strings in your screenshot, so lmk if you can't repro.

jaredcwhite:
@emilyjablonski Argh, I'm thinking we'll need to look into differences between email clients, could be some of the annoying rendering differences that come up.

emilyjablonski:
@jaredcwhite Sounds good! Fwiw I was using Chrome/Gmail on desktop, and the gmail app on an iPhone.

emilyjablonski:
@slowbot None of the bugs above are specific to dark mode as a note

emilyjablonski (approved):

Issue #2396, opened 2022-01-13
My Applications should show an indication when it is empty, and a spinner while it is loading


emilyjablonski:
**Describe the bug**
When you load my applications under a signed in profile on the public site, nothing displays if it is empty.
Additionally, regardless of many you have, the page sits for second with no feedback as it loads.

**Expected behavior**
We need to have a spinner in the content area while the applications are being fetched and an indication that you have no applications if nothing returns.

Interested in @slowbot input on how both of these states should work.

Loading Component
https://storybook.bloom.exygy.dev/?path=/story/overlays-loading-overlay--default

Blank Slate
https://www.figma.com/file/i3IAkbjv4nVPKg1d1E1g1d/bloom-public-accounts?node-id=1101%3A4139

Loading State
https://www.figma.com/file/i3IAkbjv4nVPKg1d1E1g1d/bloom-public-accounts?node-id=1101%3A4210

QA Notes:
Starting URL: https://dev-bloom.netlify.app/sign-in

Two Main Cases: 
1) Sign in to admin@example.com, click "My Applications", and observe the rendering of the spinner and "no application" message.
2) Sign in to test@example.com, click "My Applications", and observe the rendering of the spinner, loading of applications and their corresponding listings.

Note: For any of the cases, throttling the network when loading the 'My Applications' page can help to observe the frontend behavior.

kathyccheng:
LGTM

PR #2603, opened 2022-03-31
fix: no application message and loading spinner working

ColinBuyck:
# Pull Request Template

## Issue Overview

This PR addresses #issue

## Description
This PR addresses #2396 

This fix displays a message to the user when they haven't completed any applications and provides a link to view listings. 

Additionally, based on my testing the loading spinner was showing initially but there was a slight delay from when it would disappear and when the listings would show. This seemed to be because the loading state would update before the application state updated so by running the spinner until applications length property is defined as well this delay is reduced. On slower networks, however, the delay is still very noticeable.  

Both the "no applications" message and the loading spinner can be tested by going to the My Applications page with the admin@example account which has no applications. The loading spinner can also be tested on the same page with the test@example account which has multiple applications

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


emilyjablonski:
👋 I think this is ready for review based on the assigned reviewers - just a couple reminders! Make sure in the PR template you (1) add the issue number and (2) walk through the checklist and check anything relevant, and then (3) add the `needs review(s)` label to surface it to the team.


https://user-images.githubusercontent.com/65367387/161135308-aa5bc946-6acd-4cfe-a1e6-8d7c0282afef.mov

In terms of the AC, I'm seeing a delay between the loading spinner disappearing and the applications appear. The above image was from testing on the deploy preview. I noticed the delay without any throttling, but it is much easier to test and see if you throttle the network with the dev tools. My initial hypothesis is that it may be we want to check `!applications?.items.length` instead of `!applications` for the `<LoadingOverlay>` component.

ColinBuyck:
Thank you for the reminders! The PR template should be filled out correctly now. In terms of this issue, I added the length adjustment but the delay is still pretty prevalent. I can keep exploring different ideas if that'd be helpful.

Also, do you know of a way to effectively throttle the network on localhost? When I tried it with fast 3G, the page was on track to load in about 10 minutes so I had to test on the netlify preview.

emilyjablonski (approved):
I'm still seeing the loading delay, but I think it's going to be really challenging to fix before we can understand why the local stack isn't throttling correctly. I'm good to merge this with the no applications fix, and we can continue investigating the delay in another issue 👍

seanmalbert (changes requested):
@ColinBuyck and @emilyjablonski ,

I looked into this a little more and I discovered two things:
1. I believe that throttling actually works fine locally, it's that in development mode the resources are so large that under throttling the site will never realistically load. There are steps we can do to mitigate this, but I'm not sure how much it's worth it since we can still test with throttling locally by first running `yarn build` in `sites/public` and then `yarn start`. This makes the resources considerably smaller and better mimics what we're doing in the cloud environments.
2. In doing this, it was easier to see what is going on. We have the `LoadingOverlay` component and that's waiting for applications to load and does so correctly. However, that's wrapping `StatusItemWrapper`, which upon receiving applications, queries listings and this is the result in the delay between the loading overlay finishing and applications appearing. It seems that by moving `LoadingOverlay` to `StatusItemWrapper` and waiting for listings to load, we will have our bases covered.

seanmalbert (changes requested):
@ColinBuyck ,

From our last pair on this, here are the changes we discussed:

1. In a new types file under `sites/public/pages/account` export a new interface that extends Application with a new key to house the full listing. This will be used in `sites/public/pages/account/applications.tsx` and `sites/public/pages/account/StatusItemWrapper.tsx` Something like:
  ```
interface AppWithListing extends Application {
    fullListing?: Listing
}
   ```
2. in applications.tsx, update the applications useState type to be `AppWithListing[]`
3. add a new state to keep track of when listings are loading, this should default to true
4. in the return of the applications service where `setApplications` is set, update that to be `apps.items` since the new type is now `AppWithListing[]`
5. We can simplify the second useEffect you have to fetch listings and map them onto the applications. This is most easily done by using Promise.all in conjunction with applications.map. This will roughly look like:
```
void Promise.all(applications.map(async (application) => {
       // fetch listings in here and set full listing on application
       return application
    })
).then((res) => {
    // set listing loading to false
    // set applications
    // set loading to false
}
```
6. at the top of this useEffect hook, we want to return (not continue) if applications is undefined or if applications are defined and listings are not loading
7. at this point you should just be able to set the isLoading prop on LoadingOverlay to `loading`
8. Where applications is getting mapped and returning `StatusItemWrapper`, this can be cleaned up to directly map applications with `key` and `applications` being the only two props passed to `StatusItemWrapper`
9. clean up the check for `noApplicationsSection` below
10. In StatusItemWrapper, set the application key in `StatusItemWrapperProps` to be of type  `AppWithListing`, and get rid of the `listing` key.
11. update the prop values sent to `StatusItem` so that listings are now on the application

Let me know if you need clarity on any of those items.

ColinBuyck:
The most recent commit should reflect all of these directions except for the first and ninth. 

1 - Since the types cannot be alone in a file within pages they seemed more appropriate to be in the file directly than potentially alone in the src file unless it would be a helpful precedent to clean up other files.

9 - I believe the no applications check requires the length check on applications since the applications state is still defined even when none are found. If this was a must-have, then I it would be possible to perform this check in the useEffect to avoid defining applications if none are found.

ColinBuyck:
Note: The Cypress test is failing since dev has "Open" for listing status in the partners listing table instead of "Accepting Applications" which the cypress test is looking for.

seanmalbert (approved):
Looks good, @ColinBuyck . I fixed the partner cypress tests as well.

seanmalbert:
@ColinBuyck , can you please write QA notes for review on the connected issue. Then we're good to merge.

ColinBuyck:
@seanmalbert Just added them. Thanks!

PR #2620, opened 2022-04-05
fix: remove shared-helpers dependency from ui-components

emilyjablonski:
Removes shared-helper and axios dependencies from ui-components

(1) We don’t want shared-helpers to be used in ui-components at all because we need shared-helpers to have our backend as a dependency and we don’t want the backend in the component library, and (2) if we were going to add shared-helpers to the component library it would also need to be added explicitly to the ui-component package json, and since it isn’t, that is what right now is breaking consumers with errors “could not resolve shared-helpers”. The long term fix is to refactor those components to use much more generic props that don’t relate to network status, and to get our build CI job up and running.


YazeedLoonat (approved):
LGTM

Issue #2252, opened 2021-12-01
Application timeout should be made accessible


slowbot:
**Describe the solution you'd like**
Session timeout modal on the public application should be announced to screen readers

**What is the acceptance criteria/definition of done?**
- Add role "alertdialog" and aria-labelledby=title as well as an ID on the modal title set to "title" 

**QA Review Instructions**
- Enable VoiceOver
- Navigate to the short form application
- Allow the app to timeout
- Modal should be announced as alertdialog

https://cccaccessibility.org/web-1/web-developer-tutorials/accessible-timeouts

PR #2625, opened 2022-04-05
feat: application timeout screen reader accessible

adriencyberspace:
# Pull Request Template

## Issue Overview

This PR addresses #2252 

## Description

This makes the application timeout modal more accessible by being screen readable. 

1. Turn on VoiceOver (CMD+F5 on Mac)
2. Wait for application to time out and modal to appear (you can change IDLE_TIMEOUT in public .env to a smaller number to force the modal to open sooner)
3. VoiceOver should read everything on the modal

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/ui-components/src/overlays/Modal.tsx
+++ b/ui-components/src/overlays/Modal.tsx
@@ -31,13 +31,14 @@ export const Modal = (props: ModalProps) => {
   return (
     <Overlay
       ariaLabel={props.ariaLabel || props.title}
+      aria-labelledby={props.ariaLabel || props.title}
       ariaDescription={props.ariaDescription}
       open={props.open}
       onClose={props.onClose}
       backdrop={props.backdrop}
       slim={props.slim}
     >
-      <div className="modal">
+      <div className="modal" role="alertdialog" id={props.title}>
         <ModalHeader title={props.title} />
 
         <section className="modal__inner">


ui-components/src/overlays/Modal.tsx
```diff
@@ -31,13 +31,14 @@ export const Modal = (props: ModalProps) => {
   return (
     <Overlay
       ariaLabel={props.ariaLabel || props.title}
+      aria-labelledby={props.ariaLabel || props.title}
```
emilyjablonski, line 34: I _think_ we want this label on the modal div below and not on the overlay, but I could be wrong here (@slowbot )

jaredcwhite (changes requested):
@emilyjablonski @adriencyberspace 

Yeah, this is sort of a tricky one to tweeze out. FYI, both drawers and modals are types of overlays, so we should start work at the "overlay" level rather than applying a fix to just the `Modal` component.

I think the steps needed would be:

1. The `Overlay` component currently has a prop name mismatch, as `ariaLabel` isn't mapping to `aria-label` but instead is mapping to `aria-labelledby`. So let's rename that prop to `ariaLabelledBy`.
2. We'd then need whatever value is passed to that to be a valid DOM ID. In the issue it says to use "title" but that would break if we had more than one overlay on the page, so we should probably ensure the ID is unique on the page (and never use `props.title` because that's human readable text, not a DOM ID). Either that unique ID could be supplied by the consumer, or we need to manually generate a unique ID somehow just like we do for React keys.
3. Then in both the `Modal` and the `Drawer` components, the `h1` tag in the header would have an `id=` of whatever that `ariaLabelledBy` value is.
4. I'm not sure about `alertdialog` role to be honest — we already have `role="dialog"` on the wrapper div for `Overlay`, so should that change? Here's some documentation: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role
  So maybe in the case of modals `Overlay` should be using `role="alertdialog"` but for drawers it should remain `role="dialog"` – in which case we'll need a role prop which can be passed to the `Overlay` component.

Does that all make sense to you? Let me know if I missed a beat or if you have any questions - thanks!

adriencyberspace:
@jaredcwhite I believe I've made your requested changes above by creating a unique id to pass as aria-labelledby and a role prop to pass "alertdialog" from Modal and "dialog" from Drawer to Overlay.

jaredcwhite (changes requested):
This looks good @adriencyberspace! I only noticed one potential issue…it seems that the IDs change on every component re-render, which is maybe not ideal. (I'm not actually sure if that causes an issue with the screen reader to be honest.)

One thing I thought of is using a hook to keep the value stable. Seems like maybe `useRef` would work, or even just `useState`. What do you think?

slowbot (commented):
See comment

ui-components/src/overlays/Overlay.tsx
```diff
@@ -8,20 +8,23 @@ import { CSSTransition } from "react-transition-group"
...
   slim?: boolean
+  role?: string
 }
 
 const OverlayInner = (props: OverlayProps) => {
   const closeHandler = () => {
     if (props.onClose) props.onClose()
   }
 
+  console.log(props.role)
```
jaredcwhite, line 26: stray log statement

jaredcwhite (approved):
Looks good! 👍 LGTM

slowbot (approved):
@adriencyberspace LGTM

PR #2645, opened 2022-04-08
test: visual test spike

emilyjablonski:
[Spike Documentation](https://docs.google.com/document/d/1a18ZK2Y-QlSZWDHwCmQ2g7un8wacgJDXK1Wniv5_5nc/edit?usp=sharing)

### DIFF
--- a/ui-components/src/notifications/ApplicationStatus.stories.tsx
+++ b/ui-components/src/notifications/ApplicationStatus.stories.tsx
@@ -26,7 +26,7 @@ function formatDateTime(date: Date, showTime?: boolean) {
 const listing = Object.assign({}, Archer) as any
 listing.applicationOpenDate = ""
 let days = 10
-listing.applicationDueDate = dayjs().add(days, "days").format()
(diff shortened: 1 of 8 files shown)


ui-components/src/notifications/ApplicationStatus.stories.tsx
```diff
@@ -5,7 +5,6 @@ import advancedFormat from "dayjs/plugin/advancedFormat"
 import { ApplicationStatus } from "./ApplicationStatus"
 import { ApplicationStatusType } from "../global/ApplicationStatusType"
 import { t } from "../helpers/translator"
-import Archer from "../../__tests__/fixtures/archer.json"
```
emilyjablonski, line 8: Changes here remove dynamic dates, which would produce a different screenshot every time. We also aren't using the listing object in this component anymore, so these are needed updates anyway.

YazeedLoonat:
chromatic being free is really neat, I agree that only running it on updates to ui-components would be ideal

I also think that it would play nice with what we touched on last week (pulling ui-components out onto its own)!

Issue #2622, opened 2022-04-05
Throw error when min occupancy is larger than max occupancy in Add Unit modal


adriencyberspace:
In New Listing --> Add Unit, you are able to successfully add Minimum Occupancy that is larger than its Max Occupancy

The user should receive an error on Save in the Add Unit modal, if min occupancy is larger than max occupancy.

The min - max occupancy displays on the listing page, which could create confusion and misrepresentation of data:

![image](https://user-images.githubusercontent.com/60076088/161831571-889efe15-0c45-4bfa-9644-27fe30996850.png)

adriencyberspace:
QA Notes:

Entry Point: http://localhost:3001/listings/add

Go to New Listing in partners
Click the Add Unit link to open the Add Unit Modal
In the Min Occupancy dropdown, select a number
In the Max Occupancy dropdown, select a number less than the number you chose in Min Occupancy.
Press Save & Exit - you will see errors under both fields

PR #2661, opened 2022-04-13
feat: min max occupancy validation in Add Unit

adriencyberspace:
# Pull Request Template

## Issue Overview

This PR addresses #2622 

## Description

In New Listing --> Add Unit, you were previously able to successfully add a Min Occupancy larger than the Max Occupancy. Now both Min and Max Occupancy have errors when min is greater than max upon attempting to Save & New or Save & Exit.

image

Entry Point: http://localhost:3001/listings/add

1. Go to New Listing in partners
2. Click the Add Unit link to open the Add Unit Modal
3. In the Min Occupancy dropdown, select a number
4. In the Max Occupancy dropdown, select a number less than the number you chose in Min Occupancy.
5. Press Save & Exit - you will see errors under both fields

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


ui-components/src/locales/general.json
```diff
@@ -517,6 +517,8 @@
   "errors.householdTooBig": "Your household size is too big.",
   "errors.householdTooSmall": "Your household size is too small.",
   "errors.lastNameError": "Please enter a Last Name",
+  "errors.minLargerThanMaxOccupancyError": "Min Occupancy must be less than or equal to Max Occupancy",
```
emilyjablonski, line 520: Since these strings only exist in partners, we want them in the partners overrides file so that we don't pay to translate them (`sites/partners/page_content/locale_overrides/general.json`)

emilyjablonski, line 520: Since these strings only exist in partners, we want them in the partners overrides file so that we don't pay to translate them (`sites/partners/page_content/locale_overrides/general.json`)

emilyjablonski (changes requested):
This is looking close! At the moment, when I save a unit without an entry for the min or max occupancy, it saves the value as `1` instead of null (which it also looks to be doing for bathroom and floor), so I can still end up in the below state. I think we don't want to save any value if the user hasn't selected anything.

Screen Shot 2022-04-13 at 5 04 24 PM

sites/partners/src/listings/PaperListingForm/UnitForm.tsx
```diff
@@ -78,6 +80,17 @@ const UnitForm = ({ onSubmit, onClose, defaultUnit, nextId, draft }: UnitFormPro
...
+  const minOccupancy: number = useWatch({
+    control,
+    name: "minOccupancy",
+  })
+
+  const maxOccupancy: number =
+    useWatch({
+      control,
+      name: "maxOccupancy",
+    }) || numberOccupancyOptions
```
YazeedLoonat, line 92: I'm not sure this or operation here is the best way to do this, useWatch acts like other hooks, so you're setting maxOccupancy to be the result of a hook or 10.

I think instead the or should happen where the where the comparison is needed, so in this case lines ~459

YazeedLoonat (changes requested):
looks good, just needs a small tweak I think

sites/partners/src/listings/PaperListingForm/UnitForm.tsx
adriencyberspace: @YazeedLoonat this useWatch is only being used as validation for the minOccupancy field. The number of options in the Select remains constant regardless. So if the user has selected a minOccupancy but not a maxOccupancy, without this, the form throws an error, because minOccupancy > undefined. This 'or' is only to say if nothing has been selected in MaxOccupancy, don't throw an error unless minOccupancy is greater than 10 (which isn't possible in the dropdown).

adriencyberspace: @YazeedLoonat the first thing I tried was `validation={maxOccupancy && { max: maxOccupancy }}`, which threw console warnings. I then tried `validation={{ max: maxOccupancy || numberOccupancyOptions}}` which also threw console warnings. When I moved it here, they went away. I think because `register(validation)` doesn't want conditional statements passed into it. I figured it would be better to handle the logic before it hit the register function.

YazeedLoonat (changes requested):
continued convo about validation

also something I just realized, we should probably have this validation trigger as soon as users make their selections, and not wait until they hit save

emilyjablonski:
@YazeedLoonat

sites/partners/src/listings/PaperListingForm/UnitForm.tsx
```diff
@@ -454,7 +474,15 @@ const UnitForm = ({ onSubmit, onClose, defaultUnit, nextId, draft }: UnitFormPro
...
                 register={register}
                 controlClassName="control"
-                options={numberOptions(10)}
+                options={numberOptions(numberOccupancyOptions)}
+                error={fieldHasError(errors?.maxOccupancy)}
+                errorMessage={t("errors.maxLessThanMinOccupancyError")}
+                validation={{ min: minOccupancy }}
+                inputProps={{
+                  onChange: () => {
+                    trigger("maxOccupancy")
```
YazeedLoonat, line 483: I think you also need to trigger the minOccupancy as well

sites/partners/src/listings/PaperListingForm/UnitForm.tsx
```diff
@@ -440,7 +452,15 @@ const UnitForm = ({ onSubmit, onClose, defaultUnit, nextId, draft }: UnitFormPro
...
                 register={register}
                 controlClassName="control"
-                options={numberOptions(10)}
+                options={numberOptions(numberOccupancyOptions)}
+                error={fieldHasError(errors?.minOccupancy)}
+                errorMessage={t("errors.minGreaterThanMaxOccupancyError")}
+                validation={{ max: maxOccupancy || numberOccupancyOptions }}
+                inputProps={{
+                  onChange: () => {
+                    trigger("minOccupancy")
```
YazeedLoonat, line 461: I think you also need to trigger the maxOccupancy as well

YazeedLoonat (changes requested):
Looks good, the validation trigger wasn't checking both the min and max I suggested some changes that should address

seanmalbert (approved):
lgtm

Issue #2623, opened 2022-04-05
Alameda: Embark + South Alameda Preference + Program


kathyccheng:
https://docs.google.com/document/d/16-k8yoRUDJd_JD6GgZ5P5NZkRKAS6CU_OyiMO5vxFyM/edit

https://docs.google.com/document/d/15R_-W1dvl7_ZzxS0REQKvXLGvwCokskQDIV1pKEfUqQ/edit

**Notes**
* Create migration script in housingbay to add preference + program in housingbayarea main branch instead of going to core

**QA NOTES**
Navigate to https://hba-partners-dev.netlify.app/ and verify that under alameda jurisdiction the new programs and preferences are available to add to a listing 

Add program and preferences to a listing (either creating a new one or editing an existing one)

navigate to https://hba-public-dev.netlify.app/ and go through the application process for that listing, you should see the new items and the application should save correctly

kathyccheng:
@YazeedLoonat for the oakland housing authority preferences, can you set the do not show flag on it so it doesn't show on the listing detail page?

and then for the south alameda preference, can you remove this part in parenthes
![Screen Shot 2022-04-25 at 3 08 00 PM](https://user-images.githubusercontent.com/3453407/165157735-98a6e5b9-afc5-40b6-a9fd-f29d6e599419.png)
is

YazeedLoonat:
Hey @kathyccheng this is ready for your 👀 again!

kathyccheng:
LGTM thanks @YazeedLoonat

PR #2696, opened 2022-04-26
feat: adding alameda prog and prefs

YazeedLoonat:
This deploys the program and preference changes 

for: https://github.com/bloom-housing/bloom/issues/2623 
related to: https://github.com/housingbayarea/bloom/pull/398


seanmalbert (approved):

emilyjablonski:
Is this a migration we could just have on Alameda and not core?

seanmalbert:
> Is this a migration we could just have on Alameda and not core?

It would go on the new main branch in hba, but we haven't switched that yet.

PR #2700, opened 2022-04-27
fix: check for empty translations before sending to google translate

emilyjablonski:
Pulling this Detroit fix back into core: https://github.com/CityOfDetroit/bloom/pull/1191


YazeedLoonat (approved):
LGTM

Issue #2172, opened 2021-11-12
Visual Design Bugs


slowbot:
**Modal is black and should be semi transparent**
![image](https://user-images.githubusercontent.com/13915/168108787-ffd8f6bc-7299-4a27-bce1-d46e85e1ff76.png)

**Excess space on right hand side of Features, Neighborhood and Additional Information sections. Should have same space on right as Eligibility, these sections appear to have an extra 2rem applied**
![image](https://user-images.githubusercontent.com/13915/141580679-5c53b99f-c594-4459-82b7-e55ef62f4563.png)

**The is a flash at application start page to load, Would prefer if this loaded at min height and title loaded progressively**
![image](https://user-images.githubusercontent.com/13915/141587671-cc336009-6090-4bb9-a3b0-7377c3521c45.png)

**Inconsistent use of vertical spacing and font weight across checkboxes and radios. I think the tighter spacing and thicker line weight work better**
![image](https://user-images.githubusercontent.com/13915/141590116-3c9c422e-1149-43e1-bd97-f2a7448bf9bd.png)
- Name: I don't have - checkbox
- Contact: Phone - Checkboxes
- Contact: Address - Checkbox
- Contact: Do you work? - Radio
- Alternate contact type - Radio
- Preference Checkboxes, (also has inset issue, inset should match other checkbox and radio groups)

slowbot:
@kathyccheng Oldies but goodies

dominikx96:
@slowbot I fixed most of the issues, but I can't reproduce "The is a flash at application start page to load, Would prefer if this loaded at min-height and title loaded progressively" - are you sure it occurs on staging (when an app is after build)?

slowbot:
@emilyjablonski @dominikx96  Opening a new ticket for the flash element so we close this

dominikx96:
@slowbot fixes have been merged.

QA notes:
1. Open a public site and try to log in with an unconfirmed account, make sure that a modal overlay has opacity
2. Open random listing and make sure that the Features, Neighborhood, and Additional Information sections have the same padding as the Eligibility section
3. Try to fill every application step and make sure that radios/checkboxes have consistent margins and font-weight

slowbot:
@dominikx96 LGTM

PR #2737, opened 2022-05-18
2172/visual design bugs

dominikx96:
# Pull Request Template

## Issue Overview

This PR addresses #2172 #2760

## Description

Fixes visual issues:
- added modal backdrop opacity 50%
- updated checkbox input font weight to normal
- removed features section right padding
- normalized radio input spacing

Just check screenshots from the issue and make sure that the problems have been fixed.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


dominikx96:
@slowbot I fixed most of the issues, but I can't reproduce "The is a flash at application start page to load, Would prefer if this loaded at min-height and title loaded progressively" - are you sure it occurs on staging (when an app is after build)?

ui-components/src/forms/FieldGroup.tsx
```diff
@@ -93,7 +93,7 @@ const FieldGroup = ({
         />
         <label
           htmlFor={item.id}
-          className={`font-semibold ${fieldLabelClassName} ${
```
slowbot, line 96: @dominikx96 I would actually keep this font weight for checkboxes and radios. Not sure if its best to do inline or not.

slowbot (changes requested):
@dominikx96 I don't see the fix for the application start screen. Would this be better to break into a new ticket?

ui-components/src/forms/FieldGroup.tsx
dominikx96: @slowbot updated! I also think it's a good idea to break the loading issue into a new ticket.

emilyjablonski: @dominikx96 I added some additional info into the ticket about the loading issue, and I think it would be a relatively small fix in the `FormCard` component to still render the header styling even if we don't have the header yet. What do you think?

slowbot (approved):
@dominikx96 LGTM. Created a new ticket for the remaining issue

https://github.com/bloom-housing/bloom/issues/2760

ui-components/src/forms/FieldGroup.tsx
dominikx96: @emilyjablonski makes sense! I notice in a couple of places there is no header so I think I need to refactor the `FormCard` props to determine whether the header should be visible and what value should be inside.

dominikx96:
@slowbot @emilyjablonski I refactored the `FormCard` component and the blue header works properly for me. I think we should do a review round before merging.

emilyjablonski (approved):
TYSM for the changes!

Issue #2746, opened 2022-05-20
Core - New "Table Card" Component for Settings 


kathyccheng:
### Story
As a user I want to see multiple tables displayed on the same page

### Properties
- title - table card title
- heading level - title should have an editable heading element, default to H2
- table - should be able to nest a table component
- action - should be able to add additional items to the table (not sure if this is part of this component or part of implementation)

### States
- blank - should support blank slate with message "Add [items] to edit" where items is the based on the card titles
  
### Acceptance Criteria
- should support a blank slate
- should support sortable table
- should support action rows

### Notes
- based on [Minimal Table ](https://storybook.bloom.exygy.dev/?path=/story/tables-minimaltable--default)
- use new css styles

### Mockups
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=5875%3A185029

dominikx96:
QA notes:

1. Open the storybook and find a `TableCard` component
2. Compare 2 variants (blank/with content) with designs: https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=5875%3A185029

slowbot:
@dominikx96 @emilyjablonski Is there a dev storybook I can review for this?

emilyjablonski:
@slowbot This PR is just for the card itself, so 1-3 will be handled separately in #2747. The component just takes children.

PR #2781, opened 2022-06-01
feat: create table card component

dominikx96:
# Pull Request Template

## Issue Overview

This PR addresses #2746

## Description

Adds a table card component: https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=5875%3A185029

1. Open storybook
2. Compare component with designs

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


ui-components/__tests__/tables/TableCard.test.tsx
```diff
@@ -0,0 +1,47 @@
+import React from "react"
+import { render, cleanup, fireEvent } from "@testing-library/react"
+import { TableCard } from "../../src/tables/TableCard"
+
+afterEach(cleanup)
+
+describe("<TableCard>", () => {
```
emilyjablonski, line 7: Great tests!

ui-components/src/tables/TableCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+          if (!elementsQty || elementsQty === 0)
+            return <div className="table-card__blank">{t("t.addItemsToEdit")}</div>
+
+          return children
+        })()}
+      </div>
+
+      <div className="table-card__footer">
+        <Button type="button" onClick={onAddClick} data-test-id="table-card-btn-add">
+          {t("t.addItem")}
```
emilyjablonski, line 32: We're moving away from having translated strings in ui-components as any consumer might have a different translation service. An example of this change would be in the `Contact` component where we can pass a strings object as a prop.

ui-components/src/tables/TableCard.tsx
```diff
@@ -0,0 +1,39 @@
+import React from "react"
+import { Heading } from "../headers/Heading"
+import { Button } from "../actions/Button"
+import { t } from "../helpers/translator"
+import "./TableCard.scss"
+
+export interface TableCardProps {
+  title: string
+  children?: React.ReactNode
+  elementsQty: number // to show/hide an empty state
```
emilyjablonski, line 10: This prop is really tied to our use for this component - that it will add / remove items, but visually this component could be used for other things. Could we toggle that `add items` state on/off through the children? Open to other suggestions!

ui-components/src/tables/TableCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+  onAddClick: () => void
+}
+
+const TableCard = ({ title, children, elementsQty, onAddClick }: TableCardProps) => {
```
emilyjablonski, line 14: As a move toward generalizing our components, I'd suggest a name that doesn't rely on this component having a table as its children. Maybe just something like `Card`?

ui-components/src/tables/TableCard.scss
```diff
@@ -0,0 +1,21 @@
+.table-card__inner {
```
emilyjablonski, line 1: 🥳

ui-components/src/locales/general.json
```diff
@@ -831,6 +831,8 @@
   "t.accessibility": "Accessibility",
   "t.additionalAccessibility": "Additional Accessibility",
   "t.additionalPhone": "Additional Phone",
+  "t.addItem": "Add item",
```
emilyjablonski, line 834: Once these strings are moved out of the component, we can also move them into the partners overrides strings file since they won't be used on the public side.

emilyjablonski (changes requested):
Screen Shot 2022-06-01 at 10 03 47 AM
I have a couple code style comments, and then it looks like we are missing just a few styles from the Figma designs (the row dividers should reach all the way across, less padding, and ensuring it supports the ability to have that action row that has no header text). Let me know if you have any questions!

ui-components/src/tables/TableCard.tsx
dominikx96: Make sense @emilyjablonski ! I think we can follow other components and use the `StandardCard` name, what do you think?

ui-components/src/tables/TableCard.tsx
dominikx96: To make this component more reusable, I can get rid of `Add item` button and just create a prop like `footer: React.ReactNode`. You can pass everything in the parent component (e.g button), but the top padding still is controlled by this card component. What do you think?

dominikx96:
@emilyjablonski thanks for the review! 🚀 I've implemented everything listed above + decided to get rid of the `elementsQty ` prop - we can show/hide an empty state by checking in children exists. I assume that a table content will be controlled by `useState`, so we can render the table conditionally depending on state length 🤔

ui-components/src/tables/TableCard.tsx
emilyjablonski: That sounds great!

jaredcwhite:
Hey @dominikx96, awesome to see the updated 2nd gen design tokens getting worked in! I'm wondering if we could add a `.standard-card` class itself to the stylesheet and add a few specific tokens/variables for the component? For example, the border radius and border width/color are currently hardcoded into the stylesheet, rather than coming from component-specific default tokens which could be overridden by a consumer. I guess it depends on which visual aspects we think a consumer might want to customize, but judging from the direction Detroit took with a lot of the components, the answer seems to be "almost everything"… 😅

emilyjablonski (approved):
This looks great!! Would just want to checkin with Jared about his comments before merge 👍

dominikx96:
> 

Hey @jaredcwhite , make sense! I checked 2nd theming docs again and decided to update component variables + story docs. I love this approach, it's much cleaner!

jaredcwhite (approved):
Thanks @dominikx96, looks good to me! I wonder if the padding should also get tokenized, but otherwise LGTM. 👍

Issue #2801, opened 2022-06-08
Component - Multi-Select 


kathyccheng:
**Mockup**
https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=6071%3A181866

**Reference**
https://mynamesleon.github.io/aria-autocomplete/

**Elements**
- Type to search should trigger autocomplete
- Show all control provides dropdown of items
- Results for autocomplete or show all should utilize lazy loading to remain performant
- Items rendered should be dismissible

**Accessible**
- NPM package uses ARIA attributes, custom screen reader announcements, and has been tested with assistive technologies
- Progressive enhancements falls back to select or checkboxes

jaredcwhite:
@kathyccheng @slowbot I'm getting a PR ready which uses that autocomplete library, and the use case of the top row of the mockup is supported (seeing a simple list, showing selected options as tags). But the bottom row which has on/off checkboxes within the list doesn't seem to be supported, we'd have to build that manually (and likely hack the library to do so). Was that just a potential example, or do we require that style along with the top tags style?

slowbot:
@jaredcwhite The checkboxes example was a reference only. You are building the right one.

PR #2844, opened 2022-06-25
2801/multi-select combo box component

jaredcwhite:
## Issue Overview

This PR addresses #2801 

## Description

This PR implements the control in the Figma mockup linked in the issue. The multi-select field will accept a `dataSource` prop which can be a function with returns a promise, making it ideal for lazy-loading content from the server. Or you can provide a pre-loaded dataset. (Basically any of the source options allowed by the aria-autocomplete library.)

You can preview the component in Storybook: https://deploy-preview-2844--dev-storybook-bloom.netlify.app/?path=/story/forms-multiselect-field--standard

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/ui-components/__tests__/forms/MultiSelectField.test.tsx
+++ b/ui-components/__tests__/forms/MultiSelectField.test.tsx
@@ -1,0 +1,61 @@
+import React from "react"
+import { fireEvent, render, cleanup, waitFor } from "@testing-library/react"
+import { MultiSelectField, MultiSelectFieldItem } from "../../src/forms/MultiSelectField"
+import { useForm } from "react-hook-form"
+
+afterEach(cleanup)
+
+const staticData: MultiSelectFieldItem[] = []
+for (let i = 0; i < 10; i++) {
+  staticData.push({
+    value: `item${i + 1}`,
+    label: `Item ${i + 1}`,
+  })
+}
+
+const DefaultMultiSelectField = () => {
+  const { register, getValues, setValue } = useForm({ mode: "onChange" })
+
+  return (
+    <MultiSelectField
+      dataSource={staticData}
+      label="Select one or more items"
+      placeholder="Type to search"
+      id="the-value"
+      name="selectedOptions"
+      register={register}
+      getValues={getValues}
+      setValue={setValue}
+    />
+  )
+}
+
+describe("<MultiSelectField>", () => {
+  it("shows a dropdown list and a selected item", async () => {
+    const { getByLabelText, getAllByRole, findByRole } = render(<DefaultMultiSelectField />)
+
+    expect(getByLabelText("Select one or more items")).not.toBeNull()
+
+    // Click the dropdown button
+    const dropdownButton = await findByRole("button")
+    fireEvent.click(dropdownButton)
+
+    // Wait for results to render
+    await waitFor(
+      () => {
+        expect(getAllByRole("option")[0]).not.toBeNull()
+      },
+      { timeout: 1000 }
+    )
+
+    // Click the second item in the list
+    const option = getAllByRole("option")[1]
+    expect(option.textContent).toEqual("Item 2")
+    fireEvent.click(option)
+
+    // Verify the item is now shown in the selected tags area
+    const buttons = getAllByRole("button")
+    expect(buttons.length).toBe(2)
+    expect(buttons[buttons.length - 1].textContent).toEqual("Item 2")
+  })
+})
--- a/ui-components/src/forms/MultiSelectField.scss
+++ b/ui-components/src/forms/MultiSelectField.scss
@@ -6,7 +6,7 @@
 .multi-select-field {
   --field-height: 48px;
   --list-height: 325px;
-  --list-background-color: #fff; 
+  --list-background-color: #fff;
   --tag-hover-background-color: var(--bloom-color-primary-dark);
 
   .control input.input.multi-select-field__input {
@@ -114,13 +114,13 @@
   cursor: pointer;
 }
 
-.multi-select-field__option:hover[aria-disabled='true'],
-.multi-select-field__option--focused[aria-disabled='true'] {
+.multi-select-field__option:hover[aria-disabled="true"],
+.multi-select-field__option--focused[aria-disabled="true"] {
   background: #f5f5f5;
 }
 
 .multi-select-field__option--no-results,
-.multi-select-field__option[aria-disabled='true'] {
+.multi-select-field__option[aria-disabled="true"] {
   color: var(--bloom-color-gray-750) !important;
   cursor: not-allowed !important;
   text-decoration: none !important;
@@ -157,7 +157,7 @@
 }
 
 .multi-select-field__selected:after {
-  content: '';
+  content: "";
   position: absolute;
   top: 50%;
   right: 7px;
@@ -191,7 +191,7 @@
 }
 
 .multi-select-field__wrapper--loading:after {
-  content: '';
+  content: "";
   position: absolute;
   top: 0;
   right: 0;
--- a/ui-components/src/forms/MultiSelectField.scss
+++ b/ui-components/src/forms/MultiSelectField.scss
@@ -1,0 +1,247 @@
+.multi-select-field {
+  --field-height: 48px;
+}
+
+.multi-select-field__wrapper {
+  position: relative;
+}
+
+.field .control input.input.multi-select-field__input {
+  padding-right: calc(var(--field-height) - 8px);
+}
+
+.multi-select-field--hide {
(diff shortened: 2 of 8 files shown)


jaredcwhite:
@slowbot The Storybook story is at a good point to try out…I'd still want to write up the documentation before merge.

slowbot (commented):
@jaredcwhite See comments

ui-components/src/forms/MultiSelectField.tsx
```diff
@@ -0,0 +1,67 @@
...
+        cssNameSpace: "multi-select-field",
+        onChange: (selected) => {
+          setValue(
+            name,
+            selected.map((item) => item.value)
+          )
+        },
+      })
+    }
+  }, [autocompleteRef, name, setValue, props.dataSource])
```
kramduckner, line 44: I'm curious how we handle this in all our components - Without having a check to see if props.dataSource exists, is it possible that this useEffect will get run multiple times, for example, before and after data is fetched? Should there be a check to see it exists before attaching it to the ref? Or is that state handled in the parent component, and it's assumed you'll always have data on the first mount?

ui-components/src/forms/MultiSelectField.scss
```diff
@@ -0,0 +1,220 @@
...
+  padding: var(--bloom-s3) var(--bloom-s4);
+  overflow-wrap: break-word;
+}
+
+.multi-select-field__option:focus {
+  outline: none;
+  box-shadow: none;
+}
+
+.multi-select-field__option:hover,
```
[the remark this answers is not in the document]
jaredcwhite: Interesting question…arguably they're both a list of items, although in one case it's a list of links, and in another a list of menu items. Unfortunately I don't have much control over the exact class names—the aria-autocomplete library is generating a bunch of the names. But I could see us defining some higher-level design tokens and using those as variables for this and other similar lists.

ui-components/src/forms/MultiSelectField.tsx
jaredcwhite: I think the idea is that either you have data on the first mount, or you supply an async function so the data is loaded in real-time. We should make sure `useEffect` doesn't run more than once, unless there's a major change to those form control details. The `dataSource` prop is definitely required, otherwise I believe the aria-autocomplete control won't work correctly.

jaredcwhite:
Sounds good @slowbot, I'll work on those changes.

jaredcwhite:
Sorta lost track of this over the holiday break…will try to wrap it up again.

dominikx96 (changes requested):
@jaredcwhite see comments above and:

- It would be great to have tests for this component
- a tag has incorrect close icon color on hover:
Screenshot 2022-07-11 at 12 47 30

jaredcwhite:
All right, looks like all relevant tests/builds that need to be passing are passing now. I believed I've addressed all feedback except adding a unit test for the component. I'm not sure exactly the best way to do that, since nearly all interactions are actually managed by aria-autocomplete.

ui-components/src/forms/MultiSelectField.tsx
```diff
@@ -0,0 +1,92 @@
...
+    // capabilities (and it doesn't really make sense to SSR the control anyway)
+
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    ;(async () => {
+      if (autocompleteRef.current) {
+        autocompleteRef.current.value = props.getValues(name)
+        const AriaAutocomplete = (await import("aria-autocomplete")).default
+        AriaAutocomplete(autocompleteRef.current, {
+          source: props.dataSource,
+          delay: 500, // debounce for a half-second
```
emilyjablonski, line 53: There is no need to debounce on the first key press

ui-components/src/forms/MultiSelectField.scss
```diff
@@ -0,0 +1,232 @@
...
+  background-position: center center;
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' width='10' height='10' viewBox='0 0 10 10' version='1.1'%3E%3Cg id='surface1'%3E%3Cpath d='M 0.234375 10 L 5 5.21875 L 9.765625 10 L 9.988281 9.78125 L 5.21875 5 L 9.988281 0.21875 L 9.765625 0 L 5 4.78125 L 0.234375 0 L 0.0117188 0.21875 L 4.78125 5 L 0.0117188 9.78125 Z M 0.234375 10 ' style='stroke:%230077da;fill-rule:nonzero;fill:%23ffffff;fill-opacity:1;'/%3E%3C/g%3E%3C/svg%3E");
+  transform: translateY(-50%);
+}
+
+.multi-select-field__selected:hover:after {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' width='10' height='10' viewBox='0 0 10 10' version='1.1'%3E%3Cg id='surface1'%3E%3Cpath d='M 0.234375 10 L 5 5.21875 L 9.765625 10 L 9.988281 9.78125 L 5.21875 5 L 9.988281 0.21875 L 9.765625 0 L 5 4.78125 L 0.234375 0 L 0.0117188 0.21875 L 4.78125 5 L 0.0117188 9.78125 Z M 0.234375 10 ' style='stroke:%23ffffff;fill-rule:nonzero;fill:%23ffffff;fill-opacity:1;'/%3E%3C/g%3E%3C/svg%3E");
+}
+
+@media speech {
```
emilyjablonski, line 184: can you walk me through what the speech query does? not finding great resources online

ui-components/src/forms/MultiSelectField.tsx
```diff
@@ -0,0 +1,92 @@
...
+export type MultiSelectDataSourceParams<T> = (
+  query: string,
+  render: (toRender: MultiSelectFieldItem[]) => void,
+  isFirstCall: boolean
+) => T
+
+export interface MultiSelectFieldProps {
+  name: string
+  dataSource:
+    | string
```
emilyjablonski, line 20: why not just always require a string[] ?

ui-components/src/forms/MultiSelectField.tsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const MultiSelectField = (props: MultiSelectFieldProps) => {
+  const autocompleteRef = useRef<HTMLInputElement>(null)
+
+  const { name, register, setValue } = props
+  register({ name }, props.validation)
+
+  useEffect(() => {
+    // We need to dynamically import the aria-autocomplete control only on the
```
emilyjablonski, line 42: Does this make it so this component does not work out of the box for consumers?

emilyjablonski:
A few questions! I also think it is important to add a simple test (does this component render, can I add an item and remove an item) to ensure it is working appropriately and that we are consuming the api appropriately.

ui-components/src/forms/MultiSelectField.scss
jaredcwhite: @emilyjablonski I have no idea TBH. 😅 Some of this stylesheet was pulled from the aria-autocomplete example.

jaredcwhite:
@emilyjablonski I was able to add a unit test, and also I lowered the debounce value a bit (plus actually the data source in Storybook was simulating a 500ms request delay, so I trimmed that down to 100ms). Let me know what you think!

jaredcwhite:
(Oh and I added a new static data source example with a large number of items to scroll.)

ui-components/__tests__/forms/MultiSelectField.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+      id="the-value"
+      name="selectedOptions"
+      register={register}
+      getValues={getValues}
+      setValue={setValue}
+    />
+  )
+}
+
+describe("<MultiSelectField>", () => {
```
emilyjablonski, line 33: Love this testt

emilyjablonski (commented):
This is looking so awesome!! Thank you for the changes. Everything above looks great, I just have one thing I didn't catch before (sorry about that!) which is the loading state - if I check out the Storybook preview on a throttled connection here is the loading state:

Screen Shot 2022-07-20 at 12 28 28 PM

slowbot (approved):
@jaredcwhite LGTM!

jaredcwhite:
@emilyjablonski Good catch on the `undefined` issue…I was able to track that down and fix, and also make the input height stable so there's not layout shift as the control loads in. Think we're good to go now!

Issue #2792, opened 2022-06-06
Core - Filter Users based off of new roles 


kathyccheng:
**Acceptance Criteria** 
* When I am on the Users page in partners as a Super Admin user type, I should continue to see all users for any jurisdiction and the ability to add users as usual 
* As a jurisdictional admin user type, I should be able to see the Users page
* As a jurisdictional admin user type, I should be able to see all of the users in my jurisdiction only. I SHOULD NOT be able to see super admins 
* As a jurisdictional admin user type, if I have a user in my jurisdiction who is in multiple jurisdictions, then I'll see all of those listings next to their name on the table 

**Notes**
* Assume back end filtering handled in Michal's PR

**QA NOTES**
Create/edit a user and give them the jurisdictional admin role, and 1 listing (for now)

log in as that user, and verify that the user can view all the partners/jurisdictional admin that are tied to the same jurisdiction of the listing the logged in user is tied to

log out as the new user

log in as a super admin user, verify that you view all users across the jurisdictions

kathyccheng:
LGTM

PR #2848, opened 2022-06-28
2792/juris admin user list filtering

YazeedLoonat:
## Issue Overview
This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/2792)

## Description
This adds user list filtering by by jurisdictional admin/super admin rules

Create/edit a user and give them the jurisdictional admin role, and 1 listing (for now)

log in as that user, and verify that the user can view all the partners/jurisdictional admin that are tied to the same jurisdiction of the listing the logged in user is tied to

log out as the new user

log in as a super admin user, verify that you view all users across the jurisdictions 

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend/core/src/auth/services/user.service.ts
```diff
@@ -562,12 +567,27 @@ export class UserService {
...
+    // jurisdictional admins can't view super admins
+    if (targetUser.roles?.isAdmin) {
+      throw new HttpException("Forbidden", HttpStatus.FORBIDDEN)
+    }
+
+    const requesterJurisdictions = requestingUser.jurisdictions?.map((juris) => juris.id)
+    const targetJurisdictions = targetUser.jurisdictions?.map((juris) => juris.id)
+    // jurisdictional admins should only see a user if they share a jurisdiction
+    return requesterJurisdictions.some((juris) => targetJurisdictions.includes(juris))
+  }
```
seanmalbert, line 589: What do you think about leaving authorizeUserAction as it was and updating the use `list` service to add a filter to the query if the user is a jurisdictional admin to get users who are not isAdmin and to only get users who are in the same jurisdiction? And maybe that would be a separate query to get the IDs first and then query by ids.

seanmalbert (changes requested):

YazeedLoonat:
Hey @seanmalbert after we talking during our review on this, I think we got everything addressed/discussed

can you let me know what you think please!

sites/partners/cypress/fixtures/jurisdictionalAdmin.json
```diff
@@ -0,0 +1,10 @@
+{
+  "email": "alameda-admin@example.com",
```
emilyjablonski, line 2: this user is added, but is assigned no listings, which isn't a state we can ever get into in the ui since assigning at least one listing is required - can we add that?

emilyjablonski:
My interpretation of the AC in the ticket is that jurisdictional admin should not be able to add users. I can still see the button in the UI, though the request fails, so I think if that's correct we have to remove the button.

backend/core/src/auth/filters/user-query-filter.ts
```diff
@@ -19,49 +21,47 @@ export class UserQueryFilter extends BaseQueryFilter {
...
-    const userRolesColumnName = userFilterTypeToFieldMap[UserFilterKeys.isPortalUser]
-    if (filterValue == "true") {
+  private addIsPortalUserQuery(qb: WhereExpression, filterValue: string, user: User) {
+    addIsPortalUserQuery(qb, filterValue, user)
+  }
+}
+
+export function addIsPortalUserQuery(qb: WhereExpression, filterValue: string, user: User) {
+  const userRolesColumnName = userFilterTypeToFieldMap[UserFilterKeys.isPortalUser]
+  if (filterValue == "true") {
```
emilyjablonski, line 39: `===`

backend/core/src/migration/1656362047737-user-role-updates.ts
```diff
@@ -0,0 +1,25 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            UPDATE user_roles
+            SET 
+                is_admin = true,
+                is_partner = false,
+                is_jurisdictional_admin = false
+            WHERE is_admin = true AND (is_jurisdictional_admin = true OR is_partner = true);
+
+            UPDATE user_roles
```
emilyjablonski, line 15: The first update will always set jurisdictional admin to false where is_jurisdictional_admin is true, and the second is looking for records where is_jurisdiction_admin is true, which I think due to the first update will never be the case - might be missing something

sites/partners/cypress/integration/06-user-mangement.spec.ts
```diff
@@ -0,0 +1,47 @@
+describe("User Mangement Tests", () => {
+  it("as admin user, should show all users regadless of jurisdiction", () => {
```
emilyjablonski, line 2: `regardless`

sites/partners/cypress/integration/06-user-mangement.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    cy.visit("/")
+    cy.getByTestId("Users-1").click()
+    const rolesArray = [
+      "Partner",
+      "Partner",
+      "Administrator",
+      "Administrator",
+      "Jurisdictional Admin",
+    ]
+    cy.get(`.ag-center-cols-container [row-id="0"] [col-id="roles"]`)
```
emilyjablonski, line 13: No worries if this is challenging to update, but this selector is very tied to implementation details - is there a way to use a more generic selector or a test id?

sites/partners/cypress/integration/06-user-mangement.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    cy.visit("/")
+    cy.getByTestId("Users-1").click()
+    const rolesArray = ["Partner", "Partner", "Jurisdictional Admin"]
+    cy.get(`.ag-center-cols-container [row-id="0"] [col-id="roles"]`)
+      .contains(rolesArray[0])
+      .should("have.text", rolesArray[0])
+    cy.get(`.ag-center-cols-container [row-id="1"] [col-id="roles"]`)
+      .contains(rolesArray[1])
+      .should("have.text", rolesArray[1])
+    cy.get(`.ag-center-cols-container [row-id="2"] [col-id="roles"]`)
```
emilyjablonski, line 42: This test isn't 100% testing what it indicates. It is testing that with our current data, the first 3 rows aren't admins, and we see them. This test would pass if we had a fourth row that had an admin. It works for this data set, but if we hypothesize that this functionality breaks in the future on a different data set, and we do show admins on accident, and it's any future row, this test is a false positive. Is there a way we can verify not that we are successfully seeing partners / jurisdictional admins, but explicitly that we are not seeing super admins? In general testing this might make more sense as a unit test of the filtering, rather than a full integration test.

emilyjablonski (changes requested):
See notes above!

sites/partners/cypress/fixtures/jurisdictionalAdmin.json
emilyjablonski: A flow I'm seeing is that this user when they log in, no terms load - maybe because they have no jurisdiction assigned? Not sure what is happening there

sites/partners/cypress/integration/06-user-mangement.spec.ts
YazeedLoonat: I think we are somewhat limited as we are using the `ag-grid-react` package, I couldn't find a better way to target rows/column specific data

sites/partners/cypress/integration/06-user-mangement.spec.ts
YazeedLoonat: nice catch ty

backend/core/src/auth/filters/user-query-filter.ts
YazeedLoonat: nice catch ty!

YazeedLoonat:
Hey @emilyjablonski can you take another peek at this when you get another chance

emilyjablonski:
@YazeedLoonat Do you mind rebasing so that I can add users with the email fix?

emilyjablonski (approved):
Looking good! To confirm since the AC is split up, this is what I tested: adding a new jurisdictional admin & assigning them to certain listings results in them being able to see users from that jurisdiction

seanmalbert (approved):

PR #2843, opened 2022-06-25
Core Utilities/Additional Fees Update

ColinBuyck:
# Pull Request Template

## Issue Overview

This PR mirrors the work done on Detroit [here](https://github.com/CityOfDetroit/bloom/pull/1345). 

Its adds the option (not required) to indicate which utilities are included in the cost of rent. This is seen on the partners form fields, the listing detailed view on the partners side, and the detailed listing view on the public side. The public side representation in the AdditionalFees component required a slight prop refactoring so that a user can pass in an array of ReactNodes or strings to be displayed in a 2 column-format. This is done to maintain alignment between the AdditionalFees.tsx component in Detroit and the one in Core. Note, there were other misalignments between these two components such as the handling of Deposit text, but it felt outside of the scope of this PR.

On the backend, this PR required a migration to add the new 'utilities' table to the model along with multiple updates to various backend type files. Additionally, translations and tests were added to complete and verify these changes.

## What's Unique to Core's Implementation
Since no jurisdiction has requested this feature yet, I added a feature flag called enableUtilitiesIncluded. This allows for all the aforementioned functionality to be switched off based on the selected jurisdiction of the listing. Additionally, I also added the ability for jurisdictions to be updated to reflect any and all changes of the default jurisdiction data. This allows for us to quickly test out different jurisdiction configurations through seeding. 

## How Can This Be Tested/Reviewed?

Testing on Partners with Flag True:
Part 1: This can be tested by selecting a listing that is part of the Alameda jurisdiction, clicking edit, and scrolling down to ensure that the new utilities form fields are available and clickable. After selecting a variety of utilities, click 'save & exit', and then ensure that they are on the detailed view.

Part 2: Go back to edit the same listing, unselect all options for utilities, and then save the listing again. The 'Utilities Included' title and bullets should be no longer visible.

Testing on Partners with Flag False:
Part 1: This can be tested by creating a new listing and setting the jurisdiction to Detroit. Scrolling down you will notice that the new utilities form fields are unavailable and clickable

Testing on Public with Flag True:
Select the Alameda listing that you recently marked as having included utilities and observe the bulleted list in the 'Additional Fees' Component. If other 'additional fee' messages are available for the listing, the section will be in a 2-column format (with the exception of the last row which will fill if there is an odd number of footer info pieces).

Testing on Public with Flag False: 
[...]

PR #2875, opened 2022-07-11
fix: fix email from address

emilyjablonski:
At the moment on dev you are unable to send emails. This was an existing data issue surfaced by this [commit](https://github.com/bloom-housing/bloom/pull/2843/files#diff-68c2a0379eb7c1ee8cc5467d1333b6f7fe2eca6ee9a12c2b5dfe162180d358a3R69). We slightly changed how jurisdictions are being seeded such that the mock data in that file is now overwriting any jurisdictions we had added in the migrations. The migration files had the correct `emailFromAddress` but this file did not, so I just updated the values.


YazeedLoonat (approved):
LGTM

Issue #2835, opened 2022-06-22
Core - Set duplicates to run asynch 


kathyccheng:
**Acceptance Criteria**
* Runs once a day instead of on application submission 
* Will need to know when it was last run to render on front end 

Notes:

Right now we run the duplicate check on each application submission. We want to move these checks to a cron job that runs once a day (let's say 11pm). We have a couple of options on how to handle this check:
- fetch applications submitted since the last time the job finished running (we have to store this somewhere anyway per the AC)
- add a flag to the applications table which we set to true when the job has run for that application and fetch applications where the flag is set to false

The benefit of the latter option is it allows us to continue on to the next application check if something goes wrong, rather than stopping the job entirely.

PR #2889, opened 2022-07-13
feat(backend): add bull job scheduler for afs processing

pbn4:
# Pull Request Template

## Issue Overview

This PR addresses #2835 

## Description

This changes moves duplicates into an async worker. The worker is a part of the same process as backend service and uses Redis + Bull.js to schedule jobs to the queue. It uses repeatable jobs and according to docs Bull.js is smart enough to not post the same repeatable job twice into the queue, so there should be no duplicate / overlapping runs.

Processing logic (cc: @kathyccheng):
- processing runs once a day on midnight (cron string `0 0 * * *`) unless configured otherwise with env variable `AFS_PROCESSING_CRON_STRING`
- processing runs for all listings with status active
- on each processing run any existing AFS for a given Listing are removed and recreated
- processing is done in the background of web server process
- processing is also triggered by chaning Listing status from active to any (to cover the case where processing has already run, new applications were added after that and some of them are not included in AFS because processing will not run for closed Listings)

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


seanmalbert (changes requested):
@pbn4 ,

I was thinking that we could cut down the processing further by only running `updateApplicationFlaggedSetsForRule` on applications that have not been processed. We can achieve this by fetching applications that were submitted since the last time the job ran or based off of a flag on the application. In either case, we wouldn't want to remove the flagged sets and only add to them.

We also want to be able to display the last time a job was ran. It's possible we could calculate that based off of the cron string or we could create a new job table to store that.

I thought we could also stop running the jobs once a listing was no longer open, but I discussed this with @kathyccheng earlier today and it's possible that a partner portal user could add applications via the portal after a listing has closed and we would want to flag duplicates if this were the case. If we update the job to only run on newly submitted applications this shouldn't be an issue. We could also consider not running the job on closed listings and always run it on paper applications submitted through the partner portal, if you saw any issues checking all active or closed listings for new applications each job.

pbn4:
> We also want to be able to display the last time a job was ran. It's possible we could calculate that based off of the cron string or we could create a new job table to store that.

Sure what do you think about adding it as a column to the listing table? 

>I thought we could also stop running the jobs once a listing was no longer open, but I discussed this with @kathyccheng earlier today and it's possible that a partner portal user could add applications via the portal after a listing has closed and we would want to flag duplicates if this were the case. If we update the job to only run on newly submitted applications this shouldn't be an issue. We could also consider not running the job on closed listings and always run it on paper applications submitted through the partner portal, if you saw any issues checking all active or closed listings for new applications each job.

Currently job only fetches listings with state equal to active. So to summarize we want the job to run for every listing for every application that was submitted after the job was last run?

What if partner portal user updates the application in a way that triggers/cancels AFS rule? The fact that I'm running everything from scratch as long as listing is active handles this case, but handling per not processed basis makes it much more difficult because the AFS sets already exists and a change to application requires modifying existing sets to create/reverse rules triggered by application updates.

I assumed creating everything from scratch for active listings is an acceptable solutions because we were at some point aligned on the fact that frontend will not display AFS views as long as listing applications are not finished. I understand that now there is a case that someone can add an application to a Listing with status closed, right? 

We could still have it working in a batch by:
- by triggering reprocessing in such case (but that adds unnecessary f if/else logic, complicating the app submission code), 
- or having a boolean flag that controls reprocessing of AFS 
- or giving that control to the user in a form of a button

I'm trying to keep the batch processing of everything from scratch even if it's a bit not optimal because of the simplicity of the code.

seanmalbert:
> Sure what do you think about adding it as a column to the listing table?

That sounds good to me.

> Currently job only fetches listings with state equal to active. So to summarize we want the job to run for every listing for every application that was submitted after the job was last run?
[...]
I'm trying to keep the batch processing of everything from scratch even if it's a bit not optimal because of the simplicity of the code.

I see your point about applications getting updated and keeping the process as simple as possible. I think having a boolean flag that we can use to determine if a closed listing has had an updated application could solve the problem of running it on closed listings, without having to do all listings each time. What do you think?

seanmalbert (approved):
@pbn4 ,
This looks good. I started with some performance updates in how we select listings and was going to continue, but since that afs work was brought over/existing, we can do the rest of that in a follow-up issue. That'll make it easier to track the changes too.

Issue #2935, opened 2022-08-04
Step header component


slowbot:
- **Step header** - header for the individual step. should display the order of the step, total count of steps and step title.

slowbot:
LGTM @ColinBuyck

Issue #2919, opened 2022-07-27
Core - Progress Bar Component Update Eligibility Update


kathyccheng:
As a user I want to be able to see my progress through a multistep process.

Note this as an accessibility update and theming if existing component. https://storybook.bloom.exygy.dev/?path=/story/navigation-progress-nav--default
- Start with accessibility
- Minimal markup changes to ensure less refactoring on current implementation

### Properties
- **Steps** - one or more steps of the process, should have title and state

### States
- **Steps** each step should have three states
   - default
   - current
   - complete 

### Accessibility
- Code progress bar as ordered list
- **Step state** should add sr-only text of step state to the end of step title
- Add aria-current=true to the current list item
- Add aria-label=progress to the component container

### Theme and Options
   - Step 
      - Style
        - Line (new design)
        - Dot (current)
     - Line width
     - Color
        - default
        - active
        - complete
   - Type
      - Font face
      - Font size
      - Font case

### Example
- https://designsystem.digital.gov/components/step-indicator/

### Mockup
https://www.figma.com/file/pIlN9vqLwUgOZYq6I5oM9l/bloom-public-eligibility-wizard?node-id=827%3A7153

**QA Notes**
Two Paths
Start here: https://deploy-preview-2950--dev-storybook-bloom.netlify.app/
1. Search for Progress Nav
2. Test both Default(dot style) and Bar Style by using the magnifying glass to see the components within different screen sizes.

Start Here: https://dev-bloom.netlify.app/applications/start/choose-language?listingId=6365b003-1b3f-446e-b041-1b907a59e15b
1. Begin filling out the application
2. Verify that the Progress Nav component behavior is as intended through the process.

slowbot:
@ColinBuyck LGTM

PR #2939, opened 2022-08-04
2935/step header

ColinBuyck:
# Pull Request Template

## Issue Overview

This PR addresses #2935 

## Description

This PR creates a new step header component which indicates the state of the user's progress and will be a subcomponent of the broader redesign of the header+ progress bar + step header recipe in Detroit.

This can be tested by reviewing the variations of the step header component in Storybook.
## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/ui-components/src/headers/StepHeader.scss
+++ b/ui-components/src/headers/StepHeader.scss
@@ -16,6 +16,7 @@
     font-size: var(--font-desktop-size);
   }
 }
+
 .step-header__circle-number {
   display: flex;
   justify-content: center;
@@ -32,6 +33,7 @@
   margin-right: var(--title-spacing);
   border-radius: 50%;
 }
+
 .step-header__title {
   margin-left: var(--title-spacing);
   font-weight: var(--title-font-weight);
--- a/ui-components/src/headers/StepHeader.scss
+++ b/ui-components/src/headers/StepHeader.scss
@@ -1,0 +1,6 @@
+.circle-number {
+  border-width: 1px;
+  padding-left: 0.25rem;
+  padding-right: 0.25rem;
+  border-radius: 50%;
+}
--- a/ui-components/src/headers/StepHeader.scss
+++ b/ui-components/src/headers/StepHeader.scss
@@ -2,6 +2,7 @@
   --number-color: var(--bloom-color-white);
   --circle-color: var(--bloom-color-primary);
   --circle-size: var(--bloom-s6);
+  --circle-x-padding: var(--bloom-s0_5);
   --title-spacing: var(--bloom-s1_5);
   display: flex;
   flex-wrap: wrap;
@@ -11,6 +12,7 @@
   color: var(--number-color);
   min-width: var(--circle-size);
   height: var(--circle-size);
+  padding: 0rem var(--circle-x-padding);
   margin-right: var(--title-spacing);
   border-radius: 50%;
   text-align: center;
--- a/ui-components/src/headers/StepHeader.scss
+++ b/ui-components/src/headers/StepHeader.scss
@@ -1,6 +1,13 @@
 .circle-number {
-  border-width: 1px;
-  padding-left: 0.25rem;
-  padding-right: 0.25rem;
   border-radius: 50%;
+  text-align: center;
+  background-color: var(--bloom-color-primary);
+  color: var(--bloom-color-white);
+  width: var(--bloom-s6);
+  height: var(--bloom-s6);
+  margin-right: var(--bloom-s1_5);
+}
+.step-label {
+  font-weight: bold;
+  margin-left: var(--bloom-s1_5);
 }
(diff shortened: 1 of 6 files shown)


ui-components/src/headers/StepHeader.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+  title: "Headers/Step Header 🚩",
+}
+
+export const defaultTitle = () => <StepHeader currentStep={1} totalSteps={4} />
+
+export const customTitle = () => <StepHeader currentStep={1} totalSteps={4} stepTitle={"Phases"} />
+
+export const finalStep = () => <StepHeader currentStep={3} totalSteps={3} />
+
+export const doubleDigits = () => <StepHeader currentStep={10} totalSteps={40} />
```
ludtkemorgan, line 15: Are 3 digits in scope? If so, looks like those get cut off 
image.
A potential fix is to switch the circle-number css from "width" to "min-width"

ColinBuyck: Appreciate the feedback! While I couldn't envision it in the current planned usage, it seemed worth it for potential future uses. Let me know what you think about the recent changes.

ui-components/src/headers/StepHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface StepHeaderProps {
+  currentStep: number
+  totalSteps: number
+  stepTitle?: string
+}
+
+const StepHeader = ({
+  currentStep,
+  totalSteps,
+  stepTitle = t("stepHeader.defaultLabel"),
```
emilyjablonski, line 14: We can't add any references to `t` in ui-components because of consumers who don't use translations this way. We'll have to pass the default as we consume the component.

ui-components/src/headers/StepHeader.tsx
```diff
@@ -0,0 +1,25 @@
...
+}: StepHeaderProps) => {
+  return (
+    <div className="step-header">
+      <div className="circle-number">{currentStep}</div>
+      <div>{`out of ${totalSteps}`}</div>
```
emilyjablonski, line 19: We'll need to pass this string in as a prop, so that we don't have any untranslated strings

ui-components/src/headers/StepHeader.scss
```diff
@@ -0,0 +1,23 @@
+.step-header {
```
emilyjablonski, line 1: This is close but not quite the right format for styling v2. A good example is `StandardCard.scss`. We should have no nested classes, one root class with the name of the component, the following classes start with that root, and we have a number of css variables underneath that root specific to this component that we anticipate a consumer might want to override (like here maybe background color, color, min width).

emilyjablonski (changes requested):
A few code style comments, and then a few visual notes:
* Looks like the design has more padding around the circular number than I'm seeing here
* The design has different text for the numbers, it has `1 of 3` instead of `1 out of 3`

From the issue, I would have anticipated that the horizontal lines would have been a part of this issue, is that anticipated for a further issue?

ColinBuyck:
> From the issue, I would have anticipated that the horizontal lines would have been a part of this issue, is that anticipated for a further issue?
Yep, it is covered in this sprint by Issue #2919 @emilyjablonski

ColinBuyck:
Thank you for the detailed feedback @emilyjablonski @ludtkemorgan! I believe all of the suggested changes have been implemented and I also added some padding which prevents the first and last digit from being super tight to the circle edge with three digits but leaves the other cases unchanged. Let me know if I missed anything!

@slowbot Through discussions, a few design questions came up. 
1. For the case of higher step counts (particularly triple digits), three digits will be cut off by the 24 x 24 box so it currently shifts to a pill shape to contain all the digits. There is an example in storybook demonstrating that. Does that look good to you?
2. There seems to be a bit more space between the number and the circle edge in the Figma than what resulted from the implementing the same font size and circle size in Core. I believe it is related to the custom font and line-height used in Figma. Does the current spacing on single digits look good to you or would you like me to pursue closer alignment?

slowbot (dismissed):
@ColinBuyck See comments above

ColinBuyck:
@slowbot The mobile/desktop sizing has been added and the step labels are now being passed in as an array which changes for each step. I also added a check so that if a user were to pass in an array with one label but have more than one step, it repeats that label for each step. Let me know if you agree with the inclusion of that check and if there's any other functionality I can add.

emilyjablonski (approved):
LGTM once Jesse approves!

jaredcwhite (changes requested):
Looks good @ColinBuyck! Can we add a corresponding Markdown docs page for the component with a basic description and info about the customizable CSS variables? (See PageHeader, etc.)

ColinBuyck:
Good point, @jaredcwhite! I just added the documentation and let me know if there's anything else I can improve!

slowbot (commented):
@ColinBuyck LGTM

These are nits  For the story preview 
1. Can you add some spacing above and between your preview buttons. 
2. Can you use small buttons

ColinBuyck:
@slowbot Just updated!

ludtkemorgan (approved):

PR #2948, opened 2022-08-10
fix: replace usage of sanitize-html

kramduckner:
# Pull Request Template

## Issue Overview

This PR addresses #issue

## Description

A [recent pr](https://github.com/bloom-housing/bloom/pull/2914) replaced the dompurify library (which sanitizes html) with sanitize-html which broke DAHLIA (attached screenshot). This fix uses the <Markdown /> component (which we already have access to via [markdown-to-jsx](https://www.npmjs.com/package/markdown-to-jsx) lib) which does some basic sanitization - although it's a bit hard to find out exactly what sanitization it does when the disableParsingRawHTML is set to false, it at least removes script tags! 

Just because this is blocking uptakes on dahlia, it'd be nice to fast track this! I'm open to reworking this with a different, longer-term solution. Also open to any other quick, potentially temporary fixes.

Screen Shot 2022-08-10 at 3 15 25 PM

- Ensure that previews work from partners (it doesn't crash).
- Compare MultiLineAdress components in storybooks (should have a 1:1 visual parity)
- Check a couple examples of <Multiline/> usage in listings (e.g. the component that shows the address in the map on the detailed listing page)
- Updates don't introduce any breaking changes

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/ui-components/package.json
+++ b/ui-components/package.json
@@ -85,7 +85,6 @@
     "@types/react-dom": "^16.9.5",
     "@types/react-text-mask": "^5.4.6",
     "@types/react-transition-group": "^4.4.0",
-    "@types/sanitize-html": "^2.6.2",
     "ag-grid-community": "^26.0.0",
     "ag-grid-react": "^26.0.0",
     "aria-autocomplete": "^1.4.0",
--- a/ui-components/package.json
+++ b/ui-components/package.json
@@ -107,7 +107,6 @@
     "react-tabs": "^3.2.2",
     "react-text-mask": "^5.4.3",
     "react-transition-group": "^4.4.1",
-    "sanitize-html": "^2.7.1",
     "tailwindcss": "2.2.10",
     "ts-jest": "^26.4.1",
     "typesafe-actions": "^5.1.0"
--- a/ui-components/src/helpers/MultiLineAddress.tsx
+++ b/ui-components/src/helpers/MultiLineAddress.tsx
@@ -19,8 +19,8 @@ export interface MultiLineAddressProps {
 const MultiLineAddress = ({ address }: MultiLineAddressProps) => {
   if (!address) return null
 
-  const makeHtmlString = (address) => {
-    const str = ""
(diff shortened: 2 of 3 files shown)


seanmalbert (commented):
Hey @kramduckner ,

I know you got a WIP label on here so I won't comment on the cleanup that needs to happen in what you added (well, aside from that). I'm down with the change as long as there aren't any visual breaking changes. Can you make sure @dominikx96 reviews this update, since he made the original updates to MultiLineAddress.

kramduckner:
Thanks @seanmalbert !

That sounds good. Ready for clean up cleanup comments! I also realized I don't actually know how to test listing previews on a netlify deploy preview - looks like I can't create an account and I think one has to be logged in to public to view listing previews? But this also feels like it's in the territory of the code speaking for itself in terms of stability since we already use that npm package in both bloom and dahlia.

dominikx96 (changes requested):
@kramduckner from the public and partners looks great! I just noticed 2 things:
- there is a `"@types/sanitize-html": "^2.6.2"` package, it should be removed along with the `sanitize-html`.
- In a storybook you can notice `undefined` values - can you check why it renders `undefined` as a string?
Screenshot 2022-08-12 at 17 59 14

Screenshot 2022-08-12 at 17 59 21

kramduckner:
@dominikx96  good catch on the types! I'll remove that. The undefined stuff is weird. It's the way template literals are being parsed through these libraries. I'd like to rework this component and improve that (I don't like ever showing undefined to an end user!) - but I think that should require some input from design (e.g. do we even want to support an address without a city? If there's no state, should we just omit that whole line?), and since this is a blocker, I just tried to make the functionality 1:1 with what was there before. What do you think about just making sure it is 1:1 with what was there, and file a new ticket to handle these undefined states better?

seanmalbert (approved):
lgtm

Issue #2959, opened 2022-08-16
AFS Processing Performance Optimization


seanmalbert:
Created as a follow-up to #2889 

The previous issue moved the AFS checking logic to a cron job with some optimization, but there is a fair amount of performance optimization that remains, particularly in how often we query, what we are querying and how we're processing the return of those queries within and across rules.

YazeedLoonat:
closiong!

PR #2961, opened 2022-08-16
2835/async duplicates processing

seanmalbert:
Creating a new PR, since we reverted the merge of #2889. The migration, which adds a new rule_key, should just need to be updated.


seanmalbert:
The migration file needs to be updated

seanmalbert:
I updated the migration file to set the rule_key for existing AFSs.

emilyjablonski:
@seanmalbert What needs to be tested here?

backend/core/src/application-flagged-sets/application-flagged-sets-cronjob-consumer.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    const afses = await this.afsRepository
+      .createQueryBuilder("afs")
+      .leftJoin("afs.applications", "applications")
+      .select(["afs", "applications.id"])
+      .where(`afs.ruleKey = :ruleKey`, { ruleKey: this.getRuleKeyForRule(newApplication, rule) })
+      .getMany()
+
+    if (afses.length == 0) {
```
YazeedLoonat, line 139:
```suggestion
    if (afses.length === 0) {
```

backend/core/src/application-flagged-sets/application-flagged-sets-cronjob-consumer.ts
```diff
@@ -0,0 +1,293 @@
...
+      await this.afsRepository.save({
+        rule: rule,
+        ruleKey: this.getRuleKeyForRule(newApplication, rule),
+        resolvedTime: null,
+        resolvingUser: null,
+        status: FlaggedSetStatus.flagged,
+        applications: [newApplication, ...applicationsMatchingRule],
+        listing: { id: newApplication.listingId },
+      })
+    } else if (afses.length == 1) {
```
YazeedLoonat, line 149:
```suggestion
    } else if (afses.length === 1) {
```

backend/core/src/application-flagged-sets/application-flagged-sets-cronjob-consumer.ts
```diff
@@ -0,0 +1,293 @@
...
+    switch (rule) {
+      case Rule.nameAndDOB:
+        return await this.fetchDuplicatesMatchingNameAndDOBRule(application)
+      case Rule.email:
+        return await this.fetchDuplicatesMatchingEmailRule(application)
+    }
+  }
+
+  private async fetchDuplicatesMatchingEmailRule(newApplication: Application) {
+    return await this.applicationRepository.find({
```
YazeedLoonat, line 174: something pretty similar to this using the afsrepository was what was causing our backend to crash under larger application loads, so we may want to convert this into a querybuilder that we construct ourselves

backend/core/src/migration/1658992843452-add-afs-related-properties-to-listing.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    // set rule_key for existing afses
+    const afsas = await queryRunner.query(`
+      SELECT application_flagged_set_id, applications_id
+      FROM application_flagged_set_applications_applications
+    `)
+    for (const afsa of afsas) {
+      const afs = await queryRunner.query(
+        `
+        SELECT id, listing_id, rule, rule_key
```
YazeedLoonat, line 29: could this become part of the above query as an inner join so we can cut down on how many times we connect to the db

backend/core/src/migration/1658992843452-add-afs-related-properties-to-listing.ts
```diff
@@ -0,0 +1,77 @@
...
+        FROM application_flagged_set
+        WHERE id = $1
+      `,
+        [afsa.application_flagged_set_id]
+      )
+      if (afs.rule_key !== null) continue
```
YazeedLoonat, line 35: I think this can be a part of the above query in the where clause

YazeedLoonat (changes requested):
looks good, a few change requests

seanmalbert:
@emilyjablonski,

I ended up adding an endpoint so we can manually trigger the process (and one we'll need anyway), which you can call with `localhost:3100/applicationFlaggedSets/process`. You can test by intentionally adding duplicate applications. The logic is the same as before (it was just moved to a different location), where it'll match on name and date of birth or email. Once you've added applications, you can call the endpoint and it should process. The very first time you run it (or any time you run it after a reseed), it will take a while.

When `process` is called it first queries listings to check if this process was last ran before the newest application was added. So if you run it subsequent times without adding any new applications, then it should just do this query and return fairly quickly.

backend/core/src/application-flagged-sets/entities/application-flagged-set.entity.ts
```diff
@@ -18,6 +18,11 @@ export class ApplicationFlaggedSet extends AbstractEntity {
   @IsString({ groups: [ValidationsGroupsEnum.default] })
   rule: string
 
+  @Column({ enum: Rule, nullable: false, unique: true })
```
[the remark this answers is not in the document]
seanmalbert: You're right, this enum doesn't apply to this field.

backend/core/src/application-flagged-sets/application-flagged-sets-cronjob-consumer.ts
```diff
@@ -0,0 +1,293 @@
...
+      outOfDateListing.afsLastRunAt = new Date()
+      await this.listingRepository.save(outOfDateListing)
+    }
+  }
+
+  private async generateAFSesForListingRule(
+    listing: Pick<Listing, "id" | "afsLastRunAt">,
+    rule: Rule
+  ) {
+    const newApplications = await this.applicationRepository.find({
```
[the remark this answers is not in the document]
seanmalbert: Since this logic was carried over, I created a follow-up issue to handle performance optimization: #2959

backend/core/src/application-flagged-sets/application-flagged-sets-cronjob-consumer.ts
seanmalbert: Noting here for #2959

emilyjablonski:
@seanmalbert Would you be available for an in person PR review for this one?

YazeedLoonat (approved):
LGTM

backend/core/src/listings/entities/listing.entity.ts
```diff
@@ -668,6 +668,20 @@ class Listing extends BaseEntity {
   @ValidateNested({ groups: [ValidationsGroupsEnum.default], each: true })
   @Type(() => ListingUtilities)
   utilities?: ListingUtilities
+
+  @Column({ type: "timestamptz", nullable: true, default: "1970-01-01" })
```
emilyjablonski, line 672: I'm not seeing other Date columns with a default, just curious why this column has this default

backend/core/src/listings/listings.service.ts
```diff
@@ -119,6 +121,10 @@ export class ListingsService {
     })
     listingDto.unitsAvailable = availableUnits
 
+    if (listing.status == ListingStatus.active && listingDto.status === ListingStatus.closed) {
```
emilyjablonski, line 124: nbd / existing but `listing` vs `listingDto` could use some more clear names

backend/core/src/seeder/seed.ts
```diff
@@ -381,6 +384,10 @@ async function seed() {
       ])
     }
   }
+  // process duplicate check
+  console.log("processing duplicates check")
```
emilyjablonski, line 388: 🧹

emilyjablonski (approved):
Just some q's otherwise LGTM

backend/core/src/listings/entities/listing.entity.ts
seanmalbert: It has a default set to the beginning of time so that it'll be processed by the check the first time it runs.

backend/core/src/seeder/seed.ts
seanmalbert: I think it may be helpful to keep these in for looking at the logs, at least initially.

Issue #2250, opened 2021-12-01
Ala - Duplicates - Cannot open a new tab in Partners and stay signed in


emilyjablonski:
As a leasing agent, 
I might want to have multiple windows open at once in Partners 

**Acceptance Criteria**
* When I am signed in on Partners as a leasing agent, admin or jurisdictional admin, and I open a new tab that's also in Partners, then I should be logged in on all tabs 
* When I sign out of one, then I should be logged out on all 

**Notes**
* This will require some investigation into the best approach - some ideas below: 
* Move towards a session cookie, we think this currently has something to do with how we do session storage  
* We can also use local storage to persist session storage across tabs 
* Will need to update how we handle timeout for inactivity 

**QA NOTES**
The easiest way to test this is to go through a few of the different log in processes

try to log into the public site (as a non-mfa user) then try opening a new tab. You should remain logged in to both without needing to relog in

try to log into the partner site (as an mfa user and later as a nonmfa user). Try opening a new tab, and you should see that user remains logged in the new tab.

you should also try to make sure that if the user is a jurisdictional admin then the permissions should still work correctly

you should be able to log out

you should be able to create/confirm a new public user

you should be able to create/confirm as a new partner user

**QA NOTES**
The best way to test this is to use the stood up environments:
https://bloom-partners-orm.netlify.app/sign-in
and
https://bloom-core-orm.netlify.app/

testers should verify that basically all backend endpoints work correctly. This would include:

testing the flow of creating -> editing -> viewing details for -> viewing in public site for listings

testing the flow of creating -> editing -> viewing details for -> logging in as a user on both the public and parter site

testing the flow of applying for a listing on both the partner and public site

emilyjablonski:
Likely related to #2626

YazeedLoonat:
This is good to go, pending wrapping up of: https://github.com/bloom-housing/bloom/issues/3368

PR #2964, opened 2022-08-17
2250/auth cookie

dominikx96:
# Pull Request Template

## Issue Overview

This PR addresses #2250

## Description

Refactors authentication flow from session storage to a secured cookie. The reason why we should use the cookie, instead of other options is described here: https://tkacz.pro/how-to-securely-store-jwt-tokens/

I've refactored backend & frontend apps. Once a user tries to authenticate, a backend creates a cookie with httpOnly and secure options and sends along with a response. The cookie is visible in browser cookies tab, but there is no way to read the cookie using javascript.

From the frontend perspective, to authorize requests I had to add the `withCredentials` parameter to an axios config. It sends cookies along with a request.

To log out a user, I defined a new `/logout` endpoint that creates a new cookie with the current date and sends in response. Because a cookie date is set to the current date, it means it's expired to a user is logged out.

The easiest way to test this is to go through a few of the different log in processes

try to log into the public site (as a non-mfa user) then try opening a new tab. You should remain logged in to both without needing to relog in

try to log into the partner site (as an mfa user and later as a nonmfa user). Try opening a new tab, and you should see that user remains logged in the new tab. 

you should also try to make sure that if the user is a jurisdictional admin then the permissions should still work correctly 

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


seanmalbert:
@dominikx96 and @YazeedLoonat,

I made an update to this PR to use an allowlist defined as an environment variable. See `backend/core/.env.template`'s `CORS_ORIGINS` for the value to set in your `.env`

backend/core/src/auth/dto/login-response.dto.ts
```diff
@@ -2,5 +2,5 @@ import { Expose } from "class-transformer"
 
 export class LoginResponseDto {
   @Expose()
-  accessToken: string
+  success: boolean
 }
```
seanmalbert, line 6: Do we need this or can we use StatusDto in its place?

backend/core/src/auth/dto/logout-response.dto.ts
```diff
@@ -0,0 +1,6 @@
+import { Expose } from "class-transformer"
+
+export class LogoutResponseDto {
+  @Expose()
+  success: boolean
+}
```
seanmalbert, line 6: same question here, can we use StatusDto instead?

seanmalbert:
> Another problem is with refreshing a token. I'm not sure if I understand correctly, but it looks like the /auth/token is responsible for refreshing a token, am I right?

@dominikx96,
Yes, /auth/token calls `generateAccessToken`, so I think we can add the refresh token logic to something that gets hit on every request, like AuthzGuard's canActivate. Would like support from @YazeedLoonat to implement that piece here?

YazeedLoonat (changes requested):
Looks good, but two small issues I've found:
when you first hit the public and when you first hit the partner site we get an `UnauthorizedException` which I think is going to bog down papertrail on production/dev sites

dominikx96:
@seanmalbert @YazeedLoonat 

About `UnauthorizedException` - we need to know if a user is logged in to show proper navigation elements so that's why I'm calling the user data endpoint. When we had a token in local storage, we could check if a token exists, but currently, there is no way to read a secured cookie. I don't have any idea how we can prevent `UnauthorizedException` and check if a user is logged in.

I'm also not sure about refresh token logic - a refresh token should be longer than an access token, but don't know what exp date we should define for both. @YazeedLoonat can you help me with this part? I can assist you and implement the frontend part! :) 

My vision here is: once a user gets a message that the token is expired, we can call the refresh token endpoint, if a refresh token is valid, then we can update a cookie and a user is logged in, if not - we can update a cookie with exp time set to actual time, so it means - a user will be logged out.

YazeedLoonat:
So updates on here should be good to go! I got the backend there and got the front end to call the new refreshToken as well 

So I think this should all be squared away

ludtkemorgan:
@YazeedLoonat this PR doesn't start up because of an error in app.module.ts. Is that a known issue?

YazeedLoonat:
update: there is a .env change in backend/core so for reviewers please do include those changes!

backend/core/.env.template
```diff
@@ -18,3 +18,4 @@ GOOGLE_API_ID=
 GOOGLE_API_KEY=
 GOOGLE_API_EMAIL=
 PARTNERS_PORTAL_URL=http://localhost:3001
+CORS_ORIGINS=["http://localhost:3000", "http://localhost:3001", "http://localhost:9000"]
```
ludtkemorgan, line 21: just curious what the 9000 port is for?

backend/core/src/app.module.ts
```diff
@@ -41,7 +41,18 @@ import { CatchAllFilter } from "./shared/filters/catch-all-filter"
 
 export function applicationSetup(app: INestApplication) {
   const { httpAdapter } = app.get(HttpAdapterHost)
-  app.enableCors()
+  const allowList = process.env.CORS_ORIGINS
```
ludtkemorgan, line 44: might be a good idea just in case we miss a config file configuration in one of the environments we should change this line to 
`const allowList = process.env.CORES_ORIGINS || []`

backend/core/src/auth/controllers/auth.controller.ts
```diff
@@ -37,18 +42,20 @@ export class AuthController {
...
-    return mapTo(LoginResponseDto, { accessToken })
+  async login(
+    @Request() req,
+    @Response({ passthrough: true }) res: ExpressResponse
+  ): Promise<LoginResponseDto> {
+    return mapTo(LoginResponseDto, await this.authService.tokenGen(res, req.user))
   }
 
   @UseGuards(DefaultAuthGuard)
-  @Post("token")
```
ludtkemorgan, line 46: By deleting this token endpoint won't we run into the case where during the deployment all calls to the old `token` endpoint will fail. Or in other words when we deploy the backend the /auth/token endpoint will disappear so any calls to this endpoint will fail until the UI is deployed

ludtkemorgan (commented):
Also, I found a doc file called `Authentication.md` that explains the authentication process. It hasn't been updated in a while, but for this change we might want to update the documentation

backend/core/src/auth/controllers/auth.controller.ts
YazeedLoonat: I think that when we deploy this without a front end release things would not work either, since the front end would not have a cookie set to allow it to get through our auth checks

so I think either way we would need to have the front end and backend deployed at the same time for both of these

YazeedLoonat:
This should be ready, the tests that are failing I'm pretty sure are more due to flakiness (looking for something that hasn't rendered quite yet)

ludtkemorgan (approved):

emilyjablonski:
Pinged you offline, but having some issues signing in

emilyjablonski:
@YazeedLoonat Looks like it's failing to compile, can you take a look so we can review?

YazeedLoonat:
Hey @emilyjablonski can you give this another peep please I think its ready

emilyjablonski:
@YazeedLoonat 👁️ Are the remaining issues flagged in the PR description still problems?

YazeedLoonat:
Hey @ludtkemorgan do you have the new `CORS_ORIGINS` .env file in backend/core?

ludtkemorgan:
> Hey @ludtkemorgan do you have the new `CORS_ORIGINS` .env file in backend/core?

Looks like my .env file was accidentally deleted so that fixed the issue!

emilyjablonski:
https://user-images.githubusercontent.com/65367387/207986428-b5ec4af6-5587-483c-97d5-70dc64752e86.mov

Heads up, I can't log out of partners

YazeedLoonat:
Hey @emilyjablonski can you peep one more time (hopefully) and lemme know what you think

I'm pretty sure I got this sorted out

emilyjablonski (approved):
So excited about this!!!

My last question is just making sure the expiry is looking good on your end?

I also would like one additional review since there have been a lot of changes since the last approval.

ColinBuyck:
Still reviewing different cases but one that just came up is that I can log-in on the public side and then it gives me access to the partner's portal automatically. I'm assuming this is related to the admin@example account having access to both sides but unsure if it's desired behavior or a concern @YazeedLoonat 

https://user-images.githubusercontent.com/53269332/208210759-bf6e05d0-31ed-4df0-abee-498e36713ee9.mov

backend/core/src/auth/auth.module.ts
```diff
@@ -33,7 +33,7 @@ import { ListingRepository } from "../listings/db/listing.repository"
       useFactory: (configService: ConfigService) => ({
         secret: configService.get<string>("APP_SECRET"),
         signOptions: {
-          expiresIn: "10m",
+          expiresIn: "1d",
```
YazeedLoonat, line 36: flag for later

backend/core/.env.template
```diff
@@ -18,4 +18,5 @@ GOOGLE_API_ID=
 GOOGLE_API_KEY=
 GOOGLE_API_EMAIL=
 PARTNERS_PORTAL_URL=http://localhost:3001
-AFS_PROCESSING_CRON_STRING=0 * * * *
\ No newline at end of file
+CORS_ORIGINS=["http://localhost:3000", "http://localhost:3001", "http://localhost:9000"]
```
YazeedLoonat, line 21: flag for later

backend/core/src/auth/auth.module.ts
YazeedLoonat: https://github.com/auth0/node-jsonwebtoken/issues/609#issuecomment-504041869

YazeedLoonat:
Hey @ColinBuyck I got some updates out in the most recent commit:
-- confirming account on public side 

-- on old access token being used both users should get logged out

are the things that got resolved 

lemme know what you think

ColinBuyck (commented):
Talked offline but documenting here that I was unable to sign-in on either the public nor the partner's side on Safari with the most recent changes but did not have issues with Chrome

YazeedLoonat:
Hey @ColinBuyck newest update should resolve this across safari, chrome and firefox lemme know what you think

YazeedLoonat:
to add some notes the fix here was to turn off the secure flag when we are doing local development (NODE_ENV is development) 

this is because the actual request locally is done over http instead of https. Safari flags that since an http cookie can't actually be secure

this is kinda touched on: https://stackoverflow.com/a/71883080

once this gets into dev/a hosted env we can verify that the flag works properly and is marked as secure in those environments

backend/core/src/auth/constants.ts
```diff
@@ -1,3 +1,26 @@
...
 // Length of hashed key, in bytes
 export const SCRYPT_KEYLEN = 64
 export const SALT_SIZE = SCRYPT_KEYLEN
+export const TOKEN_COOKIE_NAME = "access-token"
+export const REFRESH_COOKIE_NAME = "refresh-token"
+export const ACCESS_TOKEN_AVAILABLE_NAME = "access-token-available"
+export const TOKEN_COOKIE_MAXAGE = 86400000 // 24 hours
+export const AUTH_COOKIE_OPTIONS: CookieOptions = {
+  httpOnly: true,
+  secure: process.env.NODE_ENV !== "development",
```
ludtkemorgan, line 11: A comment here might be useful for explanation on why this is needed

backend/core/src/auth/constants.ts
```diff
@@ -1,3 +1,26 @@
+import { CookieOptions } from "express"
 // Length of hashed key, in bytes
 export const SCRYPT_KEYLEN = 64
 export const SALT_SIZE = SCRYPT_KEYLEN
+export const TOKEN_COOKIE_NAME = "access-token"
+export const REFRESH_COOKIE_NAME = "refresh-token"
+export const ACCESS_TOKEN_AVAILABLE_NAME = "access-token-available"
+export const TOKEN_COOKIE_MAXAGE = 86400000 // 24 hours
```
ludtkemorgan, line 8: Maybe not for this story, but this could be a good candidate to be an environment variable in case future sites want a longer/shorter login period

ludtkemorgan (approved):
The most recent changes look good. I was able to login on safari! Thanks for sticking with this story.

With these changes I was getting an error when logging out. I think it was a 401 response and the UI handled it ok, but in dev mode an error popped up. However, after rebuilding and restarting I am no longer seeing that. Before merging can you just re-check the logout flow? Otherwise looks great!

ColinBuyck (approved):
This all looks good to me! Great work 🏆 🍪 🏆

backend/core/src/auth/constants.ts
YazeedLoonat: I agree!

YazeedLoonat (approved):
approving because at 1 point I requested changes

PR #2966, opened 2022-08-17
fix: update for migration

YazeedLoonat:
This is a migration update for the pref/program to multiselect question 

in theory this should build the translations we need without affecting the build in ci


seanmalbert (approved):
looks good, smart thinking on getting the translations from github!

emilyjablonski (approved):
LGTM once tests go green

PR #2990, opened 2022-08-23
fix: Migration hotfix 2

YazeedLoonat:
This makes some updates to the migration around programs being radios by defaults and also resolves the double `'` one

### DIFF
--- a/backend/core/src/migration/1658871879940-multiselect-question.ts
+++ b/backend/core/src/migration/1658871879940-multiselect-question.ts
@@ -132,9 +132,11 @@ export class multiselectQuestion1658871879940 implements MigrationInterface {
       const { optOutText, options } = this.resolveOptionValues(
         pref.form_metadata,
         "preferences",
-        jurisInfo?.length ? jurisInfo[0].name : ""
+        jurisInfo?.length ? jurisInfo[0].name : "",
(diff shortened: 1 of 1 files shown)


backend/core/src/migration/1658871879940-multiselect-question.ts
```diff
@@ -322,6 +336,19 @@ export class multiselectQuestion1658871879940 implements MigrationInterface {
           juris
         )
         shouldPush = false
+      } else if (
+        !isForPreferences &&
+        (!formMetaData.type || formMetaData.type !== "checkbox") &&
+        index === formMetaData.options.length - 1
+      ) {
```
ludtkemorgan, line 343: These if statements might read a bit better if you condensed the same functionality together. You could either have an if else statement with `index !== formMetaData.options.length - 1` and then an inner if statement of the two other checks. 

Or you could do the inverse of `!isForPreferences &&
        (!formMetaData.type || formMetaData.type !== "checkbox")` being the top level if statement with the index check being the inner if. 
        
However, what you have here is ok, especially since it's a migration file.

ludtkemorgan (approved):

seanmalbert (approved):

PR #3019, opened 2022-08-29
feat: visuals for 3010

emilyjablonski:
The frontend pieces for the individual flagged set page on this PR:
https://github.com/bloom-housing/bloom/pull/3019

https://www.figma.com/file/rK4OfJpwQ01qoBAwi5VWzg/Partners-Duplicates?node-id=404%3A22266

I was sort of annoyed that when the table only has 2 rows (as it is on a reseed) the table leaves a little whitespace at the bottom, but the package we are pulling in has a min height on the table and the docs indicate it's an intentional design choice & helps render the empty state smoothly ¯\_(ツ)_/¯


YazeedLoonat (approved):
LGTM

Issue #3035, opened 2022-08-31
Some rent values showing as `NaN`


emilyjablonski:
**Acceptance Criteria**
* When no rent numbers are provided, it should say "N/A" not NaN

![187236746-dade0867-e8b6-444e-a65e-f58c3098f893.png](https://images.zenhubusercontent.com/601c7ede5322cf8ac29d50c4/4e16f724-f033-4888-9cad-dce90cd91a26)
 
https://dev-bloom.netlify.app/listing/af6c3dc9-e6a1-4deb-8d7c-08675b148e3e/emily_jablonski_1380_w_12_th_ave_broomfield_co

emilyjablonski:
QA Notes: Peruse the listings on dev and ensure none are showing `NaN` as table values. Many listings on dev were already doing so.

kathyccheng:
LGTM

PR #3040, opened 2022-09-06
fix: return N/A if rent and income is NaN

KrissDrawing:
# Pull Request Template

## Issue Overview

This PR addresses #3035

## Description

While not passing rent and income values, we've sent probably `null` value that was processed anyway as valid number, so while formatting `parseFloat` returned `NaN`.
Now instead of `$NaN per month` it displays as `N/A per month` etc.
Not sure if that should also contain `$` before `N/A`

We need to have listing where in listing units we select nothing in `How is Rent Determined?`. While editing listing just remove existing values, at first it can throw error while saving, but after open listing unit and saving it again it should work fine.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/backend/core/src/shared/units-transformations.ts
+++ b/backend/core/src/shared/units-transformations.ts
@@ -245,7 +245,7 @@ const hmiData = (units: Units, maxHouseholdSize: number, amiCharts: AmiChart[]) 
 
 const getCurrencyString = (initialValue: string) => {
   const roundedValue = getRoundedNumber(initialValue)
-  if (Number.isNaN(roundedValue)) return "N/A"
+  if (Number.isNaN(roundedValue)) return "t.n/a"
   return usd.format(roundedValue)
 }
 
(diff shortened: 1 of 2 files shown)


backend/core/src/shared/units-transformations.ts
```diff
@@ -244,7 +244,9 @@ const hmiData = (units: Units, maxHouseholdSize: number, amiCharts: AmiChart[])
 }
 
 const getCurrencyString = (initialValue: string) => {
-  return usd.format(getRoundedNumber(initialValue))
+  const roundedValue = getRoundedNumber(initialValue)
+  if (Number.isNaN(roundedValue)) return "N/A"
```
emilyjablonski, line 248: If we return this string in English here, the translations won't come through. We may need to pass just the key (`"t.n/a"`) (like we are doing here: https://github.com/bloom-housing/bloom/blob/dev/ui-components/src/tables/StandardTable.tsx#L155) and translate on the frontend (which we allow here: https://github.com/bloom-housing/bloom/blob/dev/ui-components/src/tables/StandardTable.tsx#L155)

emilyjablonski (changes requested):
Just need to be able to support translations! Context above

emilyjablonski:
Screen Shot 2022-09-08 at 7 19 55 PM
translations are coming through now, thank you! the last update will be that we don't want the `per month` string in these `n/a` cases

seanmalbert (approved):

Issue #2938, opened 2022-08-04
Core - Card


slowbot:
As a user I want to know a summary of content.  

### Properties
- **Header group** - nested component
- **Content** - content section
-  **Footer** - footer section

### Mobile
- **Padding** - reduced padding

### Variants
- **Border** - has border
- **Padding** - padding sm, md, lg

### Mockups
https://www.figma.com/file/pIlN9vqLwUgOZYq6I5oM9l/bloom-public-eligibility-wizard?node-id=827%3A7471

jaredcwhite:
See more discussion along these lines in #2982

Issue #2936, opened 2022-08-04
Core - Header Group


slowbot:
As a user I want to know a summary of content.

Header group: A header and sub header element       

### Properties
- **Heading** - content title - use existing heading component
- **Sub heading** - content description
- **Heading level** - h1 - h4

### Mobile
- Font smaller on mobile 

### Mockups
https://www.figma.com/file/pIlN9vqLwUgOZYq6I5oM9l/bloom-public-eligibility-wizard?node-id=827%3A7471

slowbot:
LGTM @jaredcwhite

PR #3043, opened 2022-09-06
2938/card component

jaredcwhite:
## Issue Overview

This PR addresses #2938 and #2936

## Description

This PR adds a new flexible Card component, along with various subcomponents which can be used within a Card structure.

BREAKING CHANGE: Renamed exported interfaces in ListingCard so as to not conflict with Card and other components

**Note:** this also includes a prototype Card which uses an alternate way to "slotting" content into the container from components marked with certain types. (React doesn't offer the concept of "slots" like in web components, Vue, etc., other than simply allowing passing nodes inside of props, so this was an experiment to see how that could work. In the end, I decided on a more idiomatic React approach. But happy to discuss this further.)

You can view the Storybook examples and docs here:
https://deploy-preview-3043--dev-storybook-bloom.netlify.app/?path=/story/blocks-card--text-content

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


jaredcwhite:
The Jest backend tests failure doesn't seem to be related to this PR per se…

ui-components/src/blocks/Card.scss
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+.card__header + .card__footer {
+  margin-block-start: var(--content-padding-block);
+}
+
+.card > *:nth-child(n + 3),
+.card__footer > .card__section {
+  border-top: var(--rule-width) solid var(--rule-color);
```
ludtkemorgan, line 71: For the mixed content story the border is double the width compared to the other stories. And it seems like it is that way because it is getting the border-bottom from the header and also a border top from the main content section since it is the third child. Is this intended? 
image

ui-components/src/prototypes/Card/CardContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+const slotElements = (children: React.ReactElement[], type: string) => {
+  return children.filter((child) => child.props.__TYPE === type)
+}
+
+const CardContainer = (props: ContainerProps) => {
+  const imageChildren = slotElements(props.children, "CardImage")
+  const headerChildren = slotElements(props.children, "CardHeader")
+  const sectionChildren = slotElements(props.children, "CardSection")
+  const footerChildren = slotElements(props.children, "CardFooter")
+  console.info(sectionChildren.length)
```
ludtkemorgan, line 24: accidental console logging

ui-components/src/blocks/Card.tsx
```diff
@@ -0,0 +1,105 @@
...
+  return (
+    <hgroup role="group">
+      <Heading className={props.headingClassName} priority={props.headingPriority ?? 2}>
+        {props.heading}
+      </Heading>
+      <p className={props.subheadingClassName} aria-roledescription="subtitle">
+        {props.subheading}
+      </p>
+    </hgroup>
+  )
```
ludtkemorgan, line 52: Should the Heading Group live outside of the Card Component? I thought the intent of #2936 was to make the headerGroup as a standalone component

jaredcwhite:
@slowbot Something that was mentioned in the issue was the idea of props to adjust the padding of the component (aka sm / md / lg). In other cases we have customizable padding via CSS variables…separately we also have something like the `size` prop of Button (which uses the AppearanceSizeType enum of small / normal / big). I left off any specific padding prop this go around, but if that's still desirable for Card I'd love to come up with a common prop name + enum that we can use in a consistent fashion across the component library.

ui-components/src/blocks/Card.tsx
jaredcwhite: Oh, hmm that's a good point. I could extract that out to a separate file/className.

ui-components/src/blocks/Card.scss
jaredcwhite: Bummer, I once thought I'd resolved that issue, not sure why it's back. Will investigate!

jaredcwhite (commented):
@ludtkemorgan OK, feedback should addressed now. HeadingGroup has been split out to its own component now.

jaredcwhite:
@ludtkemorgan BTW I solved that border issue with a simple `<div>` wrapper, but ultimately it'd be better to add support for images, etc. to the card header subcomponent or something to that effect. But that's out of scope for this component design, so probably fine to punt a bit.

ludtkemorgan (approved):

ui-components/src/prototypes/Card/CardFooter.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+
+const CardFooter: React.FunctionComponent = ({ children }) => {
```
emilyjablonski, line 3: We aren't using this typing / component style elsewhere in the app, curious about the impetus for introducing it here?

jaredcwhite: Ah, so that's just in the example prototype which is not intended for production use…wouldn't normally write in that style.

jaredcwhite:
@ludtkemorgan 

> I see that it is resolved now for that story. Out of curiosity what use case is this for .card > *:nth-child(n + 3)? I don't see it applied for any of the stories

That just automatically ensures any elements (besides a footer) in the card from the third one on down have a top border. Because the first one (header) wouldn't want to add extra border, and the second one wouldn't either because the header itself has a bottom border. I think you might be right though, no stories which have multiple sections in the card to show how that looks.

ui-components/src/blocks/Card.docs.mdx
```diff
@@ -0,0 +1,60 @@
...
+import { Swatch } from "../prototypes/Swatch"
+import { Card } from "./Card"
+
+# Card
+
+<Canvas>
+  <Story id="blocks-card--text-content" />
+</Canvas>
+
+The 2nd-generation Card component is intended to serve as a flexible content container with a header, optional content sections, and a footer. Over time our intention is to "port" some other card types and content blocks in the system to use Card as their underlying structure.
```
emilyjablonski, line 11: I'm finding these docs / the way this component has been set up to be a bit confusing to me. We aren't using this kind of subcomponent structure anywhere else vs just exporting the pieces separately - is there a benefit outside of style to do it this way?

ui-components/src/blocks/Card.stories.tsx
```diff
@@ -0,0 +1,239 @@
...
+            </Card.Section>
+          </Card.Footer>
+        </Card>
+        <style>{cssVarsOverride}</style>
+      </div>
+    </>
+  )
+}
+
+export const CustomWackyCard = () => {
```
emilyjablonski, line 184: Open to others' thoughts as well, but I'd prefer to remove this story, or to edit is so that it isn't quite as jarring visually (which I get is the point ;)). We already have the Detroit story to showcase visual styling.

ui-components/src/headers/HeadingGroup.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+  id: "headers/heading-group",
+  parameters: {
+    docs: {
+      page: HeadingGroupDocumentation,
+    },
+    badges: [BADGES.GEN2],
+  },
+}
+
+export const withContent = () => (
```
emilyjablonski, line 17: Could we add a custom styling story?

ui-components/src/headers/HeadingGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+  /** A string or element to display in an `p` tag (using `aria-roledescription="subtitle"`) */
+  subheading: React.ReactNode
+  /**
+   * The heading level (1 through 6)
+   * @default 2
+   */
+  headingPriority?: number
+  /** Additional class name for the whole group */
+  className?: string
+  /** Additional class name for the heading */
```
emilyjablonski, line 17: Why do we need one specific to each heading vs letting consumers pass in a root class name with overrides? I thought we may have been moving away from extra class names

ui-components/src/prototypes/Card/CardContainer.stories.tsx
```diff
@@ -0,0 +1,75 @@
+import React from "react"
```
emilyjablonski, line 1: My understanding of the prototype folder was that it is for components that were in a beta / unreleased but with the intention to be released eventually - what's the goal w these prototypes here?

ui-components/src/blocks/Card.docs.mdx
jaredcwhite: Sure, so what I'm proposing is a bit of a stylistic shift away from supplying "subcomponent" content as extra props on a top-level component (e.g. Card), and instead using child components for those pieces. So instead of something like:

```jsx
<Card header={<HeadingGroup title="Title" subtitle="Subtitle" />} footer={<ButtonGroup ... />}>
```
etc.

We can do this:

```jsx
<Card>
  <Card.Header>
    <HeadingGroup title="Title" subtitle="Subtitle" />
  </Card.Header>
  <Card.Footer>
    <ButtonGroup ... />
  </Card.Footer>
</Card>
```
etc.

The idea being that `Card` is both a top-level component and a namespace: Card (namespace) .Header (subcomponent)

The additional code in the prototypes folder was an example of pushing this idea even further so that subcomponents could be "slotted" into specific places within the DOM structure of the top-level component. In this case we're sticking with more of a simplistic structure where all the children of Card are fine with rendering in order from top to bottom within the parent.

Exporting the pieces separately (as `CardHeader`, `CardFooter`) is done so simply for the ergonomics of supplying those "atoms" separately to a developer who wants to create a different component that is Card-like. Normally a consumer wouldn't need to use those exports standalone, as generally they don't provide much utility on their own.

I was inspired on some of these aspects by looking at other component libraries…for example Semantic UI. Here's their example of a Modal: https://react.semantic-ui.com/modules/modal/#types-modal  You can see they offer `<Modal.Header>`, `<Modal.Actions>`, etc. Contrast that with our modal which currently has a `title` prop, an `actions` prop, etc.

Functionality I suppose we're not talking about much of a difference at the end of the day, but personally I have found the use of React props for supplying lots of content and subcomponents to be rather cumbersome, and I appreciate the use of slots in other frameworks like Vue or in web components, and this sort of gets us closer to that style.

Certainly open to feedback on how all this might play into our increased focus on staying away from supplying large monolithic components.

ui-components/src/blocks/Card.stories.tsx
jaredcwhite: No worries, I'll remove that.

ui-components/src/headers/HeadingGroup.tsx
jaredcwhite: Good point, just having that "escape hatch" on the root is enough.

ui-components/src/prototypes/Card/CardContainer.stories.tsx
jaredcwhite: I can remove it, but I'm hoping first it can spark some feedback and discussion (as I mentioned in the longer comment above)…

emilyjablonski:
@jaredcwhite Thanks for updates! Hoping to get some others to chime in to make sure we're all on the same page & then we can move forward

ui-components/src/blocks/Card.docs.mdx
ludtkemorgan: I didn't realize that this would be the first component to switch to this sub component model. I'm a big fan of the way Jared is doing it here. Having used semantic-ui and several other component libraries that use this pattern I personally like this way better than having each section be a prop to the main component.

However, there are a few things that we might want to discuss as a larger team. The biggest being how strict do we want the component to be. Like do we want our consumers to be able to put the footer before the header? Or do we allow any immediate children of the Card be anything besides `Card.Header`, `Card.Section`, or `Card.Footer`. The less strict we are the more likely a bug will appear, but the more strict means potentially more requests from new use cases or consumers.

The way it is being done in the prototype is a good way to add more restrictions on how the card is used. For example a user can no longer put the footer before the header. And I like where that component is heading.

But for this story I'd recommend not going down the path of "slotting" things in and keeping with the simplicity of the Card.tsx. 

The only call out I'm a bit worried about in this doc is encouraging creating a new type of Card but still adding the `.card` class. I'm not a fan of using a class built for a component outside of the component

jaredcwhite: @ludtkemorgan Thanks for your feedback. I do like the thought of being more strict about how stuff can get slotted in like the prototype shows, but I feel like we'd only want to adapt that pattern if we think we'll use it in other places and not just the Card component.

Regarding the `.card` class, at one point I had tried something along the lines of splitting out all the CSS variables like `--background-color` from the class itself…I forget what I called it but let's say it was `.card-variables`. Then the Card component itself would use both `.card` and `.card-variables`, but some other component could opt to use only `.card-variables`. But then I felt like maybe that was overly complicating matters, and it's another pattern we haven't used anywhere else yet. Curious what you think about that.

emilyjablonski (approved):

PR #3058, opened 2022-09-13
Visuals for modal follup

YazeedLoonat:
This makes backend changes for the way duplicates works.

What this should resolve is:
```
add status to the set
    pending
    resolved

when "resolving" one
    we need to mark an application state
    we need to also mark the afs state

modify statuses on the applications
    pending <- default if duplicate set created
    pendingAndValid
    valid <- default if no duplicates
    duplicate <- everything else once valid is determined
```

**Update**
this now also includes seeding updates so each listing should have 4 application flagged sets, and each set should include 2 applications (2 afs for email, 2 afs for name + dob per listing)


YazeedLoonat:
Hey @emilyjablonski I got this updated if you want to take another peep

emilyjablonski (approved):
lgtm!

PR #3060, opened 2022-09-13
refactor: update backend fields on afs

emilyjablonski:
@YazeedLoonat I think these might be the changes we need, will touch base offline

The current issues:

The `resolve` endpoint as is takes in the set id, an array of application ids, and an applications status. The issue I'm running into is on the frontend, when I'm in the `save` modal, the information I have that I want to save is (1) an array of applications with a status on each one mapping to whether or not it is checked, and (2) a status on the set, mapping to which radio in the modal is selected.

With the current endpoint I can only set an array of applications to one status. I'd have to make two calls, to gather the selected rows and the unselected rows to set their statuses, in case I had unchecked a row. I also need to be able to update the status of the set.

Let me know what you think! I added in a piece of the frontend which calls the resolve endpoint so we can test some of it in the browser.

This PR targets: https://github.com/bloom-housing/bloom/pull/3058

### DIFF
--- a/backend/core/src/application-flagged-sets/application-flagged-sets.service.ts
+++ b/backend/core/src/application-flagged-sets/application-flagged-sets.service.ts
@@ -116,6 +116,7 @@ export class ApplicationFlaggedSetsService {
         "applications.id",
         "applications.submissionType",
         "applications.confirmationCode",
+        "applications.reviewStatus",
         "applicant.firstName",
         "applicant.lastName",
         "applicant.birthDay",
@@ -142,6 +143,7 @@ export class ApplicationFlaggedSetsService {
     return await getManager().transaction("SERIALIZABLE", async (transactionalEntityManager) => {
       const transAfsRepository = transactionalEntityManager.getRepository(ApplicationFlaggedSet)
       const transApplicationsRepository = transactionalEntityManager.getRepository(Application)
+      // TODO: For each application, set it to the status on that application, then set the set status
       const afs = await this.findOneById(dto.afsId, dto.applications)
 
       if (afs.listing.status !== ListingStatus.closed) {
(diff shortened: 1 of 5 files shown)


backend/core/src/application-flagged-sets/dto/application-flagged-set-resolve.dto.ts
```diff
@@ -14,10 +14,10 @@ export class ApplicationFlaggedSetResolveDto {
   @IsArray({ groups: [ValidationsGroupsEnum.default] })
   @ArrayMaxSize(512, { groups: [ValidationsGroupsEnum.default] })
   @ValidateNested({ groups: [ValidationsGroupsEnum.default], each: true })
-  @Type(() => IdDto)
-  applications: IdDto[]
+  @Type(() => ApplicationResolve)
+  applications: ApplicationResolve[]
```
emilyjablonski, line 18: This can just be the id

backend/core/src/application-flagged-sets/types/application-resolve.ts
```diff
@@ -0,0 +1,15 @@
+import { Expose, Type } from "class-transformer"
```
emilyjablonski, line 1: This can be removed

backend/core/src/application-flagged-sets/application-flagged-sets.service.ts
```diff
@@ -142,6 +143,7 @@ export class ApplicationFlaggedSetsService {
     return await getManager().transaction("SERIALIZABLE", async (transactionalEntityManager) => {
       const transAfsRepository = transactionalEntityManager.getRepository(ApplicationFlaggedSet)
       const transApplicationsRepository = transactionalEntityManager.getRepository(Application)
+      // TODO: For each application, set it to the status on that application, then set the set status
```
emilyjablonski, line 146: Don't need this anymore

Issue #3028, opened 2022-08-31
Core - Preferences Delete


kathyccheng:
As a partner
I might want to delete an erroneous preference that hasn't been added to a listing yet 
So that I can keep the listings list clean 

**Acceptance Criteria**
* When I am on the listings table on the settings page, then I will see the ability to delete a preference 
* When I click on delete, then I will see a modal that says Are you sure you want to delete this preference?
* When I click on delete and the listing has already been added to a listing, then I will see a modal that says "This preference has already been added to the following listings: [List of listing names]. Please disassociate this preference from those listings if you'd like to delete it." 
* When a preference gets deleted, then I'll see the toast notification with "Preference deleted" 

**Notes**
* I think we do a hard delete
* Make sure #3007 goes in first 
* Log date, preference id, user who took the action 
* Update controller and service for multi-select question to query by a preference and get back listing 
* Double check if we have cross jurisdictional preferences

**Design**
https://www.figma.com/file/ek2VsPKqCysxZN2pgznR4h/partners-admin?node-id=2%3A5537

ludtkemorgan:
@slowbot I don't know if I'm not seeing it on figma, but is there a design for the modal? This is what the two scenarios looks like without any custom css using the text provided.

![image](https://user-images.githubusercontent.com/42942267/189762194-bb9eba44-f20d-4482-b653-8f68d66bee2b.png)


![image](https://user-images.githubusercontent.com/42942267/189762128-8484df31-b879-4924-ac57-33bb95c6ae45.png)

slowbot:
@ludtkemorgan The modal https://www.figma.com/file/Y8bFd0aSz7eYJ2TCNid6s3/bloom-partners?node-id=5876%3A183489

kathyccheng:
LGTMC

PR #3075, opened 2022-09-16
feat(preferences): add delete preference button

ludtkemorgan:
# Pull Request Template

## Issue Overview #3028 

This PR addresses #issue

## Description

This PR adds the delete functionality to the "settings" page in the partners site. 

It adds the delete icon back in for each preference
image

Once clicked a modal pops up
image

And on Deletion a toast notification indicates the preference was deleted
image

There are two main use cases that should be tested. First off is the happy path of deleting a preference that doesn't have any listings attached. The second is testing on a preference that does have at least one listing attached to it. In that case the "Delete" button on the modal shouldn't be visible and there should be a different text in the modal.

You will need to run the migration script in order for this feature to work.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


ludtkemorgan:
@slowbot for this PR I made the delete modal look like the mock up in the Figma you linked in the story. However, don't know if I'm just not seeing it, were you able to get a mock up for the scenario where the preference is already attached to a listing? Right now I just kept the same design but changed the middle text and removed the "Delete" button. I'm guessing this won't be what we want when there are several listings attached
image

sites/partners/__tests__/pages/settings/index.test.tsx
```diff
@@ -0,0 +1,132 @@
+import React from "react"
```
emilyjablonski, line 1: Yayyyyyyyyyyy!

sites/partners/__tests__/pages/settings/index.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+        return res(ctx.json([]))
+      })
+    )
+
+    const { getByText, findByText } = render(<Settings />)
+
+    expect(getByText("Settings")).toBeInTheDocument()
+    expect(getByText("Preferences")).toBeInTheDocument()
+
+    await findByText("None")
```
emilyjablonski, line 32: Is this just to ensure it loads?

sites/partners/src/settings/PreferenceDeleteModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+      border={AppearanceBorderType.borderless}
+      onClick={onClose}
+      ariaLabel={t("t.cancel")}
+    >
+      {t("t.cancel")}
+    </Button>
+  )
+
+  return (
+    <>
```
emilyjablonski, line 72: 🧹

emilyjablonski (approved):
This looks great! I know we're still waiting on design but I have a question around what the header text of the delete modal should be as well:
Screen Shot 2022-09-19 at 6 12 17 PM
I'm also curious where the logging is happening?

backend/core/src/multiselect-question/multiselect-question.controller.ts
```diff
@@ -79,9 +82,21 @@ export class MultiselectQuestionsController {
...
+    const results = await this.multiselectQuestionsService.findListingsWithMultiSelectQuestion(
+      multiselectQuestionId
+    )
+    return mapTo(ListingDto, results)
+  }
+
   @Delete()
   @ApiOperation({ summary: "Delete multiselect question by id", operationId: "delete" })
   @UseGuards(OptionalAuthGuard, AdminOrJurisdictionalAdminGuard)
+  @UseInterceptors(ActivityLogInterceptor)
```
ludtkemorgan, line 99: @emilyjablonski this is where the logging is happening. This interceptor will grab the applicable fields (user and preference id) and save it to the Activity Log table

sites/partners/__tests__/pages/settings/index.test.tsx
ludtkemorgan: Yeah, it verifies that when the empty array is returned the component outputs "None". Since we have to wait for the API call to return it's safer to do the `await findBy....`

emilyjablonski:
Sounds good! Looks like we can make use of the `[MinimalTable](https://storybook.bloom.exygy.dev/?path=/story/tables-minimaltable--default)`

ludtkemorgan:
@emilyjablonski I've updated the PR to accommodate the new Modal. Feel free to re-review this

ColinBuyck (approved):
Everything looks good here @ludtkemorgan, just noticed two small things around design. In the PR, the title bolding and font size doesn't seem to match the Figma from Jesse. Additionally, it seems that the modal is taking up a larger percentage of the screen vertically than the Figma. What are your thoughts on creating a css variable for max-height defaulted to 100vh, and then override it to match this design?
PR:
Screen Shot 2022-09-26 at 11 26 09 AM
Design:
Screen Shot 2022-09-26 at 11 40 52 AM

ludtkemorgan:
> Everything looks good here @ludtkemorgan, just noticed two small things around design. In the PR, the title bolding and font size doesn't seem to match the Figma from Jesse. Additionally, it seems that the modal is taking up a larger percentage of the screen vertically than the Figma. What are your thoughts on creating a css variable for max-height defaulted to 100vh, and then override it to match this design? 

@ColinBuyck 
I didn't change the font-size of the title or the height of the modal so that I wouldn't have to update the modal component. I didn't want to make this modal different from the other modals in the application. However, I can make the changes if we think it's better for this use case. Specifically the max-height bit. I think this might be the only case besides the imageCard where the content runs into the it being too big for the modal. Thoughts?

ludtkemorgan:
@slowbot Colin mentioned that this is the [new design of the modal](https://www.figma.com/file/tUtgsJO8XnF9Ij2dwu22EF/bloom-public-general?node-id=2601%3A8746) and that is the design of the mockup you created for this delete modal. Right now the modal is more like the screenshot Colin took of my implementation. Is there a plan to update the modal to be more in line with your modal design? Specifically around how tall the modal can be and the font-size of the header.

ludtkemorgan:
@ColinBuyck @slowbot I'll merge this PR and then do a fast follow up of the modal changes to update to the new modal standard throughout. Including the max height and width options. The header change will be easier once Jared's header group is merged

PR #3101, opened 2022-09-27
fix: duplicates ux issues

emilyjablonski:
Two small frontend issues from the duplicates bash
(1) Use a translation string for the closed listing alert
(2) Accessibility color contrast issue on the application link when the row is selected


seanmalbert (approved):
lgtm!

PR #3102, opened 2022-09-27
feat: adds setsAfsLastRunAt1664300247901 migration

seanmalbert:
This sets the afsLastRunAt to "now" so that we don't have re-process everything.


YazeedLoonat (approved):
LGTM

PR #3169, opened 2022-11-11
feat: get rid of ui-component prerelease

ludtkemorgan:
This triggers the release step on merges to the `main` branch. Currently releases are currently only made when things get merged to `dev`. However, with the change of not using the dev branch we need to change it to the `main` branch. Along with this we also want to not do the pre-release and instead just do a full release of ui-components when a change is merged

### DIFF
--- a/.github/workflows/release_components.yml
+++ b/.github/workflows/release_components.yml
@@ -1,9 +1,9 @@
 name: Pre-release ui-components
 
 on:
-  # Triggers the workflow on push only for the dev branch
+  # Triggers the workflow on push only for the main branch
   push:
-    branches: [ dev ]
+    branches: [ main ]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


emilyjablonski (approved):

.github/workflows/release_components.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Version ui-components
-        run: yarn version:prerelease:ui-components
```
kramduckner, line 32: curious what specific things this will change re: uptakes? e.g. no more alpha versions?

kramduckner (approved):

.github/workflows/release_components.yml
ludtkemorgan: Correct. There will be no more alpha version. It should (if this works correct) just always do major/minor/patch versions

Issue #3191, opened 2022-11-23
bug - application flagged sets cron job is not getting triggered


ludtkemorgan:
The duplicate check for applications in application-flagged-sets-cronjob-consumer.ts is not getting triggered in Heroku.

This is due to an unknown issue with connecting to the redis version in Heroku. When running locally it works correctly for both the cron job and running it one off when a listing is closed.

Potential solutions:
* Try to fix the redis issues
* Move off of the bull/redis framework and use a framework like the built in nest scheduler
* Use a framework like Heroku Scheduler to call it outside of the code.

Things to consider:
* Can we make sure that it is only triggered once instead of once per dyno?
* Is there a way to dynamically set the interval?

*QA Notes*
No visual changes can be tested. This change will have to be monitored in the dev and stage environments. Using papertrail we can see if the afs cron job is run and also check that the database table of cron_job has been added with a new entry of name AFS_CRON_JOB after the cron job first gets triggered. Make sure the last run time is updated correctly

PR #3199, opened 2022-12-01
feat: migrate afs job off of redis

ludtkemorgan:
## Issue Overview

This PR addresses #3191 

## Description

The current setup for the application flagged sets cron job is never getting hit. This is due to its reliance on the bull nestjs library that uses redis. For an unknown reason the redis setup in heroku is connecting but nothing is getting set in it. To solve this problem this PR removes the dependence on redis and uses nestjs's scheduler functionality.

Things to note:
- The cron frequency can be set as a .env variable called `AFS_PROCESSING_CRON_STRING`
- Due to us using 3 dynos in our production environment if we were to use the nestjs cron jobs without any changes it would get run three times simultaneously and potentially crash. To solve this I have added a step where a random number between 0 and 59 is put at the beginning of the cron string. This makes it so it will run at a random second at the cron run time and theoretically the jobs will be run at a staggered interval. Also when the cron job is triggered a new timestamp is saved to the db to indicate the last run time of the job. Before running it checks to make sure that that the last run time is over 5 minutes old.
  - Sample example: Cron job is set to run at midnight everyday. dyno 1 randomly selects 24 as the seconds, dyno 2 selects 11, and dyno 3 selects 53.
    - dyno 2 triggers first at 12:00:11. It checks the db and last run was almost 24 hours ago.
    - dyno 2 updates the "last run time" to 12:00:11
    - dyno 2 runs the processing
    - dyno 1 triggers at 12:00:24. It checks the db and last run was 13 seconds ago so it doesn't run
    - dyno 3 triggers at 12:00:53. It checks the db and last run was 42 seconds ago so it doesn't run

The easiest way to test this is to add a new record to your .env file for backend core for `AFS_PROCESSING_CRON_STRING`. Sample `AFS_PROCESSING_CRON_STRING="0,10,30 * * * *"` (runs every hour at 0, 10, and 30 minutes). When you start up the application you should see logging of "running the Application flagged sets cron job" to appear within the minute you set in the environment variable. I'd recommend testing a couple of variations of the cron string. Note that you need to restart the application if you change the .env file.

The database table of `cron_job` should also be added with a new entry of name `AFS_CRON_JOB` after the cron job first gets triggered. Make sure the last run time is updated correctly

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


emilyjablonski:
Thank you for the great documentation!! I have a quick question, I think I'm not sure why the randomized seconds is necessary, does it not work to have the first job that gets there set the db flag, and have each job just check and make sure it's been at least five minutes?

ludtkemorgan:
> Thank you for the great documentation!! I have a quick question, I think I'm not sure why the randomized seconds is necessary, does it not work to have the first job that gets there set the db flag, and have each job just check and make sure it's been at least five minutes?

Without the random seconds the cron jobs will all get triggered at the exact same time. For example if the cron job is set for `0 0 * * *` (every day at midnight) all three will get triggered exactly at midnight to the second. That means that they will all try to update at the exact same time and not know that the other two are also trying to update the last run timestamp.

This is also just a theory that this random seconds method will work properly. We'll need to test it in the prod environment to actually make sure it's working as expected

YazeedLoonat (approved):
LGTM

YazeedLoonat:
@ludtkemorgan could you please add the new env to the .env template please!

ludtkemorgan:
> @ludtkemorgan could you please add the new env to the .env template please!

Added!

Issue #3103, opened 2022-09-27
Duplicates - Add Flagged as Duplicate + Marked as Duplicate columns to All Applications table 


kathyccheng:
As a leasing agent, 
I might want an indication that the application is a duplicate 

**Acceptance Criteria**
* When I am looking at the table of applications in the All Applications tab, then I should see 2 columns at the end of the table - one for Flagged as Duplicate, and one for Marked as Duplicate 

![image](https://user-images.githubusercontent.com/3453407/192888320-4c1eb760-86e1-458a-bf6e-f3bd71508ace.png)

sarahlazarich:
lgtm - QA'd at https://partners.bloom.exygy.dev/listings/55b490e3-a654-42d3-bd66-c9c3e3d9a97a/applications.

PR #3214, opened 2022-12-13
feat: add marked and flagged as duplicate to all applications table

KrissDrawing:
# Pull Request Template

## Issue Overview

This PR addresses #3103

## Description

Added two columns `flagged as duplicate`, and `marked as duplicate` to All Aplications.
Marked as duplicate has its own field. Flagged as duplicate is when reviewStatus equals 'duplicate'

Go to applications tab in listing. In All Application table scroll to the right. Last two columns should be `Flagged as Duplicate`, and `Marked as Duplicate`. On null / undefined state they should display n/a (not sure if possible)

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


sites/partners/src/applications/ApplicationsColDefs.ts
```diff
@@ -581,5 +581,30 @@ export function getColDefs(maxHouseholdSize: number, countyCode: string) {
...
+  const duplicateCols = [
+    {
+      headerName: t("applications.duplicates.flaggedAsDuplicate"),
+      field: "reviewStatus",
+      sortable: false,
+      filter: false,
+      width: 125,
+      minWidth: 100,
+      valueFormatter: ({ value }) => {
+        return formatYesNoLabel(value && value === "duplicate")
```
ColinBuyck, line 593: Only thing I noticed on this PR was that I believe this should return yes for either "pending" or "duplicate" rather than just "duplicate". The way I'm understanding the difference between these columns is that an application may be flagged as a duplicate automatically but not yet marked as a duplicate by the user which results in the pending state. Is this correct @kathyccheng?

YazeedLoonat: `flagged as duplicate` means that the application is marked as a possible duplicate
`marked as duplicate`  (when the review status === duplicate) is when the application was confirmed as a duplicate

emilyjablonski: Has this been resolved?

ColinBuyck: I don't believe so @emilyjablonski.  If I am understanding our conversations correctly @YazeedLoonat, the issue is that flagged as duplicated will become false if a flagged application is then marked as valid when using the review_status. I believe there are three cases where flagged as duplicate should be true: 1) Application is part of a duplicate group in pending review (would be covered by my comment above) 2) Application was part of a duplicate group in pending review but then marked as duplicate covered by PR currently) 3) Application was part of a duplicate group in pending review but then marked as valid which cannot currently be represented by the review_status. Let me know if this makes sense @KrissDrawing and what your thoughts on next steps would be @YazeedLoonat.

KrissDrawing:
@YazeedLoonat Thanks! 
Flagged looks good, but `markedAsDuplicate` field still doesn't seem to return correct values.
Even after resolving application (so its reviewStatus = 'duplicate') field `markedAsDuplicate` = false.
I can see in exported csv file it is properly marked as duplicate 🤔 

On frontend i can solve that just by checking reviewStatus. But then don't know what to do with that `markedAsDuplicate` field as it might be misleading in the future i guess 😅

YazeedLoonat:
Hey @KrissDrawing got this updated for ya, should be good to go now hopefully if you wanna take another look

KrissDrawing:
@YazeedLoonat Works Good 🎉 
@ColinBuyck @emilyjablonski it's ready to review

backend/core/src/applications/services/applications.service.ts
```diff
@@ -103,6 +104,19 @@ export class ApplicationsService {
...
+    )
+    const flags = await Promise.all(promiseArray)
+    result.forEach((application, index) => {
+      if (flags[index].status) {
+        application.markedAsDuplicate = flags[index].status === ApplicationReviewStatus.duplicate
+      } else {
+        application.markedAsDuplicate = false
+      }
+      application.flagged = !!flags[index].id
+    })
```
ludtkemorgan, line 118: I wonder if this should be done after the check for user authorization so we avoid doing the duplicate db call if the user isn't even authorized. @YazeedLoonat thoughts?

ludtkemorgan (changes requested):
Functionally it is working as expected. Looks like there are failing tests now though. https://app.circleci.com/pipelines/github/bloom-housing/bloom/9194/workflows/37fff7e6-8ad8-42ad-bb22-51d226b8f5b7/jobs/66264. Can you look into those?

backend/core/src/applications/services/applications.service.ts
YazeedLoonat: just to add some notes here from yesterday's discussion:
I think we do want to do that but for right now its probably okay to leave this as is. 

We more or less need to undertake an auth investigation and this can fall into that investigation bucket

ludtkemorgan (approved):

emilyjablonski:
@YazeedLoonat Is this one ready?

YazeedLoonat:
Hey @emilyjablonski I think its ready for another set of eyes I probably shouldn't be a reviewer on this since I committed code to this pr though

PR #3286, opened 2023-01-17
chore: run commit message formatter before linter

emilyjablonski:
Changes the order of our commit message linting - now runs the conventional commit formatter before the linter. Previously, if linting failed, sometimes the only way to get past [failures from ignored files](https://github.com/eslint/eslint/issues/15010) was to commit with `--no-verify` which then skipped the conventional commit check linter.

To test this, in a component file locally, add a variable you don't use and try to commit with a valid conventional commit formatted message. Linting should run and the commit should fail.

Then try to commit in a not-conventional-commit format w/out a lint error (i.e. just `test`) and that should fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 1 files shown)


ludtkemorgan (approved):

ColinBuyck (approved):
LGTM 🦞

Issue #3184, opened 2022-11-22
Upgrading TypeORM


YazeedLoonat:
**What is this feature?**
This is a ticket to figuring out a way forward on upgrading our version of TypeORM on the backend. We are currently using a version that is about 1 year old, so we should get some performance/stability gains by upping our version.

**Considerations**
Since we are pretty far behind, this may require several other packages we use to need to be upgraded as well. This could become a larger task depending on how many/what needs to happen to up all of those

sarahlazarich:
This looks good at https://partners.bloom.exygy.dev/ !

PR #3309, opened 2023-02-01
feat: resetting migrations

YazeedLoonat:
# Pull Request Template

## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3184)

## Description
This resets our set of migrations, this will make it easier to do the typeorm upgrade and also lets us clean up a migration list that doesn't need to be as large. 

**BEFORE WE RUN THIS ON PRODUCTION**
we should run
```
SELECT 
    id 
FROM migrations
ORDER BY id DESC
LIMIT 1;

INSERT INTO migrations
SELECT <id + 1>, extract(epoch from now()) * 1000, 'migrationReset1675268723433';
```

since we do not need/want this to run on any stood up environments

After doing a reseed, you should poke around the partner and public site and make sure things work

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


ludtkemorgan (approved):
I haven't been able to find any issue. Assuming you are able to fix the broken tests this looks good to me 🌵

sites/partners/cypress/integration/07-jurisdictional-admin-user-management.spec.ts
```diff
@@ -1,6 +1,6 @@
 describe("Admin User Mangement Tests", () => {
   before(() => {
-    cy.login("jurisdictionalAdmin")
+    cy.loginAndAcceptTerms("jurisdictionalAdmin")
```
emilyjablonski, line 3: Was this a change required by this PR or was this broken on dev?

backend/core/src/migration/1675268723433-migration-reset.ts
```diff
@@ -0,0 +1,717 @@
...
+        confirmMyAccount: "Confirm my account",
+      },
+      mfaCodeEmail: {
+        message: "Access token for your account has been requested.",
+        mfaCode: "Your access token is: %{mfaCode}",
+      },
+      header: {
+        logoUrl:
+          "https://res.cloudinary.com/exygy/image/upload/v1652459319/housingbayarea/163838489-d5a1bc08-7d69-4c4a-8a94-8485617d8b46_dkkqvw.png",
+        logoTitle: "Alameda County Housing Portal",
```
emilyjablonski, line 461: Maybe this was existing but I didn't think we had references to Alameda here, dev is just "Housing Portal"

emilyjablonski (changes requested):
Locally on fresh main, if I come to this branch and just run the migrations it fails. Lmk if you're seeing something different!

emilyjablonski (approved):
LGTM after running the script on staging :)

backend/core/src/migration/1675268723433-migration-reset.ts
YazeedLoonat: so I pulled these translations directly from the old migrations, I think we probably updated some of the translations directly through queries instead of through migrations 

thats a good thing to flag and another reason we probably don't want to reseed staging/production envs

nice catch!

sites/partners/cypress/integration/07-jurisdictional-admin-user-management.spec.ts
YazeedLoonat: I think this change came about because of this PR since other PRs were able to pass. I think its because of the change to no longer having many migrations and instead only having 1 that user now has to accept terms whereas before they didn't have to

Issue #3282, opened 2023-01-17
User Export


sarahlazarich:
**User Story:** 

As a jurisdictional admin OR an admin in partners portal, 
I want to be able to export my jurisdiction’s user information in CSV format,
So that I can check on user engagement. 

**Acceptance criteria:**

Given that I am an admin looking at the users page in partners portal:
- When I am looking above the user table, then I should see a button indicating that I can export to CSV.
- When I click the “export to csv” button, then my user information should download as a CSV.
- When I click the “export to csv” button, then my user export will download in a timely fashion.
- When I click the “export to csv” button, then my page shows an indication that the export is in progress.
- When my user export downloads successfully, then I will see a message indicating the download was completed. 
- When there is an issue with my download, then I will see an error message on the page.

Given that I am an admin looking at my user export:  
- When I open the user information CSV, then I will see the columns listed in [user details tab](https://docs.google.com/spreadsheets/d/1RrYNnkAdG4QPyVBC8yG1svqPaLt0V4qVQMzQgHUEIjY/edit#gid=0).
- When I look through my user export files, then I will see data that reflects the users in the portal at time of download.
- When I look through my user export files, then I will see user information for both partners and admins.

**Implementation Notes:**
Please refer to the [design asset](https://www.figma.com/file/di1NVcJveFBNSwlOemIqUV/partners-reports?node-id=203%3A2476&t=AWGzqqiTd8fWZtMv-0) for export buttons and error/success messages.

QA Notes:
Use the bloom dev partners site: https://partners.bloom.exygy.dev/users. Click on the "Export" button and a CSV should download with the appropriate columns populated.

sarahlazarich:
lgtm!

PR #3341, opened 2023-02-21
feat: add user csv export

ludtkemorgan:
# Pull Request Template

## Issue Overview

This PR addresses #3282 

## Description

This adds a new endpoint to the backend that exports all partner users to a csv file. It also adds an "export" button to the users page on the partner site (made the export and add users button the smaller size as well). 
image

The export includes all of the columns referenced here: https://docs.google.com/spreadsheets/d/1RrYNnkAdG4QPyVBC8yG1svqPaLt0V4qVQMzQgHUEIjY/edit#gid=670984196.

When running this locally and clicking on "export" you should get a csv download with all of the users shown on the users page.

An additional unit and cypress test was also added.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/backend/core/src/auth/services/user-csv-exporter.service.ts
+++ b/backend/core/src/auth/services/user-csv-exporter.service.ts
@@ -1,0 +1,39 @@
+import { Injectable, Scope } from "@nestjs/common"
+import dayjs from "dayjs"
+import { Pagination } from "nestjs-typeorm-paginate"
+import { CsvBuilder } from "../../applications/services/csv-builder.service"
+import { User } from "../entities/user.entity"
+
+@Injectable({ scope: Scope.REQUEST })
+export class UserCsvExporterService {
+  constructor(private readonly csvBuilder: CsvBuilder) {}
+
+  exportFromObject(users: Pagination<User>): string {
+    const userObj = users.items.reduce((obj, user) => {
+      const roles = []
+      if (user.roles?.isAdmin) {
+        roles.push("Administrator")
+      }
+      if (user.roles?.isPartner) {
+        roles.push("Partner")
+      }
+      if (user.roles?.isJurisdictionalAdmin) {
+        roles.push("Jurisdictional Admin")
+      }
+      obj[user.id] = {
+        "First Name": user.firstName,
+        "Last Name": user.lastName,
+        Email: user.email,
+        Role: roles.join(", "),
+        "Date Created": dayjs(user.createdAt).format("MM-DD-YYYY"),
+        Status: user.confirmedAt ? "Confirmed" : "Unconfirmed",
+        "Listing Names":
+          user.leasingAgentInListings?.map((listing) => listing.name).join(", ") || "",
+        "Listing Ids": user.leasingAgentInListings?.map((listing) => listing.id).join(", ") || "",
+        "Last Logged In": dayjs(user.lastLoginAt).format("MM-DD-YYYY"),
+      }
+      return obj
+    }, {})
+    return this.csvBuilder.buildFromIdIndex(userObj)
+  }
+}
--- a/backend/core/src/auth/services/user-csv-exporter.service.ts
+++ b/backend/core/src/auth/services/user-csv-exporter.service.ts
@@ -25,12 +25,12 @@ export class UserCsvExporterService {
         "Last Name": user.lastName,
         Email: user.email,
         Role: roles.join(", "),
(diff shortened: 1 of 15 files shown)


backend/core/src/auth/services/user-csv-exporter.service.ts
```diff
@@ -0,0 +1,39 @@
...
+import { CsvBuilder } from "../../applications/services/csv-builder.service"
+import { User } from "../entities/user.entity"
+
+@Injectable({ scope: Scope.REQUEST })
+export class UserCsvExporterService {
+  constructor(private readonly csvBuilder: CsvBuilder) {}
+
+  exportFromObject(users: Pagination<User>): string {
+    const userObj = users.items.reduce((obj, user) => {
+      const status = []
```
emilyjablonski, line 13: I found it to be a little confusing here to call this status, when it's the role and there is also a status column

shared-helpers/src/auth/AuthContext.ts
```diff
@@ -179,14 +180,18 @@ const reducer = createReducer(
 )
 
 export const AuthContext = createContext<Partial<ContextProps>>({})
-export const AuthProvider: FunctionComponent = ({ children }) => {
+export const AuthProvider: FunctionComponent<{ isTesting?: boolean }> = ({
+  children,
+  isTesting = false, // Set this to true for tests that you need to set the user
```
emilyjablonski, line 185: Totally fine if this is the best case, but would prefer to not add testing paths into app code

sites/partners/__tests__/pages/users/index.test.tsx
```diff
@@ -68,4 +69,31 @@ describe("users", () => {
     expect(getByText("09/04/2022")).toBeInTheDocument()
     expect(getByText("Confirmed")).toBeInTheDocument()
   })
+
```
emilyjablonski, line 72: 😍

sites/partners/src/lib/hooks.ts
```diff
@@ -468,26 +489,20 @@ export const useApplicationsExport = (listingId: string, includeDemographics: bo
...
       const blob = new Blob([content], { type: "text/csv" })
       const fileLink = document.createElement("a")
-      fileLink.setAttribute("download", `applications-${listingId}-${dateString}.csv`)
+      fileLink.setAttribute("download", fileName)
       fileLink.href = URL.createObjectURL(blob)
       fileLink.click()
     } catch (err) {
       setCsvExportError(true)
-      setSiteAlertMessage(err.response.data.error, "alert")
+      setSiteAlertMessage("User Export Failed", "alert")
```
emilyjablonski, line 501: Could we move this to the strings file?

emilyjablonski (changes requested):
A few code comments, and then I tested with a jurisdictional admin and I wasn't able to see the export button, but the issue indicates jurisdictional admins should be able to export

backend/core/src/auth/services/user-csv-exporter.service.ts
ludtkemorgan: oops, that was a mistake. I'll update it

shared-helpers/src/auth/AuthContext.ts
ludtkemorgan: I also would prefer to not do this. I might tinker around on the test some more to see if there is a better way

ludtkemorgan:
> A few code comments, and then I tested with a jurisdictional admin and I wasn't able to see the export button, but the issue indicates jurisdictional admins should be able to export
> 
> I'm also not clear why the ids of the listings are included, it didn't look like that was part of the google sheet, but I just found the name of that listing column to be confusing so that could be my bad!

Good catch. I missed that AC. I've updated it to allow jurisdictional admin to have the ability to use the button and download just the users for their jurisdiction. For the listing ids I took that from the "Listings they can access" column and then the listings name is second "Name" column. I'll double check with product to make sure I did that correct

shared-helpers/src/auth/AuthContext.ts
ludtkemorgan: I came up with a solution to mock out the session storage. I like it better since it doesn't touch the AuthContext anymore.

emilyjablonski (approved):
Looks good! 🗝️

ColinBuyck (approved):
LGTM 〰️

Issue #3347, opened 2023-03-04
Move from data-test-id to data-testid in both RTL and Cypress


emilyjablonski:
See context here: https://testing-library.com/docs/queries/bytestid#overriding-data-testid

Will require a lot of files changed but should mostly be a find-replace to update the name of this attribute and remove the override.

Issue #3313, opened 2023-02-02
Upgrade React to 18


ludtkemorgan:
Our public and partner sites in core are currently on version 17 of React. We should upgrade it to 18 (the current latest version)
https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

Notes:
- Need to upgrade public, partners, and UI-C
- We might need to upgrade storybook

ludtkemorgan:
QA notes:
All pages should behave the same for the following urls:
- https://deploy-preview-3360--bloom-exygy-dev.netlify.app/
- https://deploy-preview-3360--partners-bloom-dev.netlify.app/sign-in

PR #3360, opened 2023-03-15
feat: upgrade react to 18

ludtkemorgan:
# Pull Request Template

## Issue Overview

This PR addresses #3313 #3347 #3366 

## Description

This PR does the following changes:
- Upgrade React from 17 to 18. This includes pulling in the latest version of UI-Components
- Updated all instances of `data-test-id` to `data-testid`. This standardizes everything between cypress and our unit tests
- Upgrades Cypress from 9 to 12. This includes the following required changes
  - Directory changes from `integration` to `e2e`
  - Tests no longer can depend on previous tests so all tests that relied on things such as being logged in have to now do per test. Most just needed to change "before" to "beforeEach", but some also had to re-navigate to the correct page
- The mapbox address endpoints keeps returning a suggested address even with the junk address typed. Because of this I migrated that api call to a fixture in our cypress test. This will make the test more consistent as that is one of our flakiest test and also reduce our total calls to our mapbox api

This change will require testing of every page to make sure no warnings appear. I'd recommend pulling it down, reseeding, and checking every page as best you can. 

Also the UI is new for the Cypress GUI. It appears to be faster for me. However, there is now no "run all" button so if you want to test the cypress test you will need to run each desired test individually.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


sites/partners/__tests__/pages/users/index.test.tsx
```diff
@@ -110,16 +109,16 @@ describe("users", () => {
...
+    expect(exportButton).toBeInTheDocument()
+    fireEvent.click(exportButton)
     const successMessage = await findByText("The File has been exported")
     expect(successMessage).toBeInTheDocument()
   })
 
   it("should render error message csv fails", async () => {
     // set a logged in token
-    jest.useFakeTimers()
+    jest.spyOn(console, "log").mockImplementation(jest.fn())
```
emilyjablonski, line 121: Spying on a log seems a little close to implementation details, if the error message is in the document it seems like we might not need the spy? I might be missing its purpose

sites/partners/src/pages/_app.tsx
```diff
@@ -20,6 +20,12 @@ function BloomApp({ Component, router, pageProps }: AppProps) {
   const { locale } = router
   const skipLoginRoutes = ["/forgot-password", "/reset-password", "/users/confirm", "/users/terms"]
 
+  // fix for rehydation
```
emilyjablonski, line 23: "rehydration"

sites/public/cypress/e2e/pages/application/contact/address.spec.ts
```diff
@@ -25,6 +25,8 @@ describe("applications/contact/address", function () {
   })
 
   it("should provide a way to validate address via API", function () {
+    // fake the address call to the mocked data
+    cy.intercept("GET", "/geocoding/v5/**", { fixture: "address" })
```
emilyjablonski, line 29: ❤️

sites/public/cypress/support/e2e.js
```diff
@@ -12,7 +12,7 @@
 // You can read more here:
 // https://on.cypress.io/configuration
 // ***********************************************************
-import "@cypress/code-coverage/support"
+// import "@cypress/code-coverage/support"
```
emilyjablonski, line 15: Should we remove this?

emilyjablonski (commented):
Just a few code comments!

sites/partners/__tests__/pages/users/index.test.tsx
ludtkemorgan: I don't love doing this, but this is a way for the test output to be a little less cluttered. This test purposely throws an error so an error  is expected, however when it appears in the test logs it looks like an error happened in the test even though the test passes. So this is just to make it so that those logs are hidden. However, I'm ok not doing this since it does all mute all console logging in the test

ColinBuyck (approved):
LGTM 🆙

shared-helpers/src/auth/AuthContext.ts
```diff
@@ -279,7 +279,9 @@ export const AuthProvider: FunctionComponent = ({ children }) => {
         const response = await authService?.login({ body: { email, password, mfaCode, mfaType } })
         if (response) {
           dispatch(saveToken({ accessToken: response.accessToken, apiUrl, dispatch }))
-          const profile = await userService?.userControllerProfile()
+          const profile = await userService?.userControllerProfile({
```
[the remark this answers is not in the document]
ludtkemorgan: The dispatch one line above saves the token and makes it so when any downstream api calls now have a bearer token set as a header. However with one of the changes (most likely the react upgrade) it is no longer being set by the time the `userService?.userControllerProfile` is fired. My theory is that there was a slight change in how the reducer is acting. This change gets around it by manually setting the header for this specific call. Without it you are unable to log in on the partner site.

sites/public/cypress/support/e2e.js
ludtkemorgan: I was able to add it back in by upgrading the version of it. I had originally commented it out because all of the cypress tests were failing after they had finished and I traced it back to this being the culprit. However, I forgot to go back and investigate since it wasn't happening in the partner tests.

PR #3365, opened 2023-03-28
fix: fix partner unit tests

ludtkemorgan:
This fixes the two broken partner unit tests with the auth cookie change. It is pointed to the cookie update PR

1. Update the mock service worker urls to point to the new `/api/adapter/` endpoint
2. The token is no longer set as a session storage. So to simulate a logged in user we now need to set the cookie


YazeedLoonat (approved):
LGTM

PR #3382, opened 2023-04-06
fix: react type errors

ludtkemorgan:
Pulls in the latest ui-component changes. Functionally there is nothing different in ui-components, it just now accurately updates the `@types/react` to 18. Because of this we also needed to update the types in core. The small changes are mostly around how the types for 18 are more strict on what can be a child and that "children" has to now be explicitly listed in the prop types


emilyjablonski (approved):

ColinBuyck (approved):
LGTM ⚾

PR #3407, opened 2023-04-18
updates proxy to support access control allow list

seanmalbert:
This fixes the cors issue application start pages are seeing by updating Nginx to support passing in an allow list (now a required parameter).

I also updated the README to include instructions for running locally.

04/24 update:
There is also an issue with the purge requests. This was because of two reasons.
1. Newest NGINX version fails on the purge requests. My theory is that it is functionality that moved to the paid version of nginx. 
    - Resolved by downgrading nginx to a version that works (still newer than what is in prod)
2. The partner site was where we were making the call to the proxy for the purge after listing edit. This was causing a CORs error
    - Resolved by moving the purge calls to the backend.


backend/proxy/shared-location.conf
```diff
@@ -6,9 +6,8 @@
     proxy_cache webapp_cache;
     proxy_cache_use_stale error timeout http_500 http_502 http_503 http_504;
     proxy_cache_key $uri$is_args$args$http_language;
-    if ($request_method = 'PURGE') {
```
ludtkemorgan, line 9: Is there ever a case where a PURGE request won't come from the allow list? I'm curious on the original intent of allowing purges from any domain

ludtkemorgan:
@seanmalbert this is working locally, but when I deploy it to Heroku the issue still persists. I deployed it and added the allow list here: https://dashboard.heroku.com/apps/bloom-backend-main-proxy/settings. But when going to https://bloom.exygy.dev/ and applying to a listing the issue still persists

backend/proxy/shared-location.conf
seanmalbert: No, noting the TODO comment, we just never got back to it until now.

seanmalbert:
> @seanmalbert this is working locally, but when I deploy it to Heroku the issue still persists. I deployed it and added the allow list here: https://dashboard.heroku.com/apps/bloom-backend-main-proxy/settings. But when going to https://bloom.exygy.dev/ and applying to a listing the issue still persists

There's something funny going on here that I ran into locally while testing things. The header is getting set, but incorrectly: `The 'Access-Control-Allow-Origin' header contains multiple values 'https://bloom.exygy.dev, https://bloom.exygy.dev', but only one is allowed.`

ludtkemorgan:
The deployment issue has been resolved. It was applying two of the same header. Most recent change removes the header from the backend and reapplies it from the proxy.

emilyjablonski:
@ludtkemorgan would you recommend testing locally or in a netlify environment?

ludtkemorgan:
> @ludtkemorgan would you recommend testing locally or in a netlify environment?

It is probably best to test it in the deployed environment https://bloom.exygy.dev/ and https://partners.bloom.exygy.dev/ since the issue is specific to the proxy setup

backend/proxy/README.md
```diff
@@ -31,6 +52,8 @@ Now you can sign into Container Registry.
 $ heroku container:login
 ```
 
+_Note_ if you are using an Apple M1 device the following steps will not work. You will have to do the following: https://stackoverflow.com/a/67001433
```
seanmalbert, line 55: @ludtkemorgan , I have an M1, and I don't think I've had to do this in the past, FYI.

YazeedLoonat (changes requested):
Hey @ludtkemorgan I think this looks good, I did run into that problem with the listings not being purged properly

I did the following:
look at public site and pick a listing
go to partner site and update the name of that listing
save and look back at the public site

the listing wasn't updated. I did try adding localhost:3100 `
localhost:3000|localhost:3001|localhost:3100` as the ALLOW_LIST variable and I was still unable to purge the listing

backend/proxy/README.md
ludtkemorgan: @seanmalbert I ran the steps listed in this README again and I got the same error again. When it deploys to Heroku the logs show this:
```
Apr 24 10:31:38 [bloom-backend-main-proxy](https://my.papertrailapp.com/systems/bloom-backend-main-proxy/events?focus=1587209679106510850&selected=1587209679106510850) [heroku/web.1](https://my.papertrailapp.com/events?q=program%3Aheroku%2Fweb.1&focus=1587209679106510850&selected=1587209679106510850) State changed from starting to crashed
Apr 24 10:31:38 [bloom-backend-main-proxy](https://my.papertrailapp.com/systems/bloom-backend-main-proxy/events?focus=1587209679222063112&selected=1587209679222063112) [app/web.1](https://my.papertrailapp.com/events?q=program%3Aapp%2Fweb.1&focus=1587209679222063112&selected=1587209679222063112) Error: Exec format error
Apr 24 10:31:38 [bloom-backend-main-proxy](https://my.papertrailapp.com/systems/bloom-backend-main-proxy/events?focus=1587209679268962304&selected=1587209679268962304) [heroku/web.1](https://my.papertrailapp.com/events?q=program%3Aheroku%2Fweb.1&focus=1587209679268962304&selected=1587209679268962304) Process exited with status 126
```

ludtkemorgan:
@YazeedLoonat I moved the purge api call from the partner site to the backend and it is now working. Also I had to downgrade the version of nginx to solve the purge requests failing so you'll need to rebuild the proxy docker image locally again to test this.

backend/core/src/listings/listings.service.ts
```diff
@@ -143,7 +146,33 @@ export class ListingsService {
...
+    const saveResponse = await this.listingRepository.save(listing)
+    /**
+     * Send purge request to Nginx.
+     * Wrapped in try catch, because it's possible that content may not be cached in between edits,
+     * and will return a 404, which is expected.
+     * listings* purges all /listings locations (with args, details), so if we decide to clear on certain locations,
+     * like all lists and only the edited listing, then we can do that here (with a corresponding update to nginx config)
+     */
+    if (process.env.PROXY_URL) {
+      firstValueFrom(
```
YazeedLoonat, line 158: I think we need to await the `firstValueFrom()` calls or there is a chance that an error is swallowed with the function already having been returned

YazeedLoonat (changes requested):
Hey @ludtkemorgan looks good, looking for your thoughts on await the proxy call

backend/core/src/listings/listings.service.ts
ludtkemorgan: Good point. I chose not to do an await as I thought the calls shouldn't be a blocker to the main request. But that makes sense, especially since we are still catching the errors so it shouldn't block the overall call to the update endpoint

ludtkemorgan: @YazeedLoonat this has been added

YazeedLoonat (approved):
LGTM

ludtkemorgan (approved):

PR #3419, opened 2023-04-27
fix: downgrade version of nest axios

ludtkemorgan:
## Issue Overview

This PR addresses _Heroku deployment_

## Description

With the most recent change of moving the purge calls from partners to the backend the deployment to the heroku failed. This is due to the way we have our backend setup. The yarn lock file is at the base level of the bloom repo and when deploying the backend all it knows about is what is in the backend/core directory. Because of this it does not find a yarn lock file and instead tries to do a fresh install using npm and the package.json file. For the most recent change it grabs the version of `@nestjs/axios` specified in package.json (2.0.0) and tries to download all of the dependencies. We don't have a high enough version of axios that we are requesting so the install fails.

This PR solves that problem by downgrading @nestjs/axios which is ok with the older version of axios. Long term we would like to solve this by upgrading both @nestjs/axios and axios but that will require changes to Partners and Public so this is a quick fix to solve it in the mean time.

In order to test this I have deployed it to bloom backend main heroku (the one that normally deploys on merges to main). So you can test it here https://bloom.exygy.dev/ and https://partners.bloom.exygy.dev/. But it also doesn't hurt to pull it down locally to test

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/backend/core/package.json
+++ b/backend/core/package.json
@@ -40,7 +40,7 @@
   "dependencies": {
     "@anchan828/nest-sendgrid": "^0.3.25",
     "@google-cloud/translate": "^6.2.6",
-    "@nestjs/axios": "2.0.0",
+    "@nestjs/axios": "1.0.1",
     "@nestjs/cli": "^8.2.1",
     "@nestjs/common": "^8.3.1",
     "@nestjs/config": "^1.2.0",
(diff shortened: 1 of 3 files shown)


YazeedLoonat (approved):
LGTM

ColinBuyck (approved):
LGTM 🕹️

Issue #3426, opened 2023-05-05
Prisma: Schema Generation


YazeedLoonat:
Goal:
        Produce a schema that is based on the current hba/core database
        merge into it the detroit db schema
        clean up the generated schema
        
The outcome should be:
        a pr into a feature branch should be the outcome of this

PR #3429, opened 2023-05-06
feat: prisma schema generation

YazeedLoonat:
## Issue Overview

This PR addresses[ #issue](https://github.com/bloom-housing/bloom/issues/3426)

## Description
This is a prisma schema generated by aiming prisma's db introspection at both the hba/core database and the detroit database.

Those 2 schemas were then stitched together and cleaned up to make the schema more usable.

This schema clean up included:

1. fixing name spaces (everything was generated as snake cased, so this was ch
anged to make the eventual api more usable)

2. Names of relationships and models/enums were autogenerated as very verbose so they were cleaned up so as to retain understandability and be more readable

3. comments were left in the schema to flag any validation that was tied to the old typeorm schema/non-database fields that will need to be recreated as the rest of the prisma change over is built out

4. in some spots in the old typeorm model's some columns were typed as enums but stored as simple strings. This was addressed in the change over so things should be stored as the enum not just a string representation

5. There were differences between detroit and hba/core, and then there were some changes that were made to hba/core but never found their way to detroit. 
I made the following list to make that more clear:
```
model address
    property

model ami_chart
    ami_chart_item
    unit_group_ami_levels

model applications
    application_preference (no data on prod)

model jurisdictions
    jurisdictions_preferences_preferences
    jurisdictions_programs_programs

model units
    status
    property_id
    property

model user_accounts
    user_preferences

model listings 
    property_id
    hrd_id
    owner_company
    management_company
    management_website
    ami_percentage_min
    ami_percentage_max
    phone_number
    temporary_listing_id (not sure whats the deal with this)
    is_verified
    what_to_expect_additional_text
    section8_acceptance
    neighborhood_amenities_id
    verified_at
    home_type
    listing_preferences
    listing_programs
    listing_neighborhood_amenities
    unit_group
    user_preferences_favorites_listings

    region (originally off of the property table)
```

Go through the set up steps listed out in the read me

This part will be tedious:
Look through the prisma schema and make sure the prisma schema makes sense to you

Compare the prisma schema to the old typeorm models

start writing some queries against the newly generated **bloom_prisma** db that is generated. Make sure things that were previously doable (joining listings to multiselect question, and so on) is still doable

look through the new readme file, make sure everything in there makes sense so far

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/src/app.controller.ts
```diff
@@ -0,0 +1,12 @@
+import { Controller, Get } from '@nestjs/common';
+import { AppService } from './app.service';
+
+@Controller()
+export class AppController {
```
ludtkemorgan, line 5: For this PR do we need to have all of the src files? These look like the default controller, service, etc that are generated for a nestjs app. If these aren't required could we limit this PR to the bare minimum?

backend_new/README.md
```diff
@@ -0,0 +1,29 @@
+# Setup
+## Installation
+Make sure the .env file's db placement is what works for your set up, Then run the following:
```
ludtkemorgan, line 3: Can you expand on this? I'm struggling to get the db setup command to work since I don't know what the format of the .env file should look like.

backend_new/src/app.controller.ts
YazeedLoonat: I feel like getting this set up with the default nest set up will make some of the other pieces easier later down the line, I'd rather leave these in place for now

if it feels like there is too much review in this though I can defo pull it out. your call!

YazeedLoonat: I feel like having the starter nest app makes a few steps later easier, so what I ended up doing is merging that into the feature branch, which shrinks this pr down to only parts related to prisma

lemme know if that works for you!

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1102 @@
...
+  resolvingUserId       String?              @map("resolving_user_id") @db.Uuid
+  userAccounts          UserAccounts?        @relation(fields: [resolvingUserId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  listings              listings             @relation(fields: [listingId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  applications          Applications[]
+
+  @@index([listingId])
+  @@map("application_flagged_set")
+}
+
+// model application_flagged_set_applications_applications {
```
ludtkemorgan, line 149: What is the plan for these seven commented out models?
```
application_flagged_set_applications_applications
applications_preferred_unit_unit_types
jurisdictions_multiselect_questions_multiselect_questions
listings_leasing_agents_user_accounts
user_accounts_jurisdictions_jurisdictions
unit_group_unit_type_unit_types
user_preferences_favorites_listings
```

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1102 @@
...
+  userId       String?       @map("user_id") @db.Uuid
+  userAccounts UserAccounts? @relation(fields: [userId], references: [id], onUpdate: NoAction)
+
+  @@map("activity_log")
+}
+
+// Note: [place_name, city, county, state, street, street2, zip_code] formerly had max length of 64 characters
+model Address {
+  id                           String            @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt                    DateTime          @default(now()) @map("created_at") @db.Timestamp(6)
```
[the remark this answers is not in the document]
YazeedLoonat: so those types are generated by prisma itself to map our database structure, I don't think it would cause any differences since thats the prisma way of getting us the exact same db column

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1102 @@
...
+  recordId     String        @map("record_id") @db.Uuid
+  action       String        @db.VarChar
+  metadata     Json?
+  userId       String?       @map("user_id") @db.Uuid
+  userAccounts UserAccounts? @relation(fields: [userId], references: [id], onUpdate: NoAction)
+
+  @@map("activity_log")
+}
+
+// Note: [place_name, city, county, state, street, street2, zip_code] formerly had max length of 64 characters
```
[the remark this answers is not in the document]
YazeedLoonat: thats a solid question, that validation is really only important when we hit the create/update step (we don't validate data thats already in the system only data that is going to be added into the system)


my initial thoughts is to leverage DTOs to add this validation on those create/update endpoints. I think there may be a few other options but I think that research will need to be done with the create/update prisma work

backend_new/prisma/schema.prisma
YazeedLoonat: ah dang, those aren't really "models" they were part of the old structure but we can let prisma generate those the way it needs to for its relationships.

removed em

backend_new/prisma/migrations/20230506172411_init/migration.sql
ludtkemorgan, line 1: I know this is needed for now. But do you have a plan for how we will run the migrations on the existing DBs? Were you thinking we would create another migration file that just makes the changes from the existing database? Like adding the new tables pulled from Detroit and what not

ludtkemorgan (approved):
This looks great! We can discuss further down the road about the deployment plan

backend_new/prisma/migrations/20230506172411_init/migration.sql
YazeedLoonat: yeah I think the game plan would be to create an initial migration for hba/detroit (from hba/detroit db introspection)

then do a second migration that takes that starting point and gets us to this endpoint 

there may be a bit tweaking we'll need to do to make sure data is correct, but I this is probs the easiest approach

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+}
+
+datasource db {
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
+}
+
+model Accessibility {
+  id           String        @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt    DateTime      @default(now()) @map("created_at") @db.Timestamp(6)
```
emilyjablonski, line 12: createdAt / updatedAt / id in core are optional, just want to confirm this is an intentional change

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  action       String        @db.VarChar
+  metadata     Json?
+  userId       String?       @map("user_id") @db.Uuid
+  userAccounts UserAccounts? @relation(fields: [userId], references: [id], onUpdate: NoAction)
+
+  @@map("activity_log")
+}
+
+// Note: [place_name, city, county, state, street, street2, zip_code] formerly had max length of 64 characters
+model Address {
```
emilyjablonski, line 37: city, state, street, and zipCode are mandatory in core

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  applicationMailingAddress    listings[]        @relation("application_mailing_address")
+  leasingAgentAddress          listings[]        @relation("leasing_agent_address")
+  applicationPickUpAddress     listings[]        @relation("application_pick_up_address")
+  buildingAddress              listings[]        @relation("building_address")
+
+  @@map("address")
+}
+
+// Note: [type, phone_number] formerly max length 16; [other_type, first_name, last_name] formerly max length 64; [agency] formerly max length 128
+model AlternateContact {
```
emilyjablonski, line 67: email has an "isEmail" validator

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  applications     Applications?
+
+  @@map("alternate_contact")
+}
+
+model AmiChart {
+  id                 String               @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt          DateTime             @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt          DateTime             @updatedAt @map("updated_at") @db.Timestamp(6)
+  items              Json
```
emilyjablonski, line 89: this was previously typed as `AmiChartItem` is that lost here? i see we have the relationship below

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  units              Units[]
+
+  @@map("ami_chart")
+}
+
+// Note: [birth_month, birth_day, birth_year] formerly max length 8; [phone_number, phone_number_type] formerly max length 16; 
+//       [first_name, middle_name, last_name] formerly max length 64
+// Note: [first_name, last_name] formerly min length 1
+// Note: [email_address] needs to have lower case enforcement
+model Applicant {
```
emilyjablonski, line 104: email validator is missing,

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  lastName             String?       @map("last_name")
+  birthMonth           String?       @map("birth_month")
+  birthDay             String?       @map("birth_day")
+  birthYear            String?       @map("birth_year")
+  emailAddress         String?       @map("email_address")
+  noEmail              Boolean?      @map("no_email")
+  phoneNumber          String?       @map("phone_number")
+  phoneNumberType      String?       @map("phone_number_type")
+  noPhone              Boolean?      @map("no_phone")
+  workInRegion         YesNoEnum?    @map("work_in_region")
```
emilyjablonski, line 119: wondering why this isn't a boolean?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  id                    String               @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt             DateTime             @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt             DateTime             @updatedAt @map("updated_at") @db.Timestamp(6)
+  rule                  RuleEnum
+  ruleKey               String               @unique() @map("rule_key") @db.VarChar
+  resolvedTime          DateTime?            @map("resolved_time") @db.Timestamptz(6)
+  listingId             String               @map("listing_id") @db.Uuid
+  showConfirmationAlert Boolean              @default(false) @map("show_confirmation_alert")
+  status                FlaggedSetStatusEnum @default(pending)
+  resolvingUserId       String?              @map("resolving_user_id") @db.Uuid
```
emilyjablonski, line 140: this was previously a relation, just making sure that's intentional

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+
+// Note: [rule] used to be a different enum but prisma doesn't support that kind of enum yet. See: https://github.com/prisma/prisma/issues/273
+model ApplicationFlaggedSet {
+  id                    String               @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt             DateTime             @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt             DateTime             @updatedAt @map("updated_at") @db.Timestamp(6)
+  rule                  RuleEnum
+  ruleKey               String               @unique() @map("rule_key") @db.VarChar
+  resolvedTime          DateTime?            @map("resolved_time") @db.Timestamptz(6)
+  listingId             String               @map("listing_id") @db.Uuid
```
emilyjablonski, line 137: do we need listingId with the listings relation?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  @@map("application_methods")
+}
+
+// Note: [additional_phone_number, additional_phone_number_type, household_size] formerly max length 16; 
+//       [contact_preferences, income] formerly max length 64;
+//       [app_url] formerly max length 256; 
+// Note: [contact_preferences] formerly max array length of 8
+// Note: [household_member] formerly had max array lenght of 32
+// Note: missing virtual [flagged] field
+model Applications {
```
emilyjablonski, line 172: is this maybe supposed to be a singular `Application`?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  additionalPhoneNumberType    String?                       @map("additional_phone_number_type")
+  contactPreferences           String[]                      @map("contact_preferences")
+  householdSize                Int?                          @map("household_size")
+  housingStatus                String?                       @map("housing_status")
+  sendMailToMailingAddress     Boolean?                      @map("send_mail_to_mailing_address")
+  householdExpectingChanges    Boolean?                      @map("household_expecting_changes")
+  householdStudent             Boolean?                      @map("household_student")
+  incomeVouchers               Boolean?                      @map("income_vouchers")
+  income                       String?
+  incomePeriod                 IncomePeriodEnum              @map("income_period")
```
emilyjablonski, line 189: Seems like this should be optional

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  preferences                  Json
```
emilyjablonski, line 190: I know it wasn't before, but it seems like this should be optional (and recognizing a multiselection question refactor is out of scope) - did we lose typing here for both programs and preferences?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  programs                     Json?
+  status                       ApplicationStatusEnum
+  language                     LanguagesEnum
```
emilyjablonski, line 193: This was previously optional

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  submissionType               ApplicationSubmissionTypeEnum @map("submission_type")
+  acceptedTerms                Boolean?                      @map("accepted_terms")
+  submissionDate               DateTime?                     @map("submission_date") @db.Timestamptz(6)
+  // if this field is true then the application is a confirmed duplicate
+  // meaning that the record in the application flagged set table has a status of duplicate
+  markedAsDuplicate            Boolean                       @default(false) @map("marked_as_duplicate")
```
emilyjablonski, line 199: Looks like we're missing the flagged field, I see you noted this, can we not add it here?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  confirmationCode             String                        @map("confirmation_code")
+  reviewStatus                 ApplicationReviewStatusEnum   @default(valid) @map("review_status")
```
emilyjablonski, line 201: Was optional previously

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  preferredUnitTypes           UnitTypes[]
+  householdMember              HouseholdMember[]
+
+  @@unique([listingId, confirmationCode])
+  @@index([listingId])
+  @@map("applications")
+}
+
+// Note: [file_id, label] formerly max length 128
+model Assets {
```
emilyjablonski, line 228: Is this meant to be singular?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  id                            String              @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt                     DateTime            @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt                     DateTime            @updatedAt @map("updated_at") @db.Timestamp(6)
+  fileId                        String              @map("file_id")
+  label                         String
+  listingEvents                 ListingEvents[]
```
emilyjablonski, line 234: Am not seeing `listingEvents` or `listingImages` in `asset.entity`

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  @@map("assets")
+}
+
+// Note: [name] formerly max length 64
+model CronJob {
+  id          String    @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  name        String?
+  lastRunDate DateTime? @map("last_run_date") @db.Timestamptz(6)
+  createdAt   DateTime  @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt   DateTime  @updatedAt @map("updated_at") @db.Timestamp(6)
```
emilyjablonski, line 249: Wondering if these fields should also default to now

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+// Note: [how_did_you_hear, race] formerly max array length 64
+model Demographics {
+  id                String        @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt         DateTime      @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt         DateTime      @updatedAt @map("updated_at") @db.Timestamp(6)
+  ethnicity         String?
+  gender            String?
+  sexualOrientation String?       @map("sexual_orientation")
+  howDidYouHear     String[]      @map("how_did_you_hear")
+  race              String[]
```
emilyjablonski, line 264: Should be optional

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  middleName                 String?       @map("middle_name")
+  lastName                   String?       @map("last_name")
+  birthMonth                 String?       @map("birth_month")
+  birthDay                   String?       @map("birth_day")
+  birthYear                  String?       @map("birth_year")
+  emailAddress               String?       @map("email_address")
+  phoneNumber                String?       @map("phone_number")
+  phoneNumberType            String?       @map("phone_number_type")
+  noPhone                    Boolean?      @map("no_phone")
+  sameAddress                YesNoEnum     @map("same_address")
```
emilyjablonski, line 302: Should this + the similar one below be a boolean?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  multiselectQuestions        MultiselectQuestions[]
+  listings                    listings[]
+  reservedCommunityTypes      ReservedCommunityTypes[]
+  translations                Translations[]
+  user_accounts               UserAccounts[]
+
+  @@map("jurisdictions")
+}
+
+model ListingEvents {
```
emilyjablonski, line 341: Might be intended to be singular

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  id        String            @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt DateTime          @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt DateTime          @updatedAt @map("updated_at") @db.Timestamp(6)
+  type      MarketingTypeEnum
```
emilyjablonski, line 345: I don't think this is the right enum "ListingEventType"

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  startDate DateTime?         @map("start_date") @db.Timestamptz(6)
+  startTime DateTime?         @map("start_time") @db.Timestamptz(6)
+  endTime   DateTime?         @map("end_time") @db.Timestamptz(6)
+  url       String?
+  note      String?
+  label     String?
+  listingId String?           @map("listing_id") @db.Uuid
+  fileId    String?           @map("file_id") @db.Uuid
+  assets    Assets?           @relation(fields: [fileId], references: [id], onDelete: NoAction, onUpdate: NoAction)
```
emilyjablonski, line 354: I think I'm just not understanding these, an asset can't exist on more than one listing, so not sure why we would have listings here, or what assets is related to

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  acInUnit                Boolean?  @map("ac_in_unit")
+  hearing                 Boolean?
+  visual                  Boolean?
+  mobility                Boolean?
+  barrierFreeUnitEntrance Boolean?  @map("barrier_free_unit_entrance")
+  loweredLightSwitch      Boolean?  @map("lowered_light_switch")
+  barrierFreeBathroom     Boolean?  @map("barrier_free_bathroom")
+  wideDoorways            Boolean?  @map("wide_doorways")
+  loweredCabinets         Boolean?  @map("lowered_cabinets")
+  listings                listings?
```
emilyjablonski, line 384: should this be related to a single listing?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+
+  @@map("listing_features")
+}
+
+model ListingImages {
```
emilyjablonski, line 389: Likely singular

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  imageId   String   @map("image_id") @db.Uuid
+  assets    Assets   @relation(fields: [imageId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  listings  listings @relation(fields: [listingId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+
+  @@id([listingId, imageId])
+  @@index([listingId])
+  @@map("listing_images")
+}
+
+model ListingMultiselectQuestions {
```
emilyjablonski, line 401: Likely singular

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  phone       Boolean?
+  internet    Boolean?
+  listings    listings?
+
+  @@map("listing_utilities")
+}
+
+// Note: missing [referralApplication, applicationConfig, showWaitlist, unitsSummarized] virtual property
+// Note: [reserved_community_description, result_link] formerly max length 4096;
+model listings {
```
emilyjablonski, line 431: likely singular, other ones seem to be capitalized

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  unitAmenities                         String?                       @map("unit_amenities")
+  servicesOffered                       String?                       @map("services_offered")
+  yearBuilt                             Int?                          @map("year_built")
+  applicationDueDate                    DateTime?                     @map("application_due_date") @db.Timestamptz(6)
+  applicationOpenDate                   DateTime?                     @map("application_open_date") @db.Timestamptz(6)
+  applicationFee                        String?                       @map("application_fee")
+  applicationOrganization               String?                       @map("application_organization")
+  applicationPickUpAddressOfficeHours   String?                       @map("application_pick_up_address_office_hours")
+  applicationPickUpAddressType          ApplicationAddressTypeEnum?   @map("application_pick_up_address_type")
+  applicationDropOffDddressOffice_hours String?                       @map("application_drop_off_address_office_hours")
```
emilyjablonski, line 460: Address misspelled

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  applicationMailingAddressType         ApplicationAddressTypeEnum?   @map("application_mailing_address_type")
+  buildingSelectionCriteria             String?                       @map("building_selection_criteria")
+  costsNotIncluded                      String?                       @map("costs_not_included")
+  creditHistory                         String?                       @map("credit_history")
+  criminalBackground                    String?                       @map("criminal_background")
+  depositMin                            String?                       @map("deposit_min")
+  depositMax                            String?                       @map("deposit_max")
+  depositHelperText                     String?                       @map("deposit_helper_text")
+  disableUnitsAccordion                 Boolean?                      @map("disable_units_accordion")
+  leasingAgentEmail                     String?                       @map("leasing_agent_email")
```
emilyjablonski, line 471: Email validator

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  leasingAgentName                      String?                       @map("leasing_agent_name")
+  leasingAgentOfficeHours               String?                       @map("leasing_agent_office_hours")
+  leasingAgentPhone                     String?                       @map("leasing_agent_phone")
+  leasingAgentTitle                     String?                       @map("leasing_agent_title")
+  name                                  String
```
emilyjablonski, line 476: Remembering we have a different set of fields that are required for listing publish in core and Detroit, just thinking about the best way for us to manage that

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  jurisdictionId                        String?                       @map("jurisdiction_id") @db.Uuid
+  leasingAgentAddressId                 String?                       @map("leasing_agent_address_id") @db.Uuid
+  reservedCommunityTypeId               String?                       @map("reserved_community_type_id") @db.Uuid
+  resultId                              String?                       @map("result_id") @db.Uuid
+  featuresId                            String?                       @unique() @map("features_id") @db.Uuid
+  utilitiesId                           String?                       @unique() @map("utilities_id") @db.Uuid
+  // START DETROIT SPECIFIC
+  hrdId                                 String?                       @map("hrd_id")
+  ownerCompany                          String?                       @map("owner_company")
+  managementCompany                     String?                       @map("management_company")
```
emilyjablonski, line 513: I think these are duplicates + just Detroit's name for leasing agent

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  managementWebsite                     String?                       @map("management_website")
+  amiPercentageMin                      Int?                          @map("ami_percentage_min")
+  amiPercentageMax                      Int?                          @map("ami_percentage_max")
+  phoneNumber                           String?                       @map("phone_number")
+  temporaryListingId                    Int?                          @map("temporary_listing_id")
+  isVerified                            Boolean?                      @default(false) @map("is_verified")
+  marketingType                         MarketingTypeEnum             @default(marketing) @map("marketing_type")
```
emilyjablonski, line 520: Should be optional considering core doesn't use this

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  id        Int    @id() @default(autoincrement())
+  timestamp BigInt
+  name      String @db.VarChar
+
+  @@map("migrations")
+}
+
+// Note: missing [untranslatedText, untranslatedOptOutText] virtual fields
+// Note: [options] formerly had array max length 64
+model MultiselectQuestions {
```
emilyjablonski, line 568: Likely singular

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  optOutText         String?                                    @map("opt_out_text")
+  hideFromListing    Boolean?                                   @map("hide_from_listing")
+  applicationSection MultiselectQuestionsApplicationSectionEnum @map("application_section")
+  jurisdictions      Jurisdictions[]
+  listings           ListingMultiselectQuestions[]
+
+  @@map("multiselect_questions")
+}
+
+model PaperApplications {
```
emilyjablonski, line 586: Okay my b I thought only a few had the plural issue I will just note it for the whole file 😅

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  @@map("translations")
+}
+
+// Note: [name] formerly max length 256
+model UnitAccessibilityPriorityTypes {
+  id           String         @id() @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  createdAt    DateTime       @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt    DateTime       @updatedAt @map("updated_at") @db.Timestamp(6)
+  name         String
+  units        Units[]
```
emilyjablonski, line 632: Should these be relations?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  unitTypes                      UnitTypes?                      @relation(fields: [unitTypeId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  unitAccessibilityPriorityTypes UnitAccessibilityPriorityTypes? @relation(fields: [priorityTypeId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  listings                       listings?                       @relation(fields: [listingId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+
+  @@map("units_summary")
+}
+
+// Note: [mfa_code] formerly max length 16; [first_name, middle_name, last_name] formerly max length 64
+// Note: [email] formerly lower case enforced
+model UserAccounts {
```
emilyjablonski, line 738: This was previously called just "User"

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  listingId                      String?                         @map("listing_id") @db.Uuid
+  priorityTypeId                 String?                         @map("priority_type_id") @db.Uuid
+  unitTypes                      UnitTypes?                      @relation(fields: [unitTypeId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  unitAccessibilityPriorityTypes UnitAccessibilityPriorityTypes? @relation(fields: [priorityTypeId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  listings                       listings?                       @relation(fields: [listingId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+
+  @@map("units_summary")
+}
+
+// Note: [mfa_code] formerly max length 16; [first_name, middle_name, last_name] formerly max length 64
```
emilyjablonski, line 736: Some of these fields had "@expose" and others didn't, don't know if that's something we need to note

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  passwordValidForDays     Int                     @default(180) @map("password_valid_for_days")
+  resetToken               String?                 @map("reset_token") @db.VarChar
+  confirmationToken        String?                 @map("confirmation_token") @db.VarChar
+  confirmedAt              DateTime?               @map("confirmed_at") @db.Timestamptz(6)
+  email                    String                  @unique() @db.VarChar
+  firstName                String                  @map("first_name") @db.VarChar
+  middleName               String?                 @map("middle_name") @db.VarChar
+  lastName                 String                  @map("last_name") @db.VarChar
+  dob                      DateTime?               @db.Timestamp(6)
+  phoneNumber              String?                 @map("phone_number") @db.VarChar
```
emilyjablonski, line 751: We had a phone number validator

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  createdAt                DateTime                @default(now()) @map("created_at") @db.Timestamp(6)
+  updatedAt                DateTime                @updatedAt @map("updated_at") @db.Timestamp(6)
+  language                 LanguagesEnum
```
emilyjablonski, line 754: This was optional

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  mfaEnabled               Boolean                 @default(false) @map("mfa_enabled")
+  mfaCode                  String?                 @map("mfa_code") @db.VarChar
+  mfaCodeUpdatedAt         DateTime?               @map("mfa_code_updated_at") @db.Timestamptz(6)
+  lastLoginAt              DateTime                @default(now()) @map("last_login_at") @db.Timestamp(6)
+  failedLoginAttemptsCount Int                     @default(0) @map("failed_login_attempts_count")
+  phoneNumberVerified      Boolean?                @default(false) @map("phone_number_verified")
+  agreedToTermsOfService   Boolean                 @default(false) @map("agreed_to_terms_of_service")
+  hitConfirmation_url      DateTime?               @map("hit_confirmation_url") @db.Timestamptz(6)
```
emilyjablonski, line 762: Sneaky 🐍

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  applications             Applications[]
+  listings                 listings[]
+  jurisdictions            Jurisdictions[]
+  userPreferences          UserPreferences?
+  userRoles                UserRoles?
+
+  @@map("user_accounts")
+}
+
+model UserRoles {
```
emilyjablonski, line 776: We had two classes in this file, not sure if we "lost" the one that isn't an entity or need it

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  isJurisdictionalAdmin Boolean      @default(false) @map("is_jurisdictional_admin")
+  isPartner             Boolean      @default(false) @map("is_partner")
+  userId                String       @id() @map("user_id") @db.Uuid
+  userAccounts          UserAccounts @relation(fields: [userId], references: [id], onDelete: Cascade)
+
+  @@map("user_roles")
+}
+
+// START DETROIT SPECIFIC
+model AmiChartItem {
```
emilyjablonski, line 787: This is also in core

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  @@map("unit_group_ami_levels")
+}
+
+model UserPreferences {
+  sendEmailNotifications Boolean      @default(false) @map("send_email_notifications")
+  sendSmsNotifications   Boolean      @default(false) @map("send_sms_notifications")
+  userId                 String       @id() @unique() @map("user_id") @db.Uuid
+  favoriteIds            String[]     @default([]) @map("favorite_ids")
+  userAccounts           UserAccounts @relation(fields: [userId], references: [id], onDelete: Cascade, onUpdate: NoAction)
+  favoritesListings      listings[]
```
emilyjablonski, line 858: Should this be a relation?

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+  PaperPickup
+  POBox
+  LeasingAgent
+  Referral
+
+  @@map("application_methods_type_enum")
+}
+
+enum LanguagesEnum {
+  en
```
emilyjablonski, line 878: Detroit has Arabic and Bengali, not sure how to incorporate that

backend_new/prisma/schema.prisma
```diff
@@ -0,0 +1,1025 @@
...
+
+enum MarketingTypeEnum {
+  marketing
+  comingSoon
+
+  @@map("listings_marketing_type_enum")
+}
+
+enum RegionEnum {
+  Greater_Downtown
```
emilyjablonski, line 1012: This is super Detroit specific, makes me wonder if it should be in the data model this explicitly but idk

emilyjablonski:
Just a bunch of questions and findings!

backend_new/prisma/schema.prisma
YazeedLoonat: good question, this is really (almost strictly) for how things are stored in the db, if create/update has those as optional they are still defaulted/required in the database. 

That optionality will be addressed when we get to the create/update endpoints

backend_new/prisma/schema.prisma
YazeedLoonat: those are required when data is coming in from a front end/api call but aren't required in the db 

they all are of the form `@Column({ type: "text", nullable: true })`

backend_new/prisma/schema.prisma
YazeedLoonat: nice catch!

backend_new/prisma/schema.prisma
YazeedLoonat: good question, and this is now captured in a note

` @Column("jsonb")` is the db's column type, the type protection will have to be done during the create/update endpoint steps

backend_new/prisma/schema.prisma
YazeedLoonat: nice catch!

backend_new/prisma/schema.prisma
YazeedLoonat: yeah this is a bit confusing and we may want to take a pass (later on) to convert these to booleans but as it stands rn

```
  @Column({ type: "text", nullable: true })
  @Expose()
  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
  @IsIn(["yes", "no"], { groups: [ValidationsGroupsEnum.default] })
  workInRegion?: string | null
```
workInRegion is technically stored as a string in the db, with a requirement the string is either "yes", "no", or null

this enum change just makes this easier to understand since we effectively this was an enum before

backend_new/prisma/schema.prisma
YazeedLoonat: yeah this ties stuff together but:
```
  resolvingUserId       String?              @map("resolving_user_id") @db.Uuid
  userAccounts          UserAccounts?        @relation(fields: [resolvingUserId], references: [id], onDelete: NoAction, onUpdate: NoAction)
```

basically that `resolvingUserId` is used in the `userAccounts` relation to connect those 2 models

backend_new/prisma/schema.prisma
YazeedLoonat: yeah so
```
  listingId             String               @map("listing_id") @db.Uuid
listings              Listings             @relation(fields: [listingId], references: [id], onDelete: NoAction, onUpdate: NoAction)
```

listingId in this case is used by that listings relation to explain to prisma how that relationship is connected (hopefully that makes sense)

backend_new/prisma/schema.prisma
YazeedLoonat: yeah I think this may be something we want to talk about, previously our naming convention was a bit all over the place (plural vs singular)

I wonder if you, @ludtkemorgan have a preference

I didn't think to establish a convention

backend_new/prisma/schema.prisma
YazeedLoonat: fantastic catch, updated

backend_new/prisma/schema.prisma
YazeedLoonat: typing for these would occur at a higher level, I don't know if we really need it to be optional since technical an empty `{}` or `[]` would work just as well 

I feel like that would be a good thing for us to consider later though (as a sort of pass through the schema)

backend_new/prisma/schema.prisma
YazeedLoonat: excellent catch!

backend_new/prisma/schema.prisma
YazeedLoonat: technically `flagged` was never stored in the db it was a virtual field

I think I would like to (in the future) have that as a true field in the db so we don't have to keep calculating that value

for now though it'll have to stay a virtual and we can address later

that virtual field will get built out later on when we get to the application get queries

backend_new/prisma/schema.prisma
YazeedLoonat: `@Column({ enum: ApplicationReviewStatus, default: ApplicationReviewStatus.valid })` it was optional at the DTO level but on the db level it was required with a default in case of null

backend_new/prisma/schema.prisma
YazeedLoonat: https://github.com/bloom-housing/bloom/pull/3429#discussion_r1199453550

good question, establishing a good convention for this is defo a good idea

backend_new/prisma/schema.prisma
YazeedLoonat: ah good question, this is one of the differences between prisma and typeorm 

in typeorm some relationships only need to live on one side (listingEvents had assests but assets didn't need to have listingEvents)

in prisma relations need to live on both sides 

so this is making explicit what used to be an implicit bijected relation

backend_new/prisma/schema.prisma
YazeedLoonat: good question! 
`@updatedAt` is prisma's way of handling this case (basically marks a column as an updatedAt column

so we don't need to manage defaults ourselves

backend_new/prisma/schema.prisma
YazeedLoonat: this is a difference between prisma and typeorm 

in typeorm arrays can be nulled out
in prisma optional lists are not nullable

this would just become the empty [] instead of null for us

backend_new/prisma/schema.prisma
YazeedLoonat: this is a similar case to something I wrote about above, we basically created an enum but the hard way previously, this makes explicit (and in less code) that this is an enum of Yes or No

backend_new/prisma/schema.prisma
YazeedLoonat: convention convo from above

backend_new/prisma/schema.prisma
YazeedLoonat: Sheeesh
nice catch

backend_new/prisma/schema.prisma
YazeedLoonat: sure, so:
```
  fileId    String?               @map("file_id") @db.Uuid
  assets    Assets?               @relation(fields: [fileId], references: [id], onDelete: NoAction, onUpdate: NoAction)
```

that fileId is really an asset's id 
then the assets relationship uses that id to make the connection between a listing event and an asset 

because of the structure here it is a 1 to 1 relationship as opposed to a 1 to many or many to many relationship

hope that helps!

backend_new/prisma/schema.prisma
YazeedLoonat: yeah so this is a confusing part of prisma but the actual guts of the relationship are defined on the listings model 

what is here is just the other side of it. Each one is only tied to 1 listing

backend_new/prisma/schema.prisma
YazeedLoonat: convention convo

backend_new/prisma/schema.prisma
YazeedLoonat: convention convo

backend_new/prisma/schema.prisma
YazeedLoonat: convention convo

but also good 👀 it should be capitalized camel case, will fix!

backend_new/prisma/schema.prisma
YazeedLoonat: ayyyyyyyyy fixed

backend_new/prisma/schema.prisma
YazeedLoonat: good catch

backend_new/prisma/schema.prisma
YazeedLoonat: good question!

I think that the requirements will need to shared across both the front end and the backend. In the backend DTOs are likely the solution here

but I think we can cross that bridge a bit when we go to bring detroit on to backend 2.0 

you're right though that some of these fields will need to become optional in the db once that happens (that wasn't something I corrected for on my passes through this)

backend_new/prisma/schema.prisma
YazeedLoonat: interesting, both of these fields are never filled out on detroit

may want to flag that as something to dig into elsewhere but I think its best for now to keep them as part of the schema and if we decide they can be removed later we can remove them via a migration

backend_new/prisma/schema.prisma
YazeedLoonat: I think the default should be sufficient (if core doesn't use it whether its null or a defaulted value won't really make a difference)

backend_new/prisma/schema.prisma
YazeedLoonat: convention convo

backend_new/prisma/schema.prisma
YazeedLoonat: LOL dang I was actually hoping you had found all of them and we could make a convention through that, but either way yea we need to figure out whether these should be singular or plural

backend_new/prisma/schema.prisma
YazeedLoonat: as mentioned above its the other side of a relationship so the work of relation things is done in the units model

backend_new/prisma/schema.prisma
YazeedLoonat: gosh I WISH!

`@Entity({ name: "user_accounts" })` we named it user_accounts in the db 

probably because I think users is a reserved word in postgres (or at least it can be in other dbs)

backend_new/prisma/schema.prisma
YazeedLoonat: excellent call out

that will get covered when we go to create the user get endpoints

but basically @expose means that the backend is clearing that to send it  out and when its missing it means we'll never let that field leave the backend

backend_new/prisma/schema.prisma
YazeedLoonat: nice catch!

backend_new/prisma/schema.prisma
YazeedLoonat: noice catch!

backend_new/prisma/schema.prisma
YazeedLoonat: 🐫

backend_new/prisma/schema.prisma
YazeedLoonat: we'll need it but its only important for DTO purposes not db purposes

YazeedLoonat: or actually we may not need it, I haven't quite thought out how its being used but tbh we may be able to just roll that into something else/prisma may have a better answer for us

backend_new/prisma/schema.prisma
YazeedLoonat: I'm not boosted, don't mind me <insert whistling here>

backend_new/prisma/schema.prisma
YazeedLoonat: yeah so this is a many to many relationship (many listings can be tied to many users and the reverse)

so this is how that relation is defined in prisma

backend_new/prisma/schema.prisma
YazeedLoonat: huh, weirdly we don't have those defined in detroit as viable options 

I'll have to investigate this a bit when we go to bring this over to detroit

good catch

backend_new/prisma/schema.prisma
YazeedLoonat: yeah I think eventually we'll want this to be a table and a relationship instead of a enum (like for sass this doesn't make sense as an enum)

but that's a bridge I think we cross as a 2nd pass through this

YazeedLoonat:
Hey @emilyjablonski this should be good to re-review

backend_new/prisma/schema.prisma
emilyjablonski: Wondering if this and programs beneath it should be the same, but either works for me!

backend_new/prisma/schema.prisma
emilyjablonski: Oh I was looking at the name of the class, which is `User`, my b!

emilyjablonski (approved):
🐫 🥳 😎

backend_new/prisma/schema.prisma
ludtkemorgan: Hmm, I'm on the fence about this. Ideally it should be called `Application` since it is singular. However, I think the idea of this prisma change over is that the table names don't change (even if they are not the name we would have chosen today). It looks like we can change the name of the model but still keep the same table name with `@@map` at the bottom of the model. But then we run into the issue of the models don't match the table names for future reference. I could go either way on this. I'd prefer `Application`, but can see the argument for `Applications`. 

Let me know if I'm over thinking this

Issue #3463, opened 2023-05-23
Applying to Closed Listing Errors


ColinBuyck:
Even if a listing is closed, you’re still able to hit the apply url on a closed listing, and we aren’t blocking folks from filling it out (until it fails on submit). 

We need some sort of flow for hitting an application url of a closed listing. Potentially redirecting to the detail page of that listing and popping a “hey that application is closed” toast 

(credit to @emilyjablonski)

sarahlazarich:
@ColinBuyck can you add QA notes here please?

ColinBuyck:
Yes! QA Notes: 
1. setting a listing to closed or find a listing that is already closed.
2. Look to its url to find the listing id
3. Then hitting this url: https://bloom.exygy.dev/applications/start/choose-language?listingId=**listingIdThatYouJustClosed**

sarahlazarich:
this looks good!!!!! thank you.

PR #3467, opened 2023-05-24
fix: redirect and alert language

ColinBuyck:
# Pull Request Template

## Issue Overview

This PR addresses #3463 

## Description

This fix redirects a user who lands at the beginning of an application for a listing that is closed. Couple notes:

1. Did my best on the language with the intention of it being clear while avoiding the repetition of "Applications Closed" seen on the right. Also curious how translations should be handled here? @sarahlazarich @eajensenwa 
Screenshot 2023-05-24 at 12 21 06 AM
2. I recognize that this only covers the case of when a user hits the first step of an application for a closed listing directly as opposed to setting up the redirect for landing on any page. My thought is that, in the case of this quick fix, I shouldn't build in coverage for users skipping parts of our application since that seems like its own conversation/bug that isn't dependent on whether the listing is closed or not.

This can be tested by pulling it down locally, setting a listing to closed, and then hitting this url:

http://localhost:3000/applications/start/choose-language?listingId=**listingIdThatYouJustClosed**

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/shared-helpers/src/locales/general.json
+++ b/shared-helpers/src/locales/general.json
@@ -608,6 +608,7 @@
   "listings.applicationPerApplicantAgeDescription": "per applicant age 18 and over",
   "listings.applications": "Applications",
   "listings.applicationsClosed": "Applications Closed",
+  "listings.applicationsClosedRedirect": "This listing is no longer accepting applications.",
   "listings.apply.applicationWillBeAvailableOn": "Application will be available for download and pick up on %{openDate}",
   "listings.apply.applyOnline": "Apply Online",
   "listings.apply.downloadApplication": "Download Application",
--- a/sites/public/src/components/listing/ListingView.tsx
+++ b/sites/public/src/components/listing/ListingView.tsx
@@ -39,6 +39,7 @@ import {
   EventType,
   StandardTableData,
   ExpandableSection,
+  SiteAlert,
 } from "@bloom-housing/ui-components"
 import {
   cloudinaryPdfFromId,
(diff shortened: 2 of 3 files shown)


sarahlazarich:
I think the language in the error message is good, and agree w your thought on the related bug. We can translate in google for now (see below) and I'll let Izzie know to send jurisdictions an FYI. 

Translations:
- Filipino: Ang listahang ito ay hindi na tumatanggap ng mga aplikasyon
- Chinese: 此房源不再接受申请
- Vietnamese: Danh sách này không còn chấp nhận các ứng dụng
- Spanish: Esta lista ya no acepta solicitudes

eajensenwa:
language looks good to me!

ludtkemorgan (commented):
This looks good. I think it might be useful to add it to a few more pages. Specifically the other ones in the "start" folder (autofill and what-to-expect). Also it might be a good idea to do the last page of the flow so the user doesn't get the error until after they submit. Thoughts?

eajensenwa:
yep, looks great! 🎨

ludtkemorgan:
> @ludtkemorgan I wonder if we need to think more about a big picture solution before we start adding it to individual pages, I think ideally long term we just add this once, wondering if this can be our quick fix?

I briefly spoke to colin in the huddle. I think it makes sense to not add it to all of the pages so I'm ok with just having it on the choose-language page. However, I do think we would get a lot of bang for our buck if we put it on the last page as well for this quick fix. That way it will catch everyone, even if they come from a different page since we don't know exactly how this is happening at the moment. The long term fix will require a bit more considerations

emilyjablonski:
@ludtkemorgan makes sense, sounds great!

ColinBuyck:
Updated @emilyjablonski @ludtkemorgan 📟

ludtkemorgan (approved):
⏰

YazeedLoonat (approved):
LGTM

Issue #3471, opened 2023-05-24
Prisma: create all endpoints for the ami-charts


ludtkemorgan:
Create the entire controller and corresponding services for ami-charts.
 
*More details go here*

PR #3477, opened 2023-05-27
Prisma ami-charts

YazeedLoonat:
## Issue Overview

This PR addresses #[issue](https://github.com/bloom-housing/bloom/issues/3471)

## Description
This creates all the endpoints for ami charts

Follow the readme instructions for how to stand the new backend up
You should be able to test things out using the openai site at http://localhost:3101/api

you should be able to:
create, get a list of, update, get an individual, and delete ami charts

permission is not implemented yet so don't worry about testing that yet

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/src/services/ami-chart.service.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /*
+    this will create an ami chart
+  */
+  async create(incomingData: AmiChartCreate) {
+    const rawResult = await this.prisma.amiChart.create({
+      data: {
+        ...incomingData,
+        items: JSON.stringify(incomingData.items),
```
ludtkemorgan, line 89: I noticed the items field now escapes all of the quotes. It looks intentional based on this code, but do we have any concerns with this difference?
e.g this is in bloom_prisma
`"[{\"percentOfAmi\":50,\"householdSize\":3,\"income\":65850},`
vs this in the bloom db
`[{"income":65850,"percentOfAmi":50,"householdSize":3},`

backend_new/src/dtos/shared/id.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @IsUUID(4, { groups: [ValidationsGroupsEnum.default] })
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty()
+  id: string;
+
+  @Expose()
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty()
+  name?: string;
```
ludtkemorgan, line 17: The swagger docs are getting a bit confused with this on the jurisdiction field. The sample request thinks we need to send the jurisdiction name. It works without sending it. I wonder if there is a way to make the ApiProperty excluded for this. However, I'm guessing your intention is this IdDTO is suppose to be for more than just jurisdiction field. So that might not work
image

ludtkemorgan (commented):
Still making my way through it, but no big issues so far!

backend_new/prisma/seed-helpers/ami-chart-factory.ts
```diff
@@ -0,0 +1,27 @@
...
+  const toReturn = [];
+  for (let i = 0; i < numberToCreate; i++) {
+    toReturn.push({
+      percentOfAmi: i,
+      householdSize: i,
+      income: i,
+    });
+  }
+
+  return JSON.stringify(toReturn);
```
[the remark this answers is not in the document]
YazeedLoonat: good call, updated!

backend_new/src/services/ami-chart.service.ts
YazeedLoonat: I didn't catch this! I added a transform function to the DTO so if it happens to be stringified it is no longer stringified

backend_new/src/dtos/shared/id.dto.ts
YazeedLoonat: you're right adding @Apiproperty({ required: false }) fixes the swagger docs for us 

I think I would love if we don't need to add `required: false` to like every field so I'll spend some time figuring out if there is a better way forward

backend_new/prisma/seed-helpers/ami-chart-factory.ts
```diff
@@ -0,0 +1,22 @@
...
+  items: amiChartItemsFactory(i),
+  jurisdictions: {
+    connect: {
+      id: jurisdictionId,
+    },
+  },
+});
+
+const amiChartItemsFactory = (numberToCreate: number) => {
+  return JSON.stringify(
```
seanmalbert, line 17: It looks like this is returning `[null,null,null,null,null,null,null,null,null,null]` when `db:setup` is run. Unfortunately:

> callbackFn is invoked only for array indexes which have assigned values. It is not invoked for empty slots in [sparse arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Indexed_collections#sparse_arrays).

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#description, so you'll have to do something else.

It looks like the return type of `amiChartItemsFactory` could be `Prisma.JsonArray`, which also allows you to get rid of the JSON.stringify.

Something like this should work:
```
const amiChartItemsFactory = (numberToCreate: number): Prisma.JsonArray =>
  new Array(numberToCreate).fill(undefined).map((_, index) => ({
    percentOfAmi: index,
    householdSize: index,
    income: index,
  }));
```

seanmalbert (changes requested):
@YazeedLoonat ,
Got some initial feedback for you. See a couple of comments in the code. Also a question. I see that the `ami_chart` table is created with a jsonb column for items and that the seed populates those items, but there is also a `ami_chart_item` table that doesn't appear to be used. Should that go away?

backend_new/prisma/seed-helpers/ami-chart-factory.ts
seanmalbert: ooo, I think I'd prefer this over what I suggested above:
```
const amiChartItemsFactory = (numberToCreate: number): Prisma.JsonArray =>
  [...Array(numberToCreate)].map((_, index) => ({
    percentOfAmi: index,
    householdSize: index,
    income: index,
  }));
```

backend_new/prisma/seed-helpers/listing-factory.ts
```diff
@@ -357,17 +359,19 @@ const unitFactory = (
...
-          items: {},
-          name: `listing: ${i} unit: ${j} amiChart: ${j}`,
-          jurisdictions: {
-            connect: {
-              id: jurisdictionId,
+      amiChart: amiChartId
+        ? { connect: { id: amiChartId } }
+        : {
+            create: {
+              items: {},
```
seanmalbert, line 366: suggest that `items: {}` be `items: []`, since it's a JSON array if populated.

backend_new/src/controllers/ami-chart.controller.ts
```diff
@@ -0,0 +1,69 @@
...
+@ApiTags('amiCharts')
+@UsePipes(new ValidationPipe(defaultValidationPipeOptions))
+@ApiExtraModels(AmiChartCreate, AmiChartUpdate, IdDTO, AmiChartQueryParams)
+export class AmiChartController {
+  constructor(private readonly AmiChartService: AmiChartService) {}
+
+  @Get()
+  @ApiOperation({ summary: 'List amiCharts', operationId: 'list' })
+  @ApiOkResponse({ type: AmiChart, isArray: true })
+  async list(@Query() queryParams: AmiChartQueryParams) {
```
seanmalbert, line 38: suggestion: let's be thorough and specify the return type of functions.

backend_new/src/dtos/ami-charts/ami-chart-get.dto.ts
seanmalbert, line 1: suggestion: name the `get` dtos to match the class name better, so `ami-chart-get.dto.ts` => `ami-chart.dto.ts`

YazeedLoonat:
> @YazeedLoonat , Got some initial feedback for you. See a couple of comments in the code. Also a question. I see that the `ami_chart` table is created with a jsonb column for items and that the seed populates those items, but there is also a `ami_chart_item` table that doesn't appear to be used. Should that go away?

hey @seanmalbert 
I actually just mentioned this to @ludtkemorgan as it stands right now detroit uses that  table and core uses the json column. We defo want to standardize but I think some discovery on what converting the current json columns into actual tables does to row limits on our heroku dbs (don't want to blow over the limit and make the dev sites break)


my preference is to eventually switch the json into actual columns of data in that table

backend_new/package.json
```diff
@@ -22,7 +22,7 @@
     "db:migration:run": "yarn prisma migrate deploy",
     "db:seed": "yarn prisma db seed",
     "generate:client": "ts-node scripts/generate-axios-client.ts && prettier -w types/src/backend-swagger.ts",
-    "test:e2e": "yarn db:resetup && yarn db:migration:run && jest --config ./test/jest-e2e.json",
+    "test:e2e": "yarn db:resetup && yarn db:migration:run && jest --config ./test/jest-e2e.config.js --runInBand",
```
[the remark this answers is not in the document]
YazeedLoonat: thats a good callout! I ended up having to switch to runInBand because what I was running into was that tests were running simultaneously (so like creating more ami-charts or whatever) than the integration tests were expecting in their `list()` 

swiching to runInBand prevented this, I'm open to a better way for sure!

I just couldn't think of a better solution

backend_new/src/services/ami-chart.service.ts
```diff
@@ -0,0 +1,143 @@
...
+};
+
+@Injectable()
+export class AmiChartService {
+  constructor(private prisma: PrismaService) {}
+
+  /*
+    this will get a set of ami charts given the params passed in
+  */
+  async list(params: AmiChartQueryParams) {
```
seanmalbert, line 27: suggest adding return type to methods that are missing it here too

backend_new/src/services/ami-chart.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  async findOne(amiChartId: string) {
+    const amiChartRaw = await this.prisma.amiChart.findUnique({
+      include: view,
+      where: {
+        id: amiChartId,
+      },
+    });
+
+    if (!amiChartRaw) {
+      throw new NotFoundException();
```
seanmalbert, line 74: recommend adding a message that includes the id to make debugging easier

backend_new/test/integration/ami-chart.e2e-spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    const query = stringify(queryParams as any);
+
+    const res = await request(app.getHttpServer())
+      .get(`/amiCharts?${query}`)
+      .expect(200);
+
+    expect(res.body.length).toEqual(1);
+    expect(res.body[0].name).toEqual(amiChartA.name);
+
+    await cleanUpDb(
```
seanmalbert, line 73: recommend that this call gets moved to an afterEach, so a dev doesn't have to remember to add it to each test

backend_new/test/integration/ami-chart.e2e-spec.ts
```diff
@@ -0,0 +1,189 @@
...
+        },
+      });
+    }
+    for (let i = 0; i < jurisdictionIds.length; i++) {
+      await prisma.jurisdictions.delete({
+        where: {
+          id: jurisdictionIds[i],
+        },
+      });
+    }
```
seanmalbert, line 45: Prisma has a way to delete many records: https://www.prisma.io/docs/concepts/components/prisma-client/crud#delete-multiple-records

backend_new/src/controllers/ami-chart.controller.ts
```diff
@@ -0,0 +1,69 @@
...
+import { AmiChartUpdate } from '../dtos/ami-charts/ami-chart-update.dto';
+import { defaultValidationPipeOptions } from '../utilities/default-validation-pipe-options';
+import { AmiChartQueryParams } from '../dtos/ami-charts/ami-chart-query-params.dto';
+import { IdDTO } from '../dtos/shared/id.dto';
+import { SuccessDTO } from '../dtos/shared/success.dto';
+
+@Controller('/amiCharts')
+@ApiTags('amiCharts')
+@UsePipes(new ValidationPipe(defaultValidationPipeOptions))
+@ApiExtraModels(AmiChartCreate, AmiChartUpdate, IdDTO, AmiChartQueryParams)
```
seanmalbert, line 31: I know we have plans to update access control, but I think we should still add the guards we have in the mean time. Also please update the tests to include testing against someone who does not have permissions for CUD.

seanmalbert:
> > @YazeedLoonat , Got some initial feedback for you. See a couple of comments in the code. Also a question. I see that the `ami_chart` table is created with a jsonb column for items and that the seed populates those items, but there is also a `ami_chart_item` table that doesn't appear to be used. Should that go away?
> 
> hey @seanmalbert I actually just mentioned this to @ludtkemorgan as it stands right now detroit uses that table and core uses the json column. We defo want to standardize but I think some discovery on what converting the current json columns into actual tables does to row limits on our heroku dbs (don't want to blow over the limit and make the dev sites break)
> 
> my preference is to eventually switch the json into actual columns of data in that table

Can you call that out somewhere in the docs? If we don't need to filter or do any operations other than select the items as stored, I think my preference would be to stick with the jsonb column. That will probably both perform better and take up less space.

backend_new/package.json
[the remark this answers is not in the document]
YazeedLoonat: It was happening locally for me if I remember correctly its if listing e2e starts running before the amiChart e2e 

I went ahead and removed it and we can see as other prs go in if we run into race cases

backend_new/prisma/seed-helpers/listing-factory.ts
YazeedLoonat: updated!

backend_new/src/controllers/ami-chart.controller.ts
YazeedLoonat: function return types should now be specified!

backend_new/src/dtos/ami-charts/ami-chart-get.dto.ts
YazeedLoonat: cleaned up!

backend_new/src/services/ami-chart.service.ts
YazeedLoonat: thats so big 🧠 updated!

backend_new/test/integration/ami-chart.e2e-spec.ts
YazeedLoonat: ⚡

backend_new/src/controllers/ami-chart.controller.ts
YazeedLoonat: Hey Sean I think we should add guards/permissioning in a separate ticket since most of our operations require a logged in user and logging in is being handled in a later ticket 

adding the guards would more or less mean that nothing could be tested until that happens

YazeedLoonat:
Hey @ludtkemorgan Imma respond to ya here:
https://github.com/bloom-housing/bloom/pull/3477#discussion_r1218116940
I actually looked into this some more and we can now store without quotes!


https://github.com/bloom-housing/bloom/pull/3477#discussion_r1218130713
so in the old backend we had a `IdDto` which was just `id` and an `IdNameDto` which was both id and name instead of having both I felt like having an `IdDto` that covers both of those cases was best

In the case of the old backend IdDto was used primarily for when the front end is sending the backend data and IdnameDto was used mostly for the backend sending data to the front end 

so in our new case the IdDto has an optional Name filed now. Hopefully that makes some sense

backend_new/test/integration/ami-chart.e2e-spec.ts
YazeedLoonat: moved!

YazeedLoonat:
Hey @seanmalbert this is ready for you again, I ran into a race condition with using afterEach() to clear the db on the ami chart tests. I set this back I'm open to other ideas on how to use that properly!

backend_new/test/integration/ami-chart.e2e-spec.ts
seanmalbert: @YazeedLoonat , does it make sense to update this logic to use `deleteMany`?

backend_new/test/integration/ami-chart.e2e-spec.ts
seanmalbert: I don't see this updated?

backend_new/src/services/ami-chart.service.ts
```diff
@@ -0,0 +1,143 @@
...
+            id: params.jurisdictionId,
+          },
+        });
+      }
+    }
+
+    return {
+      AND: filters,
+    };
+  }
```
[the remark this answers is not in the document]
seanmalbert: I meant, not at the same time, but why support each of those one at a time? What are the cases where the ID is passed in and the name passed in? Can we narrow it down to one of them, like the id?

backend_new/package.json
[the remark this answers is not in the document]
seanmalbert: @YazeedLoonat , instead of clearing out the records before or after each test, can the tests rely on the data created within the test and then clear out the data at the very end?

backend_new/src/services/ami-chart.service.ts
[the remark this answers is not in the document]
YazeedLoonat: Oh my bad I misunderstood you were right there was no where in our codebase that was making the call using the name.

I removed the option to do so, so now its jurisdictionId or nothin

backend_new/package.json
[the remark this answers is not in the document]
YazeedLoonat: Sorry I sent you a direct message, for future reference I'm thinking its better to (in between tests) completely clear the db

I think the problem with not using runInBand is that if an amiChart test (say the list) is running and a listing test is running where an ami chart is created then there is a possibility of a conflict

I know it would be faster to remove runInBand but without doing something like hard resetting the db between tests I'm not sure if we can safely do so and guarantee no "collisions"

backend_new/src/services/ami-chart.service.ts
```diff
@@ -0,0 +1,146 @@
...
+  */
+  async findOne(amiChartId: string): Promise<AmiChart> {
+    const amiChartRaw = await this.prisma.amiChart.findUnique({
+      include: view,
+      where: {
+        id: amiChartId,
+      },
+    });
+
+    if (!amiChartRaw) {
```
emilyjablonski, line 64: I wonder if we should have the same messaging on delete? Right now looks like an internal server error happens

backend_new/src/services/ami-chart.service.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    return mapTo(AmiChart, rawResult);
+  }
+
+  /*
+    this will update an ami chart's name or items field
+    if no ami chart has the id of the incoming argument an error is thrown
+  */
+  async update(incomingData: AmiChartUpdate): Promise<AmiChart> {
```
emilyjablonski, line 97: I am able to call the update endpoint and add a jurisdiction with an ID that does not exist, might expect to get an error message in that case

emilyjablonski (commented):
Just some questions for you!

backend_new/package.json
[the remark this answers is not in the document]
YazeedLoonat: a few reasons:
1. our current backend tests also runInBand probably to address race conditions
2. our current test suite has a different paradigm. It expects the db to have data and it will modify that data set along with ending with data in the database

the new test suite is more similar to how civiform aims to do it where the db starts and ends blank between tests. 

I'm open to changing it to the old style but I think my preference is this newer way

backend_new/test/integration/ami-chart.e2e-spec.ts
YazeedLoonat: sorry yeah I tried to leave a note in a couple places, moving this into afterEach was running into race cases so I set it back

backend_new/README.md
```diff
@@ -114,3 +115,7 @@ Running the following will run all unit tests:
 ```bash
 $ yarn test
 ```
+
+# Considerations For Detroit
```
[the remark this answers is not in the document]
YazeedLoonat: yeah once we fully unify core/detroit this comment would go away as it stands right now the prisma schema has some stuff that only detroit has and some stuff that only core has but we will have both of those set up on the same backend when detroit gets this prisma update

backend_new/test/integration/ami-chart.e2e-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  it('testing create endpoint', async () => {
+    const jurisdictionA = await prisma.jurisdictions.create({
+      data: jurisdictionFactory(10),
+    });
+
+    const res = await request(app.getHttpServer())
+      .post('/amiCharts')
+      .send({
+        name: 'name: 10',
```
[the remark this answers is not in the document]
YazeedLoonat: I think it just makes it easier for me to verify that things are looking correct, I'm open to changing it

there is also a possibility that the seeding ticket will give us access to better data to test with and these tests will get updated at that time

backend_new/test/integration/ami-chart.e2e-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    });
+
+    const amiChartA = await prisma.amiChart.create({
+      data: amiChartFactory(10, jurisdictionA.id),
+    });
+
+    const res = await request(app.getHttpServer())
+      .put(`/amiCharts/${amiChartA.id}`)
+      .send({
+        id: amiChartA.id,
```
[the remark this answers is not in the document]
YazeedLoonat: ah yeah that was the main update here (or at least the first most easily visible) updating the name field 

kind of like above I think its okay to keep in place at least until we see how the seeding ticket plays out.

Defo open to changing this up though if you have a strong preference

backend_new/test/unit/services/ami-chart.service.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+  });
+
+  it('testing buildWhereClause() no params', () => {
+    const params: AmiChartQueryParams = {};
+    expect(service.buildWhereClause(params)).toEqual({
+      AND: [],
+    });
+  });
+
+  it('testing buildWhereClause() jurisdictionId param present', () => {
```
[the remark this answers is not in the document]
YazeedLoonat: sure I think we can set some time aside to chat about this (possibly this thursday) 

I'm not sure I completely understand what you mean by testing implementation details. Unit tests in this paradigm are testing our business logic and that function represents our business logic

backend_new/test/unit/services/ami-chart.service.spec.ts
```diff
@@ -0,0 +1,503 @@
...
+          householdSize: 3,
+          income: 3,
+        },
+      ],
+      name: 'ami name 4',
+      id: 'ami Id 3',
+    };
+
+    expect(await service.update(params)).toEqual({
+      id: 'ami Id 3',
```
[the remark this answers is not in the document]
YazeedLoonat: sure, much like before we are updating the name as part of the call, I'm open to making changes to this data though (especially since none of this is stored in the db)

technically the id would be a UUID in the db so any string here would suffice. I can change it to just the number if you prefer though

backend_new/test/unit/services/listing.service.spec.ts
```diff
@@ -68,6 +68,9 @@ const mockListing = (
           name: `AMI Name ${i}`,
           createdAt: date,
           updatedAt: date,
+          jurisdictions: {
```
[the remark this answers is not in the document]
YazeedLoonat: sure, that comes from the prisma schema and represents some work that will likely get standardized as part of our schema standardization ticketo

for now prisma/our old db has a mix of singular/plural so this has to match up with that

backend_new/types/src/backend-swagger.ts
```diff
@@ -180,6 +180,142 @@ export class ListingsService {
...
+
+      const configs: IRequestConfig = getConfigs(
+        'get',
+        'application/json',
+        url,
+        options,
+      );
+      configs.params = { jurisdictionId: params['jurisdictionId'] };
+
+      /** 适配ios13，get请求不允许带body */
```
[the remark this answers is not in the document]
YazeedLoonat: this is auto generated from the nest/swagger package hopefully they address this at some point but for now its probably going to be sticking around

backend_new/src/services/ami-chart.service.ts
YazeedLoonat: excellent call out! updated!

backend_new/src/services/ami-chart.service.ts
YazeedLoonat: as it stands right now we don't update the jurisdiction at all, its a good call out though. I removed it from the update DTO

YazeedLoonat:
Hey @seanmalbert @emilyjablonski I believe I got things updated lemme know what you think

emilyjablonski (approved):
LGTM! In general I do think for little details like unformatted ids or mismatching ids I'd prefer to clean those up now rather than leave them for later, but if it seems like all of that will become clean with the extra seeding ticket that works for me.

backend_new/src/dtos/shared/id.dto.ts
seanmalbert: @ludtkemorgan , is this comment thread resolved?

backend_new/package.json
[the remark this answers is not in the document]
seanmalbert: Noting here that there was some offline discussion and the db is not getting reset between tests. Instead, all tests should create and work with soley the data that was created for that particular test.

@emilyjablonski , feel free to resolve this convo once you've had all your questions answered.

seanmalbert (approved):

seanmalbert:
> LGTM! In general I do think for little details like unformatted ids or mismatching ids I'd prefer to clean those up now rather than leave them for later, but if it seems like all of that will become clean with the extra seeding ticket that works for me.

@YazeedLoonat , can you make sure this happens?

backend_new/src/dtos/shared/id.dto.ts
ludtkemorgan: The issue is still happening, but similar things are happening in the current setup. I have a ticket created to standardize our swagger docs with better sample code and standards. So this can be resolved for this PR

backend_new/test/unit/services/ami-chart.service.spec.ts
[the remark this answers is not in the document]
emilyjablonski: (@ludtkemorgan related to huddle chat rn)

PR #3489, opened 2023-05-30
fix: update backend-new circleci job

ludtkemorgan:
Fixes the broken circle ci jobs for the backend_new jest tests and linting.

A couple of call outs:
- Had to update the e2e test to clear all of the listing db tables before each run instead of doing a reseed as the `psql` command doesn't exist
- The top level linting task now does linting of all packages besides backend_new and then does backend_new separately
- Fixed several eslint errors that have existed for a long time. There are still 40ish warnings but no errors


backend_new/test/integration/listing.e2e-spec.ts
```diff
@@ -23,20 +23,19 @@ describe('Listing Controller Tests', () => {
     app = moduleFixture.createNestApplication();
     prisma = moduleFixture.get<PrismaService>(PrismaService);
     await app.init();
+    await clearAllDb();
```
YazeedLoonat, line 26: really like this change!

backend_new/src/services/listing.service.ts
```diff
@@ -117,7 +117,16 @@ export class ListingService {
     this set can either be paginated or not depending on the params
     it will return both the set of listings, and some meta information to help with pagination
   */
-  async list(params: ListingsQueryParams) {
+  async list(params: ListingsQueryParams): Promise<{
+    items: ListingGet[];
```
YazeedLoonat, line 121: is adding the type explicitly here a required change or do you feel like this is a better pattern (making the return type as explicit as possible I mean)

ludtkemorgan: This change shouldn't theoretically be needed as the type can be inferred by the returned object. However, when running the tests in ci it was saying it was "unknown" type. There is potentially a configuration change that would make it better pick up types

emilyjablonski (approved):
🥮

YazeedLoonat (approved):
LGTM

Issue #3496, opened 2023-06-02
Prisma: add translations service


ludtkemorgan:
In the existing backend there is a translations module, controller, and service. In the new setup all we really need is the service since the controller is not actually used by anything.

This ticket is to generate the translations service including all current functionality

PR #3478, opened 2023-05-28
feat: reserved community type

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3440)

- [x] This change addresses the issue in full

## Description
This creates all the endpoints for reserved community types

## How Can This Be Tested/Reviewed?
Follow the readme instructions for how to stand the new backend up
You should be able to test things out using the openai site at http://localhost:3101/api

you should be able to:
create, get a list of, update, get an individual, and delete reserved community types

permission is not implemented yet so don't worry about testing that yet

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have added QA notes to the issue with applicable URLs
- [x] I have performed a self-review of my own code
- [x] I have reviewed the changes in a desktop view
- [ ] I have reviewed the changes in a mobile view
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have assigned reviewers
- [x] I have run `yarn generate:client` and/or created a migration if I made backend changes that require them
- [x] My commit message(s) is/are polished, and any breaking changes are indicated in the message and are well-described
- [x] Commits made across packages purposefully have the same commit message/version change, else are separated into different commits

## Reviewer Notes:

Steps to review a PR:

- Read and understand the issue, and ensure the author has added QA notes
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Also review the acceptance criteria on the Netlify deploy preview (noting that these do not yet include any backend changes made in the PR)
- Either explicitly ask a clarifying question, request changes, or approve the PR if there are small remaining changes but the PR is otherwise good to go

## On Merge:

If you have one commit and message, squash. If you need each message to be applied, rebase and merge.

PR #3506, opened 2023-06-12
Prisma Translation service

ludtkemorgan:
## Issue Overview

This PR addresses #3496 

## Description

Add the translation service to the new Prisma backend. This does not require a controller so it is just the service that translates listings and gets the translations from the DB

This can only be partially tested as the one service endpoint is currently not being used.

Testing the `translateListing` endpoint:
- When running the backend locally, you can hit the get listing endpoint from swagger: http://localhost:3101/api#/listings/retrieve.
- Get a listing id from the db and hit the above endpoint with that listing id and a non-english language (e.g `es`)
- All of the translatable fields on the listing should be translated to the selected language in the output.
- Also these new translations should show up in the `generated_listing_translations` table. If you hit the endpoint again a call to google should not happen (would require console logging added to verify).

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/backend_new/src/services/translation.service.ts
+++ b/backend_new/src/services/translation.service.ts
@@ -16,32 +16,26 @@ export class TranslationService {
     language: LanguagesEnum,
     jurisdictionId: string | null,
   ) {
-    try {
-      return await this.prisma.translations.findUniqueOrThrow({
+    let translations = await this.prisma.translations.findUnique({
+      where: {
+        jurisdictionId_language: { language, jurisdictionId },
+      },
+    });
+
+    if (translations === null) {
+      console.warn(
+        `Fetching translations for ${language} failed on jurisdiction ${jurisdictionId}, defaulting to english.`,
+      );
+      translations = await this.prisma.translations.findUnique({
         where: {
-          jurisdictionId_language: { language, jurisdictionId },
+          jurisdictionId_language: {
+            language: LanguagesEnum.en,
+            jurisdictionId,
+          },
         },
       });
-    } catch (e) {
-      if (
-        e instanceof Prisma.PrismaClientKnownRequestError &&
-        e.code === 'P2025'
-      ) {
-        console.warn(
-          `Fetching translations for ${language} failed, defaulting to english.`,
-        );
-        return await this.prisma.translations.findUnique({
-          where: {
-            jurisdictionId_language: {
-              language: LanguagesEnum.en,
-              jurisdictionId,
-            },
-          },
-        });
-      } else {
-        throw e;
-      }
     }
+    return translations;
   }
 
   public async translateListing(listing: ListingGet, language: LanguagesEnum) {
--- a/backend_new/src/services/translation.service.ts
+++ b/backend_new/src/services/translation.service.ts
@@ -4,7 +4,6 @@ import { LanguagesEnum, Prisma } from '@prisma/client';
 import { ListingGet } from '../dtos/listings/listing-get.dto';
 import { GoogleTranslateService } from './google-translate.service';
 import * as lodash from 'lodash';
-import { NotFoundError } from '@prisma/client/runtime';
 
 @Injectable()
 export class TranslationService {
--- a/backend_new/src/services/translation.service.ts
+++ b/backend_new/src/services/translation.service.ts
@@ -1,21 +1,168 @@
 import { Injectable } from '@nestjs/common';
 import { PrismaService } from './prisma.service';
-import { LanguagesEnum } from '@prisma/client';
+import { LanguagesEnum, Prisma } from '@prisma/client';
+import { ListingGet } from '../dtos/listings/listing-get.dto';
+import { GoogleTranslateService } from './google-translate.service';
+import * as lodash from 'lodash';
+import { NotFoundError } from '@prisma/client/runtime';
 
 @Injectable()
 export class TranslationService {
-  constructor(private prisma: PrismaService) {}
+  constructor(
+    private prisma: PrismaService,
+    private readonly googleTranslateService: GoogleTranslateService,
+  ) {}
 
   public async getTranslationByLanguageAndJurisdictionOrDefaultEn(
     language: LanguagesEnum,
     jurisdictionId: string | null,
   ) {
-    const translations = await this.prisma.translations.findUnique({
-      where: {
-        jurisdictionId_language: { language, jurisdictionId },
+    try {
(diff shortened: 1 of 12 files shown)


backend_new/test/unit/services/translation.service.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+  'translated multiselect description',
+  'translated multiselect subtext',
+  'translated multiselect opt out text',
+];
+
+describe('Testing translations service', () => {
+  let service: TranslationService;
+  let prisma: PrismaService;
+  let googleTranslateServiceMock;
+  beforeEach(async () => {
```
YazeedLoonat, line 122: can we switch this to `beforeAll` instead

backend_new/src/services/translation.service.ts
```diff
@@ -0,0 +1,167 @@
...
+    private prisma: PrismaService,
+    private readonly googleTranslateService: GoogleTranslateService,
+  ) {}
+
+  public async getTranslationByLanguageAndJurisdictionOrDefaultEn(
+    language: LanguagesEnum,
+    jurisdictionId: string | null,
+  ) {
+    try {
+      return await this.prisma.translations.findUniqueOrThrow({
```
YazeedLoonat, line 20: how do you feel about switching this to `findUnique` and then we handle the error ourselves instead of wrapping a try catch?

backend_new/src/services/translation.service.ts
```diff
@@ -0,0 +1,167 @@
...
+          jurisdictionId_language: { language, jurisdictionId },
+        },
+      });
+    } catch (e) {
+      if (
+        e instanceof Prisma.PrismaClientKnownRequestError &&
+        e.code === 'P2025'
+      ) {
+        console.warn(
+          `Fetching translations for ${language} failed, defaulting to english.`,
```
YazeedLoonat, line 31: can we add the jurisdiction id as well to this log or do you think it would be overkill?

YazeedLoonat (changes requested):
Looks excellent a few minor changes!

thanks for tackling this one!

backend_new/test/unit/services/translation.service.spec.ts
ludtkemorgan: I am unable to switch to beforeAll because I change the mocks in each test and the `toHaveBeenCalledTimes` is inaccurate unless it is reset every test

backend_new/src/services/translation.service.ts
ludtkemorgan: `findUnique` looks a lot cleaner!

ludtkemorgan:
@YazeedLoonat this is ready to re-review

backend_new/src/services/translation.service.ts
```diff
@@ -0,0 +1,161 @@
...
+              `listingMultiselectQuestions[${index}].multiselectQuestions.options[${optionIndex}].text`
+            ] = multiselectOption.text;
+            pathsToFilter[
+              `listingMultiselectQuestions[${index}].multiselectQuestions.options[${optionIndex}].description`
+            ] = multiselectOption.description;
+            pathsToFilter[
+              `listingMultiselectQuestions[${index}].multiselectQuestions.options[${optionIndex}].description`
+            ] = multiselectOption.description;
+          },
+          // TODO: should we translate links?
```
emilyjablonski, line 107: Is this TODO still needed?

backend_new/test/integration/listing.e2e-spec.ts
```diff
@@ -92,7 +82,7 @@ describe('Listing Controller Tests', () => {
       filter: [
         {
           $comparison: Compare.IN,
-          name: 'name: 10,name: 50',
+          name: 'name: 11,name: 51',
```
emilyjablonski, line 85: Curious why this is 11 and 51?

backend_new/src/services/translation.service.ts
ludtkemorgan: It is a question for the team, it is being done in the existing translations service however I don't think it is needed. In HBA there are no multiselect questions with links or options with links. I would also assume that links shouldn't be translated in general

backend_new/test/integration/listing.e2e-spec.ts
ludtkemorgan: I took the changes in https://github.com/bloom-housing/bloom/pull/3478/files#diff-4846ef09a013a233b0132f82b567da6c19be50bb3001fb8d38b2474cb18605c4 to fix the broken test. I believe it is relatively random, but @YazeedLoonat might know better

YazeedLoonat: on line 53 of that file we created listings 10 and 50 and since we no longer clean up after each test, we needed to switch to 2 different numbers otherwise we would find 10 and 50 each time 

so we switched to 11 and 51

those numbers are all random but we did need to guarantee no collision

emilyjablonski: @YazeedLoonat was the uuid stuff we paired on wednesday a better solution to this?

backend_new/src/services/translation.service.ts
ludtkemorgan: @YazeedLoonat do you have an opinion on this?

backend_new/test/integration/listing.e2e-spec.ts
emilyjablonski: @YazeedLoonat Checking in here with you!

backend_new/src/services/translation.service.ts
YazeedLoonat: if we don't have any right now I think we're safe to remove these and if we end up having a link that needs to get translated we can do that then 

like you said I don't think we want to translate links at all

backend_new/test/integration/listing.e2e-spec.ts
YazeedLoonat: good question! We made those UUID changes for unit tests where we mock interactions with the db

in e2e tests we are creating entries in the database. If we were just using IDs then we could have created another set of 50 and 10, but 
```
$comparison: Compare.IN,
name: 'name: 11,name: 51',
```
would mean that we find 4 records (2 50s and 2 10s) 

I'm not sure if that explanation makes sense, hopefully that helps though!

YazeedLoonat (approved):
LGTM

Issue #3466, opened 2023-05-24
Prisma: Actual useful Seed Data


YazeedLoonat:
**What is this feature or what feature is this part of?**
Currently we don't have good seed data in the new prisma stuff
I think it would be funny if everyone made a listing and then I can just use that data

if people can just give me a json file or literally just this but filled out with useful data it would go a long way

sarahlazarich:
Product to make some test listings. 

We are getting close to our row limit for seed data, need to cut it down to only useful test listings.

seanmalbert:
Linking to convo from #3478: https://github.com/bloom-housing/bloom/pull/3478/files#r1230227715

sarahlazarich:
@ludtkemorgan @YazeedLoonat can you please add QA notes + tag the person responsible or close this?

ludtkemorgan:
This cannot be QA'd more than what it was during the PR process. It has been partially deployed to a testing environment in heroku and you can see the data populated here: https://bloom-prisma-96be10547db7.herokuapp.com/api. I will close this issue and add extra verification steps to later stories.

PR #3524, opened 2023-06-23
fix: add better seeded data

ludtkemorgan:
# Pull Request Template

## Issue Overview

This PR addresses #3466 

## Description

As part of the typeORM -> Prisma changeover we need to re-think how data is seeded for both manual and automated testing. This PR creates a new pattern compared to the current backend by doing the following:

- Setup the seed file to do different things based on environment:
  - Prod: Only set the minimum fields needed to have a running environment such as a jurisdiction
  - Stage: Set up data that would make a demo (staging) site usable by a user. It sets up all of the mandatory fields along with mock tables that are consistent every time it is run. Lots of this data is setup from objects with a name and a diverse set  of values.
  - Dev: Sets up values to be used in local testing with completely random values. Names are random and a mix of values are randomly chosen.
- For all environments a jurisdiction name can be passed in the command by adding `--jurisdictionName <name>` when running the command.
  - dev environment is setup with the jurisdiction name of "Bloomington" so that running it locally you can have the name set in the .env file
- For each table type there is a factory to generate rows into that table. Everything that can be randomized is along with the ability to override values.
- The factories are used for all the seed setups (with different override values sent in all cases) and within the tests. The randomized values are then used to validate the fields rather than relying on a set value

After `yarn install` in the `backend_new` directory you can run the following commands and look to see what is in the DB.
`yarn db:setup` - setups the database with completely random values
`yarn db:setup:staging` - setups the database with the pre-defined staging values

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/prisma/seed-helpers/listing-factory.ts
```diff
@@ -348,7 +492,8 @@ export const listingFactory = (
   ),
 });
 
-const unitFactory = (
+// Temporarily keeping this for reference
+const unitFactory2 = (
```
YazeedLoonat, line 496: do we still need this since we have a dedicated unit factory?

YazeedLoonat (changes requested):
Looks good, just one question

backend_new/prisma/seed-helpers/listing-factory.ts
ludtkemorgan: I was keeping this here so merge conflicts were easier every time I pulled in your changes. I can remove it now that it is ready!

ludtkemorgan: @YazeedLoonat This has been deleted

backend_new/prisma/seed-helpers/listing-factory.ts
```diff
@@ -1,415 +1,147 @@
...
-      county: `listingsBuildingAddress: ${i} county: ${i}`,
-      state: `listingsBuildingAddress: ${i} state: ${i}`,
-      street: `listingsBuildingAddress: ${i} street: ${i}`,
-      street2: `listingsBuildingAddress: ${i} street2: ${i}`,
-      zipCode: `listingsBuildingAddress: ${i} zipCode: ${i}`,
-      latitude: i * 100,
-      longitude: i * 101,
-    },
+  prismaClient: PrismaClient,
+  optionalParams?: {
```
ColinBuyck, line 15: Nice!

ColinBuyck (approved):
This LGTM 🌸 just two non-blocking thoughts. 1) In the listing data, could we vary the status? So there are active and closed listings as well. 2) Could we vary the address table data so that we can verify that the address, application pickup address, application drop off address, and application mailing address are all correctly mapping to unique values.

emilyjablonski (approved):
This LGTM just some general feedback / questions!
* The issue indicates`Product to make some test listings.` was that part of this process?
* Something that has always been really useful to me about the dev data is that it is consistent. If I need to test the state of preferences on a listing that doesn't have any, it's been really nice to have that dev data consistent and I know a listing already exists named something like "No Preferences Listing". I'm curious for your perspective on that, I'm a little worried listing that functionality is something that could slow me down in the future.
* It seems like this could lead to a number of missing states - I think it's just hard to nail them all down - maybe part of this is as new states come up not in the data that block dev we can have license to adjust the factories?

YazeedLoonat (approved):
LGTM

Issue #3504, opened 2023-06-12
Prisma: Create Get Endpoints for Applications, application flagged sets, and users


YazeedLoonat:
blocked by: https://github.com/bloom-housing/bloom/pull/3484

applications: https://github.com/bloom-housing/bloom/pull/3494 (on hold)
users: https://github.com/bloom-housing/bloom/pull/3546 (out for review)

PR #3546, opened 2023-07-12
feat: Prisma user get endpoints

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3504)

## Description
This creates all the endpoints for users

Follow the readme instructions for how to stand the new backend up
You should be able to test things out using the openai site at http://localhost:3101/api

you should be able to:
get a list of, and get an individual users

permission is not implemented yet so don't worry about testing that yet

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/prisma/seed-helpers/user-factory.ts
```diff
@@ -1,11 +1,15 @@
 import { Prisma } from '@prisma/client';
+import { randomAdjective, randomNoun } from './word-generator';
 
 export const userFactory = (
   roles?: Prisma.UserRolesUncheckedCreateWithoutUserAccountsInput,
+  firstName?: string,
+  lastName?: string,
+  email?: string,
```
ludtkemorgan, line 8: The format that I have been doing with these optional params in the factory has been making them an object.
```
optionalParams?: {
    roles?: Prisma.UserRolesUncheckedCreateWithoutUserAccountsInput;
    firstName?: string;
    lastName?: string;
    email?: string;
  }
```

That way when calling the factory it is easier to know what field is getting set `userFactory({firstName: "first", email: "example@example.com"})`

backend_new/prisma/seed-helpers/user-factory.ts
```diff
@@ -1,11 +1,15 @@
...
 ): Prisma.UserAccountsCreateInput => ({
-  email: 'admin@example.com',
-  firstName: 'First',
-  lastName: 'Last',
+  email: email || `${randomNoun()}@${randomAdjective()}.com`,
```
ludtkemorgan, line 10: If you are pulling out the default email we will need to update the seed-dev.ts and seed-staging.ts to make sure the admin@example.com account is created for testing.

ludtkemorgan (changes requested):

backend_new/prisma/seed-helpers/user-factory.ts
YazeedLoonat: curious if you think we should keep the admin@example.com account or if we should create a kinda spread of accounts instead? 

With MFA multiple users trying to use admin@example.com causes some problems so maybe having a few users that are different may be helpful?

backend_new/prisma/seed-helpers/user-factory.ts
YazeedLoonat: updated!

backend_new/prisma/seed-helpers/user-factory.ts
ludtkemorgan: That's a good point. I think `admin@example.com` should still be used for the dev seeding since theoretically that is just local or testing purposes so there should just be one person logging in. For staging I feel like we should have the admin@example.com as well but then after we seed a new site we manually go in to create accounts. I guess that is what we currently do, so I wonder if there is a better way

YazeedLoonat: I think maybe on staging after we use the admin user to create users for a few people we delete the admin user out so that its no longer accessible

the other option is we add  people straight into the seed script, but I'm hesitant putting our emails into a file thats openly accessible on github

I went ahead and have admin@example.com get created on both the dev and staging levels for now at least

emilyjablonski (approved):
🦴

backend_new/src/dtos/users/user.dto.ts
seanmalbert, line 1: @YazeedLoonat , some of the optional parameters specify the type or null and some do not, can you please make sure that these are consistent and add comments for any cases that don't follow that convention.

backend_new/src/services/user.service.ts
```diff
@@ -0,0 +1,234 @@
...
+    const rawUsers = await this.prisma.userAccounts.findMany({
+      skip: calculateSkip(params.limit, params.page),
+      take: calculateTake(params.limit),
+      orderBy: buildOrderBy(
+        ['firstName', 'lastName'],
+        [OrderByEnum.ASC, OrderByEnum.ASC],
+      ),
+      include: view,
+      where: whereClause,
+    });
```
seanmalbert, line 50: @YazeedLoonat , I was curious if we could reduce the findMany and count to a single transaction, and it looks like there is a way: https://stackoverflow.com/questions/74328100/prisma-find-many-and-count-in-one-request. Can you please test this out and make sure Prisma handles this well.

backend_new/src/services/user.service.ts
```diff
@@ -0,0 +1,234 @@
...
+  /*
+    this helps build the where clause for the list()
+  */
+  buildWhereClause(
+    params: UserQueryParams,
+    user: User,
+  ): Prisma.UserAccountsWhereInput {
+    const filters: Prisma.UserAccountsWhereInput[] = [];
+
+    if (params.search) {
```
seanmalbert, line 71: Same suggestion as past PRs with updating to return if no params, instead of wrapping this in a conditional.

backend_new/src/services/user.service.ts
```diff
@@ -0,0 +1,234 @@
...
+                      isAdmin: false,
+                    },
+                  },
+                ],
+              },
+            ],
+          });
+        }
+      }
+    });
```
seanmalbert, line 207: recommend seeing if you can clean this up to not require nested conditionals.

backend_new/test/unit/utilities/pagination-helpers.spec.ts
```diff
@@ -37,4 +38,35 @@ describe('Testing pagination helpers', () => {
...
+    });
+    it('should return 2 page for 1 full page', () => {
+      expect(buildPaginationMetaInfo({ limit: 10, page: 1 }, 20, 10)).toEqual({
+        currentPage: 1,
+        itemCount: 10,
+        itemsPerPage: 10,
+        totalItems: 20,
+        totalPages: 2,
+      });
+    });
```
seanmalbert, line 59: The descriptions for these tests are a little confusing, can you please update and also add more test coverage for different variations of params.

seanmalbert (changes requested):

backend_new/test/unit/utilities/pagination-helpers.spec.ts
YazeedLoonat: updated and added a few more tests

backend_new/src/services/user.service.ts
YazeedLoonat: I moved some stuff around, lemme know what you think

backend_new/src/services/user.service.ts
YazeedLoonat: Hey Sean I do that below this line with `if (!params.filter?.length)` we just return otherwise we go through the param filter's array and build each filter

for this case params has an optional search and an optional filter so I think we need this if statement to capture if params.search exists then we build the filter for it

even if we did that opposite I think we would basically have the same thing with 

if (!params.search) { // do nothing} < insert the building of the search filter here> 

hopefully that makes some sense?

backend_new/src/services/user.service.ts
YazeedLoonat: So after doing some digging I found some interesting comments:

https://github.com/prisma/prisma/issues/7550#issuecomment-1594572700
and  https://www.prisma.io/blog/how-prisma-supports-transactions-x45s1d5l0ww1

from my reading of it $transaction still is happening sequentially, I'll switch to a promise.all style solution since the results of 1 query doesn't affect the 2nd query at all

It seems like eventually prisma will have a built in solution for this but for the time being we can run with the promise.all approach for stuff like this or transactions in other cases!

backend_new/src/dtos/users/user.dto.ts
YazeedLoonat: I went ahead and removed the `| null` and will be consistent with that going forward

 nice catch ty! (and when I go back to do permissioning on each of these modules I'll clean up the existing ones)

backend_new/src/controllers/user.controller.ts
```diff
@@ -0,0 +1,65 @@
...
+
+@Controller('user')
+@ApiTags('user')
+@UsePipes(new ValidationPipe(defaultValidationPipeOptions))
+@ApiExtraModels(IdDTO)
+export class UserController {
+  constructor(private readonly userService: UserService) {}
+
+  @Get()
+  profile(@Request() req): User {
```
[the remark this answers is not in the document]
YazeedLoonat: nice catch, updated

backend_new/src/services/user.service.ts
seanmalbert: Hey @YazeedLoonat , I see that there are two ways to filter for if a user is a partner, by `UserFilterKeys.isPartner` and `UserFilterKeys.isPortalUser` and then adding to the filter `userRoles.isPartner`. Can we align on one of those approaches instead of using both?

backend_new/src/services/user.service.ts
```diff
@@ -0,0 +1,234 @@
...
+                  {
+                    userRoles: {
+                      isAdmin: null,
+                    },
+                  },
+                  {
+                    userRoles: {
+                      isAdmin: false,
+                    },
+                  },
```
seanmalbert: Are we missing adding filters for `isJurisdictionalAdmin` here?

backend_new/prisma/seed-helpers/user-factory.ts
seanmalbert: @YazeedLoonat and @ludtkemorgan , is this conversation resolved?

backend_new/prisma/seed-helpers/user-factory.ts
seanmalbert: @YazeedLoonat and @ludtkemorgan, is this conversation resolved?

seanmalbert (changes requested):
@YazeedLoonat , a few conversations left to resolve. Thank you for all the updates!

backend_new/prisma/seed-helpers/user-factory.ts
YazeedLoonat: yeah I think the resolution was re-adding the admin@example.com email and then the "how do we use this in staging" question is something we can address when we do the migration work and set up staging on prisma

resolve convo!

backend_new/src/services/user.service.ts
YazeedLoonat: wow really good catch, thats a bug from the current typeorm backend
so one less bug being brought over!

backend_new/src/services/user.service.ts
YazeedLoonat: after digging through some code I think both of those conditions are unneeded went ahead and cleaned that up!

YazeedLoonat:
Hey @seanmalbert nice call outs, got those updates in!

seanmalbert (approved):

PR #3552, opened 2023-07-14
fix: add correct command to run on heroku

ludtkemorgan:
These are the required changes in order to get the prisma backend to work in Heroku.

- Add Procfile so that the correct start command is done (by default it tries to do npm start)
- The start:prod command was point to the wrong directory.

Also of note, Heroku had to add this buildpack https://github.com/timanovsky/subdir-heroku-buildpack and add the PROJECT_PATH env variable

### DIFF
--- a/backend_new/Procfile
+++ b/backend_new/Procfile
@@ -1,1 +1,1 @@
-web: yarn start
+web: yarn start:prod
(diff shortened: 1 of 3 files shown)


YazeedLoonat (approved):
LGTM

Issue #3434, opened 2023-05-08
Uptake Toast Component in to Bloom-Core


sarahlazarich:
[ZeroHeight Docs](https://zeroheight.com/5e69dd4e1/p/99531d-toast/b/78684d)
[Seeds PR](https://github.com/bloom-housing/ui-seeds/pull/17)

AC:
Every instance of toasts that are consumed directly in Bloom core are updated to the new seeds Toast component.

PR #3553, opened 2023-07-18
feat: uptake toast component from ui-seeds

ColinBuyck:
# Pull Request Template

## Issue Overview

This PR addresses #3434

## Description

Pulls in the toast component from ui-seeds and replaces all places the ui-component version of toast was used with the new one.

Notes:
- In an attempt to standardize how the toast is used and displayed, a message context was added. This allows us to set a message anywhere on the site and it will appear on the toast in the same place of the page.
  - The places where `<Toast>` is placed in the app is reduced
- The users page had a somewhat big rewrite in order to prevent unnecessary re-renders. Previously after a new user was added and the modal closed, a call to the backend is triggered which triggered the whole page to re-render. This is is because the api call was getting done at too high of a level. The new format makes it so only the users table area of the page is re-rendered
  - This follows best practices but also prevents the toast from getting rendered and then immediately disappearing after new api call is done

There are many Toasts throughout the site
- Invite a new user
- Delete a User
- Update a listing
- Export CSV on listings and users pages
- ~Confirm account~

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


ludtkemorgan:
@ColinBuyck I fixed the user add functionality by pulling out the part of the index file that needs data. That way when the swr triggers a new api request it doesn't cause a full page refresh. No guarantees that I didn't break a different toast flow. Do you mind taking a look to see if this makes sense? I will do another pass at the confirm user flow as well

sites/partners/src/pages/index.tsx
```diff
@@ -149,6 +150,8 @@ export default function ListingsList() {
         <title>{t("nav.siteTitlePartners")}</title>
       </Head>
       <SiteAlert type="success" timeout={5000} dismissable sticky={true} />
+      <Toast variant="success"></Toast>
```
jaredcwhite, line 153: I think maybe this was left here by mistake?

ludtkemorgan:
@ColinBuyck I did some updates to the PR. I spoke to Jesse and he agreed that the confirm user use case shouldn't be a toast. Most likely it should be a success alert, but we should do it as a separate PR. Can you take a look at this to see if any of the other use cases is still broken?

sites/partners/src/components/users/FormUserConfirm.tsx
```diff
@@ -33,6 +32,7 @@ const FormUserConfirm = () => {
   const router = useRouter()
   const { mutate, isLoading: isConfirmLoading, isError, reset: resetMutation } = useMutate<Status>()
   const { userService, loadProfile, loading } = useContext(AuthContext)
+  const { setToast } = useContext(MessageContext)
```
ColinBuyck, line 35: Looks like this line and MessageContext on line 16 are unused in this file

ColinBuyck:
I'm noticing that after the first "invite sent" toast on the user's page, another toast won't show despite new input

https://github.com/bloom-housing/bloom/assets/53269332/6a04a825-1768-4045-bcf9-c1cdb4fd189a

ColinBuyck:
I think at some point we agreed that some minor placement changes were okay but want to get designs input on this toast location on the public side.
Existing:
Screenshot 2023-09-25 at 1 17 21 PM
New:
Screenshot 2023-09-25 at 1 16 49 PM

ludtkemorgan:
@ColinBuyck I think that welcome back message shouldn't be a toast. It's probably more appropriate to be a success alert. Especially since the user is getting routed to this page rather than just having a toast appear on the page you are on. Do you think that would make sense?

ColinBuyck:
Yeah! That sounds like a better plan to me @ludtkemorgan

shared-helpers/src/utilities/MessageContext.ts
```diff
@@ -0,0 +1,59 @@
+import React, { FunctionComponent, createElement, useRef, useState } from "react"
+
+export enum alertTypes {
```
emilyjablonski, line 3: This should be typed according to what's available in seeds, it looks like that actually isn't available as an exported type though - just some cleanup for later that's on my radar.

sites/partners/src/layouts/index.tsx
```diff
@@ -63,9 +65,13 @@ const Layout = (props) => {
           siteHeaderWidth={"wide"}
           homeURL={"/"}
         />
-
-        <main>{props.children}</main>
-
+        <main>
+          {/* KEY is set to a random number to force a re-render. This is not ideal and I do not recommend it */}
+          <Toast {...getToastProps()} testId="toast-alert" key={Math.random()}>
```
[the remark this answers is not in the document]
emilyjablonski: I imagine the toasts need a unique key, and that we can't use the text itself for consecutive toasts of the same type, IMO Math.random() or new Date() seem like good solutions.

sites/partners/src/pages/reset-password.tsx
```diff
@@ -35,9 +35,9 @@ const ResetPassword = () => {
 
     try {
       const user = await resetPassword(token.toString(), password, passwordConfirmation)
-      setSiteAlertMessage(t(`authentication.signIn.success`, { name: user.firstName }), "success")
       await router.push("/")
       window.scrollTo(0, 0)
+      setToast(t(`authentication.signIn.success`, { name: user.firstName }), "success")
```
ColinBuyck, line 40: This might be one of the cases we want to remove. When following through the forgot password flow, I saw a flashing toast after resetting my password

ColinBuyck:
This also still includes the updates to the log in experience which I don't think are intended.
Dev:
Screenshot 2023-10-03 at 5 06 34 PM
PR:
Screenshot 2023-10-03 at 4 53 51 PM

sites/public/src/components/account/ConfirmationModal.tsx
```diff
@@ -37,7 +38,7 @@ const ConfirmationModal = (props: ConfirmationModalProps) => {
     try {
       await resendConfirmation(email)
 
-      setSiteAlertMessage(t(`authentication.createAccount.emailSent`), "success")
+      setToast(t(`authentication.createAccount.emailSent`), { variant: "success" })
```
ColinBuyck, line 41: Hard to test this one since doesn't seem like its working (dev or in PR)
https://github.com/bloom-housing/bloom/assets/53269332/93a011fd-3776-45e9-b3b6-7027bf5fffa6

ColinBuyck (commented):
LGTM 🟢 just a few call outs

ludtkemorgan:
@ColinBuyck Can you expand on this? `Also, noticing that the invite sent now stretches beyond the width of the text with these changes but it previously fit the text as the other toasts on the page do.`. Looking at the storybook for toast it looks like all toasts are the same length: https://storybook-ui-seeds.netlify.app/index.html?path=/story/blocks-toast--standard-variants. That is what I'm seeing in this invite sent toast

sites/partners/src/pages/reset-password.tsx
ludtkemorgan: I agree that this should be removed. I'll make a note that this should be an alert instead

ColinBuyck:
Yep @ludtkemorgan! I was just noting the visual change from the screenshot below from dev but seems like an intended change so all good.
Screenshot 2023-10-05 at 3 18 58 PM

sites/partners/src/pages/users/UserDataTable.tsx
emilyjablonski, line 1: Files in the pages directory should be navigable pages on the site / URLs - we should move this to components/

sites/public/src/pages/index.tsx
```diff
@@ -56,7 +58,10 @@ export default function Home(props: IndexProps) {
       <MetaTags title={t("nav.siteTitle")} image={metaImage} description={metaDescription} />
       <div className="flex absolute w-full flex-col items-center">
         <SiteAlert type="alert" className={alertClasses} />
-        <SiteAlert type="success" className={alertClasses} timeout={30000} />
+        <SiteAlert type="success" />
+        <Toast {...getToastProps()} className={alertClasses} hideTimeout={30000}>
```
emilyjablonski, line 62: Sometimes this is 3,000 and sometimes 30,000, not sure which is preferred?

emilyjablonski:
Screenshot 2023-10-16 at 9 26 27 PM
The width and placement here feels unanticipated, is this expected behavior?

ludtkemorgan:
@emilyjablonski the placement is determined by the Toast component. It is in a fixed position so that it is always in the top right. We are able to override how far from the top in the overrides.scss but moving it further left dynamically is harder.

But can you expand on the concern you have with the "width"? That is the default width we get from the toast component https://storybook-ui-seeds.netlify.app/index.html?path=/story/blocks-toast--standard-variants

emilyjablonski:
@ludtkemorgan I think imo the ideal result would be that it aligns on the right with the navigation links and is only as long as the content, wdyt? I know those are seeds changes, but just seeing the default behavior in action here seems like it might be a little funky

sites/public/src/pages/index.tsx
emilyjablonski: Agreed, @slowbot @eajensenwa cuurious if there are UX considerations here?

emilyjablonski:
I'd recommend addressing the above issues in seeds before merging this in its current state

jaredcwhite:
This is looking pretty solid now, but still want to see if I can get the top of the toast stack to align better with the site header. So small TODO remaining…

jaredcwhite:
@ColinBuyck @ludtkemorgan This PR was superseded by the newer one that got merged in (#3962). We should probably close this?

Issue #3468, opened 2023-05-24
CORS_ORIGINS config variable is breaking deploy previews


emilyjablonski:
At the moment on our core public site deploy previews, you can't apply for applications. We think the issue is bc the CORS_ORIGINS is looking for certain URLS and the deploy preview URLs are different every time.

Not sure what the solution should be (can we regex the value, or remove the check from dev?)

YazeedLoonat:
I think there is a pastability to pull the CORS_ORIGIN check when the env is DEV and keep it around for when the ENV is prod 

I think the other option is to change 
`app.module.ts` line ~49 to work with a reg ex solution

emilyjablonski:
Not ready for QA yet post-merge

emilyjablonski:
QA Notes
Ensure that the first page of starting an application after clicking `Apply Online` fully loads in a deploy preview (ex: https://deploy-preview-3611--bloom-demo-public.netlify.app)

emilyjablonski:
@sarahlazarich we don't have mandatory accounts anywhere, that behavior is what happens when the preview isn't working - our apps having the listing image and apply buttons on that page - not sure why the preview isn't working anymore will take a look!

emilyjablonski:
sorry it was just the wrong link! this should work: https://deploy-preview-3564--bloom-exygy-dev.netlify.app/applications/start/choose-language?listingId=285b6ec8-16fe-4561-b44f-b2c438d3e76c

sarahlazarich:
awesome, that works!

PR #3564, opened 2023-07-25
fix: cors origin deploy preview issues

emilyjablonski:
# Pull Request Template

## Issue Overview

This PR addresses #3468

## Description

We currently have an issue where applications aren't working on deploy previews as a result of a change to how we are allow-listing URLs. The issue is that our deploy preview URLs change every time, so our static allowlist doesn't work. This is a simple change to just remove that check in non-prod environments.

Tricky to test locally, but there's a new environment variable in the backend that takes in an array of regex expressions that will also be checked against the origin. In netlify environments we can set expressions like `https://deploy-preview-[A-Za-z0-9]+--bloom-exygy-dev\.netlify\.app/` to allow for the deploy previews to fully work.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/backend/core/src/app.module.ts
+++ b/backend/core/src/app.module.ts
@@ -46,7 +46,7 @@ export function applicationSetup(app: INestApplication) {
       credentials: true,
       origin: false,
     }
-    if (allowList.indexOf(req.header("Origin")) !== -1) {
+    if (allowList.indexOf(req.header("Origin")) !== -1 || process.env.NODE_ENV !== "production") {
       options.origin = true
     }
     cb(null, options)
--- a/backend/core/src/app.module.ts
+++ b/backend/core/src/app.module.ts
@@ -46,7 +46,11 @@ export function applicationSetup(app: INestApplication) {
       credentials: true,
       origin: false,
     }
-    if (allowList.indexOf(req.header("Origin")) !== -1 || process.env.NODE_ENV !== "production") {
+
+    if (
+      allowList.indexOf(req.header("Origin")) !== -1 ||
+      (process.env.CORS_REGEX && new RegExp(process.env.CORS_REGEX).test(req.header("Origin")))
+    ) {
       options.origin = true
     }
     cb(null, options)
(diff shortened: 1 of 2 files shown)


backend/core/src/app.module.ts
```diff
@@ -46,7 +46,7 @@ export function applicationSetup(app: INestApplication) {
       credentials: true,
       origin: false,
     }
-    if (allowList.indexOf(req.header("Origin")) !== -1) {
+    if (allowList.indexOf(req.header("Origin")) !== -1 || process.env.NODE_ENV !== "production") {
```
emilyjablonski, line 49: There's space to do something more sophisticated here, I somewhat worry this will remove test coverage from our dev environments of this code path, but curious for perspectives on whether or not that is a blocking issue / vibes on if that feels like an issue at all

YazeedLoonat: I think I'm a little hesitant to blanket turn off cors checks on non prod environments

After looking at the preview links this pr generated:
https://deploy-preview-3564--bloom-demo-public.netlify.app/
https://deploy-preview-3564--bloom-exygy-dev.netlify.app/

I think I'm more in favor of changing this from a `.indexOf()` check to a regex that checks to see from where the request is originating from.

I think knowing that the hosted urls are:
https://partners.bloom.exygy.dev/ and https://bloom.exygy.dev/

I'm wondering if we could construct a regex that covers all 4 possibilities (we also need to cover the proxy and the backend url)

YazeedLoonat (changes requested):
hey @emilyjablonski I think this would for sure work, I think I've become more in favor of a regex style solution that would maintain cors checks across all hosted envs

YazeedLoonat (approved):
LGTM

backend/core/src/app.module.ts
```diff
@@ -41,12 +41,20 @@ import { CatchAllFilter } from "./shared/filters/catch-all-filter"
 export function applicationSetup(app: INestApplication) {
   const { httpAdapter } = app.get(HttpAdapterHost)
   const allowList = process.env.CORS_ORIGINS || []
+  const allowListRegex = process.env.CORS_REGEX !== "" ? JSON.parse(process.env.CORS_REGEX) : []
```
ludtkemorgan, line 44: The tests are failing because if `CORE_REGEX` doesn't exist in the env this check passes as true since `undefined !== ""`.

I think we could get away with just `const allowListRegex = process.env.CORS_REGEX ? JSON.parse(process.env.CORS_REGEX) : []`

ludtkemorgan (changes requested):

ludtkemorgan (approved):

Issue #3510, opened 2023-06-14
Add backend validation for URL fields


emilyjablonski:
Currently on the listing form, field validation is passed backend --> frontend. We currently don't have any validation on URL fields. We should add the IsUrl decorator to ensure URL fields are properly formatted including requiring the protocol.

sarahlazarich:
@emilyjablonski to add fields that need validation in

emilyjablonski:
* Open house event
* Building selection criteria
* Settings - preference url
* Settings - preference option url
* Online application url

KrissDrawing:
QA Notes: For 
- Open house event
- Building selection criteria
- Settings - preference url
- Settings - preference option url

it should work as before cause there is frontend validation, so without tricks we shouldn't be able to get backend error.

Only digital application will change it's behaviour, and should throw same errors as other fields (so 1 for missing https:// and other for invalid page).
So on partners page https://partners.bloom.exygy.dev pick a listing and edit it (or create new one) in application process select `yes` in `is there digital application` and then `no` in common application field. The field that appear will now have validation triggered on listing save. Also now it won't reset on save and will send correct application type.

sarahlazarich:
this looks good!

PR #3568, opened 2023-07-31
fix: add backend url validation for url fields

KrissDrawing:
# Pull Request Template

## Issue Overview

This PR addresses #3510

## Description

Adds backend validation for url fields. It should make a difference only for digital application url, as rest has it's own frontend validation, so technically we can't push invalid url field.
Also for it to work i had to fix an issue with application types section reseting on submit (when error). So on edit it took value from listing, even though form value was different.
Also fixed an issue where we could have no digital application selected, but field for url was visible.

On partners site go to edit listing. In application process select `yes` in `is there digital application`, in url field provide invalid url. After submit it should highlight error and display error message. 
I suspect that mocked listings might have mutually exclusive values, so like digitalApplication might have application of "internal" type, but field for `commonDigitalApplication: false` which should result in `externalLink` type (cause thats only case it's not working for me, if i change it and save then it works fine).
Also now after selecting `is there a digital application` -> yes, then `Are you using the common digital application?` -> no, then `is there a digital application` -> no, it should not display url field (i believe that change shouldn't broke apps without common digital application, but worth to double check)

Check other fields if they work fine (to check if backend works i commented out "type="url" prop from fields):
- Open house event
- Building selection criteria
- Settings - preference url
- Settings - preference option url

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


ColinBuyck:
Just a heads up that there seems to be some testing issues to resolve here 🌵

KrissDrawing:
i fixed the test issues, but i believe there is something wrong with whole application type section. Cause it may pass wrong application type, so it won't trigger validation, even though visually it has proper fields selected.

ColinBuyck (commented):
I don't think I'm fully following the PR test description but I got the behavior below for an invalid url and when editing a new listing (not one from seeded data), I get the same thing. Is this expected?

https://github.com/bloom-housing/bloom/assets/53269332/71d10f32-38e7-476e-be05-49cb03f1d7e7

KrissDrawing:
Right, good catch, i thought it's impossible to recreate 😅. So listings might have `commonDigitalApplicationChoice` as `false` (no) even when no digital application, and after switching to digital application default method is set to Internal (in this case we got `Internal` method but `commonDigitalApplicationChoice` was false, so we could type link). So to fix that i used application method type as indicator which default value chose.

emilyjablonski (approved):
Screenshot 2023-08-11 at 8 29 58 AM
Screenshot 2023-08-11 at 8 29 52 AM

This LGTM, my only comment is that we should standardize the error message across all the fields
We only have the "valid URL" version on the custom digital application field

ColinBuyck:
@emilyjablonski Curious of your thoughts on this since I also found this error experience on the preference option url flow where both errors are used depending on the issue with the URL. 

https://github.com/bloom-housing/bloom/assets/53269332/03f909a7-7fd3-41ab-a33c-1d52c0cddfae

KrissDrawing:
I can try some custom validators instead of `IsUrl`, that will return different backend errors for different case (without https, and for invalid url, but with https). So it will work similar to frontend validation. If that will take too much effort or sth i'll let you know

emilyjablonski (approved):
This looks great! Loving the consistency of the experience

ludtkemorgan:
@ColinBuyck can you take another look at this PR?

ColinBuyck (approved):
So good! Thanks for making these updates 🐧

PR #3569, opened 2023-07-31
feat: Prisma heartbeat

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3570)

## Description
This creates a heartbeat endpoint so we can have an easy way to verify the application is stood up

This also moves around the app module, controller, service and tests to where the should live

This also removes the old backend tests from running on the prisma branches (they are flaky and also add to time in CI)

The endpoint is available to test on http://localhost:3101/api 

e2e and unit tests were added as well

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/src/services/app.service.ts
```diff
@@ -0,0 +1,11 @@
+import { Injectable } from '@nestjs/common';
+import { SuccessDTO } from '../dtos/shared/success.dto';
+
+@Injectable()
+export class AppService {
+  heartbeat(): SuccessDTO {
+    return {
+      success: true,
```
ludtkemorgan, line 8: at some point it would be ideal to have the heartbeat endpoint also make sure that the db is accessible. I wonder if we should just do it now?

.circleci/config.yml
```diff
@@ -148,9 +146,6 @@ workflows:
       - jest-shared-helpers:
           requires:
             - setup
-      - jest-backend:
```
ludtkemorgan, line 151: 🎉

backend_new/src/services/app.service.ts
YazeedLoonat: Nice call out! I got this updated so we do the most simple call possible, lemme know what you think

ludtkemorgan (approved):
This is also a minor thing, but I think our usage of this endpoint is more closely related to a "health check" rather than a "heartbeat" as we need to call it instead of it sending a value. At least that is how I've used the two terms in the past

backend_new/src/services/app.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@nestjs/common';
+import { SuccessDTO } from '../dtos/shared/success.dto';
+import { PrismaService } from './prisma.service';
+
+@Injectable()
+export class AppService {
+  constructor(private prisma: PrismaService) {}
+
+  async heartbeat(): Promise<SuccessDTO> {
+    await this.prisma.$queryRaw`SELECT 1`;
```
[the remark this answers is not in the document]
YazeedLoonat: Right! also I made the switch from heartbeat -> healthCheck 
lemme know if that works for ya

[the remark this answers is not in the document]
ludtkemorgan: Looks great!

Issue #3505, opened 2023-06-12
Prisma: Create C.U.D. endpoints for remaining modules


YazeedLoonat:
We need to create the Create, Update, and Delete endpoints for the following:

Listings: merged
Applications: merged
Users: merged
Auth: merged
Application Submission: (no longer exists)
User-Profile: (no longer exists)
Assets (only the createPresignedUploadMetadata): merged

PR #3591, opened 2023-08-10
feat: prisma user write

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3505)

## Description
This creates the endpoints for user writes, updates, and deletes as well as user profile since effectively they are the same

Unit and Integration tests were added, but you can also use the openapi site at http://localhost:3101/api/ to test further

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/src/services/user.service.ts
```diff
@@ -24,12 +41,21 @@ const view: Prisma.UserAccountsInclude = {
...
+type findByOptions = {
+  userId?: string;
+  email?: string;
+};
+
 @Injectable()
 export class UserService {
-  constructor(private prisma: PrismaService) {}
+  constructor(private prisma: PrismaService) {
+    dayjs.extend(advancedFormat);
```
ludtkemorgan, line 52: Should we add this advancedFormat extension at the app level rather than at the services that need them?

backend_new/src/services/user.service.ts
```diff
@@ -24,12 +41,21 @@ const view: Prisma.UserAccountsInclude = {
...
+  }
 
   /*
     this will get a set of users given the params passed in
+    Only users with a user role of admin or jurisdictional admin can get the list of available users. 
+    This means we don't need to account for a user with only the partner role when it comes to accessing this function
```
ludtkemorgan, line 58: I could be missing it, but looks like this functionality wasn't added yet. Should we add a todo to make sure that is added?

backend_new/prisma/seed-helpers/user-factory.ts
```diff
@@ -7,7 +7,9 @@ export const userFactory = (optionalParams?: {
   lastName?: string;
   email?: string;
 }): Prisma.UserAccountsCreateInput => ({
-  email: optionalParams?.email || `${randomNoun()}@${randomAdjective()}.com`,
+  email:
+    optionalParams?.email?.toLocaleLowerCase() ||
+    `${randomNoun().toLowerCase()}@${randomAdjective().toLowerCase()}.com`,
   firstName: optionalParams?.firstName || 'First',
   lastName: optionalParams?.lastName || 'Last',
   // TODO: update with passwordService hashing when that is completed
```
ludtkemorgan, line 15: Now that the password-helper is created, can you update the passwordHash below with something like `passwordToHash("abcdef")`

backend_new/src/services/user.service.ts
```diff
@@ -215,21 +241,519 @@ export class UserService {
...
+        confirmationToken,
+      );
+      // TODO: email service
+    } else if (
+      forPartners &&
+      existingUser &&
+      'userRoles' in dto &&
+      existingUser?.userRoles?.isPartner &&
+      dto?.userRoles?.isPartner &&
+      this.jurisdictionMismatch(dto.jurisdictions, existingUser.jurisdictions)
```
ludtkemorgan, line 645: What is this use case?

backend_new/src/services/user.service.ts
YazeedLoonat: that is handled through auth, basically a partner role only user wouldn't be able to hit the user list end point

backend_new/src/services/user.service.ts
YazeedLoonat: As of right now only the user service uses this extension (in both the prisma and typeorm backend) so I think we are okay doing this at the app level and then if we find a use case later on in another service we can make this app wide

backend_new/src/utilities/password-helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+    scrypt(password, salt, SCRYPT_KEYLEN, (err, key) =>
+      err ? reject(err) : resolve(key.toString('hex')),
+    ),
+  );
+};
+
+export const passwordToHash = async (password: string): Promise<string> => {
+  const salt = generateSalt();
+  const hash = await hashPassword(password, salt);
+  // TODO: redo how we append the salt to the hash
```
emilyjablonski, line 31: Curious where this TODO will be addressed, is there a specific followup?

backend_new/test/integration/user.e2e-spec.ts
```diff
@@ -27,7 +38,8 @@ describe('User Controller Tests', () => {
     await app.close();
   });
 
-  it('should get no users from list() when no params and no data', async () => {
+  // without clearing the db between tests or test suites this is flakes because of other e2e tests
+  it.skip('should get no users from list() when no params and no data', async () => {
```
emilyjablonski, line 42: Same as above

backend_new/test/unit/services/user.service.spec.ts
```diff
@@ -170,41 +172,1017 @@ describe('Testing user service', () => {
...
+
+  it('should verify that there is a jurisdiciton mismatch', () => {
+    const res = service.jurisdictionMismatch(
+      [{ id: 'id a' }],
+      [{ id: 'id 1' }],
+    );
+    expect(res).toEqual(true);
+  });
+
+  it('should verify that there is no a jurisdiciton mismatch', () => {
```
emilyjablonski, line 236: `not`

backend_new/test/integration/user.e2e-spec.ts
YazeedLoonat: yeah basically same comment as above, we have some ideas but need to try em out before a discussion + implementation can happen

backend_new/test/unit/services/user.service.spec.ts
YazeedLoonat: updated!

backend_new/package.json
```diff
@@ -38,6 +38,8 @@
     "class-validator": "^0.14.0",
     "class-transformer": "^0.5.1",
     "cloudinary": "^1.37.3",
+    "dayjs": "^1.11.9",
+    "jwt-simple": "^0.5.6",
```
seanmalbert, line 42: Is this package still maintained? The last publish was four years ago. Are there alternatives that are still active? I know it doesn't mean it's better,  [jsonwebtoken](https://www.npmjs.com/package/jsonwebtoken) looks to be active and has way more weekly downloads.

backend_new/src/controllers/user-profile.controller.ts
seanmalbert, line 1: @YazeedLoonat , why do we need this controller?

backend_new/src/decorators/match-decorator.ts
seanmalbert, line 1: Please add comments to to the function and class explaining the purpose of these.

backend_new/src/dtos/users/email-and-app-url.dto.ts
seanmalbert, line 1: Please add a comment making it clear the purpose of this DTO. For example, should the email be the user's email or a no-reply email?

backend_new/src/modules/user-profile.module.ts
seanmalbert, line 1: As with the controller, do we need a separate user-profile module?

backend_new/test/integration/user.e2e-spec.ts
```diff
@@ -101,4 +117,416 @@ describe('User Controller Tests', () => {
...
+      where: {
+        id: userA.id,
+      },
+    });
+
+    expect(userPostResend.email).toBe(userA.email);
+    expect(userPostResend.confirmationToken).not.toBeNull();
+  });
+
+  it('should succeed when trying to resend confirmation but not update record when user is already confirmed', async () => {
```
seanmalbert, line 213: What are the use cases where we need to resend the confirmation if the user is already confirmed?

seanmalbert (changes requested):
@YazeedLoonat , got mainly questions for you and can you update all of the areas that have "TODO"s with a comment referencing the issue in which these will be done?

Can you also make sure your conversations with @ludtkemorgan and @emilyjablonski are resolved?

backend_new/package.json
YazeedLoonat: Yeah I was looking at it, it seemed stable, but I think you're right

switched to the new package!

backend_new/src/controllers/user-profile.controller.ts
YazeedLoonat: yeah that's a good question, I originally thought it was a difference around the auth guards, but looking at it again I was wrong. 

Removed!

backend_new/src/decorators/match-decorator.ts
YazeedLoonat: updated

backend_new/src/dtos/users/email-and-app-url.dto.ts
YazeedLoonat: updated!

backend_new/test/integration/user.e2e-spec.ts
YazeedLoonat: Hey sean we don't actually resend the the email or modify the user record at all. Instead we just return success as true and do nothing else. 

If I had to guess its probably to hide if a user is already confirmed or not

YazeedLoonat:
Hey @seanmalbert got all of your comments addressed! lemme know what you think

backend_new/src/services/user.service.ts
ludtkemorgan: Sorry, one additional question, do we also need this to cover the case of the existing user is a "jurisdictional Admin" and not just a partner?

YazeedLoonat: that's a good question, right now a jurisdictional admin can only be tied to one jurisdiction (or rather the front end only supports them being tied to one jurisdiction) 

So my guts tell me not to support that case right now, and address it when we look at giving jurisdictional admin access to several jurisdictions if that makes sense

backend_new/src/decorators/match-decorator.ts
seanmalbert: Thank you!

seanmalbert (changes requested):
@YazeedLoonat , updates look great, thank you! It looks like the tests are passing fine in the pipeline, but when I run `yarn test` and `yarn test:e2e` locally, I get failing tests in both cases. The former has tests failing for user service and the latter for the user controller.  It could be something to do with my setup, but want to make sure before approving. Let me know if it's easier to get in a huddle.

backend_new/src/controllers/user.controller.ts
```diff
@@ -63,4 +74,83 @@ export class UserController {
...
+  }
+
+  @Post('resend-confirmation')
+  @ApiOperation({
+    summary: 'Resend public confirmation',
+    operationId: 'resendConfirmation',
+  })
+  @ApiOkResponse({ type: SuccessDTO })
+  async confirmation(@Body() dto: EmailAndAppUrl): Promise<SuccessDTO> {
+    return await this.userService.resendConfirmation(dto, false);
```
ludtkemorgan, line 130: I think you got this and the `resend-partner-confirmation` calls backwards. The `forPublic` flag is set to false here but should be true and vis-versa for `resend-partner-confirmation`

ludtkemorgan (changes requested):

backend_new/src/controllers/user.controller.ts
YazeedLoonat: Nice catch, got this updated!

backend_new/prisma/seed-helpers/user-factory.ts
ludtkemorgan: Awesome! We can remove this TODO comment now

ludtkemorgan (approved):
🪨

PR #3597, opened 2023-08-15
refactor: use new seeds tag component

emilyjablonski:
# Pull Request Template

## Issue Overview

This PR addresses [#307](https://github.com/metrotranscom/doorway/issues/307)

## Description

Uptakes the ui-seeds Tag component in order to be able to match the pill designs for listings approval (the CSS overrides worked great!) - I updated to the new seeds Tag everywhere we were consuming UIC Tag in partners so that it stays visually consistent (locations below).
-- Listing form pages (Tag is the listing status)
-- Application table pages (Tag is also the listing status)
-- Individual application form (Tag is the submission status of the application)
-- Application flagged set resolution page (Tag is the resolution status of the set)
-- User edit form (Tag is the confirmed status of the user) - (previously this was a Tag size that was inaccessible / we no longer support it, so now it's a little bigger)

We had duplicated logic around toggling which pill to show for listings, so I consolidated that into a new helper file.

There was one instance of Tag within some old flagged set files I went to update to the seeds version to discover we weren't using the files at all, so I removed them.

Check out the new pill visually in the places it was updated.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


sites/partners/src/components/applications/PaperApplicationForm/PaperApplicationForm.tsx
```diff
@@ -163,18 +163,16 @@ const ApplicationForm = ({ listingId, editMode, application }: ApplicationFormPr
...
-          tagStyle={
-            application?.status == ApplicationStatus.submitted
-              ? AppearanceStyleType.success
-              : AppearanceStyleType.primary
-          }
-          tagLabel={
-            application?.status
+        <StatusBar>
+          <Tag
+            variant={application?.status == ApplicationStatus.submitted ? "success" : "primary"}
```
ColinBuyck, line 168: Would it make sense to create a new Variant enum for the Seeds components?

emilyjablonski: If I'm following (lmk if this isn't what you mean) I think that variant should be an enum in seeds (that we should change [this line over there](https://github.com/bloom-housing/ui-seeds/blob/b5bb785a807824640356b0329b3cff2be2351191/src/text/Tag.tsx#L9))

ColinBuyck: Oh I think that could be helpful! My main concern was just leaving us open to mistyping variants or if there are any variants that were carried over from UI-C but aren't in seeds. But with the linked code, I'm realizing now that it would still throw a type error! I don't have very strong feelings on the seeds enum and appreciate the context.

ColinBuyck:
Noticed just a tiny discrepancy between the application edit experience on dev vs local

https://github.com/bloom-housing/bloom/assets/53269332/3e7a3bcf-8221-4130-be94-0a14075d4765

sites/partners/styles/overrides.scss
```diff
@@ -2,4 +2,14 @@
...
+
+  .seeds-tag {
+    text-transform: uppercase;
+    letter-spacing: var(--seeds-letter-spacing-wider);
+    width: 100%;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    text-align: center;
+  }
```
ludtkemorgan, line 14: This makes sense to me that these overrides are needed in order to make the tag look correct. However, I'm a bit surprised how much of it is custom css and not token overriding. Especially since you are overriding it for all tags on all of partners

ludtkemorgan (commented):
Since you are updating all the instances in partners, does it make sense to also do it for public in this PR?

emilyjablonski:
@ludtkemorgan I don't think there are any instances of the Tag in public core

ludtkemorgan:
@emilyjablonski I assumed this was the tag component 
image
But I see now that it is within the ListingCard component

sites/partners/styles/overrides.scss
emilyjablonski: https://github.com/bloom-housing/ui-seeds/blob/main/src/text/Tag.scss The existing tokens are more for styling of the tag itself which is expected, these styles here are mostly for how it fills the container. The text transform and letter spacing are technically not styles that I think we would recommend anyone to change / we are eventually moving away from uppercasing text, but this gets us to parity.

ludtkemorgan: The letter spacing makes sense as well as the uppercase. I'm just wondering if we will run into issues down the line with these overrides being on the root level of partners. 
For example in the future we might want a tag that is not 100% width. I guess for now this is probably ok to keep here until we have that use case come up.

ColinBuyck:
LGTM :electron: Thanks for making those updates

ColinBuyck (approved):

ludtkemorgan (approved):

PR #3606, opened 2023-08-17
feat: custom exception filter

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3608)

## Description
This creates a simple starting custom exception handler. This will eventually act as a way to handle errors that we don't want making it to the front end

There really isn't a way to test this. You could create an endpoint locally that throws an error and make sure that error finds its way to the front end, but for now this acting just as a pass through

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/src/utilities/custom-exception-filter.ts
```diff
@@ -0,0 +1,19 @@
...
+    front end to be exposed to this will grow
+*/
+@Catch()
+export class CustomExceptionFilter extends BaseExceptionFilter {
+  catch(exception: any, host: ArgumentsHost) {
+    console.error({
+      message: exception?.response?.message,
+      stack: exception.stack,
+      exception,
+    });
```
ludtkemorgan, line 16: I wonder if we should use the nestjs logger instead of console.error

emilyjablonski:
@YazeedLoonat If there isn't an issue can you make one? Just want to make sure product has an eye on untracked work

backend_new/src/utilities/custom-exception-filter.ts
YazeedLoonat: I went ahead and switched to using the logger, I think there are some limitations (like the logger can't be injected into the exception filter, so I created an instance for the exception handler

lemme know what you think

backend_new/src/utilities/custom-exception-filter.ts
seanmalbert, line 1: @YazeedLoonat , is there a good way to test this?

YazeedLoonat: unfortunately at the moment there isn't really. Since its just a pass through right now if it works then we shouldn't see anything different, but as cases come up this will become more visible. 

Off the top of my head I can imagine a world where for things like Prisma errors or errors from other libraries we might just log the error and then send the front end a simple InternalServerError or something, but for now I think we're unlikely to really run into those while testing

ludtkemorgan (approved):

emilyjablonski (approved):

Issue #3503, opened 2023-06-12
Prisma: add email service


YazeedLoonat:
We need to convert the email service to use prisma

ludtkemorgan:
re-putting this on hold until the remaining user endpoints are completed

PR #3607, opened 2023-08-17
fix: add email-service

ludtkemorgan:
# Pull Request Template

## Issue Overview

This PR addresses #3503

## Description

This add all of the endpoints in the EmailService and connects them up to appropriate use cases throughout the app.

Changes from the current backend's Email service:
- Jurisdiction is no longer acquired from the request header and rather from the context within the function calling the email service such as the user account.
- Figuring out which translation to get has been moved to the translation service
- Email Service is no longer injected at Request Scope. This could potentially cause issues if two people are requesting emails at the same time. It will require additional testing in a better environment to know if that is the case.

Note:
- Send MFA code and application confirmation emails functions are setup, but not connected to any controller since they are dependent on other services that are not completed

Make sure the following env variables are added to the .env file:
```
PARTNERS_PORTAL_URL
EMAIL_API_KEY
```
Also make sure you do a reseed to get the translations in the database. 

The following should be tested:
- Create Public User using the `/user` post endpoint. You should receive the "Welcome" email
- Create Partner user by using the `/user/invite` endpoint. You should receive the "Invite" email
- Resend public confirmation email using the `/user/resend-confirmation` endpoint.
  - You can use the user created in the first use case as long as you haven't confirmed the user
- Resend the partner user confirmation email using the `/user/resend-partner-confirmation` endpoint
   - You can use the user created in the second use case as long as you haven't confirmed the user
- Get a forgot password email when hitting the `/user/forgot-password` endpoint
- Change a user email by updating the user via the `/user/{id}` put endpoint. Make sure you put a new email in the request

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/prisma/seed-helpers/translation-factory.ts
```diff
@@ -0,0 +1,130 @@
...
+const translations = (jurisdictionName?: string) => ({
+  t: { hello: 'Hello', seeListing: 'See Listing' },
+  footer: {
+    line1: `${jurisdictionName || 'Bloom'}`,
+    line2: '',
+    thankYou: 'Thank you',
+    footer: `${jurisdictionName || 'Bloom Housing'}`,
+  },
+  header: {
+    logoUrl:
```
emilyjablonski, line 12: I'm surprised to see a URL in translations, do you know if we ever attempt to translate this?

backend_new/src/services/email.service.ts
```diff
@@ -0,0 +1,345 @@
...
+      jurisdiction?.id,
+      language,
+    );
+    this.polyglot.replace(translations);
+  }
+
+  private async getJurisdiction(
+    jurisdictionIds: IdDTO[] | null,
+  ): Promise<Jurisdiction | null> {
+    // TODO: figure out multiple jurisdictions
```
emilyjablonski, line 130: What's the next step for this TODO / issue?

backend_new/prisma/seed-helpers/translation-factory.ts
ludtkemorgan: This is here so that different jurisdictions can use different logos in the email. The translation can either be connected to a jurisdiction or have null jurisdiction. And the ones tied to the jurisdiction is the override to the generic one. There is no other way to change the content dynamically in the emails so that is why it is here. Nothing in here is translated, but rather a new entry is put into the database when a new language/jurisdiction combo is added. Not the greatest system at the moment

backend_new/src/services/user.service.ts
```diff
@@ -4,6 +4,7 @@ import {
   NotFoundException,
   UnauthorizedException,
 } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
```
YazeedLoonat, line 7: I'm not sure I like using the config service here. In the current backend we use the config service to sorta hardcode some env variables in. 

I think my preference is to just use process.env and put defaults like that into the template.

Hopefully that makes some sense, but it saves us a package I don't think we really need, and may may it easier to understand what exactly needs to be in the .env file for the app to work

backend_new/src/modules/email.module.ts
```diff
@@ -0,0 +1,29 @@
+import { SendGridModule } from '@anchan828/nest-sendgrid';
```
YazeedLoonat, line 1: I think my preference is to use: https://www.npmjs.com/package/@sendgrid/mail and build our own service and module kinda like what I did https://github.com/bloom-housing/bloom/pull/3617 for twilio (sms service)

I feel like its safer with the official sendgrid package, since this package doesn't seem to be officially part of nestjs

lemme know your thoughts though!

YazeedLoonat (changes requested):
Looks good, just a few personal preferences

backend_new/src/modules/email.module.ts
ludtkemorgan: I'm cool with making our own. I can basically steal the code from the package we are currently using. It is of note that the package we are using is updated frequently and using the @sendgrid/mail package under the covers.

backend_new/src/services/user.service.ts
ludtkemorgan: I think it could be useful to setup the ConfigModule so that there is a validationSchema against it. That way we don't have to rely on the template file and there are failures on startup if we don't have them all in the file. Similar to how we are using it shared.module.ts in the old backend. Thoughts?

backend_new/src/services/email.service.ts
ludtkemorgan: I decided to remove the comment since the way it works with my latest change is that if more than one jurisdiction exists we only send the generic translations (null jurisdiction)

backend_new/src/modules/email.module.ts
ludtkemorgan: I've pulled out the @anchan828 package. It simplified some things, so I like it!

YazeedLoonat (changes requested):
Hey @ludtkemorgan this looks good! thanks for making those updates!

This is mostly for posterity, but we had a discussion offline about the config service and we decided that we are going to move to the config service from using process.env 

We'll spin up an issue to address it through the new backend since up to this point it was mostly written for process.env



One last minor change then I think this is good to go:
in the service files can you add a lil comment blurb at the top describing what the function is for? I know a bunch of em are obvious but I think it'll help down the line when we come back to this file like 9months from now

YazeedLoonat (approved):
LGTM

backend_new/src/views/partials/simple-footer.hbs
```diff
@@ -0,0 +1,4 @@
+<p>
+  {{t "footer.thankYou"}}<br/>
+  {{t "footer.footer"}}
+</p>
```
seanmalbert, line 4:
```suggestion
</p>

```

backend_new/src/views/partials/leasing-agent.hbs
```diff
@@ -0,0 +1,12 @@
...
+</p>
+<p>
+  {{listing.leasingAgentName}}<br />
+  {{listing.leasingAgentTitle}}<br />
+  {{listing.leasingAgentPhone}}<br />
+  {{listing.leasingAgentEmail}}<br />
+</p>
+
+<h4>{{t "leasingAgent.officeHours"}}</h4>
+<p>{{listing.leasingAgentOfficeHours}}</p>
```
seanmalbert, line 12:
```suggestion
<p>{{listing.leasingAgentOfficeHours}}</p>

```

seanmalbert (changes requested):

backend_new/src/services/user.service.ts
ludtkemorgan: An issue has been created to go back and update all places to use a config service

backend_new/src/services/translation.service.ts
```diff
@@ -12,27 +13,68 @@ export class TranslationService {
...
+          LanguagesEnum.en,
+          jurisdictionId,
+        );
+    }
+
+    const genericDefaultTranslations =
+      await this.getTranslationByLanguageAndJurisdictionOrDefaultEn(
+        LanguagesEnum.en,
+        null,
+      );
```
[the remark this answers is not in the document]
ludtkemorgan: This has been updated to be called all at once

backend_new/src/services/email.service.ts
```diff
@@ -0,0 +1,342 @@
...
+          );
+          if (retry > 0) {
+            void this.send(to, from, subject, body, retry - 1);
+          }
+        }
+      },
+    );
+  }
+
+  private async loadTranslations(
```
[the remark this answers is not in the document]
ludtkemorgan: I think that is valid. We don't have a pattern setup yet in this new backend for memoization, so I'll hold off on doing that as part of this issue

backend_new/src/services/translation.service.ts
```diff
@@ -12,27 +13,68 @@ export class TranslationService {
...
         `Fetching translations for ${language} failed on jurisdiction ${jurisdictionId}, defaulting to english.`,
       );
-      translations = await this.prisma.translations.findUnique({
-        where: {
-          jurisdictionId_language: {
-            language: LanguagesEnum.en,
-            jurisdictionId,
-          },
-        },
+      translations = await this.prisma.translations.findFirst({
```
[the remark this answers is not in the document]
ludtkemorgan: I added the check to make sure the language isn't english

ludtkemorgan:
@seanmalbert this is ready to be re-reviewed

backend_new/src/services/email.service.ts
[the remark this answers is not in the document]
seanmalbert: Please create a new issue and reference with a TODO here.

seanmalbert (approved):

PR #3655, opened 2023-09-25
fix: remove electron

ludtkemorgan:
This dependency is frequently flagged by security checks and in the process of upgrading it, I noticed it doesn't appear to be used anywhere. I would also assume if it was used it would be in the public package file as well.


emilyjablonski (approved):

ColinBuyck (approved):
LGTM 👠

Issue #3540, opened 2023-07-10
Prisma AFS C.U.D and Process


YazeedLoonat:
This is a ticket to convert the AFS create, update, delete, and process endpoints over to prisma.

There is a known issue where if a flagged set is resolved it updates the last updated dates on the applications, which then when a new application, or an application is edited these "resolved" flagged sets go back to unresolved

as we go through and do the conversion we should look to fix that

PR #3674, opened 2023-10-10
feat: prisma application flagged sets

YazeedLoonat:
# Pull Request Template

## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3540)

## Description
This creates the services and end points for application flagged sets.

**Part of the ask on this one is to rework our current process flow**
We have been having issues where the current (typeorm version) of the process endpoint/cronjob is causing issues for existing flagged sets. This PR reworks that flow entirely to avoid that problem. That's probably the part of this pr I think needs the heaviest review to make sure no edge cases snuck through

Unit and Integration tests were added, but you can also use the openapi site at http://localhost:3101/api/ to test further

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend_new/test/unit/services/application-flagged-set.service.spec.ts
```diff
@@ -0,0 +1,2431 @@
...
+    });
+  });
+
+  it('should grab meta data', async () => {
+    const mockCount = jest.fn();
+    mockCount.mockResolvedValueOnce(1);
+    mockCount.mockResolvedValueOnce(2);
+    mockCount.mockResolvedValueOnce(3);
+    mockCount.mockResolvedValueOnce(4);
+    mockCount.mockResolvedValueOnce(5);
```
ludtkemorgan, line 529: I don't know if it is any cleaner, but jest.fn() is a chainable call so this could be 
```
    const mockCount = jest
      .fn()
      .mockResolvedValueOnce(1)
      .mockResolvedValueOnce(2)
      .mockResolvedValueOnce(3)
      .mockResolvedValueOnce(4)
      .mockResolvedValueOnce(5);

File: backend_new/src/services/application-flagged-set.service.ts [hunk 2]
```diff
@@ -0,0 +1,951 @@
...
+      totalPendingCount,
+      totalNamePendingCount,
+      totalEmailPendingCount,
+    };
+  }
+
+  /**
+    helper that builds the meta functions queries
+  */
+  metaHelper(
```
ludtkemorgan, line 248: I wonder if renaming this to something more applicable about getting counts would make sense. I'm not great with names, but maybe something like "afsCountQuery"

ludtkemorgan (commented):
No big issues have come up yet. I still need to do some functional testing though

YazeedLoonat:
From convo with Morgan, add back in the process on listing update (but do it only for that specific listing instead of for every listing)

ludtkemorgan:
@YazeedLoonat another thing I thought of was it would be nice to have some of the seeded data be applications that get flagged. That doesn't need to be part of this PR though

backend_new/src/dtos/application-flagged-sets/afs-query-params.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+  listingId: string;
+
+  @Expose()
+  @ApiPropertyOptional({
+    enum: View,
+    enumName: 'AfsView',
+    example: 'active',
+  })
+  @IsEnum(View, { groups: [ValidationsGroupsEnum.default] })
+  view?: View;
```
ColinBuyck, line 21: Surprised to see the afs get endpoint fail (Bad Request) without a view selected even though its marked as optional here
Screenshot 2023-10-13 at 11 46 35 AM

backend_new/.env.template
```diff
@@ -34,4 +34,6 @@ TWILIO_AUTH_TOKEN=
 PARTNERS_PORTAL_URL=http://localhost:3001/
 # sendgrid email api key
 EMAIL_API_KEY=SG.ExampleApiKey
+# controls the repititon of the afs cron job
```
emilyjablonski, line 37: `repetition`

backend_new/src/dtos/application-flagged-sets/afs-meta.dto.ts
```diff
@@ -0,0 +1,35 @@
...
+  @IsNumber()
+  @IsOptional()
+  @ApiPropertyOptional()
+  totalResolvedCount?: number;
+
+  @Expose()
+  @IsNumber()
+  @IsOptional()
+  @ApiPropertyOptional()
+  totalPendingCount?: number;
```
emilyjablonski, line 22: can / should this be a computed value based on the two below? somehow seems redundant

backend_new/src/services/application-flagged-set.service.ts
```diff
@@ -0,0 +1,951 @@
...
+import { OrderByEnum } from '../enums/shared/order-by-enum';
+import { View } from '../enums/application-flagged-sets/view';
+import { buildPaginationMetaInfo } from '../utilities/pagination-helpers';
+import { AfsResolve } from '../dtos/application-flagged-sets/afs-resolve.dto';
+import { User } from '../dtos/users/user.dto';
+import { Application } from '../dtos/applications/application.dto';
+import { IdDTO } from 'src/dtos/shared/id.dto';
+
+/*
+  this is the service for unit types
```
emilyjablonski, line 36: is this comment accurate?

backend_new/src/services/application-flagged-set.service.ts
```diff
@@ -0,0 +1,951 @@
...
+      application.applicant.birthDay,
+      ...(application.householdMember
+        ? application.householdMember.map((member) => member.birthDay)
+        : []),
+    ];
+    const birthYears = [
+      application.applicant.birthYear,
+      ...(application.householdMember
+        ? application.householdMember.map((member) => member.birthYear)
+        : []),
```
seanmalbert, line 727: suggestion (non-blocking): you could create a function to handle each of these that accepted a "key" parameter like "lastName" and returned the array.

backend_new/src/services/application-flagged-set.service.ts
```diff
@@ -0,0 +1,951 @@
...
+        },
+      });
+    }
+  }
+
+  /**
+    tests application to see if its a duplicate
+    if it is then its either added to a flagged set or a new flagged set is created
+  */
+  async testApplication(
```
seanmalbert, line 547: question: I think this all checks out, but could you spend some time thinking about how this function could be made more readable?

seanmalbert (changes requested):

backend_new/.env.template
```diff
@@ -34,4 +34,6 @@ TWILIO_AUTH_TOKEN=
...
+AFS_PROCESSING_CRON_STRING=0,3,15 * * * *
```
seanmalbert, line 38: question: for clarification this means it will run at minute 0, 3 and 15, correct?

suggestion: if my understanding is correct, recommend having cron run at more even intervals like "0, 15, 30, 45" or something like "15, 45", or just once an hour. Not sure what the optimal frequency will be that won't cause performance issues under heavy load.

backend_new/package.json
```diff
@@ -38,12 +38,14 @@
...
     "@nestjs/platform-express": "^8.0.0",
     "@nestjs/swagger": "~7.1.12",
+    "@nestjs/schedule": "^3.0.4",
     "@prisma/client": "^5.0.0",
     "@sendgrid/mail": "7.7.0",
     "class-transformer": "~0.5.1",
     "class-validator": "~0.14.0",
     "cloudinary": "^1.37.3",
     "cookie-parser": "~1.4.6",
+    "cron": "^2.4.3",
```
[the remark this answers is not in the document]
YazeedLoonat: removed!

backend_new/prisma/seed-helpers/application-factory.ts
```diff
@@ -16,6 +16,7 @@ export const applicationFactory = (optionalParams?: {
   applicant?: Prisma.ApplicantCreateWithoutApplicationsInput;
   overrides?: Prisma.ApplicationsCreateInput;
   listingId?: string;
+  householdMemeber?: Prisma.HouseholdMemberCreateWithoutApplicationsInput;
```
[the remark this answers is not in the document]
YazeedLoonat: hehe fixed

backend_new/test/integration/ami-chart.e2e-spec.ts
```diff
@@ -25,6 +26,11 @@ describe('AmiChart Controller Tests', () => {
     prisma = moduleFixture.get<PrismaService>(PrismaService);
     await app.init();
 
+    const schedulerRegistry =
+      moduleFixture.get<SchedulerRegistry>(SchedulerRegistry);
+    // we stop the cron job since we don't want the cron job to run during tests
+    schedulerRegistry.getCronJob('AFS_CRON_JOB').stop();
```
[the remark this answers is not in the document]
YazeedLoonat: ooo that's such a good idea! fixed

backend_new/test/integration/application-flagged-set.e2e-spec.ts
```diff
@@ -0,0 +1,1521 @@
...
+        applications: true,
+      },
+    });
+
+    expect(afs.length).toEqual(1);
+    expect(afs[0].rule).toEqual(RuleEnum.email);
+
+    await prisma.applications.update({
+      data: {
+        acceptedTerms: true,
```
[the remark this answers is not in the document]
YazeedLoonat: yeah this is just a field to update so the update date gets auto updated

backend_new/test/integration/application-flagged-set.e2e-spec.ts
```diff
@@ -0,0 +1,1521 @@
...
+        allApplicationCount++;
+      } else if (
+        flaggedSet.ruleKey.indexOf(`${listing}-email3@email.com`) >= 0
+      ) {
+        const applications = flaggedSet.applications.map((app) => app.id);
+        expect(applications).toContain(appC.id);
+        allApplicationCount++;
+        expect(applications).toContain(appD.id);
+        allApplicationCount++;
+      }
```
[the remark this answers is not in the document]
YazeedLoonat: updated!

backend_new/test/integration/application-flagged-set.e2e-spec.ts
```diff
@@ -0,0 +1,1521 @@
...
+      },
+      include: {
+        applications: true,
+      },
+    });
+
+    expect(afs.length).toEqual(1);
+    expect(afs[0].rule).toEqual(RuleEnum.email);
+  });
+});
```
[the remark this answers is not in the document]
YazeedLoonat: added!

backend_new/test/unit/services/application-flagged-set.service.spec.ts
YazeedLoonat: updated!

backend_new/src/services/application-flagged-set.service.ts
YazeedLoonat: I rolled with `metaDataQueryBuilder` hopefully thats more descriptive

backend_new/src/dtos/application-flagged-sets/afs-query-params.dto.ts
YazeedLoonat: So view in this case isn't always passed from the front end, so I think making it required would be kinda bad 

I'm not sure what change I made that seems to have fixed this, but it seems like its fixed from what I can tell

backend_new/.env.template
YazeedLoonat: ty!

backend_new/src/dtos/application-flagged-sets/afs-meta.dto.ts
YazeedLoonat: I think we could for sure calculate it that way, but if we need this data in more than just the partner site (like for a future reporting tool or something) its probably for the best we just calculate that on the backend and let the front end know what the value is 

hopefully that makes some sense

backend_new/src/services/application-flagged-set.service.ts
YazeedLoonat: my secrets have been revealed copy and pasting from other services lol

fixed!

backend_new/src/services/application-flagged-set.service.ts
```diff
@@ -0,0 +1,951 @@
...
+          where: {
+            name: CRON_JOB_NAME,
+          },
+        });
+        // To prevent multiple overlapped jobs only run if one hasn't started in the last 5 minutes
+        if (
+          !currentCronJob ||
+          currentCronJob.lastRunDate <
+            dayjs(new Date()).subtract(5, 'minutes').toDate()
+        ) {
```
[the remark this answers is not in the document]
YazeedLoonat: from convo name has been made unique in the db so collisions should be less likely/impossible

backend_new/src/services/application-flagged-set.service.ts
```diff
@@ -0,0 +1,951 @@
...
+        filters.push({
+          status: FlaggedSetStatusEnum.pending,
+          rule: RuleEnum.email,
+        });
+      } else if (params.view === View.resolved) {
+        filters.push({
+          status: FlaggedSetStatusEnum.resolved,
+        });
+      }
+    }
```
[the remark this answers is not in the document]
YazeedLoonat: updated!

backend_new/src/services/application-flagged-set.service.ts
```diff
@@ -0,0 +1,951 @@
...
+        params.listingId,
+        FlaggedSetStatusEnum.pending,
+        RuleEnum.nameAndDOB,
+      ),
+      this.metaHelper(
+        params.listingId,
+        FlaggedSetStatusEnum.pending,
+        RuleEnum.email,
+      ),
+    ]);
```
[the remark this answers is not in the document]
YazeedLoonat: 2 less queries! updated 

I think the backend should still be responsible for the calculation so I kept it as part of the dto, we just no do the calculation on the fly instead of reading from the db again!

backend_new/src/services/application-flagged-set.service.ts
```diff
@@ -0,0 +1,951 @@
...
+
+    if (afs.listings.status !== ListingsStatusEnum.closed) {
+      throw new BadRequestException(
+        `Listing ${afs.listings.id} must be closed before resolving any duplicates`,
+      );
+    }
+
+    const selectedApps = dto.applications?.length
+      ? afs.applications.map((app) => app.id)
+      : [];
```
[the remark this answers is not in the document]
YazeedLoonat: fixed, good call out

backend_new/src/services/application-flagged-set.service.ts
```diff
@@ -0,0 +1,951 @@
...
+    const afs = await this.prisma.applicationFlaggedSet.findFirst({
+      where: {
+        AND: filter,
+      },
+      include: {
+        listings: true,
+        applications: {
+          where: {
+            id: {
+              in: dto.applications.map((application) => application.id),
```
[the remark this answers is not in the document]
YazeedLoonat: updated!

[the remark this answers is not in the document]
YazeedLoonat: yeah so its kinda an interesting thing that prisma does
the one in the where clause would only get flagged sets where the application is attached (applications.some.id.i)

the other one (include.applications.where.id.in) is so that when we get the list of application flagged sets back the applications that are returned as part of that call are only those that meet the criteria

without that second one we would just be getting the entire set of applications on the flagged set. Hopefully that makes some sense 

https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#filter-a-list-of-relations

is the documentation for it

backend_new/src/services/application-flagged-set.service.ts
YazeedLoonat: nice call out! updated

backend_new/src/services/application-flagged-set.service.ts
YazeedLoonat: Hey sean, I'm not really sure how to condense or pull this function apart further. I thought about making some more helpers but I feel like it might make this file/function a bit "spaghetti-y" with having to jump around to figure out whats its trying to do 

I made some small tweaks, but lemme know if you have anything in mind to really clean this up

backend_new/.env.template
YazeedLoonat: That's a really good call out, on production we use `0 0 * * *` as the value 
which we mean every day at midnight server time this would execute 
we add the seconds as a random number 

Right now we have this run so frequently locally so that we can catch any issues with the process job, I can change the suggestion in the .env.template though if that would make things more clear

YazeedLoonat:
Hey @seanmalbert for: https://github.com/bloom-housing/bloom/pull/3674#discussion_r1370854777

I think we need to keep the view enum as optional. Since the front end makes requests to endpoints without a view in a couple places. I did fix the example though

YazeedLoonat:
Hey @ludtkemorgan @seanmalbert @emilyjablonski @ColinBuyck 
this should be ready for 👀 again!

ludtkemorgan (approved):

backend_new/.env.template
seanmalbert: ah, thanks for the clarification. I think it's ok to keep what's in the template, but maybe add to the comment what is recommended for production environments.

backend_new/src/services/application-flagged-set.service.ts
seanmalbert: Thanks, @YazeedLoonat ,I'm not sure I do without making it harder to follow in other ways like you're saying (at least not without spending more time).

backend_new/src/services/application-flagged-set.service.ts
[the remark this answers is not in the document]
seanmalbert: Thanks, @YazeedLoonat . Agreed on keeping it part of the DTO.

seanmalbert (approved):

PR #3686, opened 2023-10-24
fix: increase the number of random words

ludtkemorgan:
We have run into a few times where the random generation of words caused duplicates. This PR increases the number of words for both noun and adjectives. I used chatGPT to come up with words you'd see as an apartment name so some are better than others.


backend_new/prisma/seed-helpers/word-generator.ts
```diff
@@ -2,210 +2,441 @@ import { randomInt } from 'crypto';
...
-  'Sea',
-  'Tent',
-  'Minute',
-  'Territory',
-  'Sock',
-  'Bag',
-  'Revolution',
-  'Law',
-  'Smell',
+  'Sapphire',
```
YazeedLoonat, line 35: this is the best one

YazeedLoonat (approved):
LGTM

Issue #3687, opened 2023-10-24
Geocoding - Showing Name and Relationship on Preference Screens


kathyccheng:
As an applicant 
I want to be able to input name of the address holder and relationship to the address holder if the preference needs it 
So that I can provide the information necessary for an application 

**Acceptance Criteria** 
* When I am on the preference screen for a preference sub-option that needs Name of Address Holder and/or Relationship to Address Holder, then I will see those questions when I click on the preference sub-option 
* When I answer one or both questions, then I can click next and it will get saved to my application 

**Notes**
* Should be required if set by the preference
* Place these fields above the address
* Add a label to the address "Qualifying Address"
* The work to add those questions to the preference sub-option was handled in https://github.com/bloom-housing/bloom/issues/3673

**Designs** 
https://www.figma.com/file/Ne8jOyPpCBvlAhyaRlXgTq/Geocoding?type=design&node-id=1209-20272&mode=design&t=0X1n9eNM7LIKhP6B-0

KrissDrawing:
@kathyccheng  Should those fields be added to summary? <img width="549" alt="image" src="https://github.com/bloom-housing/bloom/assets/9580543/4fef7938-2281-47d2-9412-623be6eb4b05">

kathyccheng:
@KrissDrawing yes we should add those to the summary as well! Thanks for the catch!

KrissDrawing:
@slowbot Could you take a look if this approach to add additional fields to summary looks good?

<img width="530" alt="image" src="https://github.com/bloom-housing/bloom/assets/9580543/169f9fcf-2846-4bde-bd31-d1b1936a6587">

slowbot:
@KrissDrawing generally this looks good. I would expect additional spacing between each preference to make it easier to read. It looks like 16px/1rem now so maybe 24px/1.5rem

KrissDrawing:
to QA after whole geocoding flow is ready.

PR #3703, opened 2023-11-06
feat: geocoding - showing name and relationship on preference screens

KrissDrawing:
# Pull Request Template

## Issue Overview

This PR addresses #3687 

## Description

Adds fields for preferences address holder name and relationship to application.

Set preference option to have address verification `Yes` (partners -> `/settings`), and then test if address holder name / relationship, will appear respectively in application process (verification method doesn't matter for now). 

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend/core/src/applications/types/application-multiselect-question-option.ts
```diff
@@ -19,6 +19,18 @@ export class ApplicationMultiselectQuestionOption {
...
+  @Expose()
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty({ required: false })
+  addressHolderName?: string
+
+  @Expose()
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty({ required: false })
```
ludtkemorgan, line 31: It works to put this data at the option level. But I think this data might be better in the "extraData" array field below since it is more dynamic. You could use "addressHolderRelationship" and "addressHolderName" as the extraData keys and type of "TextInput"

ludtkemorgan (commented):
Let me know if this suggestion makes sense or not

backend/core/src/applications/types/application-multiselect-question-option.ts
KrissDrawing: @ludtkemorgan  It make sense, but extraData behaves strangely atm. First of all to extract address from it we either rely on it being first item on array (which rather should be still true after putting there more items) or by testing it's type. Keys from extraData objects for some reason are overwritten by parent `key` (so instead for example `address` its `work in county`) I believe that i can change that, assuming that the key is redundant for address.

ludtkemorgan: I think we can get away with changing the key of the extra data object to something that is more appropriate. So in your example the "extraData" object of type "address" will also have the key of "address" instead of "work in county". I don't think we use that key anywhere at the moment so it should be ok unless the frontend is using it somewhere I can't find.

sites/public/src/lib/helpers.tsx
```diff
@@ -202,7 +202,9 @@ export const untranslateMultiselectQuestion = (
 
           if (option.extraData) {
             option.extraData.forEach((extra) => {
-              extra.key = selectedOption.untranslatedText ?? selectedOption.text
+              if (!extra.key) {
+                extra.key = selectedOption.untranslatedText ?? selectedOption.text
+              }
```
KrissDrawing, line 207: that one was overwriting key's.
Tbh not sure what's the idea behind, but i assumed that overwriting it wasn't the goal here, it was more like add missing key.

sites/public/src/components/shared/FormSummaryDetails.tsx
```diff
@@ -88,21 +88,23 @@ const FormSummaryDetails = ({
...
-          `
+        acc += `${item.value.street} ${!item.value.street2 ? "," : ""} ${
+          item.value.street2 ? `${item.value.street2},` : ""
+        } ${item.value.city}, ${item.value.state} ${item.value.zipCode}`
       }
 
       return acc
     }, "")
+
+    const name = extraData.find((field) => field.key === "addressHolderName")?.value as string
```
ludtkemorgan, line 103: we should probably put "addressHolderName" and "addressHolderRelationship" as an enum somewhere since it is use in multiple places

sites/public/src/lib/helpers.tsx
ludtkemorgan: This change was added here because if you submit with a different language we want to make sure the option key is in English (untranslatedText). But I don't think the key of the extraData needs to be updated. I like what you did here and we can look later to see if this forEach is needed at all

ludtkemorgan (approved):
Thanks for making those changes! just one minor comment.

Issue #3702, opened 2023-11-03
Geocoding - wrap the feature in a feature flag


ludtkemorgan:
The geocoding feature should be behind a feature flag so that we do not release it until it is fully ready and only for the jurisdictions that want/need it.

The feature flag can be done by adding a new column to the jurisdiction table. Users should only be able to add these new preferences if the jurisdiction has the feature flag set to true

In HBA, we can plan to release to all

sarahlazarich:
@ludtkemorgan is there QA here?

ludtkemorgan:
@sarahlazarich this doesn't need any additional QA work

PR #3711, opened 2023-11-14
fix: add feature flag for geocoding

ludtkemorgan:
# Pull Request Template

## Issue Overview

This PR addresses #3702 

## Description

The geocoding feature should not be released yet so we want to have it behind a feature flag.

Also there is a grid problem with the page. This is partially fixed with this PR

Some additional unit tests have been added

turn on the new feature flag in the jurisdiction table
```
UPDATE jurisdictions
SET enable_geocoding_preferences = TRUE
where id = (select id from jurisdictions where name = 'Alameda')
```

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/sites/partners/src/components/settings/PreferenceDrawer.tsx
+++ b/sites/partners/src/components/settings/PreferenceDrawer.tsx
@@ -94,7 +94,7 @@ const PreferenceDrawer = ({
 
   const optOutQuestion = watch("canYouOptOutQuestion")
 
-  const isAdditionalDetailsEnabled = profile.jurisdictions.some(
+  const isAdditionalDetailsEnabled = profile?.jurisdictions?.some(
     (jurisdiction) => jurisdiction.enableGeocodingPreferences
   )
 
--- a/sites/partners/src/components/settings/PreferenceDrawer.tsx
+++ b/sites/partners/src/components/settings/PreferenceDrawer.tsx
@@ -94,9 +94,14 @@ const PreferenceDrawer = ({
 
   const optOutQuestion = watch("canYouOptOutQuestion")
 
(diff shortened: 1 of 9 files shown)


sites/partners/src/components/settings/PreferenceDrawer.tsx
```diff
@@ -94,9 +94,14 @@ const PreferenceDrawer = ({
 
   const optOutQuestion = watch("canYouOptOutQuestion")
 
+  const isAdditionalDetailsEnabled = profile?.jurisdictions?.some(
```
ColinBuyck, line 97: Question/Comment/Thought: Do you mind expanding on using "some" here? I remember hearing that for release we'd want to hide or show for all jurisdictions but surprised to see that we wouldn't still use the jurisdiction of the preference if we're putting it on the jurisdiction table as opposed to env. Seems like we could leave that option open code-wise while updating the db as expected. What are your thoughts?

ColinBuyck (approved):
LGTM 🛴

sites/partners/src/components/settings/PreferenceDrawer.tsx
ludtkemorgan: Discussed offline.

The issue with this page is that the jurisdiction of the preference is not known until the user selects it from the dropdown. And that can be done before or after the rest of the fields are filled out. So we can't rely on it

Also, we are planning on only releasing this to all jurisdictions at once so the "some" shouldn't be a problem. However, it is something to note for future expansions. We could potentially remove the collect geocoding information if they select a jurisdiction that doesn't have it.

PR #3739, opened 2023-11-29
feat: listing process

YazeedLoonat:
## Issue Overview

This PR addresses #3741 

## Description
This creates the listing status change over cronjob and an endpoint under listings to fire it off 

Tests were added, but you can also fire the endpoint off through http://localhost:3101/api/

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


emilyjablonski:
I absolutely could be misunderstanding how this works now, but I didn't think we were using a cron job for closing listings - as soon as the deadline including the time hits, we need to scoot the listing into the closed section so that folks aren't applying for a closed listing. Curious if I'm missing something?

backend_new/src/services/listing.service.ts
```diff
@@ -1108,4 +1161,77 @@ export class ListingService {
     });
     return mapTo(Listing, listingsRaw);
   };
+
+  /**
+    runs the job to auto close listings that are passed their due date
+    will call the the cache purge to purge all listings as long as updates had to be made
+  */
+  async process(): Promise<SuccessDTO> {
```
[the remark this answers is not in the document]
ludtkemorgan: Ope, I see you named it the same as the new controller endpoint. I guess that makes sense to keep these in sync. So I'm less worried about this name

backend_new/src/services/listing.service.ts
```diff
@@ -122,21 +131,59 @@ views.details = {
...
     private httpService: HttpService,
     private afsService: ApplicationFlaggedSetService,
     private emailService: EmailService,
     private configService: ConfigService,
+    @Inject(Logger)
+    private logger = new Logger(ListingService.name),
+    private schedulerRegistry: SchedulerRegistry,
   ) {}
 
+  onModuleInit() {
```
ludtkemorgan, line 154: This function looks basically identical to the afs cron job one. Could you pull it out to a helper function and use it in both places?

backend_new/test/integration/listing.e2e-spec.ts
```diff
@@ -613,6 +613,34 @@ describe('Listing Controller Tests', () => {
     expect(res.body.name).toEqual(val.name);
   });
 
+  it('should successfully process listings that are passed due', async () => {
```
ludtkemorgan, line 616: I think it might be useful to add a second listing in this test where the applicationDueDate is in the future. That way we can test that only the one listing is updated

ludtkemorgan:
@emilyjablonski this is how the current backend is "closing" listings. The cronjob is run every hour to see if any listings need to be closed. It's done this way for two reasons:
1. There is no easy automated way to update the status in the DB of a listing just based on the close listing timestamp without some external job.
2. We need to purge the cached listings so that the new closed listings are removed from the listing page.

One thing we could do is increase the frequency so that it is every 30 or even 15 minutes. I don't think this is a very taxing operation so more frequent shouldn't cause any performance issues

backend_new/src/services/listing.service.ts
YazeedLoonat: good call, I pulled both out for a shared function, lemme know what you think

backend_new/test/integration/listing.e2e-spec.ts
ludtkemorgan: @YazeedLoonat have you had a chance to look at this comment?

YazeedLoonat: sorry I updated this with a new test to cover this case, lemme know what you think!

ludtkemorgan (approved):

YazeedLoonat:
Note to Yazeed: Merge this in after https://github.com/bloom-housing/bloom/pull/3742 is merged

PR #3763, opened 2023-12-08
fix: application multiselect questions

seanmalbert:
## Issue Overview

This PR adds the multiselectQuestionId to `ApplicationMultiselectQuestion`, so that we can gain efficiency in tying application responses with preferences and programs to the multiselect question. The end result should be that when users submit an application, their preference and program responses will be saved with the multiselectQuestionId.


ludtkemorgan (approved):
Looks good. I'm seeing the multiselectQuestion show up in the jsonb in an application submit. But it does seem like some tests are failing

YazeedLoonat:
Hey @ludtkemorgan I think I got the tests updated, once ci passes I think this is good to merge @seanmalbert

PR #3771, opened 2023-12-13
Applications CSV (WIP)

seanmalbert:
## Issue Overview

This PR addresses #3733

- [ ] This change addresses the issue in full
- [x] This change addresses only certain aspects of the issue – it does not yet add to frontend
- [ ] This change is a dependency for another issue
- [ ] This change has a dependency from another issue

## Description

This PR adds the applications csv export functionality to the backend. It is a complete rewrite from the old csv export functionality.

TODOs:
- move csv endpoint to its own service or move bulk of function outside of application.service
- cleanup and dry export function
- add demographic mapping to export (seeds also need updating for this)
- tests
- this currently zips and streams the file (which should be way faster than the previous method and have a smaller file size). Need to discuss option with @ludtkemorgan , since it should send an email.

## How Can This Be Tested/Reviewed?

Provide instructions so we can review.

Describe the tests that you ran to verify your changes. Please also list any relevant details for your test configuration.

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have added QA notes to the issue with applicable URLs
- [ ] I have performed a self-review of my own code
- [ ] I have reviewed the changes in a desktop view
- [ ] I have reviewed the changes in a mobile view
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have assigned reviewers
- [ ] I have run `yarn generate:client` and/or created a migration if I made backend changes that require them
- [ ] My commit message(s) is/are polished, and any breaking changes are indicated in the message and are well-described
- [ ] Commits made across packages purposefully have the same commit message/version change, else are separated into different commits

## Reviewer Notes:

Steps to review a PR:

- Read and understand the issue, and ensure the author has added QA notes
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Also review the acceptance criteria on the Netlify deploy preview (noting that these do not yet include any backend changes made in the PR)
- Either explicitly ask a clarifying question, request changes, or approve the PR if there are small remaining changes but the PR is otherwise good to go

## On Merge:

If you have one commit and message, squash. If you need each message to be applied, rebase and merge.

Issue #3725, opened 2023-11-22
Prisma - Connect new backend to partners listings page


ludtkemorgan:
Connect all of the new backend's endpoints used on the listing page. All functionality on this page should work as expected

PR #3526, opened 2023-06-23
fix: timezone adjustment

ColinBuyck:
# Pull Request Template

## Issue Overview

This PR addresses #3518 

- [ ] This change addresses the issue in full
- [x] This change addresses only certain aspects of the issue (needs to be pulled into hba)
- [ ] This change is a dependency for another issue
- [ ] This change has a dependency from another issue

## Description

For this fix, I followed the pattern established by this [Detroit PR](https://github.com/CityOfDetroit/bloom/pull/1612) so that the UTC time would be converted to the exporters local timezone. Only additional changes, made in relation to manual-entry paper applications where the timing is manually entered and isn't affected by timezone adjustments of the db so the previous logic is maintained for those pieces. 

## How Can This Be Tested/Reviewed?

The most challenging component to the PR proved to be the replicating the DB's responses and time storage locally. Despite changing my databases timezone to UTC, I could not repeat this issue locally. However, the two screenshots below outline the DB responses seen on dev and I **added testing that directly mirrors these responses** to demonstrate this approach's effectiveness. 

Screenshot 2023-06-27 at 8 04 47 AM
Screenshot 2023-06-27 at 8 04 19 AM

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have added QA notes to the issue with applicable URLs
- [x] I have performed a self-review of my own code
- [x] I have reviewed the changes in a desktop view
- [ ] I have reviewed the changes in a mobile view
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have assigned reviewers
- [ ] I have run `yarn generate:client` and/or created a migration if I made backend changes that require them
- [x] My commit message(s) is/are polished, and any breaking changes are indicated in the message and are well-described
- [ ] Commits made across packages purposefully have the same commit message/version change, else are separated into different commits

## Reviewer Notes:

Steps to review a PR:

- Read and understand the issue, and ensure the author has added QA notes
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Also review the acceptance criteria on the Netlify deploy preview (noting that these do not yet include any backend changes made in the PR)
[...]

PR #3168, opened 2022-11-10
feat: adding last run to pending page

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3097)

- [x] This change addresses the issue in full

## Description
Exposes the last run date & time of the application flagged set processor to the partner's front end 

## How Can This Be Tested/Reviewed?
Head to partner's site
go to a listing's details page then navigate to the application tab
head to the application tab for pending review

notice at the bottom below the table is a time stamp for the afslastrun 

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have added QA notes to the issue with applicable URLs
- [x] I have performed a self-review of my own code
- [x] I have reviewed the changes in a desktop view
- [x] I have reviewed the changes in a mobile view
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [ ] I have assigned reviewers
- [ ] I have run `yarn generate:client` and/or created a migration if I made backend changes that require them
- [ ] I have exported any new pieces added to ui-components
- [x] My commit message(s) is/are polished, and any breaking changes are indicated in the message and are well-described
- [x] Commits made across packages purposefully have the same commit message/version change, else are separated into different commits

## Reviewer Notes:

Steps to review a PR:

- Read and understand the issue, and ensure the author has added QA notes
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Also review the acceptance criteria on the Netlify deploy preview (noting that these do not yet include any backend changes made in the PR)
- Either explicitly ask a clarifying question, request changes, or approve the PR if there are small remaining changes but the PR is otherwise good to go

## On Merge:

If you have one commit and message, squash. If you need each message to be applied, rebase and merge.

PR #3777, opened 2023-12-19
fix: partner listings applications

ludtkemorgan:
# Pull Request Template

## Issue Overview

This PR addresses #3725 

## Description

This is the second of two PR for #3725, it does the remaining steps to get the listing application page on the partner site working.

It includes the view applications for a listing along with the the application flag set flow (seeing and resolving duplicates).

*NOTES*: 
- this does not have the application csv export functionality since that is dependent on #3771 
- There are issues with the layout of some buttons/links that are tied to the recent seed changes. Those have not been address as part of this

After reseeding the new backend and starting up both the new backend and partner site you can test the following pages and functionality:
- http://localhost:3001/listings/{listingId}/applications
- http://localhost:3001/application/{applicationId}
- http://localhost:3001/listings/{listingId}/applications/add
- http://localhost:3001/application/{afsId}/review

Also the following flows should work:
- submitting an application from the public site should have that application appear in the partner site
- Closing a listing should trigger the afs job and pending flags should appear

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/sites/partners/src/pages/listings/[id]/applications/pending/index.tsx
+++ b/sites/partners/src/pages/listings/[id]/applications/pending/index.tsx
@@ -1,6 +1,9 @@
 import React from "react"
 import { useRouter } from "next/router"
 import Head from "next/head"
+import dayjs from "dayjs"
+import utc from "dayjs/plugin/utc"
+dayjs.extend(utc)
 import {
   AgTable,
   t,
@@ -14,7 +17,6 @@ import { ListingStatusBar } from "../../../../../components/listings/ListingStat
 import Layout from "../../../../../layouts"
 import { ApplicationsSideNav } from "../../../../../components/applications/ApplicationsSideNav"
 import { formatDateTime } from "@bloom-housing/shared-helpers"
-import dayjs from "dayjs"
 import { NavigationHeader } from "../../../../../components/shared/NavigationHeader"
 
 const ApplicationsList = () => {
@@ -114,10 +116,13 @@ const ApplicationsList = () => {
     formatLinkCell: isListingOpen ? formatDisabledCell : formatEnabledCell,
   }
 
-  const dayjsDate = dayjs(listingDto?.afsLastRunAt)
-  const afsLastRun = {
-    date: dayjsDate.utc().format("MM/DD/YY"),
-    time: dayjsDate.utc().format("hh:mma"),
+  let afsLastRun
+  if (listingDto?.afsLastRunAt) {
+    const dayjsDate = dayjs(listingDto?.afsLastRunAt)
+    afsLastRun = {
+      date: dayjsDate?.format("MM/DD/YY"),
+      time: dayjsDate?.format("hh:mma"),
+    }
   }
 
   return (
--- a/backend_new/prisma/migrations/20231030230627_init/migration.sql
+++ b/backend_new/prisma/migrations/20231030230627_init/migration.sql
@@ -228,7 +228,7 @@ CREATE TABLE "applications" (
     "submission_date" TIMESTAMPTZ(6),
     "marked_as_duplicate" BOOLEAN NOT NULL DEFAULT false,
     "confirmation_code" TEXT NOT NULL,
-    "review_status" "application_review_status_enum" NOT NULL DEFAULT 'valid',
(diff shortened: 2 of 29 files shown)


sites/partners/src/pages/listings/[id]/applications/pending/index.tsx
```diff
@@ -114,10 +116,13 @@ const ApplicationsList = () => {
     formatLinkCell: isListingOpen ? formatDisabledCell : formatEnabledCell,
   }
 
-  const dayjsDate = dayjs(listingDto?.afsLastRunAt)
-  const afsLastRun = {
-    date: dayjsDate.utc().format("MM/DD/YY"),
-    time: dayjsDate.utc().format("hh:mma"),
```
Comment (ludtkemorgan) [inline sites/partners/src/pages/listings/[id]/applications/pending/index.tsx:120]: I noticed the timestamp was being displayed in UTC time which is confusing since we aren't displaying the timezone. I removed that, but i could be convinced to put this back

ColinBuyck (commented):
Still reviewing but noticing that the delete application functionality isn't working. It just redirects me to the applications table where that application is still present.

ColinBuyck (commented):
When filtering by first name on the applications table, the site crashes and shows "An error has occurred." This is the error on the backend
Screenshot 2023-12-20 at 4 01 34 PM

sites/partners/src/pages/listings/[id]/applications/pending/index.tsx
Comment (ColinBuyck) [reply sites/partners/src/pages/listings/[id]/applications/pending/index.tsx:120]: Seems like this is coming from design-matching discussed [here](https://github.com/bloom-housing/bloom/pull/3168). I'd vote to show it in the user's local time to be consistent with some of the time formatting work [here](https://github.com/bloom-housing/bloom/pull/3526/files)

ludtkemorgan:
@ColinBuyck Nice catches. The filtering and sorting functionality has been fixed. Also the delete functionality I think is doing what it is supposed to be doing. It was properly deleting but the table was getting returned all applications so I have updated it to only return non-deleted listings

ColinBuyck (commented):
Nit but the afs page has applications checked by default which is not the experience on dev
Screenshot 2023-12-21 at 9 56 23 AM

ColinBuyck (commented):
Also noticing that the resolved group names are showing differently and incorrectly compared to the pending review group names
Screenshot 2023-12-21 at 10 06 52 AM

This can also happen when I save a pending group as "pending review" rather than resolved.
Screenshot 2023-12-21 at 10 21 15 AM

emilyjablonski:
Just a question: If I have a duplicate set on email, I resolve the set, and then add a third application through the paper flow with the same email, should I expect the AFS job to unresolve that set?

ludtkemorgan:
@ColinBuyck I've updated it so that things aren't checked by default. This was done because the review status was getting set to "valid" instead of "pending". It will only be apparent for digital applications as we are explicitly setting paper applications to valid which is the behavior of the old backend as well

@ColinBuyck Do you remember the steps you took to reproduce the weird group names?

@emilyjablonski If I'm following the steps correctly, it looks like the existing backend/frontend does not clear the flag set when submitting a paper application. Is that not what you are seeing?

ColinBuyck (approved):
I am pleased to announce that I can no longer reproduce those null group names 🥳 This LGTM, just found one tiny case. 

On the PR: If I go into a pending review afs and delete one of the applications, the total number of applications decreases by 1 but that application is still viewable within the afs
https://github.com/bloom-housing/bloom/assets/53269332/4b6cddfc-1f8b-4b23-a46a-1dfc232ff3b8

Doing this same flow on dev returns an error on deletion but going back to the summary view shows the application gone from the all applications number and no longer present within the afs. The erroring is obviously not ideal but I think something closer to this so users don't get confused seeing an application that has been deleted.

ludtkemorgan:
@ColinBuyck great catch. I agree both flows are not ideal. I think we might want to tackle that in a separate ticket since it will need product input. We do have a ticket https://github.com/bloom-housing/bloom/issues/3762 to fix this flow in the old backend, I think we can probably rework it to make it just a prisma change

YazeedLoonat (approved):
LGTM

PR #3819, opened 2024-01-17
fix: perm and login fixes

YazeedLoonat:
## Issue Overview
This PR fixes 2 overall issues we found in prisma bug bash:
1. User permissioning issues
a. users with the partner role were unable to accept the terms
b. public site users (users without a role) were unable to modify their user data

2. Some users were able to enter the wrong password but the correct email address and end up logging in instead of being locked out 

## Description
invite a user to the partner site, give them the partner role. As that user attempt to log in, you should now be able to log in

create a user on the public site. As that user you should be able to edit yourself via the public site

create or use an existing user. In the db change their lastLoginAt date to something well in the past (01/01/2024 works for example). Incorrectly enter the user's password when attempting to log in. You should now accurately have your account locked instead of being let through.

see above!

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/backend_new/src/passports/mfa.strategy.ts
+++ b/backend_new/src/passports/mfa.strategy.ts
@@ -171,9 +171,15 @@ export class MfaStrategy extends PassportStrategy(Strategy, 'mfa') {
   }
 
   async updateFailedLoginCount(count: number, userId: string): Promise<void> {
+    let lastLoginAt = undefined;
+    if (count === 1) {
+      // if the count went from 0 -> 1 then we update the lastLoginAt so the count of failed attempts falls off properly
+      lastLoginAt = new Date();
+    }
     await this.prisma.userAccounts.update({
       data: {
         failedLoginAttemptsCount: count,
+        lastLoginAt,
       },
       where: {
         id: userId,
--- a/backend_new/src/permission-configs/permission_policy.csv
+++ b/backend_new/src/permission-configs/permission_policy.csv
@@ -5,7 +5,7 @@ p, anonymous, application, true, submit
 
 p, admin, user, true, .*
 p, admin, userProfile, true, .*
-p, user, user, r.sub == r.obj.id, read
+p, user, user, r.sub == r.obj.id, (read|update)
 p, user, userProfile, r.sub == r.obj.id, (read|update)
 p, anonymous, user, true, create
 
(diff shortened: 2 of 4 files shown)


backend_new/src/passports/mfa.strategy.ts
```diff
@@ -171,9 +171,15 @@ export class MfaStrategy extends PassportStrategy(Strategy, 'mfa') {
   }
 
   async updateFailedLoginCount(count: number, userId: string): Promise<void> {
+    let lastLoginAt = undefined;
```
emilyjablonski, line 174: Curious about this part, outside of the context of the bug. I may be misunderstanding but it seems like now (1) if a user enters the wrong credentials we delete the last login date, but it seems like that data is unrelated and still valid? Also (2) if it's the first time a user fails to login we set their last login to be now, which seems like it also isn't valid?

ludtkemorgan: Correct me if I'm wrong @YazeedLoonat, but on the `.update` function if something is passed in as undefined it is not updated.

ludtkemorgan (approved):
⚡

backend_new/src/passports/mfa.strategy.ts
YazeedLoonat: so @ludtkemorgan is right that `undefined` in an `.update()` or `.create()` for prisma just means "make no changes" so we won't be dumping their current last login time every time

@emilyjablonski is right though that updating the last login time to be `new Date()` also isn't totally correct, I think we should look to add "last time the user attempted to log in" and using that for comparisons to see whats going on there. 

I don't think we really use the last log in time for anything though so at least for the time being I think we can keep this change set and address this post launch of prisma

YazeedLoonat: I'll hold off on merging this in @emilyjablonski if you have other concerns lemme know, or lemme know if I'm good to merge too!

emilyjablonski: @YazeedLoonat it's come up a handful of times in the support channel of wanting to know that data, so I think we need to keep it accurate - but if you'd rather make a ticket for a followup that's all good!

YazeedLoonat: Hey @emilyjablonski it took me a bit to figure it out, but it turns out the current implementation (typeorm) uses a similar trick. So the data that we've been providing won't change in terms of accuracy. 

So I went ahead and spun up: https://github.com/bloom-housing/bloom/issues/3824

and will move forward with merging this in

Issue #3814, opened 2024-01-16
Geocoding - Temporary Fixed Layer Queries for Pilot 


kathyccheng:
As an engineer on Bloom
I need a way to run the fixed layer query and store the results in the CSV export 
So that we have a way to run the pilot listing while we're waiting on MTC to get back to us 

**Acceptance Criteria**
* Update the map layer database table set as part of #3783 to have a new column with the geojson map layer
* On application submit use the turfjs helper https://turfjs.org/docs/#pointsWithinPolygon to determine if the preference is within the geojson layer. Similar logic to the radius functionality can be used
* ~~* The listings set up with a geocoding preference should still be able to be published (this can be done either by manually adding the one layer we need or by not having the select UI in place)~~
~~* Applicants should be able to provide the necessary information (address and other fields) in the preference application step, and the application should still be submitted and saved with no issue (double check even if the listing is set up with geocoding, the application submission is still fine)~~
~~* We need a script to use the address provided for these listings and check whether they're in the boundary or not (with our own ESRI set up)~~ 
~~* We need a way to populate those results into the CSV export - this can be done manually in a one time fashion, we will let the property managers and developers know not to download for geocoding information until it's been run~~

**Notes**
* This requires the map layers to be in the geojson format and manually added by the engineers. It is possible to convert shape files to geojson, but each one might need verification and also there will be a max size limit on what it can hold.
  * The pilot listing's map is already in geojson format https://dsl.richmond.edu/panorama/redlining/data
* It should be setup in such a way so that if we decide to switch to a different API long term we just have to replace the validation internal logic but still be able to use the rest of the setup.
* It would also be useful to have a way to run all applications against the verification in case this is not released when the listing goes live. But that could be a separate issue.

sarahlazarich:
Caveat: 
- might not work long term, not all jurisdictions will know how to work with a geo json file, could also run into size barriers 
- But, this solution meets the need for upcoming properties
- (sep ticket) There should also be a way to run applications after the fact in case we don't get this out in time, which would give us a bit more time to work on this, make + pull in this if time is looking tight 
- ML covered a good chunk of this work in proof of concept

sarahlazarich:
Questions 
- will we do in both prisma and typeORM (both, typeorm first prisma as a quick follow up)
- ML to create second ticket for prisma  
- testing coverage may be complicated

kathyccheng:
LGTM

PR #3825, opened 2024-01-19
fix: add map layer geocoding checks

ludtkemorgan:
# Pull Request Template

## Issue Overview

This PR addresses #3814 

## Description

This adds the validation step for all map layer preferences when an application is submitted. It uses the [turfjs pointsWithinPolygon](https://turfjs.org/docs/#pointsWithinPolygon) function to compare the lat long of the submitted application to the feature collection now saved in the mapLayer table.

A sample geojson layer has been provided in the seeded data. It is also the map layer that will be used for the pilot listing when this goes live. 

If you want to see what the layer looks like you can go to this site https://geojson.io/#map=9.32/37.6979/-122.3332 and paste the "featureCollection" portion into it.

Steps to test:
1. In partners - settings page: Create a new preference that has at least one of the option collect address and has the `Yes, check with ArcGIS map` value selected with the appropriate map layer
2. In partners - Listings page: go to a listing, attach that preference to the listing, and click save
3. In public site apply to that listing make sure that you take note of the address you are using or the preference is either inside or outside of the map layer.
4. Go back to partner site and export the application csv for that property
5. The CSV should have the appropriate "yes" or "no" or the application

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


backend/core/src/map-layers/entities/map-layer.entity.ts
```diff
@@ -20,4 +20,9 @@ export class MapLayer {
   @Column()
   @IsString({ groups: [ValidationsGroupsEnum.default] })
   jurisdictionId: string
+
+  @Column("jsonb")
+  @Expose()
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  featureCollection: any
```
YazeedLoonat, line 27: I feel like this creates some problems if we ever move to expose this as part of a DTO

as long as that won't be the case then this is okay!

YazeedLoonat (commented):
Just some questions on here! looks good

backend/core/src/applications/services/geocoding.service.ts
```diff
@@ -48,6 +52,43 @@ export class GeocodingService {
...
+    try {
+      if (preferenceAddress.latitude && preferenceAddress.longitude) {
+        const preferencePoint = point([
+          Number.parseFloat(preferenceAddress.longitude.toString()),
+          Number.parseFloat(preferenceAddress.latitude.toString()),
+        ])
+
+        // Convert the features to the format that turfjs wants
+        const polygonsFromFeature = []
+        featureCollectionLayers.features.forEach((feature) => {
```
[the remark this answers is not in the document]
ludtkemorgan: Good question. I think the `unknown` response is probably good enough and we should be logging the error on line 85. However, I'll get confirmation from product

backend/core/src/map-layers/entities/map-layer.entity.ts
ludtkemorgan: I wasn't happy about this as well. But my thought was that the `.entity` file is not needed in the Prisma backend and if we want to explicitly type it at that point it might be better to create a DTO in the new backend

ColinBuyck:
This seems to work well with new preferences but when I added the arcGIS collection to an existing preference the column shows blank in the export despite the application form reflecting the updates. Are you able to reproduce this @ludtkemorgan?

ludtkemorgan:
@ColinBuyck Are those applications that were submitted before you edited the preference? Or were they new applications?

emilyjablonski:
I'm having a similar issue, what I think the pattern is, is if I claim a preference where the answer doesn't require an address, it is showing up blank

ColinBuyck:
@ludtkemorgan The steps I followed to reproduce this now was:
1) Yarn and reseed
2) In partners, edit live/work in county - alameda, and update live in county preference to collect address and use arcGIS
3) Save
4) Go to the "No Preferences" listing and add this preference
5) Fill out app on public side
6) export applications on partner's side

Let me know if that works!

emilyjablonski:
Not sure if it matters for this release, but something I'm noticing is that the fields don't appear in the frontend for preferences like HOPWA that are not checkboxes but radios.

ludtkemorgan:
@emilyjablonski Good catch, I'll notify product that might have been a missed use case.

@ColinBuyck I think I tested it on main and the same thing happens for any preference that is added. Looks like at least one person has to claim the option for it to appear.

emilyjablonski:
@ludtkemorgan Here's a flow for the bug for me
1. Using the seeded FCFS listing
2. The displacee preference is collecting an address
3. Update the live/work preference to collect with the map layer
4. Fill out an application and claim both with an address
5. The first preference comes through claimed / checked / with an address, the second is empty even though I claimed it

ludtkemorgan:
@emilyjablonski @ColinBuyck Ah! turns out it's not the fact that it's an existing preference, but rather that there are more than one preference on the listing. The second one is not getting added to the DB. I'll work on fixing it. Nice catch!

ColinBuyck (approved):
LGTM 🍼

emilyjablonski (approved):
Looks great!

PR #3840, opened 2024-01-29
fix: bug bash fixes

ludtkemorgan:
Fixes the following bugs for Prisma found during bashes:

1. When exporting applications deleted application are no in the file
2. New migrations will no longer get run out of order (notice the name change)
3. The unitsAvailable value is set on listing creation rather than just updates and seeded data reflects this as well
4. Listing export works properly even with query params passed in.

Testing:
1. submit an application for a listing and then delete it in the partner UI. Go to the applications tab for that listing and click "export". Only the non-deleted applications should appear
2. Not super testable, but reseeding the database should properly run the migrations
3. On the listing details page the top right of the page should show the correct number of "vacant units". This should also be the case for a newly created listing as long as it is not a waitlist listing
4. Listing export button should work on the UI and the downloaded file should be able to be opened


YazeedLoonat (approved):
LGTM

Issue #3769, opened 2023-12-12
On application pages, check if user is signed-in


cade-exygy:
To prevent the possibility of a user being directed to the middle of an application, we should ensure the user is signed in when they visit every application page.

If they are not signed in at any point, redirect them to the beginning of the application.

cade-exygy:
If not signed in, redirect to the beginning of the application.

cade-exygy:
Application will get cleared when the user clicks begin

cade-exygy:
We are seeing users get to different parts of the application without supposedly required information

PR #3850, opened 2024-01-31
feat: add MA application check for signed-in user

cliu02:
# Pull Request Template

## Issue Overview

This PR addresses #3769.

- [X] This change addresses the issue in full

## Description

As part of the Mandated Accounts MVP, this PR redirects users to the sign-in page if they access an application page while not signed in. This prevents the case where signed-out users are directed to a page in the middle of an application.

To review this PR:

1. Ensure that you are not signed in. 
2. Check that when you access a page in the middle of an application (see [urls](https://github.com/bloom-housing/bloom/blob/9f35e7e1a995c63dd6bbf23a23ca8ece049b3843/sites/public/src/lib/applications/ApplicationConductor.ts#L41)), you are redirected to the `/sign-in` page. For example, `http://localhost:3000/applications/start/what-to-expect` or `http://localhost:3000/applications/review/summary`.

Note: This ticket does not include redirection for the case where a signed-in user access the middle of an application, as it applies specifically to signed-out users. 

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/shared-helpers/src/auth/UseRequireLoggedInUser.ts
+++ b/shared-helpers/src/auth/UseRequireLoggedInUser.ts
@@ -6,11 +6,11 @@ import { NavigationContext } from "@bloom-housing/ui-components"
  * Require a logged in user. Waits on initial load, then initiates a redirect to `redirectPath` if user is not
  * logged in.
  */
-function useRequireLoggedInUser(redirectPath: string, disable?: boolean) {
+function useRequireLoggedInUser(redirectPath: string) {
   const { profile, initialStateLoaded } = useContext(AuthContext)
   const { router } = useContext(NavigationContext)
 
-  if (!disable && initialStateLoaded && !profile) {
+  if (initialStateLoaded && !profile) {
     void router.push(redirectPath)
   }
   return profile
--- a/shared-helpers/src/auth/UseRequireLoggedInUser.ts
+++ b/shared-helpers/src/auth/UseRequireLoggedInUser.ts
@@ -1,6 +1,6 @@
 import { useContext } from "react"
 import { AuthContext } from "./AuthContext"
-import { NavigationContext } from "@bloom-housing/ui-components"
+import { useRouter } from "next/router"
 
 /**
  * Require a logged in user. Waits on initial load, then initiates a redirect to `redirectPath` if user is not
@@ -8,9 +8,9 @@ import { NavigationContext } from "@bloom-housing/ui-components"
  */
 function useRequireLoggedInUser(redirectPath: string) {
   const { profile, initialStateLoaded } = useContext(AuthContext)
-  const { router } = useContext(NavigationContext)
+  const router = useRouter()
 
-  if (initialStateLoaded && !profile) {
+  if (process.env.showMandatedAccounts && initialStateLoaded && !profile) {
     void router.push(redirectPath)
   }
   return profile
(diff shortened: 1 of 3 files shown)


ColinBuyck (commented):
I'm noticing that if I hit the choose language page directly, the sign in page flashes but I end up on the home page because the redirect below it line 62 is still called.

This also got me thinking that redirecting to the homepage, whether it is mandated accounts or not, may be better since I felt confused that signing in doesn't take me back to the application as it does with so many other flows. I think it would be great to have that redirect back to the application work but if that's out of scope, I would vote for the homepage for consistency with the existing non-MA redirects. CC: @emilyjablonski @cade-exygy for other thoughts

shared-helpers/src/auth/UseRequireLoggedInUser.ts
```diff
@@ -1,16 +1,16 @@
...
+import { useRouter } from "next/router"
 
 /**
  * Require a logged in user. Waits on initial load, then initiates a redirect to `redirectPath` if user is not
  * logged in.
  */
 function useRequireLoggedInUser(redirectPath: string) {
   const { profile, initialStateLoaded } = useContext(AuthContext)
-  const { router } = useContext(NavigationContext)
+  const router = useRouter()
```
emilyjablonski, line 11: Why was this change made?

shared-helpers/src/auth/UseRequireLoggedInUser.ts
```diff
@@ -1,16 +1,16 @@
...
 
-  if (initialStateLoaded && !profile) {
+  if (process.env.showMandatedAccounts && initialStateLoaded && !profile) {
```
emilyjablonski, line 13: If we add this check here, this function now only works if mandated accounts is on. The functionality should be separate from the mandated accounts toggle.

emilyjablonski:
Let me know if I'm incorrect here on the level of effort, but I think we just need to be passing the listing id as a param in the URL in order to have the application redirect?

shared-helpers/src/auth/UseRequireLoggedInUser.ts
cade-exygy: I believe it was because the next router is what we have been using for redirects. For example in forgot-password.tsx or confirmation.tsx. And I don't this hook was utilized anywhere else in the webapp. Do you the difference between using the next router and NavigationContext router? I noticed there are more available function calls on the next router, but wasn't sure why different routers were used in different places.

shared-helpers/src/auth/UseRequireLoggedInUser.ts
emilyjablonski: In general, a helper / function like this should not include business logic. You should be able to conditionally call the function based on the feature toggle without duplicating it - lmk if that doesn't work!

cade-exygy: I think we had issues conditionally calling this because it is a hook

ColinBuyck (approved):
LGTM 🏁

cade-exygy (approved):
Those changes look good!

ColinBuyck:
For context @emilyjablonski, I approved with the useFormConductor approach because we found an edge case where a user could sign in, start an application, sign out mid application, and then (if hit a direct application link url) access that same listing application due to the id still being in session storage

emilyjablonski (approved):

PR #3879, opened 2024-02-08
chore(deps): upgrade nestjs

ludtkemorgan:
Tackles the dependabot alert https://github.com/bloom-housing/bloom/security/dependabot/178.

This is two major versions updates of the nestjs primary packages so there is a non-zero chance that this can cause an issue. However reading through the breaking changes it appears like we don't use any of that functionality. https://docs.nestjs.com/migration-guide and https://trilon.io/blog/nestjs-9-is-now-available

All tests pass. But it is recommended that you start it up locally and make sure calls to the backend still work


YazeedLoonat (approved):
LGTM

PR #3897, opened 2024-02-22
fix: null check on option links

ludtkemorgan:
When a preference collects address but doesn't have a description or any links a null pointer exception happens when clicking the "preview" button when selecting a preference on a listing.

To test:
Assuming there is a preference that has at least one option that "collects address" and doesn't have a description or any links.
- In edit mode of a listing click "add preference".
- On the above mentioned preference click "preview"
- The preview should properly show


YazeedLoonat (approved):
LGTM

Issue #3900, opened 2024-02-23
Passwordless - new login endpoint


YazeedLoonat:
From the technical doc: https://docs.google.com/document/d/18DuWgsrZszJCO90knNSQ4yXj7fvwzZ9ZvHmV0rGe2nk/edit#heading=h.u360s4xybd8z

we need to create an endpoint for logging in via the single use code

it will take in email, single use code, and the jurisdiction off of the public site
If login is successful it will log a user in

PR #3928, opened 2024-03-07
feat: new single use code login endpoint

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3900)

## Description
This creates a new login via single use codes endpoint

unit and integration tests were added but the way to test this is to add requesting a single use code, and login via single use code to the public site locally (in place of the current login in authcontext call we use)

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


api/src/passports/single-use-code.strategy.ts
```diff
@@ -0,0 +1,199 @@
...
+      throw new BadRequestException(
+        'jurisdictionname is missing from the request headers',
+      );
+    }
+
+    const jurisName = req.headers['jurisdictionname'];
+    const juris = await this.prisma.jurisdictions.findFirst({
+      where: {
+        name: {
+          in: Array.isArray(jurisName) ? jurisName : [jurisName],
```
ludtkemorgan, line 55: I could be missing it. But why can jurisdictionname header be an array?

api/src/passports/single-use-code.strategy.ts
```diff
@@ -0,0 +1,199 @@
...
+        },
+        allowSingleUseCodeLogin: true,
+      },
+    });
+    if (!juris) {
+      throw new BadRequestException(
+        'Single use code login is not setup for this jurisdiction',
+      );
+    }
```
ludtkemorgan, line 64: I think it would be useful to know if this error happens because the jurisdiction exists or if somebody passes in an invalid jurisdiction. Also it looks like we don't actually use the jurisdiction object. Would it make sense to fetch the jurisdiction using `findFirst` but just select the `allowSingleUseCodeLogin`. Then if `juris` doesn't exist that means the call was because it was a jurisdiction that doesn't exist and if `allowSingleUseCodeLogin` is not true we would throw the BadRequestException you have here.

Also as a side note, it would be worthwhile to log the jurisdiction name for these errors

api/src/passports/single-use-code.strategy.ts
```diff
@@ -0,0 +1,199 @@
...
+      passwordField: 'singleUseCode',
+      passReqToCallback: true,
+    });
+  }
+
+  /*
+    verifies that the incoming log in information is valid
+    returns the verified user
+  */
+  async validate(req: Request): Promise<User> {
```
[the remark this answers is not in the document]
ludtkemorgan: +1

ludtkemorgan (changes requested):

api/src/passports/single-use-code.strategy.ts
YazeedLoonat: I think it might just be the default value for the request headers typing, I changed it to be cast as a string

api/src/passports/single-use-code.strategy.ts
YazeedLoonat: updated!

api/src/passports/single-use-code.strategy.ts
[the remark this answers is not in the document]
YazeedLoonat: I pulled some logic into a shared utilities file, lemme know what ya think

YazeedLoonat:
Hey @ludtkemorgan @emilyjablonski this is ready for re-review

emilyjablonski:
I'm still seeing a lot of duplication between mfa.strategy and single-use-code.strategy, ex `updateFailedLoginCount`, is there more we can do here?

YazeedLoonat:
@emilyjablonski 
ah I see! I think my thoughts here is that: even though its duplicated code, the code ends up being cleaner not pass the prisma service around

so I think my preference is to leave those functions in both strategy files. We only pass the prisma service around in 1 spot (cronjob creation) and I think that should be treated as a one off rather than a consistent pattern

emilyjablonski (approved):

api/src/utilities/passport-validator-utilities.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @param failedLoginAttemptsCount the number of times the user failed to log in (stored in db)
+ * @param maxAttempts the maximum number of attempts before user is considered locked out (env variable)
+ *
+ * @returns throws error if user is already locked out
+ */
+export function isUserLockedOut(
+  lastLoginAt: Date,
+  failedLoginAttemptsCount: number,
+  maxAttempts: number,
+  cooldown: number,
```
ludtkemorgan, line 15: Do you anticipate the maxAttempts and cooldown numbers to change per request type? We could just grab the env variables in this method instead of passing them in. But if you want it to be more versatile then I'm cool with this as well

ludtkemorgan (approved):

api/src/utilities/passport-validator-utilities.ts
YazeedLoonat: yeah I think its best to keep this versatile as just a helper function

PR #3940, opened 2024-03-11
test: update seeds + fix listing delete

emilyjablonski:
# Pull Request Template

## Description

Updates the seed command so that when you run `yarn setup:dev`, the following is now true:
* If you were to go into partners and save a listing, all the required fields to publish are filled out, which includes developer, leasing agent details, application details, and an image
* Giving reserved community type a 50% chance of being included instead of always being included
* Ensuring there are units on every listing
* Ensuring there are at least 4 open listings added, and that every other listing has a digital app, so that when you run setup:dev there are guaranteed to be at least two listings with a digital application

You can run `yarn setup:dev` in `api/` and check it out, the randomization will result it in being slightly different every time.

### DIFF
--- a/api/prisma/migrations/08_updating_on_delete_cascades_listings_and_applications/migration.sql
+++ b/api/prisma/migrations/08_updating_on_delete_cascades_listings_and_applications/migration.sql
@@ -1,0 +1,110 @@
+-- DropForeignKey
+
+ALTER TABLE "application_methods"
+DROP CONSTRAINT "application_methods_listing_id_fkey";
+
+-- AddForeignKey
+
+ALTER TABLE "application_methods" ADD CONSTRAINT "application_methods_listing_id_fkey"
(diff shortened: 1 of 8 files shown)


mcgarrye (approved):

api/prisma/migrations/08_updating_on_delete_cascades_listings_and_applications/migration.sql
```diff
@@ -0,0 +1,110 @@
+-- DropForeignKey
+
+ALTER TABLE "application_methods"
```
ludtkemorgan, line 3: These changes aren't referenced in the PR description. But I think I remember Yazeed talking about these changes. These are all just to allow proper deletion of listings?

emilyjablonski: Yep! It's the cascading delete issues.

ludtkemorgan (approved):

Issue #3433, opened 2023-05-08
Uptake Alert Component in Bloom-Core 


sarahlazarich:
Description: Uptake component into Bloom Core

Links:

Acceptance criteria:

QA review instructions

This component is primarily used on:

- TBD

Ensure the display of data is consistent between production and the PR review site in these key areas.

jaredcwhite:
This issue is currently blocked by both bloom-housing/ui-seeds#80 and bloom-housing/ui-seeds#82.

PR #3962, opened 2024-03-19
feat: uptake of Seeds Toast component

jaredcwhite:
## Issue Overview

This PR addresses #3434
(and builds on/supersedes #3553)

## Description

We had a version of the old UIC AlertBox which offered a "sticky" option that sort of looked like a "toast notification" but it really didn't work quite like how that works in most other design systems and didn't offer automatic stacking of multiple notifications as well as timeouts.

This PR switches to using the Toast component from Seeds, and adds a new `MessageContext` which provides a mechanism to add new toasts to the stack. This doesn't replace all use of AlertBox across the codebase because in other cases those alerts weren't sticky. We should evaluate as part of #3433 how we can update those to Seeds Alert or Message, or Toast if they should in fact be presented like toasts.

This also removes some previous bespoke mechanisms to preserve alert messages across page navigations via `sessionStorage`. We don't need that anymore because just using the Next router + the Toast stack works fine.

It's a little hard to test because toasts require actions of some kind, but you could try the following in public and partners:

* Sign in, sign out
* Create & update users
* Save data (listings and applications on partners)
* Timing out when filling out a common application while signed in
* Any other feature where you feel you'd expect a toast confirmation after performing an action

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


sites/partners/src/components/applications/PaperApplicationForm/PaperApplicationForm.tsx
```diff
@@ -147,11 +142,11 @@ const ApplicationForm = ({ listingId, editMode, application }: ApplicationFormPr
       setLoading(false)
 
       if (result) {
-        setSiteAlertMessage(
+        addToast(
           editMode
             ? t("application.add.applicationUpdated")
             : t("application.add.applicationSubmitted"),
-          "success"
+          { variant: "success" }
```
mcgarrye, line 149: I think it would be nice to have a frontend enum for all the variant types. Not a blocking request though

mcgarrye (approved):
Looks good! lgtm

sites/partners/src/components/applications/PaperApplicationForm/PaperApplicationForm.tsx
jaredcwhite: @mcgarrye so about that…I forget the exact reasoning but in the new Seeds components we dropped explicit TS enum types and just went with string unions inside the props interface. I'll double-check this PR though as I know we can import individual prop types if we need to for Core usage, and not need to re-type "success" | "warning" | etc.

jaredcwhite: Yeah I double-checked — we're pulling in the variant terms directly from the TS interface, so it should work for autocomplete/type-checking.

shared-helpers/src/utilities/MessageContext.ts
```diff
@@ -0,0 +1,51 @@
+import React, { FunctionComponent, createContext, createElement, useState, useRef } from "react"
+import { CommonMessageProps } from "@bloom-housing/ui-seeds/src/blocks/shared/CommonMessage"
+
+// TODO: this should be exportable from seeds directly
```
emilyjablonski, line 4: Can we link a ticket for this work?

jaredcwhite: @emilyjablonski Good point! I just wrote one up here: https://github.com/bloom-housing/ui-seeds/issues/83

sites/partners/styles/overrides.scss
```diff
@@ -14,6 +14,10 @@
     text-align: center;
   }
 
+  #seeds-toast-stack {
+    --toast-stack-inset-top: 7rem;
```
emilyjablonski, line 18: Should this be --seeds-s28?

sites/public/src/pages/account/edit.tsx
```diff
@@ -162,8 +161,6 @@ const Edit = () => {
           headingPriority={1}
         >
           <>
-            <SiteAlert type="notice" dismissable />
```
emilyjablonski, line 165: Curious why this wasn't replaced?

emilyjablonski:
I would have anticipated after Jesse's comment [here](https://github.com/bloom-housing/bloom/pull/3553#issuecomment-1766821830) that the toast would only go as far right as the content, as opposed to all the way to the right edge of the screen as seen below on this PR

Screenshot 2024-04-15 at 1 37 30 PM

sites/public/src/pages/account/edit.tsx
jaredcwhite: @emilyjablonski So these were essentially containers to load in the old "toast" (aka AlertBox sticky) messages and display them if necessary. Since we're handling that now at the layout level, we don't need an equivalent to SiteAlert on individual pages.

jaredcwhite:
@emilyjablonski I guess I interpreted that as hug the "viewport", as that's what it looks like to me in Figma. 🤷🏻‍♂️ Technically I'm not even sure how I'd align toasts to a max-width content container in a clean manner, as the toast stack sits right at the `<body>` level…but if we need to we need to. Thoughts @slowbot?

jaredcwhite:
New idea: because of the prevalence of mobile, I think a fair number of examples out there of toasts actually have them centered and at the bottom of the screen (or even right-aligned but still bottom). I wonder if that potentially solves some issues?

Issue #3950, opened 2024-03-13
Apply rate limiting


YazeedLoonat:
We want to apply rate limiting to the API

we determined the best approach here is: https://docs.nestjs.com/security/rate-limiting#proxies

using the nestjs throttle guard and customizing it a bit to suit our needs

this ticket is to create that custom throttle guard, tests against it, and applying it to the various endpoints/controllers in our app

PR #3967, opened 2024-03-20
fix: Proxy attempt3

YazeedLoonat:
This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3950)

## Description
this changes the logging at bit more to see if I can get the original IP instead of the proxy ip


ColinBuyck (approved):
LGTM 🧦

Issue #3948, opened 2024-03-13
Geocoding - Adjust feature flag 


kathyccheng:
As Exygy, 
We want to limit the ability to use the within radius option until Oakland is able to pay us for their portion of the work 
So we get into situations where we do work for free 

When I go to the preference set up, 
then I should no longer see the within radius radio selection 

**Notes** 
* I think we have a feature flag on the whole Do you need help validating your address section, we want to adjust that to just cover the within radius radio option
* We should have it turned off for HBA + Doorway 

**Design**
<img width="1146" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHZaQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--43b78cca71a4a4180d100bf0eac5076a9738334e/image.png" alt="image.png" />

KrissDrawing:
@kathyccheng If there is no jurisdiction selected in preference form select, should that show this option or not ?

kathyccheng:
@KrissDrawing don't show in that case

KrissDrawing:
QA Notes: radius option is hidden behind `enableGeocodingRadiusMethod` flag in jurisdiction. By default it's hidden (false)

PR #3974, opened 2024-03-25
fix: add flag to hide geocoding validation by radius

KrissDrawing:
# Pull Request Template

## Issue Overview

This PR addresses #3948 

## Description

Adds feature flag that is default to `true` (so full geocoding functionality). On `false` it hides `"validate radius"` option, and on submit it removes it's fields (as there is possibility to create option on one jurisdiction (with radius validation) and then switch to other with pre-populated fields)

Switch flag to false for jurisdiction of choice. 
- in partners settings create/edit preference.
- Add option with jurisdiction of your choice (it should not have validate by radius radio box).
- Create options for jurisdiction that has flag on `true` then switch for jurisdiction of choice (with false) and submit -> it should switch all validationRadius to `will check manually`.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


api/prisma/migrations/09_add_geocoding_flag_radius_method/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
+ALTER TABLE "jurisdictions" ADD COLUMN     "enable_geocoding_radius_method" BOOLEAN NOT NULL DEFAULT true;
```
mcgarrye, line 2: Can we change the folder prefix to 10 as 09 is already in use?

mcgarrye (changes requested):
Two things:
1. I feel like we should be defaulting to `false`. Based on the ticket, we want this turned off in all existing jurisdictions until further notice. So we should only be switching it to true on a per case basis going forward.

2. With the flag turned to true, I was able to see the field when editing existing option but NOT when creating new options. Not familiar with this code but maybe it is missing from the new option flow?

KrissDrawing:
@mcgarrye thanks for the review!

1. Yes, it probably makes more sense to default it to `false`.
2. Not sure if it's what i think it is, but this flow is slightly counterintuitive. As when you create new `preference` then you have no jurisdiction selected (so no radius option present in `options`). Then when you select jurisdiction that has that flag turned on the option will appear. Only thing that i can improve here is to add this option initially (so when no jurisdiction selected) when every jurisdiction has flag turned on. Let me know if thats the issue, or is it something else 🤔

mcgarrye:
> @mcgarrye thanks for the review!
> 
> 1. Yes, it probably makes more sense to default it to `false`.
> 2. Not sure if it's what i think it is, but this flow is slightly counterintuitive. As when you create new `preference` then you have no jurisdiction selected (so no radius option present in `options`). Then when you select jurisdiction that has that flag turned on the option will appear. Only thing that i can improve here is to add this option initially (so when no jurisdiction selected) when every jurisdiction has flag turned on. Let me know if thats the issue, or is it something else 🤔

In response to #2, that is definitely what I am seeing. If the user selects a jurisdiction before adding an option it does appear properly. I would just flag it to product but I think it is fine to leave it as is since all jurisdictions will have it turned off for now.

api/prisma/seed-helpers/jurisdiction-factory.ts
```diff
@@ -17,5 +17,6 @@ export const jurisdictionFactory = (
   enableAccessibilityFeatures: true,
   enableUtilitiesIncluded: true,
   enableGeocodingPreferences: true,
+  enableGeocodingRadiusMethod: true,
```
mcgarrye, line 20: Can we flip this to false too please

KrissDrawing: @mcgarrye Do you have any specific reason for that?
It seem like locally it's not a problem to have it switched on.
It also fails tests (which i can change, but probably they will never appear later 😄 ).
So just let me know it you see any problems with that beeing `true` here.

mcgarrye: Generally we want the seed data to match the actual environment to the best of our ability. And would be good to have a test around the flag

KrissDrawing: It makes sense! I switched it to `false`, and modified test to use second option (so map layer validation). Is that ok ?

mcgarrye (approved):
As long as product is aware of the new setting flow when the flag is true that we discussed, this looks good to go. lgtm

YazeedLoonat (approved):
LGTM 🍝

Issue #3935, opened 2024-03-11
AC: Application Export not Working


iph-97:
Our partners in AC reported that they are unable to export applications data for Ancora Place, it has consistently thrown up an error. I was able to replicate this issue. Potentially related, other listings' data is exporting via a download in browser, rather than sending via email. This is a high priority issue.
<img width="1684" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMHJOQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--96e70c774538b4b63bc0b9b8a8b651081b716c94/image.png" alt="image.png" />

sarahlazarich:
2 issues: 
- shouldn't export until preference issue is fixed, need to check on geocoding behavior.
- Memory issue 
- 10 second timeout happened here, not the memory call. Browser has a timeout, need to check on that

sarahlazarich:
test to see how long the call is locally - if it's a 10sec timeout and this is taking 12 could be an easy fix

PR #3978, opened 2024-03-26
fix: cuts time to get an app export

YazeedLoonat:
## Issue Overview

This PR addresses [#issue](https://github.com/bloom-housing/bloom/issues/3935)

- [x] This change addresses part of the issue
## Description
This should speed up the time to get an application export pretty significantly. 
This should also keep our memory footprint lower than it used to be (in line with the recent change for better memory management)

- Easiest way to check this is to use a listing with >9k applications tied to it
- request an application export for that listing
- you should notice the export shows up much faster and importantly the data in that csv **should **not** have changed
- you should also see the memory footprint remain low (around 200MB on my machine locally)

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/api/src/services/application-csv-export.service.ts
+++ b/api/src/services/application-csv-export.service.ts
@@ -41,6 +41,8 @@ export const typeMap = {
   fiveBdrm: 'Five Bedroom',
 };
 
+const NUMBER_TO_PAGINATE_BY = 500;
+
 @Injectable()
 export class ApplicationCsvExporterService
   implements CsvExporterServiceInterface
@@ -88,6 +90,11 @@ export class ApplicationCsvExporterService
     const applications = await this.prisma.applications.findMany({
       select: {
         id: true,
+        householdMember: {
+          select: {
+            id: true,
+          },
+        },
       },
       where: {
         listingId: queryParams.listingId,
@@ -101,10 +108,13 @@ export class ApplicationCsvExporterService
         queryParams.listingId,
       );
 
-    // get maxHouseholdMembers or associated to the selected applications
-    const maxHouseholdMembers = await this.maxHouseholdMembers(
-      applications.map((application) => application.id),
-    );
+    // get maxHouseholdMembers associated to the selected applications
+    let maxHouseholdMembers = 0;
+    applications.forEach((app) => {
+      if (app.householdMember?.length > maxHouseholdMembers) {
+        maxHouseholdMembers = app.householdMember.length;
+      }
+    });
 
     const csvHeaders = await this.getCsvHeaders(
       maxHouseholdMembers,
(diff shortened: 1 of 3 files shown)


api/src/services/application-csv-export.service.ts
```diff
@@ -131,131 +141,120 @@ export class ApplicationCsvExporterService
...
+                              race: true,
+                            },
+                          }
+                        : false,
+                    },
+                    where: {
+                      listingId: queryParams.listingId,
+                      deletedAt: null,
+                    },
+                    skip: i,
```
emilyjablonski, line 172: Curious why this changed from (i * num_to_paginate_by) to just i

YazeedLoonat: ah yeah that has to do with how I was iterating earlier 
in the for loop I was trying to count by batch number basically (0,1,2,..n) but I realized its easier to just count by where we are in the batch (0, 500, 1k, 1.5k, n) instead so we don't need to figure out how many to skip ourselves anymore

api/src/services/application-csv-export.service.ts
```diff
@@ -131,131 +141,120 @@ export class ApplicationCsvExporterService
...
+                let row = '';
+                paginatedApplications.forEach((app) => {
+                  let preferences: ApplicationMultiselectQuestion[];
+                  csvHeaders.forEach((header, index) => {
+                    let multiselectQuestionValue = false;
+                    let parsePreference = false;
+                    let value = header.path.split('.').reduce((acc, curr) => {
+                      // return preference/program as value for the format function to accept
+                      if (multiselectQuestionValue) {
+                        return acc;
```
mcgarrye, line 185: How does line 184 ever evaluate to true?

YazeedLoonat: as we iterate through in the `reduce` first `parsePreference` can become true, and then if that becomes true `multiselectQuestionValue` becomes true

YazeedLoonat: as we iterate through with the `reduce()` first `parsePreference` (line 206) can become true, and then later `multiselectQuestionValue` (200) can become true

mcgarrye (approved):
CSV downloaded in roughly 6s

ColinBuyck:
@YazeedLoonat Do you think there's room in this PR to expand seeding functionality to better support testing things at scale? I'm thinking about like a yarn setup 10000 that would seed 10,000 applications on one of the listing so we could more regularly test performance and ensure any other application/export changes keep response times reasonable (<10 sec)

YazeedLoonat:
Hey @ColinBuyck I'm not sure if adding that would add much

The 10 seconds isn't really a constraint outside of netlify (infrastructural constraint) 

and I think a better/more sustainable way to test this is to use an existing dataset with have access too (production data for example) and anonymizing that and setting that up in a test env that we can qa things against (like setting that data in staging)

where we also have similar enough infrastructure to production to catch timeouts and stuff like that

ColinBuyck (approved):
LGTM 🧷 I saw a 10 second improvement on one with 10k+ applications, very exciting! One nit but in the future, could you explain **why** changes improve performance in the PR description rather than just noting the areas of improvement. Also for documentation sake, the testing conversation above boiled down to this work per offline discussion #3995

PR #3999, opened 2024-04-08
fix: email updates w pwdless

emilyjablonski:
# Pull Request Template

## Issue Overview

This PR addresses #issue

## Description

Please include a summary of the change and which issue(s) is addressed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


api/src/passports/single-use-code.strategy.ts
```diff
@@ -87,6 +87,7 @@ export class SingleUseCodeStrategy extends PassportStrategy(
       },
     });
     if (!rawUser) {
+      // this is failing
```
emilyjablonski, line 90: Have struggled w the most secure thing to do here @YazeedLoonat

api/src/services/email.service.ts
```diff
@@ -326,7 +326,11 @@ export class EmailService {
     );
   }
 
-  public async sendSingleUseCode(user: User, singleUseCode: string) {
+  public async sendSingleUseCode(
+    user: User,
+    singleUseCode: string,
+    template?: string,
```
YazeedLoonat, line 332: wondering if using an enum here would make template easier to work with (less searching for what the options are if we ever use this function for something else)

api/src/services/user.service.ts
```diff
@@ -193,7 +193,7 @@ export class UserService {
         },
       });
 
-      if (juris.allowSingleUseCodeLogin) {
+      if (juris?.allowSingleUseCodeLogin) {
```
YazeedLoonat, line 196: I feel like there shouldn't be a case where juris is undefined. What was the case that made this  change necessary?

api/src/passports/single-use-code.strategy.ts
YazeedLoonat: maybe we sync up on this, but I think whats happening is:

- user has account with email
- they go to update their email to new email
- we send an email to their new email with a code, but we don't immediately update their user to have the new email. Instead we save the new email in the confirmation token in the db and once they successfully log in we do the email update
- the front end is sending the new email + code to the backend to be logged in. But that new email doesn't exist yet

I think the front end needs to send the original email + code and then the new email will take

hopefully that helps!

YazeedLoonat (commented):
looks good!

emilyjablonski:
I think after reflecting on this I'm not sure we should be sending a code, will touch base w design

PR #4085, opened 2024-05-14
fix: reading a listing shouldn't be protected

YazeedLoonat:
removes API pass key from listing read endpoints

### DIFF
--- a/api/test/integration/api-key-guard.e2e-spec.ts
+++ b/api/test/integration/api-key-guard.e2e-spec.ts
@@ -1,12 +1,16 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { INestApplication } from '@nestjs/common';
 import request from 'supertest';
+import cookieParser from 'cookie-parser';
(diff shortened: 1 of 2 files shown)


api/test/integration/api-key-guard.e2e-spec.ts
```diff
@@ -23,23 +23,25 @@ describe('API Key Guard Tests', () => {
     await app.close();
   });
 
-  it('should succeed when correct header is present', async () => {
+  it('should error but because of unauthorized (got passed the pass key)', async () => {
```
emilyjablonski, line 26: should we keep a succeed test?

emilyjablonski (approved):

api/test/integration/api-key-guard.e2e-spec.ts
YazeedLoonat: you right, I was being lazy with it. Updated the test to be logged in now and we look for a success

PR #4096, opened 2024-05-21
fix: restores getStaticProps functionality

YazeedLoonat:
This fixes an issue where during the getStaticProps call we were asking for the request but getStaticProps does not have access to the request as its a static generator

this fix should keep the public's listings list page working, the public listing detail page working, and shouldn't throw an error in the public site during build/initially when the public site has `yarn dev` ran


emilyjablonski (approved):
looks good locally 🫡

Issue #4081, opened 2024-05-09
Application Translations Not Being Updated


ColinBuyck:
When a user selects the language of their application, the api checks if there are translations saved for that language and listing and if so, those are used rather than a fresh translation of the listing. This misses the case where a listing is edited while it is live, closed and reopened, or the preference attached to the listing is updated. Ideally, if the preferences or listings have an updated at value after the translation created at field, it should fetch new translations.

Updates: Based on a team discussion, it seems adding a new field that tracks when the most recent listing content updates by partner portal users occurred would be helpful for this case and others. 

QA: https://bloom.exygy.dev/ and https://partners.bloom.exygy.dev/ 

1. Find a listing with an online application and without preferences (ie. Blue Sky Apartments)
2. Click Apply online
3. Click "Empezar"
4. Click back and you should be on the choose language page again
5. On the partner's portal, go to that listing and add a preference.
6. Return to the choose langauge tab and click empezar again.
7. See that the options in the application are translated
8. Go back once again to the choose-language step
9. On the partner's portal, edit one of the preferences that is attached to the listing (ie change the option text)
10. Back on the public side, click empezar for the last time and you'll see the updated translations for the preference question

sarahlazarich:
yay! this lgtm

Issue #4080, opened 2024-05-09
Application Translated Text Fallback


ColinBuyck:
It was discovered that when there isn't a translation for a preference option, it shows as blank rather than falling back to the english text. Preference options along with any other translated text on the application should always fallback since blank options are confusing and can block the user from progressing through the application.

Technical Notes:
- Fetch response shows untranslated text exists but text field doesn't exist

QA: See #4081

sarahlazarich:
Qa on other one passed

PR #4105, opened 2024-05-22
fix: fallback and refresh translations

ColinBuyck:
# Pull Request Template

## Issue Overview

This PR addresses #4080 and #4081

## Description

This PR does two things:
1) Creates a fallback in case there is no translation for a string for any reason
2) Adds logic to the translation service to check if any updates have been made to the listing or the associated preferences and will fetch fresh translations if those updates were made after the most recent translation entry.

Since the second fix above should prevent the use of the fallback, testing requires a bit of commenting and uncommenting.

1. Reseed on core.
2. Find a listing with an online application and without preferences (ie. Blue Sky Apartments)
3. Click Apply online
4. Open the network tab while looking at the choose language page
5. Click "Empezar"
6. In the network tab you should see the listing id and clicking it with preview, the response.
7. Click back and you should be on the choose language page again
8. Comment out line 205 to 225 in translation service

9. On the partner's portal, go to that listing and add a preference.
10. Return to the choose langauge tab and click empezar again.
11. This time in the network tab, you'll notice both the text field and untranslatedText fields are filled out with english for the multiselect question options. Previously, the text field would be missing resulting in blank fields for the user and a blocked application.
12. You can go through the application to ensure the preferences questions show and do not block the user
13. Go back to the choose-language of the same listing.
14. Uncomment out line 205 to 225 in translation service
15. Click empezar and see that the options in the network response are translated in the text field.
16. Go back once again to the choose-language step
17. On the partner's portal, edit one of the preferences that is attached to the listing (ie change the option text)
18. Back on the public side, click empezar for the last time and in the network tab you'll see the updated translations

Feel free to ping me though if a walkthrough of this would be easier!

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:


mcgarrye (approved):

api/src/services/translation.service.ts
```diff
@@ -192,9 +194,33 @@ export class TranslationService {
...
-      where: { listingId: listing.id, language: language },
+    const res = await this.prisma.generatedListingTranslations.findFirst({
+      where: {
+        listingId: listing.id,
+        language: language,
+      },
+    });
+
+    //determine if listing or associated preferences have changed since translation creation
+    let mostRecentUpdatedAt = listing.updatedAt;
```
YazeedLoonat, line 205: I think an issue we are going to run into is that when an application is created for a listing, the listing gets an update to one of its fields, and that'll change the updatedAt value as well.

I think this'll create a cycle where every application will cause listings to get re-translated 

I think we can use this and only redo translations if its been at least 1 day difference on line 213

YazeedLoonat (changes requested):
Looks good, but I think this creates a cycle we want to avoid

api/src/services/translation.service.ts
ColinBuyck: Hmmm I tried adding a paper app and applying online and both times, the updatedAt field was not changed. Do you see the opposite when you run this locally?

ColinBuyck: Per offline discussion, I've updated this logic to fetch new translations if the existing translations are a day old or if there has been a change to the preferences associated with the listing since the last translations. Lmk how that looks to you @YazeedLoonat

api/src/services/translation.service.ts
```diff
@@ -192,9 +194,39 @@ export class TranslationService {
...
+    listing.listingMultiselectQuestions?.forEach((multiselectObj) => {
+      const multiselectUpdatedAt =
+        multiselectObj.multiselectQuestions?.updatedAt;
+      if (mostRecentPreferenceUpdate < multiselectUpdatedAt) {
+        mostRecentPreferenceUpdate = multiselectUpdatedAt;
+      }
     });
+    // refresh translations if preference recently updated or translations are a day old
+    const yesterday = new Date();
+    yesterday.setDate(yesterday.getDate() - 1);
```
emilyjablonski, line 215: Totally understand I think there was some offline convo about this, but it seems like this would result in us re-translating more often than we would need to?

ColinBuyck: Yeahhh I share that worry. It seems like the most sustainable solution is to shift the way we handle translations to be triggered within the listing service (ie. on update) as opposed to when the user selects a language for the application. Feel free to add additional context @YazeedLoonat but he was also making the point that google translations are updated frequently and this temporary solution would help to provide the best experience for the user and that the added cost is worth that benefit. At this point, I think it'd be best to loop in product to get their thoughts on this question.

api/src/services/translation.service.ts
```diff
@@ -192,9 +194,39 @@ export class TranslationService {
...
-      where: { listingId: listing.id, language: language },
+    const res = await this.prisma.generatedListingTranslations.findFirst({
+      where: {
+        listingId: listing.id,
+        language: language,
+      },
+    });
+
+    //check if associated preferences have changed since translation creation
+    let mostRecentPreferenceUpdate = new Date(2000, 1);
```
ludtkemorgan, line 205: Curious why Feb 1st in the year 2000 was chosen? would `new Date(0)` not work?

api/src/services/translation.service.ts
ludtkemorgan: I also think we might want to discuss this more. I don't have all of the context, but getting new translations every day seems like it could be more excessive than need be

ColinBuyck:
@emilyjablonski @YazeedLoonat @ludtkemorgan @mcgarrye Made those updates we discussed in Core Sync, let me know what you all think of this approach 🌾

api/prisma/migrations/12_add_content_updated_at/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- Add field to capture most recent listing content update time
+ALTER TABLE "listings" ADD COLUMN "content_updated_at" TIMESTAMPTZ(6);
```
mcgarrye, line 2: nit EOF

api/src/services/translation.service.ts
```diff
@@ -192,9 +194,34 @@ export class TranslationService {
     listing: Listing,
     language: LanguagesEnum,
   ) {
-    return this.prisma.generatedListingTranslations.findFirst({
-      where: { listingId: listing.id, language: language },
+    const res = await this.prisma.generatedListingTranslations.findFirst({
```
mcgarrye, line 197: nit: can we give this a more descriptive name?

api/src/services/translation.service.ts
```diff
@@ -192,9 +194,34 @@ export class TranslationService {
...
+    });
+    //refresh translations if application content changed since translation creation
+    if (res && res.createdAt < mostRecentUpdate) {
+      await this.prisma.generatedListingTranslations.delete({
+        where: {
+          id: res.id,
+        },
+      });
+      console.log('get new translations');
+      return undefined;
```
mcgarrye, line 221: Does something need to be called? How does the translation update occur?

ColinBuyck: Good q! It's called on [line 158](https://github.com/bloom-housing/bloom/blob/869697adc070b8ccc3c21bf29b0f5b1fbe1dc690/api/src/services/translation.service.ts#L158) and if it is defined we use them but if it is undefined we fetch new translations from the google translation service

ColinBuyck:
Thanks for the review @mcgarrye! Believe I've addressed everything in the most recent commit 🦃

mcgarrye (approved):
LTGM

emilyjablonski (approved):
🩰

Issue #4119, opened 2024-06-04
Public users can log into any jurisdictional site


ColinBuyck:
Specifically in HBA, an alameda public user is able to log in to the san jose site and see applications from alameda in the san jose portal. These changes are important to have in core though to support the multiple public site model in the future.

Notes:
- These checks should include updating the checks of the forgot password flow. Likely in the isUserSiteMatch logic
- Remember to account for the case where multiple jurisdictions are captured in one site (ie Doorway)

Issue #4089, opened 2024-05-16
Prevent Forgot Password Workaround


ColinBuyck:
The forgot password flow allows for public users to get surface level info on the partner side (just listing names) and partner users on to the partner side. We block these users on the sign in flow so it should be consistent.

sarahlazarich:
I want to confirm that I took the right steps here before closing - 

- Made a public site account on https://bloom.exygy.dev/sign-in 
- Tried to log in with that accouny at https://partners.bloom.exygy.dev/
- Got an "Oops! Something went wrong" Error 
- Tried to reset my password from partners using that account, got no email

ColinBuyck:
Sorry for not putting QA Notes! That is how the public account flow should be tested but this should also be tested with a partner account trying to use the forgot password on the public side. Also a note that and this PR only addresses the forgot password part of the flow so if there is anything you notice while trying to log in/error messaging, that would be a separate issue.

PR #4120, opened 2024-06-04
fix: looser isUserSiteMatch

ColinBuyck:
# Pull Request Template

## Issue Overview

This PR addresses #4089 

## Description

This PR incorporates some points that @mcgarrye came up in discussing the previous [PR](https://github.com/bloom-housing/bloom/pull/4090) addressing this issue. Specifically in HBA, an alameda public user is able to log in to the san jose site and see applications from alameda in the san jose portal. Matching fully on the public url from the users jurisdiction would mean the forgot password experience is incorporating security that users aren't experiencing on the sign in page itself leading to a confusing experience. This issue is captured here #4119 

As a result of this, this PR just checks that a public user is not using the partners portal. **Additionally, this PR updates the env and env template to just include http://localhost:3001 to actually match the window.location.origin checks along with our staging and production sites who's urls do not include this trailing slash.**

**Update your env to match the template**

This can be tested by making a public user and partner user and with each account following the forgot password flow on the public and partners account. The flow should work as expected on the matching site but should not send an email for the mismatch.

## Checklist:

## Reviewer Notes:

Steps to review a PR:

## On Merge:

### DIFF
--- a/api/src/services/user.service.ts
+++ b/api/src/services/user.service.ts
@@ -407,8 +407,7 @@ export class UserService {
       storedUser.userRoles?.isPartner;
     const isUserSiteMatch =
       (isPartnerPortalUser && dto.appUrl === process.env.PARTNERS_PORTAL_URL) ||
-      (!isPartnerPortalUser &&
-        dto.appUrl === storedUser.jurisdictions?.[0]?.publicUrl);
+      (!isPartnerPortalUser && dto.appUrl !== process.env.PARTNERS_PORTAL_URL);
     // user on wrong site, return neutral message and don't send email
     if (!isUserSiteMatch) return { success: true };
 
--- a/api/.env.template
+++ b/api/.env.template
@@ -31,7 +31,7 @@ TWILIO_ACCOUNT_SID=
 # account auth token for twilio
 TWILIO_AUTH_TOKEN=
 # url for the partner front end
-PARTNERS_PORTAL_URL=http://localhost:3001/
+PARTNERS_PORTAL_URL=http://localhost:3001
 # sendgrid email api key
 EMAIL_API_KEY=SG.ExampleApiKey
 # controls the repetition of the afs cron job


ColinBuyck:
(Requested two reviews to ensure I'm not missing any potential side effects of these logic and env updates)

api/src/services/user.service.ts
```diff
@@ -407,8 +407,7 @@ export class UserService {
       storedUser.userRoles?.isPartner;
     const isUserSiteMatch =
       (isPartnerPortalUser && dto.appUrl === process.env.PARTNERS_PORTAL_URL) ||
-      (!isPartnerPortalUser &&
-        dto.appUrl === storedUser.jurisdictions?.[0]?.publicUrl);
+      (!isPartnerPortalUser && dto.appUrl !== process.env.PARTNERS_PORTAL_URL);
```
emilyjablonski, line 410: Still pondering if there are potential security issues here, where the appUrl doesn't need to be the/a publicUrl, it can be anything, as long as it's not the partners URL. I'm not sure I'm following the downside of a tighter match, if we were to .find on all of a user's jurisdictions

ColinBuyck: The issue we'd run into here is that on hba a user is only connected to the jurisdiction that they created their account with but they can still log in to all of the sites. That whole situation is captured in https://github.com/bloom-housing/bloom/issues/4119 but I figured we'd want to handle that at once with some investigation around how common it is.

emilyjablonski: Ohhhh interesting. Even from a UX standpoint, I think as a public user, I would expect that I would have to create different accounts for each public portal, but we can't really do that as is 🤔 Could we maybe check if it is equal to any of the appUrls on our stored jurisdictions?

YazeedLoonat (changes requested):
If this is a complex ask we can skip, but I think it would help me understand if you added a few unit tests for the different cases we could run into here 

like admin + logging into partner portal vs public user + public portal if that makes sense

ColinBuyck:
Let me know if there are other tests you'd like to see than those in the previous PR @YazeedLoonat: https://github.com/bloom-housing/bloom/pull/4090/files

api/src/services/user.service.ts
ColinBuyck: Ooo I really like that idea! My only hesitancy is that that data isn't available so it would require hitting the db again. Do you have any worries on the tradeoff between security and performance here @YazeedLoonat? If not, I'll make that update!

YazeedLoonat: the jurisdiction table is pretty small so look ups against it should be quick I don't think it'd be a problem to do that!

ColinBuyck: Just updated, thanks for the feedback y'all 🌿

emilyjablonski (approved):
🐈

mcgarrye (commented):
I am not receiving a forgot password on the partners portal with a partner account. Attempted using a confirmed admin account and a confirmed partner account

mcgarrye (approved):
LGTM

ColinBuyck:
@YazeedLoonat Just wanted to check with you pre-merge since you weighed in and requested changes ⏹️

PR #4121, opened 2024-06-06
fix: saving user account info

emilyjablonski:
This PR addresses https://github.com/metrotranscom/doorway/issues/631

## Description

This PR fixes two bugs on the user account screen:

1. Edit a field on the user account page, visit another page like the dashboard, go back in the browser and your change was not there
2. Edit a field on the user account page, edit another field, refresh, and only the second change persisted

It seemed like the problem with the first bug was that we are setting the form field defaults if the user profile is present, and when you go back in the browser, the profile is present and wasn't refetching, so I am fetching the user on page load, and after every user update I am re-setting the user object. Because I'm fetching on page load and that data is required in the UI, I pulled in our loading overlay over the card.

We were also having validation issues because we have separate forms on the page but we were not actually setting up the forms as separate as react-hook-forms expects - validation for all fields was triggering on submitting just one of them - so I separated these out into four separate forms as well.

Follow the steps above and ensure they work!

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/cypress/e2e/account/account.spec.ts
+++ b/sites/public/cypress/e2e/account/account.spec.ts
@@ -1,0 +1,63 @@
+import { publicUser, updatedPublicUser } from "../../mockData/userData"
+
+describe("User accounts", () => {
+  it("should allow users to update their account information", () => {
+    cy.visit("/sign-in")
+    cy.signIn(publicUser.email, publicUser.password)
+
+    // Assert that a user's data shows up on their settings page
+    cy.getByID("account-dashboard-settings").click()
+    cy.getByTestId("loading-overlay").should("not.exist")
+    cy.getByTestId("account-first-name").should("have.value", publicUser.firstName)
+    cy.getByTestId("account-middle-name").should("have.value", publicUser.middleName)
+    cy.getByTestId("account-last-name").should("have.value", publicUser.lastName)
+    cy.getByTestId("dob-field-month").should("have.value", publicUser.birthMonth)
+    cy.getByTestId("dob-field-day").should("have.value", publicUser.birthDay)
+    cy.getByTestId("dob-field-year").should("have.value", publicUser.birthYear)
+    cy.getByTestId("account-email").should("have.value", publicUser.email)
+
+    // Change the name fields
+    cy.getByTestId("account-first-name").clear().type(updatedPublicUser.firstName)
+    cy.getByTestId("account-middle-name").clear().type(updatedPublicUser.middleName)
+    cy.getByTestId("account-last-name").clear().type(updatedPublicUser.lastName)
+    cy.getByID("account-submit-name").click()
+    cy.getByTestId("alert-box").contains("Name update successful")
+    cy.get("[aria-label='close alert']").click()
+    cy.getByTestId("alert-box").should("not.exist")
+
+    // Change the birthday field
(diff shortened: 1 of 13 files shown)


sites/public/cypress/e2e/account/account.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { publicUser, updatedPublicUser } from "../../mockData/userData"
```
emilyjablonski, line 1: We can't test changing the email here because it requires email verification, but there is this issue I logged with the email flow: https://github.com/bloom-housing/bloom/issues/4122

sites/public/src/components/shared/FormSummaryDetails.tsx
```diff
@@ -291,69 +289,66 @@ const FormSummaryDetails = ({
...
-              {editMode && !validationError && (
-                <Link href="/applications/contact/alternate-contact-type">{t("t.edit")}</Link>
-              )}
-            </Card.Header>
-
-            <Card.Section className={styles["summary-section"]}>
-              <p className={styles["summary-note-text"]}>
-                {t(`application.alternateContact.type.description`)}
-              </p>
+      {application.alternateContact.type && application.alternateContact.type !== "noContact" && (
```
emilyjablonski, line 292: Apart from typing this, this whole file is prettier changes

sites/public/src/pages/account/edit.tsx
```diff
@@ -72,9 +111,10 @@ const Edit = () => {
     const { firstName, middleName, lastName } = data
     setNameAlert(null)
     try {
-      await userService.update({
-        body: { ...profile, firstName, middleName, lastName },
+      const newUser = await userService.update({
```
mcgarrye, line 114: nit: `updatedUser` instead of `newUser`

mcgarrye (approved):
Looks good!

sites/public/src/pages/account/edit.tsx
```diff
@@ -48,10 +76,12 @@ const Edit = () => {
   const [birthdateLoading, setBirthdateLoading] = useState(false)
   const [emailLoading, setEmailLoading] = useState(false)
   const [passwordLoading, setPasswordLoading] = useState(false)
+  const [user, setUser] = useState(null)
+  const [loading, setLoading] = useState(true)
 
-  const MIN_PASSWORD_LENGTH = 8
+  const MIN_PASSWORD_LENGTH = 12
```
ludtkemorgan, line 82: This would be a good value to have pulled in from environment variable in the future. Also, are all jurisdictions ok with increasing the min password length? If so do we have the right language listed everywhere? And also this would make the min password length different between public and partner accounts.

ludtkemorgan (approved):
Love the new test! The change looks good. Just have one question on the password min-length

sites/public/src/pages/account/edit.tsx
emilyjablonski: Good q! We updated our password regex and security measures a bit ago to 12 characters, and one upper / one lower / one number / one special character, so this was a miss. BUT I do agree it should be pulled from one place.

emilyjablonski: Updated both to 12 🫡

ludtkemorgan (commented):
This change looks good. I'm guessing the issue I found is not related to this change but we will want to get that fixed before the release today

PR #4128, opened 2024-06-11
fix: switch to HTML text inputs with numeric validation

jaredcwhite:
This PR addresses https://github.com/metrotranscom/doorway/issues/658

## Description

**Note:** [see comment below](https://github.com/bloom-housing/bloom/pull/4128#issuecomment-2197522427)

The working theory is the issue with numbers changing randomly in the unit editing is due to `input type="number"` weirdness around mouse/trackpad scrolling. This PR switches to `input type="text"` and uses React Hook Form validation to verify integer input.

Try editing units for a listing, income, or the miles radius on preferences.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/components/settings/PreferenceDrawer.tsx
+++ b/sites/partners/src/components/settings/PreferenceDrawer.tsx
@@ -704,10 +704,10 @@ const PreferenceDrawer = ({
                           name="radiusSize"
                           label={t("settings.preferenceValidatingAddress.howManyMiles")}
                           register={register}
-                          validation={{ required: true, min: 0 }}
+                          validation={{ required: true, min: 0, pattern: /^[0-9.]+$/ }}
                           error={errors.radiusSize}
                           errorMessage={t("errors.requiredFieldError")}
-                          type="number"
+                          type="text"
                           readerOnly
                           defaultValue={optionData?.radiusSize ?? null}
                           dataTestId={"preference-option-radius-size"}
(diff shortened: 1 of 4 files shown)


jaredcwhite:
Cypress test failing seems to just be a glitch

sites/partners/src/components/settings/PreferenceDrawer.tsx
```diff
@@ -700,10 +700,10 @@ const PreferenceDrawer = ({
                         name="radiusSize"
                         label={t("settings.preferenceValidatingAddress.howManyMiles")}
                         register={register}
-                        validation={{ required: true, min: 0 }}
+                        validation={{ required: true, min: 0, pattern: /^[0-9.]+$/ }}
```
emilyjablonski, line 703: What might min do here now that it's text?

emilyjablonski (changes requested):
After this change, I'm able to type letters into these numeric inputs, which I think we ideally want to avoid.

jaredcwhite:
I've updated the PR with a combination of currency-specific fields and text fields with numeric validation. Only the currency fields expressively forbid typing letters, and they also have the byproduct of forcing decimal places (aka `500.00` rather than `500`) which I'm concerned by cause trickle-down issues with display in other places. So…I'm not entirely sure we should consider this mergable without a broader convo.

mcgarrye:
Is this PR still valid? If so, can we get it caught up to main and I can review Monday

PR #4130, opened 2024-06-11
fix: allow removal of mailing address

ludtkemorgan:
This PR addresses [#638](https://github.com/metrotranscom/doorway/issues/638)

## Description

When updating a listing and removing the mailing, pickup, or drop-off address the existing address was still sticking around. This is happening due to the Prisma change over. Specifically because the field names of those were updated in Prisma and the new change was missed in one place in the partner site. This also exposed an issue with how we were deleting addresses, the address needs to be disconnected from the listing before it can be deleted from the listing table.

The following steps can be tested:

1. Add "another" mailing address, pick-up address, and drop-off address
image
2. Save the listing.

Both of these use cases should be tested for all three fields:
1. Select "at the leasing agent address" and then re-save the listing
image
2. Remove the address option altogether by selecting "No" on the above question and then re-save the listing
image

When reloading the listing the original saved listing from step one should be removed

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/listing.service.ts
+++ b/api/src/services/listing.service.ts
@@ -982,6 +982,19 @@ export class ListingService implements OnModuleInit {
           },
         });
       } else {
+        // in order to delete the old address we first need to disconnect it from the listing
+        await this.prisma.listings.update({
+          data: {
+            [field]: {
+              disconnect: {
+                id: existingListing[field].id,
+              },
+            },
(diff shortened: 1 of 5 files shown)


emilyjablonski (approved):
LGTM! Do you mind adding QA notes to the issue? 🧀

Issue #639, opened 2020-09-11
Listing: Add Lottery Results State to Listings


slowbot:
Goal: As a user I want to know the results of the application process

Acceptance Criteria:
- After applications close listings should have three states.
   - **Awaiting results** (Replaces Closed listings)
      - This is the period of time it takes leasing agents to resolve any duplicate applications and run a lottery if there is one.
      - **Pre Lottery**: Sidebar provides the date applications closed, if there is a lottery the date and time of the lottery is posted as well.
      - **Post Lottery**: Sidebar provides the date applications closed and information on when the results will be posted. 
   - **Results posted**
      - Once ranking order is set the results are available on the Portal 
      - Sidebar provides a link to "Download Results". Download is a list of ranked applications with confirmation numbers. If there is a lottery the list displays lottery rank.

![image](https://user-images.githubusercontent.com/13915/106067620-ec038c00-60b3-11eb-9732-3bc9105c42ea.png)
![image](https://user-images.githubusercontent.com/13915/106067418-7ac3d900-60b3-11eb-9b48-fcb7c01f6c4f.png)
![image](https://user-images.githubusercontent.com/13915/106178397-9203d480-614e-11eb-98c0-8cfa9b143b1d.png)
![image](https://user-images.githubusercontent.com/13915/106067445-88795e80-60b3-11eb-83c1-e20d9269cfd8.png)

ludtkemorgan:
@slowbot @sarahlazarich is this still valid?

sarahlazarich:
no, closing

PR #4134, opened 2024-06-11
fix: add error message for incorrect listing approval

ludtkemorgan:
This PR addresses [#639](https://github.com/metrotranscom/doorway/issues/639)

## Description

When submitting a listing for approval not every field needs to be filled out. This can cause a side effect of when an admin approves the listing they then have to resolve the missing fields. The issue is made even worse if the admin chooses to approve from the details view rather than the edit view as only a generic error message happens (which also was broken). As a remediation this PR adds an additional error banner on the listing details view when this error happens so that the user knows the next step that needs to happen 

- As a partner user submit a listing that is missing at least one required field.
- Log in as an admin and click the "Approve & Publish" button for that listing (note: not on the edit view, but on the page that shows when first going to a listing in partners)
- New error message should show

## Author Checklist:

- [N/A] Reviewed in a mobile view

## Review Process:


mcgarrye (approved):
Nice!

Issue #4137, opened 2024-06-12
Move Cypress application tests to unit tests


emilyjablonski:
We've had a series of Cypress tests on the public site for a long time that visit every page of an application in isolation out of actually applying, and check to see if the text and fields load, if the fields are required before moving on, and occasionally some other smaller tests on the way the form fields work.

These aren't really Cypress tests - these are conceptually closer to unit tests, and we already have two tests that go through the entire application flow.

This issue is to delete those Cypress tests but maintain the coverage by adding them all as unit tests.

The reason I did this now, is because the release for the https://github.com/metrotranscom/doorway/pull/670 has required that we turn on mandated accounts in the Cypress suite in Doorway, but when mandated accounts is on, you actually can't visit the application pages directly outside of the context of submitting anymore, because you need to be logged in (and logging in doesn't work either).

I was spending a bit of time trying to get these tests to pass in Doorway, but it felt somewhat useless knowing that we've always wanted to delete them. I didn't want to just delete them though and lose that coverage right before applications go live.

QA Notes
There is nothing to QA on the product side

PR #4138, opened 2024-06-12
test: move application page cypress tests to unit tests

emilyjablonski:
This PR addresses #4137

## Description

We've had a series of Cypress tests on the public site for a long time that visit every page of an application in isolation out of actually applying, and check to see if the text and fields load, if the fields are required before moving on, and occasionally some other smaller tests on the way the form fields work.

These aren't really Cypress tests - these are conceptually closer to unit tests, and we already have two Cypress tests that go through the entire application flow.

This PR deletes those Cypress tests but maintains the coverage by adding them all as unit tests.

The reason I did this now, is because the release for the [account saving bug in Doorway](https://github.com/metrotranscom/doorway/pull/670) has required that we turn on mandated accounts in the Cypress suite in Doorway, but when mandated accounts is on, you actually can't visit the application pages directly outside of the context of submitting anymore, because you need to be logged in (and logging in doesn't work either).

I was fighting a lil trying to get these tests to pass in Doorway, but it felt somewhat pointless knowing that we've always wanted to delete them. I didn't want to just delete them though and lose that coverage right before applications go live.

With the Cypress removals and a config adjustment on the public/partners/shared-helpers unit tests, here's how the job times changed:
Cypress public 10m --> 7m
Unit public 10m --> 3m
Unit partners 12m --> 2m
Unit shared-helpers 10m --> 2m

You can check out the new public site unit tests!

## Author Checklist:

## Review Process:


sites/partners/jest.config.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   globals: {
     "ts-jest": {
       tsconfig: "tsconfig.test.json",
+      isolatedModules: true,
```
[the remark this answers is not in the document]
emilyjablonski: https://huafu.github.io/ts-jest/user/config/isolatedModules

Compiles the test files in isolation, instead of doing full type-checking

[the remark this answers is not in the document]
mcgarrye: Coooolllll

sites/public/jest.config.js
```diff
@@ -5,7 +5,16 @@ process.env.TZ = "UTC"
...
+  collectCoverageFrom: ["**/*.ts", "**/*.tsx"],
+  coveragePathIgnorePatterns: [
+    "cypress",
+    "types",
+    "__tests__",
+    "next-env.d.ts",
+    "sentry.client.config.ts",
+    "sentry.server.config.ts",
+    "sentry.edge.config.ts",
+  ],
```
mcgarrye, line 17: Should this be added to partners and shared-helpers?

mcgarrye (approved):
LGTM!

sites/public/jest.config.js
emilyjablonski: These files are specific to the public site, and looks like code cov is already setup on partners/shared-helpers

sites/public/__tests__/pages/applications/contact/address.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+      expect(getByTestId("app-primary-additional-phone")).toBeInTheDocument()
+      expect(getByTestId("app-primary-address-street")).toBeInTheDocument()
+      expect(getByTestId("app-primary-address-street2")).toBeInTheDocument()
+      expect(getByTestId("app-primary-address-city")).toBeInTheDocument()
+      expect(getByTestId("app-primary-address-state")).toBeInTheDocument()
+      expect(getByTestId("app-primary-address-zip")).toBeInTheDocument()
+      expect(getByTestId("app-primary-send-to-mailing")).toBeInTheDocument()
+      expect(getAllByTestId("app-primary-contact-preference")).toHaveLength(4)
+      expect(getByTestId("app-primary-work-in-region-yes")).toBeInTheDocument()
+      expect(getByTestId("app-primary-work-in-region-no")).toBeInTheDocument()
```
ludtkemorgan, line 46: getByTestId works correctly. However, to get the most accurate experience a user would have, especially when considering accessibility it should be the last choice. https://testing-library.com/docs/queries/about/#priority. For form fields we should be using `getByLabelText`. For example instead of ` expect(getByTestId("app-primary-phone-number-type")).toBeInTheDocument()` we could do `expect(getByLabelText("What type of number is this?")).toBeInTheDocument()`. That way we not only test that the field exist, but also that we are displaying the correct language and it has a proper label tag.

sites/public/__tests__/pages/applications/contact/address.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+      fireEvent.click(getByText("Next"))
+      expect(
+        await findByText("There are errors you'll need to resolve before moving on.")
+      ).toBeInTheDocument()
+      expect(getByText("Please enter a phone number")).toBeInTheDocument()
+      expect(getByText("Please enter a phone number type")).toBeInTheDocument()
+      expect(getByText("Please enter an address")).toBeInTheDocument()
+      expect(getByText("Please select at least one option.")).toBeInTheDocument()
+      expect(getByText("Please select one of the options above.")).toBeInTheDocument()
```
ludtkemorgan, line 60: Should there also be a part of this test to resolve the errors and re-click "Next" to make sure errors are removed properly?

ludtkemorgan (commented):
Love these changes. So glad to move these tests to unit tests. Also the faster tests is amazing!

I don't want my comments to block this getting into Doorway, but I do think we should probably get off using using the test id as much as possible. I'm open to merging this now and then quickly following up with updating the tests to the other matchers so that Doorway is not blocked. Just let me know what you think.

sites/public/__tests__/pages/applications/contact/address.test.tsx
emilyjablonski: Eventually that sounds great, it just wasn't a part of the current coverage and didn't want to bloat this PR

sites/public/__tests__/pages/applications/contact/address.test.tsx
emilyjablonski: Great point! Would love to fast-follow with that

sites/public/__tests__/pages/applications/contact/address.test.tsx
ludtkemorgan: Good call out. It's definitely not a blocker. I'll approve this PR and we can create some tickets for additional coverage

ludtkemorgan (approved):
🧪

PR #4141, opened 2024-06-17
fix: application export fix

YazeedLoonat:

## Description
This fixes an issue with the csv export of applications
specifically the preferences columns weren't building properly and this pr should set up to work as intended

Create a new preference that has the address verifications set
Add that preference to a listing
Submit an application to that listing
verify the export of that's applications looks correct

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/application-csv-export.service.ts
+++ b/api/src/services/application-csv-export.service.ts
@@ -201,8 +201,7 @@ export class ApplicationCsvExporterService
                         parsePreference = false;
                         // there aren't typically many preferences, but if there, then a object map should be created and used
                         const preference = preferences.find(
-                          (preference) =>
-                            preference.multiselectQuestionId === curr,
+                          (preference) => preference.key === curr,
                         );
                         multiselectQuestionValue = true;
                         return preference;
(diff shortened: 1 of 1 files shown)


emilyjablonski:
Can you link the issue, and is there a way to add tests that would have caught this?

YazeedLoonat:
Hey @emilyjablonski I added the issue thanks for the reminder! 

and for testing, I think that'd be a lil tricky the current way mostly just does it in lambda/arrow functions so testing the pieces would be kinda hard 

I'd say we don't add new tests for this, but instead look at moving the arrow functions into helper functions that can be unit testable

api/src/services/application-csv-export.service.ts
```diff
@@ -568,9 +567,17 @@ export class ApplicationCsvExporterService
...
           label: `Preference ${question.text}`,
-          format: (val: boolean): string => (val ? 'claimed' : ''),
+          format: (val: any): string => {
+            const claimedString: string[] = [];
+            val?.options?.forEach((option) => {
+              if (option.checked) {
+                claimedString.push(option.key);
+              }
+            });
+            return claimedString.length ? claimedString.join(', ') : '';
```
emilyjablonski, line 579: You don't need to check for length here, can just do the join

emilyjablonski:
I tested this with both preferences and programs to hit multiselect questions, and the programs didn't show up at all in the CSV

api/src/services/application-csv-export.service.ts
```diff
@@ -639,6 +656,83 @@ export class ApplicationCsvExporterService
           });
       });
 
+    // add programs to csv headers
```
emilyjablonski, line 659: A lot of this looks like it might be copy-pasta from above and just changing the `preferences` --> `programs`? Could be nice to share this piece

YazeedLoonat: you right, updated!

emilyjablonski (approved):
❄️ 🏂 ❄️

ludtkemorgan (approved):
Looks good!

Issue #4144, opened 2024-06-18
AFS job is failing to run


ludtkemorgan:
The AFS job is failing to run in all environments because of this error:
```ERROR AFS_CRON_JOB failed to run. TypeError: Cannot read properties of null (reading 'toLowerCase')```

This is due to either the first or last name of an applicant not existing.

ludtkemorgan:
This has been merged in core and the error is no longer in the logs. PR is ready to be merged in Doorway and HBA

sarahlazarich:
@ludtkemorgan is there product QA here?

sarahlazarich:
talked to ML, closing

PR #4145, opened 2024-06-18
fix: add null check to afs rule key

ludtkemorgan:
This PR addresses #4144 

## Description

The AFS job is failing in both bloom and Doorway due to paper applications that don't have a first or last name. This adds a null check to make sure the AFS job won't run into this error when running

Add a paper application that doesn't have first and last name to a property. Then close the listing and run the AFS job

## Author Checklist:

## Review Process:

### DIFF
--- a/api/test/unit/services/application-csv-export.service.spec.ts
+++ b/api/test/unit/services/application-csv-export.service.spec.ts
@@ -559,7 +559,8 @@ describe('Testing application CSV export service', () => {
     expect(readable).toContain(firstApp);
   });
 
-  it('should build csv with submission date defaulted to PST', async () => {
+  it('should build csv with submission date defaulted to TIME_ZONE variable', async () => {
+    process.env.TIME_ZONE = 'America/Los_Angeles';
     jest.useFakeTimers();
     jest.setSystemTime(new Date('2024-01-01'));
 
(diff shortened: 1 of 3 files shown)


emilyjablonski:
I think if an application doesn't have a first/last name we probably don't want to include it in the duplicate sets for first/last name at all, wdyt?

ludtkemorgan:
> I think if an application doesn't have a first/last name we probably don't want to include it in the duplicate sets for first/last name at all, wdyt?

@emilyjablonski I agree, I was worried to add additional complexity to this PR since that would be a change in functionality. However, since theoretically all applications should have a name associated it probably wouldn't be an issue to add that additional check

api/test/unit/services/application-csv-export.service.spec.ts
```diff
@@ -559,7 +559,8 @@ describe('Testing application CSV export service', () => {
     expect(readable).toContain(firstApp);
   });
 
-  it('should build csv with submission date defaulted to PST', async () => {
+  it('should build csv with submission date defaulted to TIME_ZONE variable', async () => {
+    process.env.TIME_ZONE = 'America/Los_Angeles';
```
ludtkemorgan, line 563: This makes this test not reliant on the TIMEZONE set in the local env file.

ludtkemorgan:
@emilyjablonski I have updated this PR to not create an afs rule if the first or last name doesn't exist

emilyjablonski (approved):

Issue #4055, opened 2024-04-28
Lotteries - Export with stored general rank ordinal (BE)


emilyjablonski:
As a user running the lottery on Doorway, 
I want to be able to generate a randomized rank for each valid application, and export into a csv
So that I can use this spreadsheet to review lottery results. 

Notes: 
- This should follow the format of the ["General Rank" tab](https://docs.google.com/spreadsheets/d/1dX9ehAq3pdMpmhNq1arWV2bwjHBFMy6xAbcPGhLVHOY/edit#gid=520192447)
- Final form of this export will be a spreadsheet, but we will address that in a later ticket 
- This is just for the raw rank generation + tab in the export. The preference filtering will happen in a [later ticket. ](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4056)
- As a part of this ticket, we should store the lottery run timestamp for areas where we need to store an indication of where the lottery was run [like this one](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4047)
- The file generation tickets should also include the connection to the front end, 

AC: 
- When I generate a lottery, I want each applicant to be assigned a random "raw rank"
- I want applications marked as duplicated to be excluded from the ranking / lottery results
- I want the results of the randomization to be stored (so that they can be surfaced in the applicant account eventually)
- I want to be able to export my results into a spreadsheet (csv for now) by clicking the "export" button from the lottery tab. 
- I want the timestamp of the lottery run to be stored as well 

**QA NOTES**
- create applications on a listing (or use a listing with existing applications)
- navigate to the listings' lottery tab
- you should be able to click the run lottery button and lottery results should get generated
- you should be able to click the export results button and get a csv spreadsheet with the lottery results

sarahlazarich:
Questions: 
https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/3728
https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/metrotranscom/doorway/449

Can we consolidate and close these?

sarahlazarich:
For this ticket, just do a CSV - Emily has spreadsheet stuff into a separate ticket

sarahlazarich:
Duplicates shouldn't be assigned ordinals 

For design: 
- for duplicates, what would an applicant see in their account?

[...]

PR #4159, opened 2024-06-25
feat: adding randomization tooling to export service

YazeedLoonat:
This PR addresses https://github.com/bloom-housing/bloom/issues/4055

## Description
This adds the randomizer for lottery to the csv export 
it doesn't expose a way to do this, but there are unit tests to demo how it'll work

the endpoints and stuff are coming in a part 2, this is just the randomization logic and unit tests 

## Author Checklist:

## Review Process:

### DIFF
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -222,7 +222,7 @@ model Applications {
   demographics                 Demographics?                 @relation(fields: [demographicsId], references: [id], onDelete: NoAction, onUpdate: NoAction)
   preferredUnitTypes           UnitTypes[]
   householdMember              HouseholdMember[]
-  ApplicationLotteryPositions  ApplicationLotteryPositions[]
+  applicationLotteryPositions  ApplicationLotteryPositions[]
 
   @@unique([listingId, confirmationCode])
   @@index([listingId])
@@ -585,7 +585,7 @@ model Listings {
   requestedChangesDate                  DateTime?                     @default(dbgenerated("'1970-01-01 00:00:00-07'::timestamp with time zone")) @map("requested_changes_date") @db.Timestamptz(6)
   requestedChangesUserId                String?                       @map("requested_changes_user_id") @db.Uuid
   requestedChangesUser                  UserAccounts?                 @relation("requested_changes_user", fields: [requestedChangesUserId], references: [id], onDelete: NoAction, onUpdate: NoAction)
-  ApplicationLotteryPositions           ApplicationLotteryPositions[]
+  applicationLotteryPositions           ApplicationLotteryPositions[]
 
   @@index([jurisdictionId])
   @@map("listings")
@@ -625,7 +625,7 @@ model MultiselectQuestions {
   applicationSection          MultiselectQuestionsApplicationSectionEnum @map("application_section")
   jurisdictions               Jurisdictions[]
   listings                    ListingMultiselectQuestions[]
-  ApplicationLotteryPositions ApplicationLotteryPositions[]
+  applicationLotteryPositions ApplicationLotteryPositions[]
 
   @@map("multiselect_questions")
 }
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -222,6 +222,7 @@ model Applications {
   demographics                 Demographics?                 @relation(fields: [demographicsId], references: [id], onDelete: NoAction, onUpdate: NoAction)
   preferredUnitTypes           UnitTypes[]
   householdMember              HouseholdMember[]
+  ApplicationLotteryPositions  ApplicationLotteryPositions[]
 
   @@unique([listingId, confirmationCode])
   @@index([listingId])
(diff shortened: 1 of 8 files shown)


api/prisma/schema.prisma
```diff
@@ -597,19 +612,20 @@ model Migrations {
...
+  subText                     String?                                    @map("sub_text")
+  description                 String?
+  links                       Json?
+  options                     Json?
+  optOutText                  String?                                    @map("opt_out_text")
+  hideFromListing             Boolean?                                   @map("hide_from_listing")
+  applicationSection          MultiselectQuestionsApplicationSectionEnum @map("application_section")
+  jurisdictions               Jurisdictions[]
+  listings                    ListingMultiselectQuestions[]
+  ApplicationLotteryPositions ApplicationLotteryPositions[]
```
mcgarrye, line 628: nit: capitalized when it shouldn't be

api/prisma/schema.prisma
```diff
@@ -571,6 +585,7 @@ model Listings {
   requestedChangesDate                  DateTime?                     @default(dbgenerated("'1970-01-01 00:00:00-07'::timestamp with time zone")) @map("requested_changes_date") @db.Timestamptz(6)
   requestedChangesUserId                String?                       @map("requested_changes_user_id") @db.Uuid
   requestedChangesUser                  UserAccounts?                 @relation("requested_changes_user", fields: [requestedChangesUserId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  ApplicationLotteryPositions           ApplicationLotteryPositions[]
```
mcgarrye, line 588: nit: Same here

api/prisma/schema.prisma
```diff
@@ -222,6 +222,7 @@ model Applications {
   demographics                 Demographics?                 @relation(fields: [demographicsId], references: [id], onDelete: NoAction, onUpdate: NoAction)
   preferredUnitTypes           UnitTypes[]
   householdMember              HouseholdMember[]
+  ApplicationLotteryPositions  ApplicationLotteryPositions[]
```
mcgarrye, line 225: Here too

api/prisma/schema.prisma
```diff
@@ -424,6 +425,19 @@ model ListingMultiselectQuestions {
   @@map("listing_multiselect_questions")
 }
 
+model ApplicationLotteryPositions {
+  ordinal               Int
+  listingId             String                @map("listing_id") @db.Uuid
+  listings              Listings              @relation(fields: [listingId], references: [id], onDelete: Cascade, onUpdate: NoAction)
```
mcgarrye, line 431: General question: why is this pluralized? Is it returning a list of just one listing?

api/src/services/application-csv-export.service.ts
mcgarrye, line 1: IMO I feel like the additions to this file could live in their own lottery helper file

mcgarrye (approved):
Actual logic and code looks great! Left some comments about code location and property capitalization

api/prisma/schema.prisma
YazeedLoonat: yeaaa thats sort of a hold over/problem we have throughout our schema. I think it would be better to leave this as plural to match up with other similar models but we do have a backend TODO to clean up this plural/singular issues

api/src/services/application-csv-export.service.ts
YazeedLoonat: Good call out! I think there are 2 things here

1. I think we may end up create a new service and controller to do this work so I think it may be better to leave this here and see what ends up being the case when we attach this to an endpoint

2. In general its better to not have to pass the prisma service around too much, NestJS provides us with Dependency Injection and I think its better to keep services as the "holders" of the prisma service generally. In this case the `lotteryRandomizer()` does write the DB a couple of times so moving that to a helper I think wouldn't be the best call

YazeedLoonat:
Hey @mcgarrye made updates and left you some comments back, lemme know what you think

mcgarrye:
> Hey @mcgarrye made updates and left you some comments back, lemme know what you think

Looks great

ludtkemorgan (approved):
Functionally this looks good. My only concern is that, if I'm understanding the logic correctly, for every application there would be a record in this new table and if there are like 5 preferences selected there will also be five additional records in this table. I don't think this should be an issue, but we should be careful on the total number of rows especially in heroku

api/src/services/application-csv-export.service.ts
```diff
@@ -783,4 +784,102 @@ export class ApplicationCsvExporterService
...
+      let possibleRandomPosition =
+        Math.floor(Math.random() * numberOfUniqueApplications) + 1;
+      // ensure random value not already in array
+      while (ordinalArray.some((val) => val === possibleRandomPosition)) {
+        possibleRandomPosition =
+          Math.floor(Math.random() * numberOfUniqueApplications) + 1;
+      }
+      // push unique random value into array
+      ordinalArray.push(possibleRandomPosition);
+    });
```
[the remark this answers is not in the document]
YazeedLoonat: hey @ludtkemorgan you're right! I made updates if you wanna take another peek!

api/test/unit/services/application-csv-export.service.spec.ts
```diff
@@ -687,4 +689,162 @@ describe('Testing application CSV export service', () => {
 
     expect(readable).toContain('EST');
   });
+
+  it('should generate random ordinal arrays', () => {
```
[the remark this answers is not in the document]
YazeedLoonat: nice call out, I made moved just the new tests for now and we can discuss cleaning up our tests later

YazeedLoonat:
Hey @ludtkemorgan made some updates per your comments, if you wanna take a lil peek and then we can switch the label and merge this in!

api/src/services/application-csv-export.service.ts
```diff
@@ -863,21 +863,23 @@ export class ApplicationCsvExporterService
...
   lotteryRandomizerHelper(filterApplicationsArray: Application[]): number[] {
     // prep our supporting array
-    const numberOfUniqueApplications = filterApplicationsArray.length;
     const ordinalArray: number[] = [];
 
+    const indexArray: number[] = [];
+    filterApplicationsArray.forEach((_, index) => {
+      indexArray.push(index + 1);
+    });
+
```
ludtkemorgan, line 872: nit: instead of forEach you could do `map` and not have to do a push

ludtkemorgan (approved):
Looks great

PR #4162, opened 2024-06-26
fix: lottery rebase issues

emilyjablonski:
I ran into two issues with the Doorway release of the lottery empty state ([this core PR](https://github.com/bloom-housing/bloom/commit/4dbb6e3d10beeefa90798c904c5c797c497c8ee4))

* We weren't adding the API pass key on the lottery page
* The new lottery tab was missing from the applications page

These were added into the [Doorway release branch](https://github.com/metrotranscom/doorway/pull/706) already

### DIFF
--- a/sites/partners/src/components/shared/NavigationHeader.tsx
+++ b/sites/partners/src/components/shared/NavigationHeader.tsx
@@ -66,7 +66,7 @@ const NavigationHeader = ({
       },
     ]
 
-    if (process.env.showLottery && tabs?.lotteryLabel) {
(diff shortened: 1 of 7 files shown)


mcgarrye (commented):
Looks good but doesn't seem to be hidden by the feature flag

emilyjablonski:
@mcgarrye Are you seeing that while testing? In the Navigation component the tab is only added if the label exists and if the toggle is on

mcgarrye (approved):
Phenomenal

PR #4167, opened 2024-06-27
feat: lottery status (released to partners / retracted)

mcgarrye:
This PR addresses [#(4057)](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4057)

## Description

Create lottery status enums and status change endpoint. Setup business logic for releasing to partners and retracting actions.

To understand how the status flow works, check out this [Miro board](https://miro.com/app/board/uXjVK4sw5X4=/)

Run migration and run backend in the terminal. Use the swagger open api for the following flows.
Flow 1: Login as an admin, attempt to change the status of a active listing to releasedToPartners, you should receive an error since listing is not closed.
Flow 2: Login as an admin, attempt to change the status of a closed listing to releasedToPartners, you should receive an error since it must be in approved.
Flow 3: Change the status manually in the db to approved, login as an admin, attempt to change the status of a closed listing to releasedToPartners, confirm change was successful, attempt to change status to `ran` (this is retracting the lottery), confirm success.
Flow 4: Attempt flow 3 as a partner or juris admin, confirm you receive permission errors.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/controllers/listing.controller.ts
+++ b/api/src/controllers/listing.controller.ts
@@ -186,12 +186,12 @@ export class ListingController {
     summary: 'Change the listing lottery status',
     operationId: 'lotteryStatus',
   })
-  @ApiOkResponse({ type: Listing })
+  @ApiOkResponse({ type: SuccessDTO })
   @UseGuards(ApiKeyGuard)
   async lotteryStatus(
     @Request() req: ExpressRequest,
     @Body() dto: ListingLotteryStatus,
-  ): Promise<Listing> {
+  ): Promise<SuccessDTO> {
     return await this.listingService.lotteryStatus(
       dto,
       mapTo(User, req['user']),
--- a/api/src/controllers/listing.controller.ts
+++ b/api/src/controllers/listing.controller.ts
@@ -181,7 +181,7 @@ export class ListingController {
     return await this.listingService.process();
   }
 
-  @Put('lotteryStatus/:id')
+  @Put('lotteryStatus')
   @ApiOperation({
     summary: 'Change the listing lottery status',
     operationId: 'lotteryStatus',
--- a/api/src/controllers/listing.controller.ts
+++ b/api/src/controllers/listing.controller.ts
@@ -186,7 +186,7 @@ export class ListingController {
     summary: 'Change the listing lottery status',
     operationId: 'lotteryStatus',
   })
-  @ApiOkResponse({ type: SuccessDTO })
+  @ApiOkResponse({ type: Listing })
   @UseGuards(ApiKeyGuard)
   async lotteryStatus(
     @Request() req: ExpressRequest,
@@ -201,6 +201,7 @@ export class ListingController {
   @Put(':id')
   @ApiOperation({ summary: 'Update listing by id', operationId: 'update' })
   @UsePipes(new ListingCreateUpdateValidationPipe(defaultValidationPipeOptions))
+  @ApiOkResponse({ type: Listing })
   @UseGuards(ApiKeyGuard)
   async update(
     @Request() req: ExpressRequest,
--- a/api/src/controllers/listing.controller.ts
+++ b/api/src/controllers/listing.controller.ts
@@ -26,34 +26,35 @@ import {
   ApiOkResponse,
 } from '@nestjs/swagger';
 import { Request as ExpressRequest, Response } from 'express';
-import { ListingService } from '../services/listing.service';
-import { defaultValidationPipeOptions } from '../utilities/default-validation-pipe-options';
-import { ListingsQueryParams } from '../dtos/listings/listings-query-params.dto';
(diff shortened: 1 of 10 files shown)


mcgarrye:
Tests to come

api/src/services/listing.service.ts
```diff
@@ -1618,4 +1622,118 @@ export class ListingService implements OnModuleInit {
...
+      );
+    }
+
+    const isAdmin = requestingUser.userRoles?.isAdmin;
+    const currentStatus = storedListing.lotteryStatus;
+
+    // TODO: remove when all status logic has been implemented
+    let res;
+
+    switch (dto?.lotteryStatus) {
```
ColinBuyck, line 1657: Question: I think this is clean but I am curious about the team's preferences for switch statements versus just using if-logic (ie. listingApprovalNotify logic uses if). Should we aim to be consistent or is a balance of both fine?

api/test/unit/services/listing.service.spec.ts
```diff
@@ -2943,4 +2944,259 @@ describe('Testing listing service', () => {
...
+        permissionActions.update,
+        {
+          id: 'example id',
+        },
+      );
+
+      expect(prisma.listings.update).not.toHaveBeenCalled();
+    });
+
+    it.todo('should update status to errored');
```
ColinBuyck, line 3034: Love these it.todos

ColinBuyck (approved):
LGTM 🍏 Thanks for walking me through it in a huddle. Left some notes and documented the few things we discussed:
- Wondering about best return value for the front (ie just status, listing detail body)
- Noticed that example response didn't match the return type (addressed)

api/src/services/listing.service.ts
ludtkemorgan: I think this is a good use case for a switch statement since each enum has a different action to do.

api/src/services/listing.service.ts
```diff
@@ -1618,4 +1622,120 @@ export class ListingService implements OnModuleInit {
...
+      case LotteryStatusEnum.ran: {
+        if (!isAdmin) {
```
ludtkemorgan, line 1659: Will there be any status that a non admin can do? If there isn't we could just put the check for `!isAdmin` before the switch statement

api/src/controllers/listing.controller.ts
```diff
@@ -180,9 +181,27 @@ export class ListingController {
     return await this.listingService.process();
   }
 
+  @Put('lotteryStatus')
```
ludtkemorgan, line 184: I'm ok with this pattern and if you/the team like it better lets stick with this. But typically in a put endpoint like this I would have the listing id as a path parameter. So it would look like `@Put('lotteryStatus/:listingId')` and then update the ListingLotteryStatus dto to just have the new status

api/src/services/listing.service.ts
```diff
@@ -1618,4 +1622,120 @@ export class ListingService implements OnModuleInit {
...
+          });
+        } else {
+          // TODO: add ran to history
+        }
+        break;
+      }
+      case LotteryStatusEnum.errored: {
+        // TODO
+        break;
+      }
```
ludtkemorgan, line 1684: My guess is this status wont be changed via the endpoint but rather during the lottery run function. But it's cool to keep it here as a placeholder here for now

ludtkemorgan (approved):
Looks good. My comments are all just questions or suggestions so approving it but feel free to make the suggested changes if you want

api/src/controllers/listing.controller.ts
mcgarrye: @YazeedLoonat recommended we don't do that. Yazeed can you explain why that was again because I can't remember exactly what you said

api/src/services/listing.service.ts
mcgarrye: Yes, partners will be able to publishedToPublic

api/src/services/listing.service.ts
mcgarrye: I figure we can use the service function, not the endpoint, for wherever the lottery is run to set `errored` and `ran` but it can always be broken out later if we want to handle it a different way

api/src/controllers/listing.controller.ts
YazeedLoonat: Adding a bit more context @ludtkemorgan :
- if you look at endpoints that currently take ids via the path params we almost always also pass the id via the dto and then in the service we only use the value in the dto

- in nestjs we can validate those path params to make sure they are uuid's but to do that everywhere would probably get messy. Whereas when doing that via the dto method we can leverage the already built in field validation

- this is minor but the path params make it so that the order the endpoints show up in controller matter

Issue #692, opened 2020-09-19
Component: Status Messages


slowbot:
**Goal**: As a user I want to receive a high level message presented with emphasis. 

**Acceptance Criteria:**
Organization:
- Locate in ui-components/src/notifications

Engineering Notes:
- Stub with default data

Props: 
- status - relevant message status
- body - text content of message, _string of text_
- timestamp - time message created

Markup Prototype: https://storybook.bloom.exygy.dev/?path=/story/prototypes-statusmessages--status-messages

PR #4169, opened 2024-07-01
fix: prevent listings without AMI charts

KrissDrawing:
This PR addresses [#692](https://github.com/metrotranscom/doorway/issues/692)

## Description

It's a change from task comment. It will throw an error when trying to add unit without ami chart

Create new unit without touching `AMI Chart` dropdown.
Try to save -> It should throw validation error (prevent from saving)

## Author Checklist:

## Review Process:


mcgarrye:
I think this needs more discussion with product. Units can be added without AMIs and hiding the table but not changing how the Household Maximum Income section and Listings page/search to Listing flow confusing.
Cases to note with the change this PR makes:
- A unit is added without AMI chart and without fixed/percent rent and other fully filled out units exist. This causes n/a's on the Listings page and then when the user goes to the listing they don't see the unit at the top with the other units. Household Maximum Income has a strange NaN% column.
- A unit is added without AMI chart but with fixed/percent rent and other fully filled out units exist. This unit info is filled in on the Listings page and then when the user goes to the listing they don't see the unit at the top with the other units. Household Maximum Income has a strange NaN% column.
- Only units without AMI charts are added. The table appears with no label at the top. Household Maximum Income has an empty table.
- Only units without one AMI percent are added. The table appears with no label at the top.

This creates a confusing user experience where the user may not understand what units are available, what the rent is for those units and if they are eligible.
@sarahlazarich

sarahlazarich:
Good points @mcgarrye - let me confirm with Izzie if we should just block this entirely. I'll follow up.

ColinBuyck:
Looks like partners cypress tests are failing. Do you mind checking this out @KrissDrawing?

mcgarrye (approved):
LGTM

Issue #4174, opened 2024-07-02
AC: AMI charts from SJ/SMC Showing Up


iph-97:
We heard from AC that they are seeing SJ/SMC AMI charts appear as options for their local listings. This has not always been the case. Can we work to exclude those again? Low priority request.

QA Notes:
- Log into the partner site as an admin
- Add a new listing.
- Go to add a new unit within the listing and make sure that all AMI charts are accessible in the dropdown
- Set a jurisdiction to the listing and then try to add a unit. Only the AMI charts for that jurisdiction are accessible

PR #4177, opened 2024-07-03
fix: filter ami charts by jurisdition

ludtkemorgan:
This PR addresses #4174 

## Description

When making the backend call to get AMI charts on the partner site the jurisdiction id was not getting set. This was due to a change in Prisma where the jurisdiction field was changed to "jurisdictions" so the `watch` function was not looking for the correct field

*Note*: a new ami chart was added to the stage seeded data tied to the second jurisdiction so I'd recommend a reseed before testing
- Log into the partner site as an admin
- Add a new listing.
- Go to add a new unit within the listing and make sure that all AMI charts are accessible in the dropdown
- Set a jurisdiction to the listing and then try to add a unit. Only the AMI charts for that jurisdiction are accessible

## Author Checklist:

## Review Process:

### DIFF
--- a/api/prisma/seed-staging.ts
+++ b/api/prisma/seed-staging.ts
@@ -117,6 +117,10 @@ export const stagingSeed = async (
   const amiChart = await prismaClient.amiChart.create({
     data: amiChartFactory(10, jurisdiction.id),
   });
+  await prismaClient.amiChart.create({
+    data: amiChartFactory(8, additionalJurisdiction.id),
+  });
+  // Create map layers
   await prismaClient.mapLayers.create({
     data: mapLayerFactory(jurisdiction.id, 'Redlined Districts', redlinedMap),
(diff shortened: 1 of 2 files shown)


mcgarrye (approved):
Looks good!
While we have this front of mind, maybe we want to consider the case where a unit is created using an AMI from another jurisdiction and then the jurisdiction is set/changed. Just some food for though.

ludtkemorgan:
> Looks good! While we have this front of mind, maybe we want to consider the case where a unit is created using an AMI from another jurisdiction and then the jurisdiction is set/changed. Just some food for though.

That's a good point! There are a decent amount of units in HBA staging that have AMI charts that aren't associated to the correct jurisdiction. I don't know how common in prod it is, but I think it's something we probably should prevent. I'll create a ticket.

Issue #4161, opened 2024-06-25
Lotteries - Email Notification that results have been released to Partners


sarahlazarich:
As a Partner user on a listing with a lottery 
I want to receive an email notification when results are released to my portal,
to that I can access the lottery results for publication. 

AC: 
- When lottery results are released by an admin, partners who have access to that listing should receive an email in the correct format.

sarahlazarich:
Product/Design to get the email copy and template and link

sarahlazarich:
Email results can go to partners and admins

emilyjablonski:
QA Notes
Add a partner user and add them to a closed listing that has a lottery. As an admin, run and release the lottery. Ensure you receive the lottery released email. There isn't confirmed copy somewhere yet - this is filler / hopefully soon to be approved copy sent offline.

sarahlazarich:
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMUV6Qmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--94222592d1f1974aaa7af9ef5065dc3782bb8815/image.png" alt="image.png" />
 
It doesn't look like the translations are coming through here

sarahlazarich:
Ok, got this!!!!! and it looks good.

PR #4186, opened 2024-07-10
feat: lottery released email

emilyjablonski:
This PR addresses #4161

## Description

Adds an email notification when a lottery is released to partners. It should send to everyone w access to the listing.

Re-seed the database to get the new translations, in other environments we will use the script runner. Add a partner user with your email and add them to a closed listing that has a lottery. Additionally, update an application to have one of your emails. As an admin, run and release the lottery. Ensure the partner receives the lottery released email. Login as the partner user, and publish the lottery. Ensure the partner receives the lottery published email and the applicant receives their version of the lottery published email.

Copy lives [here](https://docs.google.com/document/d/130OuxoQnKnukoVr_SPQq57A0JZ4vycdcS9q0EX2Gkso/edit#heading=h.l124kyiwnbpp).

Screenshot 2024-07-17 at 11 08 19 AM
Screenshot 2024-07-17 at 11 08 30 AM
Screenshot 2024-07-17 at 11 08 44 AM

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/email.service.ts
+++ b/api/src/services/email.service.ts
@@ -574,7 +574,9 @@ export class EmailService {
       await this.send(
         emails,
         jurisdiction.emailFromAddress,
-        this.polyglot.t('lotteryReleased.header'),
+        this.polyglot.t('lotteryReleased.header', {
+          listingName: listingInfo.name,
+        }),
         this.template('lottery-released')({
           appOptions: { listingName: listingInfo.name },
           appUrl: appUrl,
--- a/api/src/services/email.service.ts
+++ b/api/src/services/email.service.ts
@@ -560,6 +560,33 @@ export class EmailService {
     );
   }
 
+  public async lotteryReleased(
+    user: User,
+    listingInfo: listingInfo,
+    emails: string[],
+    appUrl: string,
+  ) {
+    try {
+      const jurisdiction = listingInfo.juris
+        ? await this.getJurisdiction([{ id: listingInfo.juris }])
+        : user.jurisdictions[0];
+      void (await this.loadTranslations(jurisdiction));
+      await this.send(
+        emails,
+        jurisdiction.emailFromAddress,
+        this.polyglot.t('lotteryReleased.header'),
+        this.template('lottery-released')({
+          appOptions: { listingName: listingInfo.name },
+          appUrl: appUrl,
+          listingUrl: `${appUrl}/listings/${listingInfo.id}`,
+        }),
+      );
+    } catch (err) {
+      console.log('lottery released email failed', err);
+      throw new HttpException('email failed', 500);
+    }
+  }
+
   formatLocalDate(rawDate: string | Date, format: string): string {
     const utcDate = dayjs.utc(rawDate);
     return utcDate.format(format);
--- a/api/src/services/email.service.ts
+++ b/api/src/services/email.service.ts
@@ -643,6 +643,7 @@ export class EmailService {
               appUrl: jurisdiction.publicUrl,
             },
             appUrl: jurisdiction.publicUrl,
+            // These two URLs are placeholders and must be updated per jurisdiction
             notificationsUrl: 'https://www.exygy.com',
             helpCenterUrl: 'https://www.exygy.com',
           }),
--- a/api/src/services/email.service.ts
+++ b/api/src/services/email.service.ts
@@ -589,6 +589,64 @@ export class EmailService {
     }
   }
 
+  public async lotteryPublishedAdmin(
+    user: User,
+    listingInfo: listingInfo,
+    emails: string[],
+    appUrl: string,
+  ) {
+    try {
+      const jurisdiction = listingInfo.juris
+        ? await this.getJurisdiction([{ id: listingInfo.juris }])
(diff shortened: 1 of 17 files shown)


api/src/services/script-runner.service.ts
```diff
@@ -195,6 +195,53 @@ export class ScriptRunnerService {
...
   }
 
+  /**
+   *
+   * @param req incoming request object
+   * @param jurisdiction should contain jurisdiction id
+   * @returns successDTO
+   * @description adds lottery released translations to the database
+   */
+  async addLotteryReleasedTranslations(
```
ludtkemorgan, line 205: This solution works and it follows our current practices for the script runner. I'm wondering if we will run into annoyances having to write a new script every time we need to add new translations. And then on top of that having to remember to run the script.

This is just an option and I'm open to the team discussing this as a solution but what do you think of this idea:

- Have the translations for each language and jurisdiction along with a default one as well in a json file. So for example for core it would be what is in the translation-factory but just pulled out to a separate file.
  - As a note, then we can pull that file directly into the translation-factory so it always has the up to date translations.
- Create a script runner endpoint that takes in an optional jurisdiction name. Then in this file have the service find the appropriate file and fully update the translations for either the default translations or the passed in jurisdiction. 
- Then whenever we want to add new or update existing translations we just update those files and run the script runner.

This would require the script to be run more than once but we would just handle that by having the script name be something like `update translations ${jurisdictions} <new date>`

api/src/views/lottery-released.hbs
```diff
@@ -0,0 +1,53 @@
...
+            </p>
+            </td>
+        </tr>
+        </tbody>
+    </table>
+    <table role="presentation" border="0" cellpadding="0" cellspacing="0">
+        <tbody>
+        <tr>
+            <td align="center">
+            <a
```
ludtkemorgan, line 25: I probably just missed it, but I don't see the button in the copy so don't know what it is supposed to look like. The "View Listing" button seems to be lower down than I'd expect. In other words there is a bigger gap between the text above it and the "Thank you" below it. Don't know if this is intended
image

ludtkemorgan (approved):
It works great! I just have a question about how we want to do db translations for long term sustainability which is probably a question for the larger team

ludtkemorgan:
I also want to keep the 1 review needed tag on this PR so others can take a look at it

api/src/views/lottery-released.hbs
emilyjablonski: This is just copy/pasta format of our current approval emails which was the request from product (sorry I didn't mention that!)

mcgarrye (approved):
Looks good! If we could just add formal responses to Morgan's comments from our previous discussion about the translations/script portion for future reference please :)

emilyjablonski:
@mcgarrye Yazeed recorded that here! https://docs.google.com/document/d/168nRndz1cLsiuRe-EY2cxw0RwPaKtsi9neSHQhbiuy8/edit

emilyjablonski:
Just waiting on product/design confirmation on the emails

api/src/services/email.service.ts
```diff
@@ -560,6 +560,100 @@ export class EmailService {
...
+            listingName: listingInfo.name,
+          }),
+          this.template('lottery-published-applicant')({
+            appOptions: {
+              listingName: listingInfo.name,
+              appUrl: jurisdiction.publicUrl,
+            },
+            appUrl: jurisdiction.publicUrl,
+            notificationsUrl: 'https://www.exygy.com',
+            helpCenterUrl: 'https://www.exygy.com',
```
ludtkemorgan, line 647: Is the exygy url just a placeholder url? Or is it intended to be here because in core we don't have a specific url?

emilyjablonski: Sort of both, we don't have anything to set this to in core

emilyjablonski:
Confirmed emails w design

api/src/services/email.service.ts
ludtkemorgan: I wonder if we should make those parts of the email conditional on if the urls exist. So for example if `notificationsUrl` doesn't exist we won't show the sentence "You can sign up to receive notifications of new application opportunities here"

emilyjablonski: I think my only light concern with that is that I think ideally the feature is fully demo-able and testable in core, and then core wouldn't reflect the full emails.

ludtkemorgan: In doorway will this be determined by the env variable? I'm assuming the notifications url will match the `NOTIFICATIONS_SIGN_UP_URL` variable in public?

I'm ok with this being here for now. Maybe we should put a comment here that these are just placeholder urls. I'd prefer to have these urls not hardcoded here long term when it will have to be dynamic for each jurisdiction/environment. But to get this in a demo-able state no need to update that immediately.

ludtkemorgan (approved):

api/src/services/email.service.ts
emilyjablonski: Great q on the public site env variable I will ask! So your ideal long term is that these pieces won't be set in core and won't show? Will add a comment here

PR #4190, opened 2024-07-10
fix: new lottery copy

emilyjablonski:
## Description

There isn't a ticket for this work, we just had [copy edits](https://docs.google.com/document/d/130OuxoQnKnukoVr_SPQq57A0JZ4vycdcS9q0EX2Gkso/edit#heading=h.z2we68wtdnff) and some tiny design changes come in from Doorway on existing work.

They aren't called out explicitly in the doc, but I've listed them below.

* Slight string adjustments to wording and removing one date
* Removing references to Doorway / BAHFA in the strings so we don't have them in core, but adding them as custom strings we can change in Doorway
* Making the tabs lowercase
* Updating the add application modal to check the new listingLotteryStatus
* Changing the run with duplicates button back to an alert color

## Author Checklist:

## Review Process:


sites/partners/page_content/locale_overrides/general.json
```diff
@@ -221,29 +221,33 @@
...
+  "listings.lottery.exportFile": "This file includes the lottery raw rank and preferences data for all applications.",
+  "listings.lottery.exportFileNoPreferences": "This file includes the lottery raw rank for all applications.",
   "listings.lottery.history": "History",
   "listings.lottery.noData": "No lottery data",
   "listings.lottery.noDataDescription": "It looks like you haven't run a lottery for this listing yet.",
-  "listings.lottery.noDataDescriptionPartner": "Doorway staff will run the lottery and perform a quality check on the results before making the lottery results accessible to you on %{date}. Before viewing the lottery results, you will be prompted to publish the lottery data, which will email results to applicants. Please email partners@housingbayarea.org with any questions.",
+  "listings.lottery.noDataDescriptionPartner": "Admins will run the lottery and perform a quality check on the results before making the lottery results accessible to you on the scheduled lottery date. Before viewing the lottery results, you will be prompted to publish the lottery data, which will email results to applicants. Please email",
```
mcgarrye, line 230: This doesn't match what is stated in `listings.lottery.publishContent` about what is emailed to applicants

sites/partners/src/pages/listings/[id]/applications/index.tsx
```diff
@@ -182,7 +182,7 @@ const ApplicationsList = () => {
                   <div className="flex gap-2 items-center">
                     <Button
                       onClick={() => {
-                        if (listingDto.lotteryLastRunAt) {
+                        if (process.env.showLottery === "TRUE" && !!listingDto.lotteryStatus) {
```
Comment (mcgarrye) [inline sites/partners/src/pages/listings/[id]/applications/index.tsx:185]: We do have an `errored` status for the lottery that we might want to ignore for this.
Also, for partners/jadmins that don't know the lottery has been run yet, would this message be confusing?

sites/partners/page_content/locale_overrides/general.json
emilyjablonski: It's the confirmed copy in the doc after the latest round of changes, but I hear ya

sites/partners/src/pages/listings/[id]/applications/index.tsx
Comment (emilyjablonski) [reply sites/partners/src/pages/listings/[id]/applications/index.tsx:185]: I think partners/jadmins would still need to know, but also in Doorway I don't think partners/jadmins will be adding applications

Agreed on ignoring the `errored` state! When does the lottery end up in `errored` as opposed to `null`?

Comment (mcgarrye) [reply sites/partners/src/pages/listings/[id]/applications/index.tsx:185]: In the rare case that when the lottery run is triggered something goes wrong and fails to complete it would be set to `errored` and would need to be rerun. At least that was my understanding @YazeedLoonat

mcgarrye (approved):
Looks good!

Issue #4113, opened 2024-06-03
AC: HCV Reserved Community Type


iph-97:
we got a request from Alameda for a new reserved community type for voucher holders. This is for a property going live in the next 2-3 weeks. They want the tag to be “HCV/Section 8 Voucher”, then the banner to say “Reserved for HCV/Section 8 Voucher Holder”. 

QA Notes: https://housing.acgov.org/ and https://partners.housingbayarea.org/
Log in to the partner site and add the new community type to a listing
Go to the public site and make sure it comes through properly (and is translated correctly)

sarahlazarich:
Script for this in script runner - would just need to pull it out of Doorway

sarahlazarich:
lgtm

PR #4192, opened 2024-07-11
fix: add translations for hcv section 8 voucher

ludtkemorgan:
This PR addresses #4113 

## Description

This pulls the translation strings from HBA for the HCV/Section 8 Voucher. This Reserved community type is only used in Alameda at this point, but adding it to core will help for future jurisdictions

Also, there was a failing test unrelated to this change. A test was added to main while a PR was in review and then that PR merged made the test fail. I fixed the null pointer spot but also updated the test to have some preference data so that it better tests the page and these fields.


sites/public/__tests__/pages/applications/preferences/all.test.tsx
```diff
@@ -2,6 +2,14 @@ import React from "react"
 import { setupServer } from "msw/lib/node"
 import { mockNextRouter, render } from "../../../testUtils"
 import ApplicationPreferencesAll from "../../../../src/pages/applications/preferences/all"
+import { AppSubmissionContext } from "../../../../src/lib/applications/AppSubmissionContext"
```
mcgarrye, line 5: nit: imports aren't alphabetized

mcgarrye (approved):
LGTM

sites/public/__tests__/pages/applications/preferences/all.test.tsx
ludtkemorgan: Is that a new standard we are following? I typically don't alphabetize the imports, but I can start doing that if we want to.

mcgarrye: Idk if we ever formalized it but I know Emily and I have been doing that lately. This is fine to go out and we can discuss later

ludtkemorgan: Sounds good!

PR #4196, opened 2024-07-15
feat: lottery endpoints

YazeedLoonat:
This PR addresses #4055 

## Description
This creates the lottery results generation endpoint and the lottery export as csv

another ticket will cover converting that export to a spreadsheet and include the preference sheets

- create applications on a listing (or use a listing with existing applications)
- navigate to the listings' lottery tab
- you should be able to click the run lottery button and lottery results should get generated
- you should be able to click the export results button and get a csv spreadsheet with the lottery results

## Author Checklist:

## Review Process:


emilyjablonski:
@YazeedLoonat Do you mind filling out the PR template and adding QA notes?

api/test/integration/application.e2e-spec.ts
```diff
@@ -1543,4 +1544,134 @@ describe('Application Controller Tests', () => {
...
+      const res = await request(app.getHttpServer())
+        .put(`/applications/generateLotteryResults`)
+        .set({ passkey: process.env.API_PASS_KEY || '' })
+        .send({
+          listingId: listing1Created.id,
+        })
+        .set('Cookie', cookies)
+        .expect(400);
+      expect(res.body.message).toEqual(
+        `Listing ${listing1Created.id}: the lottery was attempted to be generated but it was already run previously`,
```
emilyjablonski, line 1673: How does this allow for the re-run flow?

sites/partners/src/pages/listings/[id]/lottery.tsx
```diff
@@ -296,35 +296,16 @@ const Lottery = (props: { listing: Listing }) => {
               <Button
                 variant="primary"
                 onClick={async () => {
-                  setLoading(true)
```
Comment (emilyjablonski) [inline sites/partners/src/pages/listings/[id]/lottery.tsx:299]: We will still want the loading state

sites/partners/src/pages/listings/[id]/lottery.tsx
```diff
@@ -445,8 +426,9 @@ const Lottery = (props: { listing: Listing }) => {
               <Button
                 variant="primary"
                 onClick={() => {
-                  // export lottery
+                  const res = onExport()
```
Comment (emilyjablonski) [inline sites/partners/src/pages/listings/[id]/lottery.tsx:429]: Curious why we don't await this

emilyjablonski:
On a fresh reseed I get a 500 when trying to run the lottery
```
const res = await this.prisma.applications.findUnique(
 BACKEND_API Inconsistent column data: Error creating UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-zA-Z], found `g` at 1
 BACKEND_API PrismaClientKnownRequestError:
 BACKEND_API Invalid `this.prisma.applications.findUnique()` invocation in
```

shared-helpers/src/types/backend-swagger.ts
```diff
@@ -1427,6 +1427,64 @@ export class ApplicationsService {
...
+      includeDemographics?: boolean
+      /**  */
+      timeZone?: string
+    } = {} as any,
+    options: IRequestOptions = {}
+  ): Promise<any> {
+    return new Promise((resolve, reject) => {
+      let url = basePath + "/applications/generateLotteryResults"
+
+      const configs: IRequestConfig = getConfigs("get", "application/json", url, options)
```
ludtkemorgan, line 1476: When testing this out it was failing from the frontend. Looks like this should be a "put" call instead of a "get". When I ran the generate:client it fixes that issue. Still didn't generate the file so looking into that. But adding this comment here so others aren't blocked on reviewing

api/test/integration/application.e2e-spec.ts
YazeedLoonat: So I think the re-run flow would need to be a 2 part process:
1. Set the lottery status back to `null` and clear the rows in the lottery position table
2. then re-run the lottery

I think we may want a stand alone endpoint to do that resetting, but defo open to feedback/thoughts here

ludtkemorgan:
@YazeedLoonat @emilyjablonski I got this to work by running `yarn generate:client` and then updating the call to the backend in lottery.tsx to be ` await applicationsService.lotteryGenerate({ body: { listingId: listing.id } })` (It needs "body")

sites/partners/src/pages/listings/[id]/lottery.tsx
Comment (YazeedLoonat) [reply sites/partners/src/pages/listings/[id]/lottery.tsx:299]: oh nice call out! Sorry I shouldn't have fiddled with the react logic so much updated

to answer your question, yes it does! Or rather the work is all wrapped in a try/catch block with only `setCsvExportLoading()` switching from true to false being outside that try catch. you can peep that at `/src/lib/hooks.ts -> useCsvExport()`

sites/partners/src/pages/listings/[id]/lottery.tsx
Comment (YazeedLoonat) [reply sites/partners/src/pages/listings/[id]/lottery.tsx:429]: tbh I copied the logic from the application csv export `partners/src/pages/listings/[id]/applications/index.ts` ~line 203

where we don't await the export call. Defo down to change it to an async await though if you think it'd be better

YazeedLoonat:
> On a fresh reseed I get a 500 when trying to run the lottery
> 
> ```
> const res = await this.prisma.applications.findUnique(
>  BACKEND_API Inconsistent column data: Error creating UUID, invalid character: expected an optional prefix of `urn:uuid:` followed by [0-9a-zA-Z], found `g` at 1
>  BACKEND_API PrismaClientKnownRequestError:
>  BACKEND_API Invalid `this.prisma.applications.findUnique()` invocation in
> ```

Hey @emilyjablonski I think you may be missing migrations if you try running `yarn setup` instead of just the reseed are you still getting the errors?

emilyjablonski:
@YazeedLoonat sorry i meant `yarn setup:dev`!

ludtkemorgan:
@YazeedLoonat @emilyjablonski That 500 error was the error I was getting before making the change I referenced above

YazeedLoonat:
Hey @ludtkemorgan ugh my bad I checked the front end and then was fiddling with the api tests when I realized it should be a put instead of a get and just didn't think to double check the ui 

updated! 

and in general @emilyjablonski @ludtkemorgan updates should be out and ready for ya both!

api/src/services/application-csv-export.service.ts
```diff
@@ -815,15 +698,380 @@ export class ApplicationCsvExporterService
...
   async lotteryRandomizer(
     listingId: string,
     applications: Application[],
     preferencesOnListing: MultiselectQuestion[],
   ): Promise<void> {
     // remove duplicates
-    let filteredApplications = applications.filter(
-      (application) => !application.markedAsDuplicate,
-    );
+    let filteredApplications = applications;
```
ludtkemorgan, line 1074: This isn't the line that is failing, but when I use HBA prod data it is failing on line 1091 with the createMany because of "Unique constraint failed on the fields: (`listing_id`,`application_id`)"

ludtkemorgan: Is this because when there are more than one multiselect question on the listing there would be two records with the same listing id and application id?

ludtkemorgan (changes requested):
Still looking through the code, but looks like the lottery doesn't work with more complex listings. Will need more investigating to figure out what the problem is.

sites/partners/src/pages/listings/[id]/lottery.tsx
```diff
@@ -296,35 +296,17 @@ const Lottery = (props: { listing: Listing }) => {
...
-                    await listingsService.lotteryStatus({
-                      body: {
-                        listingId: listing.id,
-                        lotteryStatus: LotteryStatusEnum.ran,
-                      },
-                    })
+                    setLoading(true)
+                    await applicationsService.lotteryGenerate({ body: { listingId: listing.id } })
                     setRunModal(false)
-                    setLoading(false)
```
Comment (emilyjablonski) [inline sites/partners/src/pages/listings/[id]/lottery.tsx:319]: We need these setLoading calls as well

api/src/services/application-csv-export.service.ts
YazeedLoonat: ooooo you're right, I'll adjust the schema and generate a new migration to add `multiselectQuestionId` as part of the id 

the other option is I can add a unique id column that we use as the primary id and then add indices for `listing id` and `application id` for rapid look up

api/src/services/application-csv-export.service.ts
```diff
@@ -815,15 +698,380 @@ export class ApplicationCsvExporterService
...
+   * @param req
+   * @returns generates the lottery results for a listing
+   */
+  async lotteryGenerate<QueryParams extends ApplicationCsvQueryParams>(
+    req: ExpressRequest,
+    res: Response,
+    queryParams: QueryParams,
+  ): Promise<SuccessDTO> {
+    const user = mapTo(User, req['user']);
+    await this.authorizeCSVExport(user, queryParams.listingId);
```
ludtkemorgan, line 809: I could be misremembering. But isn't it by default only admins that are allowed to generate the lottery? So do we need to have a separate check for permissions than this `authorizeCSVExport` endpoint?

api/src/services/application-csv-export.service.ts
```diff
@@ -815,15 +698,380 @@ export class ApplicationCsvExporterService
...
+        (multiselectQuestion) =>
+          multiselectQuestion.applicationSection ===
+          MultiselectQuestionsApplicationSectionEnum.preferences,
+      ),
+    );
+
+    await this.prisma.listings.update({
+      data: {
+        lotteryLastRunAt: new Date(),
+        lotteryStatus: LotteryStatusEnum.ran,
```
ludtkemorgan, line 876: We also created a LotteryStatusEnum of "errored". Should this function be wrapping everything above this listings update in a try catch function where we can update the lotteryStatus to errored?

api/test/integration/application.e2e-spec.ts
```diff
@@ -1543,4 +1544,134 @@ describe('Application Controller Tests', () => {
...
+    it('should generate results when no previous attempt to run lotteries has happened', async () => {
+      const unitTypeA = await unitTypeFactorySingle(
+        prisma,
+        UnitTypeEnum.oneBdrm,
+      );
+      const jurisdiction = await prisma.jurisdictions.create({
+        data: jurisdictionFactory(),
+      });
+      await reservedCommunityTypeFactoryAll(jurisdiction.id, prisma);
+      const listing1 = await listingFactory(jurisdiction.id, prisma);
```
ludtkemorgan, line 1558: Can you add some multiselect questions to this listing and also update the applications in this test to have them selected/not selected? That way we can capture the multiselect logic for the lottery generation

api/src/services/application-csv-export.service.ts
```diff
@@ -815,15 +698,380 @@ export class ApplicationCsvExporterService
...
+      where: {
+        id: queryParams.listingId,
+      },
+    });
+    if (listing?.lotteryStatus) {
+      // if lottery has been run before
+      throw new BadRequestException(
+        `Listing ${queryParams.listingId}: the lottery was attempted to be generated but it was already run previously`,
+      );
+    }
```
mcgarrye, line 826: Why would we want to prevent re-runs in the `ran` state? If Doorway doesn't approve and hasn't released they should be able to re-run

api/src/services/application-csv-export.service.ts
mcgarrye: I would also consider calling the listing.service.lotteryStatus function here so the logic isn't duplicated and we only have to implement the lottery history logic in one location

api/src/services/application-csv-export.service.ts
YazeedLoonat: ah you're right, I thought that the permissions would be the same between export and lottery gen but that shouldn't be the case

I switched this to throw a forbidden exception if the user is not an admin

api/src/services/application-csv-export.service.ts
YazeedLoonat: I think that we will want another endpoint to handle "undoing" a lottery run that isn't the lottery generate endpoint

that other endpoint will be responsible for setting the lotteryStatus back to null and clearing the stored lottery positions for this listing

if we want to handle it all in the same call we could then call the lottery generate function after that work is cleaned up but I think that can be tackled a little separately

api/test/integration/application.e2e-spec.ts
YazeedLoonat: I went ahead and added a second test that has preferences and a mix of claimed/unclaimed to test that logic, lemme know what you think

YazeedLoonat:
Hey @ludtkemorgan @emilyjablonski @mcgarrye this is re-ready for 👀

api/src/services/application-csv-export.service.ts
mcgarrye: Well wouldn't a re-run just update the stored data? I'm not sure what usecase there is for "undoing" where we set everything to null/cleared.

mcgarrye: I am just seeing Emily's same comment. Maybe we discuss this in huddle today

ludtkemorgan: I agree we should talk about this in the huddle. I think there is potentially a scenario where the job made it half way through and errored out leading to some applications having data and others not so we would want to start from scratch by clearing it out and starting over. But at the same time the re-run would just overwrite it so I could go either way

api/src/services/application-csv-export.service.ts
```diff
@@ -815,15 +699,389 @@ export class ApplicationCsvExporterService
...
+                      deletedAt: null,
+                    },
+                    skip: i,
+                    take: NUMBER_TO_PAGINATE_BY,
+                  });
+                if (forLottery) {
+                  paginatedApplications = paginatedApplications.sort(
+                    (a, b) =>
+                      a.applicationLotteryPositions[0].ordinal -
+                      b.applicationLotteryPositions[0].ordinal,
```
ludtkemorgan, line 977: When I ran this with production data the export failed on this line because "ordinal" didn't exist on some applications. And looking at the exported data there are several chunks where the rank doesn't show up 
image
I'll keep investigating on why this might be.

YazeedLoonat: Hey @ludtkemorgan I think I figured it out, chunks where the rank is missing is probably tied with those being marked as duplicate which shouldn't show in the export 

I made an adjustment that I'll be pushing up that should remove duplicates from the lottery export completely and solve both the ordinal is missing problem in the sort and removes this unranked chunks from the export

YazeedLoonat:
create new controller/service/module specific for lotteries move generate lottery logic there 

will move export generation logic there once spreadsheet switch over happens

api/src/services/application-csv-export.service.ts
emilyjablonski: The confirmation string here might also need a fix - I think it shows correctly for an applicant, so if this was the confirmation string used in a lottery, it wouldn't match.

emilyjablonski: We also want duplicates to be in the lottery / allow for people to do that

ludtkemorgan: This issue no longer exists, but I think the sorting is not properly working. It appears to maybe just be sorting the applications in the chunks that it works on. So it sorts 500 at a time. 
image

YazeedLoonat:
hey @emilyjablonski @ludtkemorgan this is ready for ya again!

sites/partners/src/lib/hooks.ts
```diff
@@ -521,6 +521,15 @@ export const useApplicationsExport = (listingId: string, includeDemographics: bo
   )
 }
 
+export const useLotteryExport = (listingId: string) => {
+  const { applicationsService } = useContext(AuthContext)
+
+  return useCsvExport(
+    () => applicationsService.lotteryResults({ listingId, timeZone: dayjs.tz.guess() }),
```
ludtkemorgan, line 528: Looking at the sample export, looks like we want to have demographic info on the export https://docs.google.com/spreadsheets/d/1dX9ehAq3pdMpmhNq1arWV2bwjHBFMy6xAbcPGhLVHOY/edit?gid=1191380628#gid=1191380628. You can just add it as a query param here

api/src/services/application-csv-export.service.ts
```diff
@@ -727,6 +589,18 @@ export class ApplicationCsvExporterService
       );
     }
 
+    // if its for the lottery insert the lottery position
+    if (forLottery) {
+      headers.unshift({
+        path: 'applicationLotteryPositions',
+        label: 'Lottery Position',
```
ludtkemorgan, line 596: The sample file calls this column "Raw Lottery Rank". Talking to Sarah it sounds like we want to have it be that instead of "Lottery Position"

ludtkemorgan (approved):
Thanks for making all of those changes. It all looks good. Two small things that will make it match the sample file. But works great so approving it

sites/partners/src/lib/hooks.ts
YazeedLoonat: if its always going to be on I think it may be better to force it on the api side so I updated the call to have `includeDemographics: true` always

mcgarrye (approved):
Looks good! LGTM

emilyjablonski:
Let's do it 🫡

emilyjablonski (approved):

PR #4228, opened 2024-07-25
feat: spreadsheetification of lottery export

YazeedLoonat:

## Description

## Author Checklist:

## Review Process:


sites/partners/src/lib/hooks.ts
```diff
@@ -522,12 +522,44 @@ export const useApplicationsExport = (listingId: string, includeDemographics: bo
...
+      const url = window.URL.createObjectURL(blob)
+      const link = document.createElement("a")
+      link.href = url
+      const now = new Date()
+      const dateString = dayjs(now).format("YYYY-MM-DD_HH-mm")
+      link.setAttribute("download", `${listingId}-${dateString}-test.zip`)
+      document.body.appendChild(link)
+      link.click()
+      link.parentNode.removeChild(link)
+      //addToast(t("t.exportSuccess"), { variant: "success" })
```
YazeedLoonat, line 551: why commented out?

sites/partners/src/lib/hooks.ts
```diff
@@ -522,12 +522,44 @@ export const useApplicationsExport = (listingId: string, includeDemographics: bo
...
 
 export const useLotteryExport = (listingId: string) => {
-  const { applicationsService } = useContext(AuthContext)
+  const { lotteryService } = useContext(AuthContext)
 
-  return useCsvExport(
-    () => applicationsService.lotteryResults({ listingId, timeZone: dayjs.tz.guess() }),
-    `lottery-${listingId}-${createDateStringFromNow()}.csv`
-  )
+  // return useCsvExport(
```
YazeedLoonat, line 527: 🧹

ColinBuyck (approved):
LGTM 💵 I did notice some strange behavior on adding an application after the lottery has been run and re-running the lottery but I'm assuming that that is part of future work.

api/src/services/lottery.service.ts
```diff
@@ -224,4 +257,325 @@ export class LotteryService {
...
+    const zipFilePath = join(
+      process.cwd(),
+      `src/temp/lottery-listing-${queryParams.listingId}-applications-${
+        user.id
+      }-${new Date().getTime()}.zip`,
+    );
+
+    await this.createLotterySheet(workbook, {
+      ...queryParams,
+      includeDemographics: true,
```
ludtkemorgan, line 293: Curious if this "includeDemographics" should be passed in from the frontend as part of the queryParams?

api/src/services/lottery.service.ts
```diff
@@ -224,4 +257,325 @@ export class LotteryService {
...
+   */
+  async createLotterySheet<QueryParams extends ApplicationCsvQueryParams>(
+    workbook: Excel.Workbook,
+    queryParams: QueryParams,
+  ): Promise<void> {
+    let applications = await this.prisma.applications.findMany({
+      select: {
+        id: true,
+        preferences: true,
+        householdMember: {
```
ludtkemorgan, line 333: could you just get householdSize rather than going to the householdMember table? If I'm reading this correctly you just need this to determine the headers

api/src/services/lottery.service.ts
```diff
@@ -224,4 +257,325 @@ export class LotteryService {
...
+        queryParams.listingId,
+      );
+
+    // get maxHouseholdMembers associated to the selected applications
+    let maxHouseholdMembers = 0;
+    applications.forEach((app) => {
+      if (app.householdMember?.length > maxHouseholdMembers) {
+        maxHouseholdMembers = app.householdMember.length;
+      }
+    });
```
ludtkemorgan, line 370: Assuming that the householdSize field is available you could have this be a reduce function
```
    const maxHouseholdMembers = applications.reduce((maxValue, application) => {
      return Math.max(maxValue, application.householdSize);
    }, 0);

File: api/src/services/lottery.service.ts [hunk 6]
```diff
@@ -224,4 +257,325 @@ export class LotteryService {
...
+                  // curr should equal the preference id we're pulling from
+                  if (!programs) {
+                    programs =
+                      app.programs as unknown as ApplicationMultiselectQuestion[];
+                  }
+                  parsePreference = false;
+                  // there aren't typically many programs, but if there, then a object map should be created and used
+                  const program = programs.find(
+                    (preference) => preference.key === curr,
+                  );
```
ludtkemorgan, line 528: Just for safety reasons we should port over the null check that was added here https://github.com/metrotranscom/doorway/pull/768

sites/partners/src/lib/hooks.ts
```diff
@@ -522,12 +522,44 @@ export const useApplicationsExport = (listingId: string, includeDemographics: bo
...
+  //   () => lotteryService.lotteryResults({ listingId, timeZone: dayjs.tz.guess() }),
+  //   `lottery-${listingId}-${createDateStringFromNow()}.xlsx`,
+  //   true
+  // )
+  const [csvExportLoading, setCsvExportLoading] = useState(false)
```
ludtkemorgan, line 532: nit: if this is no longer a csv could we rename this to just "exportLoading"?

ludtkemorgan (approved):
This all looks good. Just minor comments.

I'm not suggesting doing a big refactor at this point. But I'm wondering if you looked into having both the application csv and the lottery export both use exceljs?

Looks like they have a `workbook.xlsx.writeFile` as well as `workbook.csv.writeFile`. So we could basically use the same functions but just write it to different workbooks. And of course make it so the new columns are only in the lottery one

I didn't do any verification though of if that would work on our use case

ColinBuyck:
Talked offline about the adding application weirdness with @YazeedLoonat and the steps I took were to 

1. run lottery
2. get export
3. add application
4. try to get export again (500 response)

I am assuming that blocking this flow is part of upcoming FE work but I suggested that the error should be better handled at some point

api/src/services/lottery.service.ts
YazeedLoonat: thats a good question, this is how the old csv export was calculating it so I kept that logic in tact 

But after looking at prod data with:
```

SELECT
    SUM(1)
FROM applications app1
    JOIN (
        SELECT
            a.id,
            SUM(1) as sm
        FROM applications a
            LEFT JOIN household_member hm ON hm.application_id = a.id
        GROUP BY a.id
    ) app2 ON app2.id = app1.id
        AND COALESCE(app2.sm, 0) != COALESCE(app1.household_size, 0);


SELECT
    SUM(1)
FROM applications app1
    JOIN (
        SELECT
            a.id,
            SUM(1) as sm
        FROM applications a
            LEFT JOIN household_member hm ON hm.application_id = a.id
        GROUP BY a.id
    ) app2 ON app2.id = app1.id
        AND COALESCE(app2.sm, 0) = COALESCE(app1.household_size, 0);
        ```
        
        what I'm seeing is that household_size doesn't seem to always match up with the actual house hold size so switching that would probably cause problems

api/src/services/lottery.service.ts
YazeedLoonat: thats a good question, my read was that the export would always have the demographic fields, but maybe I'm wrong there 

I'll check in with sarah and reply here with what she says!

YazeedLoonat:
@ColinBuyck from convo in the slack channel it seems like this wasn't really accounted for but with the toasts being readded when it errors it does let the UI know. Lemme know if thats enough for now 

@ludtkemorgan left a few comments back based on your comments, but for your top level comment there is a ticket to switch the application export to be a spreadsheet export as well so I think it'd be work to switch the csv export up when it seems likely it'll go away so leaving that in place here. 

Although maybe in that ticket we can do some investigation

api/src/services/lottery.service.ts
ludtkemorgan: Interesting. Okie dokie lets keep it as the householdmembers instead of the size then. I wonder if the inconsistency is due to a bug or just older data isn't accurate.

ludtkemorgan:
Looks good. I'm putting this comment down here so we hopefully don't forget, but there are a few changes in the columns outputted on application export in Doorway. And since this is a big refactor when we pull it over to Doorway we'll need to remember to add those new fields to both application and lottery exports

Issue #4208, opened 2024-07-18
Duplicates - Cron job


sarahlazarich:
Blocked by [this](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4199)

This should encompass the re-write of the duplicate chron job following the steps [here](https://docs.google.com/document/d/1h7zK3b466M2mfgj5vMjTaQKSL8STKUrH0t3MW-C1T4U/edit)

1. Get all duplicate keys from the view. 
2. Go through the above response and filter it into two groups, one with unique IDs only and one with repeats (where the same group matches both email and name/dob).
3. Go through each group and see if a flagged set exists. 

Notes:

TDD [here](https://docs.google.com/document/d/1h7zK3b466M2mfgj5vMjTaQKSL8STKUrH0t3MW-C1T4U/edit)

sarahlazarich:
We entered some flagged sets on Binky Binky, and it all looks good here!

PR #4230, opened 2024-07-30
feat: new cron job for duplicates

ludtkemorgan:
This PR addresses #4208 

## Description

Creates a new and improved duplicates (application flagged set) cron job. This tackles the issues identified [here](https://docs.google.com/document/d/1h7zK3b466M2mfgj5vMjTaQKSL8STKUrH0t3MW-C1T4U/edit#heading=h.r0j0xbeefnoe). 

How the duplicates are grouped were decided on via a slack poll. Option 2 won https://exygy.slack.com/archives/C01Q4QG5R8Q/p1724708380283439. 

Since this will interfere with the listings that have had the old cron job run there is a new date that we set via env variable. Any listing that closes before that date/time will have the old cron job run against it whereas any listing closed after it will have the new one. This way any partner that has resolved applications but a new application comes through they will not have to start from the beginning again. And if the new close date is in the future (or not set) only the old job will run.

Questions:
- If a new application comes through that matches an existing flagged set do we need to set the flagged set status back to "pending" so that it needs to be re-resolved?

Notes:
- A new view is created that has all of the possible flagged sets in the system. A query is done against this to find the duplicates.
- We previously required the rule_key to be unique. The way we ensured that it was unique was that each key was prefixed with the listing id. This seemed unnecessary since we are already storing the listing id. But in order to make sure the rows are unique this PR also removes the unique constraint and adds a compound unique constraint to the listing_id and rule_key

This can be tested both purely with the backend as well as using the partner site.

Prerequisites for both ways:
- Reseed the backend (more applications have been added to the District View listing)
- Add new environment variables to .env for backend
  - DUPLICATES_CLOSE_DATE="2024-07-28 00:00 -08:00"
  - DUPLICATES_PROCESSING_CRON_STRING=0 * * * * 

Using the partner site:
- Sign into the partner site and go to the District View listing.
- Close the listing and go to the "applications" tab
- Notice the pending duplicate sets and resolve the applications as seen fit

Using the backend:
- After a reseed either wait for the cron job to run or go to the swagger doc and execute the cron job 
- Look in the DB and verify that the appropriate Application Flagged sets exist and along with the connected applications

## Author Checklist:

## Review Process:


ColinBuyck:
Am I remembering our conversation that I should expect this to be flagged as nameEmailAndDOB or something acknowledging that they match on all three cases?
Screenshot 2024-07-31 at 1 20 47 PM

ludtkemorgan:
@ColinBuyck are you still seeing the above? I pushed up a change with the UI change mentioned via slack

ludtkemorgan:
@ColinBuyck and team I did some investigation. When submitting a paper application we are setting the reviewStatus to "valid" which I'm assuming is intentional. So when you go to resolve a pending application flagged set that has a paper application the paper application will be preselected as "valid".

ColinBuyck:
Nit but I'm surprised to see these two references to the combined rule differ
Screenshot 2024-08-01 at 10 30 24 AM
Screenshot 2024-08-01 at 10 30 18 AM

ColinBuyck:
Do these serve different roles? Also I keep getting traffic not from a known source despite signing in before clicking execute. Am I missing something?
Screenshot 2024-08-01 at 10 45 11 AM

api/prisma/migrations/16_add_duplicates_view/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+)
+UNION
+(
+    SELECT
+        a.email_address as "key",
+        app.listing_id,
+        app.id as "application_id",
+        'email' as "type"
+    from
+        applications app,
```
YazeedLoonat, line 21: really cool querying syntax! looks to be as performant as traditional joins too!

api/prisma/migrations/16_add_duplicates_view/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+        household_member hm
+    where
+        hm.application_id = app.id
+);
+
+ALTER TYPE "rule_enum"
+ADD
+    VALUE 'emailAndNameAndDOB';
```
YazeedLoonat, line 43: I don't remember what version of postgres we run in prod but may want to peep: https://stackoverflow.com/questions/1771543/adding-a-new-value-to-an-existing-enum-type

and double check the versioning to make sure this will work in our various production envs

api/src/services/application-flagged-set.service.ts
```diff
@@ -470,13 +468,229 @@ export class ApplicationFlaggedSetService implements OnModuleInit {
...
+            ],
+          },
+        ],
+      },
+    });
+    this.logger.warn(
+      `running duplicates check on ${outOfDateListings.length} listings`,
+    );
+    for (const listing of outOfDateListings) {
+      type PossibleFlaggedSetQuery = {
```
YazeedLoonat, line 524: should we define this type outside this function?

api/src/services/application-flagged-set.service.ts
```diff
@@ -470,13 +468,229 @@ export class ApplicationFlaggedSetService implements OnModuleInit {
...
+        FROM application_flagged_set_possibilities
+        WHERE listing_id = ${listing.id}::UUID
+        GROUP BY key, type
+        HAVING count(application_id) > 1;`;
+
+      // group all of the groups by the applicationIds
+      const reduced = flaggedApplicationGrouped.reduce(
+        (accumulated, applicationGroup) => {
+          const key = applicationGroup.applicationids
+            .sort((a, b) => a.localeCompare(b))
```
YazeedLoonat, line 542: localeCompare is real cool! I'd never seen/heard of it before

api/src/services/application-flagged-set.service.ts
```diff
@@ -470,13 +468,229 @@ export class ApplicationFlaggedSetService implements OnModuleInit {
...
+          accumulated[key] = values;
+          return accumulated;
+        },
+        {} as {
+          [key: string]: PossibleFlaggedSetQuery[];
+        },
+      );
+
+      const applicationFlaggedSetsInDB =
+        await this.prisma.applicationFlaggedSet.findMany({
```
YazeedLoonat, line 555: I think on listings with large numbers of application flagged sets this may cause problems? 
I would suggest testing against prod data locally and keeping an eye on your activity monitor to see what happens to mem usage

YazeedLoonat (commented):
Looks good! just some questions/comments

ColinBuyck:
I'm also noticing the following behavior and feeling a bit confused
Reseed and run, Close District View Apartments
Add an application with first3, last3, [user3@example.com](mailto:user3@example.com), and 01/01/1970
Run process_duplicates
I see two separate duplicate sets. Is that expected?
Screenshot 2024-08-05 at 10 42 00 AM
Screenshot 2024-08-05 at 10 41 50 AM

emilyjablonski:
To your note in the PR description about applications across multiple sets, that seems to be a significant cause of mistakes when resolving sets, as described in the multiple set resolution section in [this doc](https://docs.google.com/document/d/1wiR5INEHPrUofwLkdeRPHwj27huLQCK9tvXpaaR-Y7o/edit#heading=h.2m1o7sozhaeh). I don't see a solution to that captured in the TDD, do you have a sense of what a solution for that might be?

emilyjablonski:
@ludtkemorgan is this on hold pending the conversations in slack or is it ready for review?

ludtkemorgan:
> @ludtkemorgan is this on hold pending the conversations in slack or is it ready for review?

@emilyjablonski I'd recommend to not review this until it is decided which pattern we will be doing since it could be a pretty big refactor if we choose the second route

api/prisma/migrations/16_add_duplicates_view/migration.sql
ludtkemorgan: I checked and all of our production environments support this.

api/src/services/application-flagged-set.service.ts
ludtkemorgan: I didn't notice any spike in my memory when using the Fremont property. It does run in only a few seconds so potentially the spike is too short lived to notice. I think we'll want to monitor it though in heroku to better simulate the memory availability

api/prisma/seed-staging.ts
```diff
@@ -497,18 +498,114 @@ export const stagingSeed = async (
...
-      // has applications that are the same email
+      // has applications that are the same email and also same name/dob
       applications: [
+        await applicationFactory(),
+        await applicationFactory(),
         await applicationFactory({
-          applicant: { emailAddress: 'user1@example.com' },
+          submissionType: ApplicationSubmissionTypeEnum.paper,
+        }),
+        await applicationFactory({
```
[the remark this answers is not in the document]
ludtkemorgan: Dev is supposed to be more randomly generated whereas staging is mostly consistent every time. But honestly I never use dev for testing so didn't think to do it. I can add some more applications to dev and will think of the best way to randomly get duplicates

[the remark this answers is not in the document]
emilyjablonski: Oh interesting! I almost only ever use dev. No worries, I think this works on just staging j curious.

emilyjablonski:
Screenshot 2024-08-28 at 12 33 20 PM
Screenshot 2024-08-28 at 12 33 32 PM

emilyjablonski:
I know we talked about doing it separately, but since we're going to end up with more across multiple sets I think we need to release this with a frontend update as well. This was Em's suggestion from the workshop: (1) See if there are ways we can remove them from the UI if the data would remain accurate (2) Otherwise, disable/lock resolved duplicates in the UI. I found it to be really confusing to go into a second set with applications already marked as a duplicate. I knew I had marked them elsewhere, but when trying to see what was a duplicate in a second set, that duplicate application looked valid in that context.

api/prisma/schema.prisma
```diff
@@ -996,6 +1007,7 @@ enum YesNoEnum {
 enum RuleEnum {
   nameAndDOB
   email
+  emailAndNameAndDOB
```
YazeedLoonat, line 1010: I know we chatted about this but adding it here:
since can act as a catch all, maybe renaming this enum?

ludtkemorgan:
@emilyjablonski Good catch with that application combo. There should just be 2 sets with those 8 applications. One with `Mira and Boba` for `email` and the other with all of the others. Due to the way I'm looping through the applications it's not properly deduping some. I'll work on a fix. If there are only two sets with no overlap between them, do you still think a front-end change is required?

YazeedLoonat (changes requested):
Hey @ludtkemorgan looks really close! Found a few edge cases while messing around:


- create an application on the partner site with no email it will match everything with what I think is a null email match

- delete an application in a dupe set
   - application still shows in the dupe set list on the partner site
   - you can click into that application
   - this is true even after a re-run of the job

- create dupe sets on listings -> delete all applications -> re-run job ->dupe sets still present

- create a dupe set with 3 records -> mark it as resolved -> edit an application so it wouldn't be in the dupe set at all -> re-run job -> job errors out

ludtkemorgan:
This PR is ready to be re-reviewed

@emilyjablonski I have done a refactor to catch the scenario you mentioned. It ended up being a fairly big change so I'd recommend doing a full e2e test of it. I also added a test called `should create multiple flag sets with chaining of flags` that uses the same setup of apps you created (replacing dog names with numbers).

@YazeedLoonat 
- I added a check in the view to ignore deleted apps and those scenarios are no longer happening.
- The scenario of rerunning after resolution and editing an application I was not able to reproduce. Potentially the other changes I did
- I have fixed the null scenario. It still flags applications that are missing name or email if there is more than one of them, but that should be intended. It just shows up as "null null" on the flagged set view on  partners.

emilyjablonski:
How are you executing the job from the swagger docs with `"Traffic not from a known source"` errors?

emilyjablonski:
All of the examples from the doc look awesome! I know it would be quite a lift to set up, but I'm wondering what you think about running this on the Fremont Family set? It's a little challening to test all the edge cases

ludtkemorgan:
@emilyjablonski That guard only happens if you have the env variable `API_PASS_KEY` set. I have just been testing with not having that environment variable.

I tested it with the Albany property. But Fremont is a good idea! I'll do that

emilyjablonski (approved):
This LGTM! It catches all the cases we manually identified we had missed w Fremont. Yay!!!

YazeedLoonat:
Hey @ludtkemorgan I'm still seeing some weird behavior if email is missing or birth day is missing 

when I was looking at the database what I was seeing was the the key became `NULL` which was causing more matching and condensing of dupe sets into 1 large dupe set instead of many smaller sets like I was expecting 

after reading some documentation I think its a quirk of postgres where the `||` operator we are using in the view causes the string to go to NULL if a null value is found. https://www.postgresqltutorial.com/postgresql-string-functions/postgresql-concat-function/ 

it may be better for us to switch to the `concat` function in the view to concat strings together. 

I also think for the emails doing something like `COALESCE()` for the email creating the email key may make the null email case less all encompassing like it is right now

lemme know what ya think!

ludtkemorgan:
@YazeedLoonat I'm open to switching to concat and/or coalesce. But first I want to make sure I'm able to reproduce what you are seeing. I tried creating applications with both missing email and name/dob, I see the null entries in the database however on the frontend it appears to be handling everything correctly. Can you lay out the steps you took and what you are seeing after the job runs?

YazeedLoonat:
Hey @ludtkemorgan okay here are the steps I took:

- New Listing (no prio apps)
- submit app 1 through the partner portal with `Name Match` as both the first and last name. No email, no DOB
- submit app 2 through partner portal with `Name Match` as both the first and last name. No email, no DOB
- submit app 3 through partner portal with `Name 1` as both first and last name. email is `emailMatch@gmail.com` no DOB
- submit app 4 through partner portal with `Name 2` as both first and last name. email is `emailMatch@gmail.com` no DOB

after running the job I would expect there to be 2 different groups of duplicates. 1 for the name match and 1 for the email match, but I suspect because of the null dob playing into it its creating 1 mega group

YazeedLoonat (approved):
LGTM

Issue #4056, opened 2024-04-28
Lotteries - Export with filtered preference sheets (BE)


emilyjablonski:
Blocked by: https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4055

As a user running the lottery on Doorway on a listing with prefrences,
I want to generate a randomized AND preference rank for each valid application, and export it into a spreadsheet
So that I can use this spreadsheet to review lottery results

Notes:
- The additional preference bucket should follow the format of the preference bucket tabs [here](https://docs.google.com/spreadsheets/d/1dX9ehAq3pdMpmhNq1arWV2bwjHBFMy6xAbcPGhLVHOY/edit#gid=1191380628)
- The preference bucket tabs should just be the raw rank tab, filtered for that specific preference. We don't need to do any additional randomization within these buckets. 
- The "Preference Bucket Rank" column should just be 1-the number of applicants in that column. It is just the rank of each  applicant within that preference. 
- The FE & BE should be connected as a part of this ticket

AC: 
- When I generate a lottery for a listing with a preference:
- I want to be able to export my results into a spreadsheet by clicking the "export" button from the lottery tab.
- I want to to download a spreadsheet instead of a CSV
- I want the raw rank tab to follow the AC [here](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4055)
- I want there to be a tab for each prefrence, which is the raw rank sheet filtered down to only applicants that claimed that preference
- I want there to be a new "preference rank" column, that is just the rank of the applicant within the prefrence bucket

ludtkemorgan:
@sarahlazarich this is in Bloom core and can be tested. I created a listing with preferences and ran the lottery here https://partners.bloom.exygy.dev/listings/0967a809-6ca6-414c-87e5-87a9e4b3c612.

PR #4233, opened 2024-08-01
feat: adding preference sheets to lottery export

YazeedLoonat:
This PR addresses #4056 

## Description
This adds the preference sheets to the lottery export as well as a few other things:

- preference sheets
    - added
- demographics
    - now based on role just like application csv export
- columns
    - added preference rank
    - still need to double check column order
- permissions
    - UI is now hooked up to do a partner export

- head to the partners site
- on a closed listing with applications and preferences on the listing (and to make it easier claimed preferences on at least some of the applications)
- generate the lottery results
- get the spreadsheet export

you should notice individual sheets for each preference as well as the original raw sheet

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/lottery.service.ts
+++ b/api/src/services/lottery.service.ts
@@ -639,11 +639,9 @@ export class LotteryService {
         (header) => header.header === 'Raw Lottery Rank',
       );
 
-      res[indx].header = `${preference.name} Rank`;
-
       res.splice(indx + 1, 0, {
         key: 'Raw Lottery Rank',
-        header: 'Raw Lottery Rank',
+        header: `${preference.name} Rank`,
       });
     }
     return res;
--- a/api/src/services/lottery.service.ts
+++ b/api/src/services/lottery.service.ts
@@ -289,7 +289,7 @@ export class LotteryService {
       }-${new Date().getTime()}.zip`,
     );
 
-    await this.createLotterySheet(workbook, {
+    await this.createLotterySheets(workbook, {
       ...queryParams,
     });
 
@@ -320,9 +320,9 @@ export class LotteryService {
    *
    * @param filename
    * @param queryParams
-   * @returns generates the lottery sheet
+   * @returns generates the lottery sheets
    */
-  async createLotterySheet<QueryParams extends ApplicationCsvQueryParams>(
+  async createLotterySheets<QueryParams extends ApplicationCsvQueryParams>(
     workbook: Excel.Workbook,
     queryParams: QueryParams,
   ): Promise<void> {
@@ -388,7 +388,7 @@ export class LotteryService {
     );
 
     const mappedApps = mapTo(Application, applications);
-    await this.lotteryExportHelper(
+    await this.generateSpreadsheetData(
       workbook,
       mappedApps,
       columns,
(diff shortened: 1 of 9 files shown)


mcgarrye (dismissed):
Raw rank on the preference sheet is the same for all applicants
Screenshot 2024-08-06 at 10 36 54 AM

ludtkemorgan:
@mcgarrye the bug with all showing the same rank has been fixed. There was a `=` instead of `===`. I'll also add some tests around that

emilyjablonski:
The [sample export](https://docs.google.com/spreadsheets/d/1dX9ehAq3pdMpmhNq1arWV2bwjHBFMy6xAbcPGhLVHOY/edit?gid=520192447#gid=520192447) has a slightly different column order with the raw rank coming first

api/src/services/lottery.service.ts
```diff
@@ -426,22 +447,46 @@ export class LotteryService {
...
     queryParams: ApplicationCsvQueryParams,
     forLottery = false,
+    preference?: IdDTO,
   ): Promise<void> {
-    // create raw rank spreadsheet
-    const rawRankSpreadsheet = workbook.addWorksheet('Raw');
-    rawRankSpreadsheet.columns = this.buildExportColumns(csvHeaders);
+    // create a spreadsheet. If the preference is passed in use that as a title otherwise 'raw'
+    const spreadsheet = workbook.addWorksheet(
+      preference ? preference.name : 'Raw',
```
emilyjablonski, line 463: The sample export has the title as `Raw Lottery Rank`

sites/partners/src/pages/listings/[id]/lottery.tsx
```diff
@@ -44,7 +44,21 @@ const Lottery = (props: { listing: Listing }) => {
...
-  const { onExport, exportLoading } = useLotteryExport(listing?.id)
+
+  const listingJurisdiction = profile?.jurisdictions?.find(
+    (jurisdiction) => jurisdiction.id === listing?.jurisdictions.id
+  )
+
+  const includeDemographicsPartner =
+    profile?.userRoles?.isPartner && listingJurisdiction?.enablePartnerDemographics
+
+  const { onExport, exportLoading } = useLotteryExport(
```
Comment (emilyjablonski) [inline sites/partners/src/pages/listings/[id]/lottery.tsx:55]: Couldn't we do this all on the backend by looking at the requesting user?

Comment (ludtkemorgan) [reply sites/partners/src/pages/listings/[id]/lottery.tsx:55]: This is where we call the backend. Are you referring to the includeDemographics logic? The include demographics query param follows the same pattern as the csv export. We could move it to the backend but then we would also need to look up the jurisdiction to know if partnerdemographics is enabled

Comment (emilyjablonski) [reply sites/partners/src/pages/listings/[id]/lottery.tsx:55]: Yeah we don't have to make this change here, but I think it makes more sense to do all of that on the backend

api/src/services/lottery.service.ts
ludtkemorgan: fixed!

mcgarrye: It seems like you changed the wrong header name :/
Screenshot 2024-08-12 at 11 43 23 AM

ludtkemorgan:
@mcgarrye nice catch. I have fixed the columns to have the right data

Issue #4053, opened 2024-04-28
Lotteries - Public list view (FE)


emilyjablonski:
As an applicant in my account view, 
I want to be able to see all the listings I've applied to in my account with updated statuses 
So that I have more insight into the status of my application. 

As an applicant in my account view, 
I want to be able to see some key dates surfaced on listings in certain statuses,
So that I have more insight into what important events are coming up on the listing. 

As an applicant in my account view, 
I want to be able to see different buttons based on the status of the listing, 
So that I am able to access key information for each listing. 

[Design](https://www.figma.com/design/PAeCVANDmGdTz2Y53ouErM/Lotteries?node-id=3345-7278&t=XeYvwRFLR4nl0xia-0)

AC: 

The following listing statuses are surfaced in my account: 
- "Accepting applications" should map to an open listing 
- "Applications closed" should map to a closed listing. On lottery listings, this should be before a lottery is published.
- On a lottery listing, the status should say read "Lottery run" after a lottery is **published** 

These key dates are surfaced in my account, depending on the status of the listing: 
- Application close date should be surfaced on listings with the status "accepting applications" 
- Before a lottery is published, the lottery date from the listing should appear on lottery listings with "applications closed" status 
- After a lottery is published, the lottery published date should appear on lottery listings with the "Lottery run" status 

These buttons are surfaced in my account, depending on the status of the listing: 
- When a status is set to “Accepting applications”, two buttons should be visible: 1) View your application, 2) See listing
- When a status is set to “Lottery run”, two buttons should be visible: 1) View lottery results, 2) See listing. If the lottery is retracted, the “View lottery results” button should disappear .  
- In Doorway only, when a status is set to “Applications closed”, only one button should be visible: 1) See listing. The application is no longer visible to the user. In Core, the user should still be able to view their application. 

QA Notes: https://bloom.exygy.dev/, https://partners.bloom.exygy.dev/
This can be tested by signing into the public site and applying to a few different listings. Then, update the listings to show as open without a due date, open with a due date, closed without a lottery date, closed with a lottery date, and lottery ran.

emilyjablonski:
Will want to store the date of lottery results publication on a listing instead of having to get that from the activity log

emilyjablonski:
In Doorway you can only view your application on open listings - for a closed listing, you won't have the `view application` button in the my applications page, but will be able to see it in core

sarahlazarich:
What should the empty state be for tabs? 
[...]

PR #4237, opened 2024-08-07
feat: application side nav

ColinBuyck:
This PR addresses https://github.com/metrotranscom/doorway/issues/437

## Description

This PR utilizes the tab component so a user can filter the applications they would like to see. 

Note One: This implementation results in applications being re-fetched each time they select a new category. Based on the performance of this, I am wondering if there is a more reduced fetch call that only provides the required information.

Note Two: I made some assumptions on the mobile experience of this section, let me know what y'all think!

Note Three: I was looking into expanding the test coverage and was considering ctypress tests but know we have been cautious in adding more cypress. Additionally, seeing how limited my-applications.spec.ts is would it be worth it to expand it? Planning on expending integration test coverage in #4053 

This can be tested by signing into the public site and first testing the empty states of the different sections. If there are no applications at all, the message will be to browse more listings but if you have applied but none of the applications are in the selected state, it will be a custom message. Then apply to multiple listings. Lastly, change the status of those listings and observe the applications move to the corresponding section

## Author Checklist:

## Review Process:


sites/public/src/components/account/StatusItem.tsx
```diff
@@ -24,7 +24,7 @@ interface StatusItemProps {
 
 const StatusItem = (props: StatusItemProps) => {
   return (
-    <Card.Section className={accountStyles["account-card-applications-section"]}>
+    <Card.Section className={applicationsViewStyles["account-card-applications-section"]}>
```
ColinBuyck, line 27: Is it preferred to duplicate this class instead of this crossover @emilyjablonski ?

shared-helpers/src/locales/general.json
```diff
@@ -10,10 +10,18 @@
   "account.createAccount": "Create Account",
   "account.errorFetchingApplications": "Error fetching applications",
   "account.haveAnAccount": "Already have an account?",
+  "account.allMyApplications": "All my applications",
```
emilyjablonski, line 13: Do you mind alphabetizing these? W lines selected, I use command + shift + p --> sort lines ascending in VSCode

sites/public/src/components/account/ApplicationsView.module.scss
```diff
@@ -0,0 +1,44 @@
+.applications-section-background {
+  background: var(--seeds-color-gray-300);
```
emilyjablonski, line 2: Where possible it's preferable to use semantic tokens for theming, for example this could be `--seeds-bg-color-canvas` which is defaulted in our theme to `--seeds-color-gray-300`

sites/public/src/components/account/ApplicationsView.module.scss
```diff
@@ -0,0 +1,44 @@
...
+  border-color: var(--seeds-color-gray-450);
```
emilyjablonski, line 3: `--seeds-border-color`

emilyjablonski:
Not sure if it matters, but in the lottery designs, the order of the last two tabs is switched, the design goes all --> lottery run --> applications closed --> accepting applications

emilyjablonski:
When I test this in the deploy preview, I relatively quickly and am getting the issue where the listing data stops appearing, and it looks like what is happening is that I'm getting throttled. We may need to update the threshold. I'm interested in separately moving forward w a smaller view, created a ticket [here](https://github.com/bloom-housing/bloom/issues/4240) 🏹

sites/public/src/components/account/StatusItemWrapper.tsx
emilyjablonski, line 1: Ty!

sites/public/src/components/account/StatusItem.tsx
emilyjablonski: This lgtm!

emilyjablonski:
Screenshot 2024-08-07 at 7 34 30 PM
Seeing this button when I am already on that page felt confusing, we maybe just need to remove it from that view.

emilyjablonski (changes requested):
Some feedback above!

sites/public/src/components/account/ApplicationsView.tsx
```diff
@@ -0,0 +1,230 @@
...
+        setLoading(false)
+      })
+      .catch((err) => {
+        console.error(`Error fetching applications: ${err}`)
+        setError(err)
+        setLoading(false)
+      })
+  }, [applications, listLoading, listingsService])
+
+  const selectionHandler = (index: number) => {
```
jaredcwhite, line 114: FWIW, this definitely leads me to believe that what we really need is the to-be-developed Seeds Menu component. The Tabs component is intended to switch views already loaded on-page without changing routes. I'm not sure if the a11y on this quite makes sense.

sites/public/src/components/account/ApplicationsView.tsx
```diff
@@ -0,0 +1,230 @@
...
+    if (profile && loading) {
+      pushGtmEvent<PageView>({
+        event: "pageView",
+        pageTitle: `My Applications - ${ApplicationsFilterEnum[props.filterType]}`,
+        status: UserStatus.LoggedIn,
+      })
+      applicationsService
+        .list({ userId: profile.id })
+        .then((apps) => {
+          apps?.items?.length > 0 ? setApplications(apps.items) : setLoading(false)
```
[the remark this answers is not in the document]
ColinBuyck: This was carried over from the previous implementation and controls the loading overlay. I think the idea is that if there are no applications, we don't need to wait for the listings fetch to occur and we can replace the overlay with the message but if there are applications we should keep the loading overlay until the listings fetch completes.

shared-helpers/src/locales/general.json
ColinBuyck: Learn something new everyday

sites/public/src/components/account/ApplicationsView.module.scss
ColinBuyck: Oh nice, I haven't noticed the canvas one before!

ColinBuyck:
Thanks for the review @emilyjablonski ! I've fixed most of these but just have a couple notes. I changed the throttle envs to match hba prod but would like to bring it up to the team on Monday to discuss what value seems right. I'm a little worried about undermining security (if the throttle limit extends to prod) based on this one feature. Secondly, I explored the view thing we discussed and should've noted it in the description but my understanding is that we're using views to make complex queries (ie. joins) easier rather than get a subset of fields in a single table. For example, "fundamentals" view still returns all the fields below
Screenshot 2024-08-08 at 2 32 04 PM

Might still be worth it to do a custom query like getPublicUserEmailInfo() if we think the number of fields is driving the slowdown. Alternatively if its the number of calls on the db we could batch the ids and then make a single call. Orrr explore using context

Finally, do you mind providing the steps to repdouce the "See all applications" state. That should read as "See listings" but I am also noticing the lack of numbers showing so I'm wondering if its an edge case caused by the throttling

ColinBuyck:
@emilyjablonski Also updating to 1 review needed since some of this logic is altered for the next lottery FE ticket.

emilyjablonski (approved):
🥣

sites/public/src/components/account/ApplicationsView.tsx
```diff
@@ -0,0 +1,229 @@
...
+            open++
+            if (props.filterType === ApplicationsFilterEnum.Open) displayApplications.push(app)
+          } else if (app.fullListing?.lotteryStatus === LotteryStatusEnum.publishedToPublic) {
+            lottery++
+            if (props.filterType === ApplicationsFilterEnum.Lottery) displayApplications.push(app)
+          } else {
+            closed++
+            if (props.filterType === ApplicationsFilterEnum.Closed) displayApplications.push(app)
+          }
+        })
```
ludtkemorgan, line 100: I don't know if this is the spot that is causing the issue. But the placement of applications seems to be incorrect. For example in this screenshot the "Accepting applications" shows a count of 1 but the application is showing up in the "Applications closed"

image

shared-helpers/src/locales/general.json
ludtkemorgan: One call out here, since this whole file was alphabetized we will have to be careful when pull this over to the jurisdictions. There is a chance that we might not catch if one of the moved keys is a different translation (this has happened in the past). And I'd also recommend that the alphabetizing is also run on those jurisdiction's repos so that any keys that are unique to that repo are also placed in the right place.

sites/public/src/components/account/ApplicationsView.tsx
ludtkemorgan: To add on to this, do we need to fully change routes when switching tabs? In the past when I've worked on a page like this I used a context to store the data so when switching between tabs I just grab the appropriate data. It's fairly performant, but having the whole page load when you click on any of the tabs on the left side makes the page feel a little slower and jarring than it needs to be.

And if we need to keep track of what tab is selected in the URL we could use a query param rather than a path param.

sites/public/src/components/account/ApplicationsView.tsx
ludtkemorgan: This has been fixed

ludtkemorgan (approved):
Thanks for making those changes. Colin and I talked about this offline, but the throttling issue is due to the large number of api calls. It is recommended to create a new backend endpoint that returns the appropriate data rather than calling it once for every application and every listing. This will be addressed at a later point

Issue #4248, opened 2024-08-12
Minimize BE calls in Lottery Public FE experience


ColinBuyck:
With the new applications filtering, the existing application/listing fetch performance is being exacerbated by re-fetching the data each time the user selects a different type of application to view. For example, 10 applications would result in 11 calls to the BE (1 to list applications and 10 to fetch each associated listing). Additionally, a ton of additional data is being sent that isn't necessary and it requires significant FE business logic to map the results to the user's experience.

This ticket is intended to simplify this process to a single call and minimize the FE business logic so that user's have a more performant experience and they're significantly less likely to encounter throttling in their use of this feature.

QA (Handled by Eng @ColinBuyck )
This can be tested by signing into an account, applying to more than one listing, and seeing in the network tab that only one call is being made no matter how many listings there are.

PR #4252, opened 2024-08-15
feat: feature flag and new endpoint

ColinBuyck:
This PR addresses #4248 

## Description

This PR adds **feature flagging** to hide or show the side nav and creates a custom call to reduce the load time for accounts with numerous applications.

Add SHOW_LOTTERY=TRUE to the public side

One note that you will see ApplicationsIndexEnum and ApplicationsFilterEnum are very similar but based on the current implementation requiring tabs, offline convo landed on using both would be best to maintain strong typing on the BE and easy tab selection based on index + better readability on the FE.

This can be tested by signing into an account, applying to more than one listing, and seeing in the network tab that only one call is being made no matter how many listings there are. Additionally, you can rerun the application with the new feature flag set to false, and you will no longer see the Lottery run tab but everything else should work as expected.
## Author Checklist:

## Review Process:


api/src/dtos/applications/public-apps-view-params.dto.ts
```diff
@@ -0,0 +1,24 @@
...
+  userId: string;
+
+  @Expose()
+  @ApiPropertyOptional({
+    enum: ApplicationsFilterEnum,
+    enumName: 'ApplicationsFilterEnum',
+    example: 'all',
+  })
+  @IsEnum(ApplicationsFilterEnum, { groups: [ValidationsGroupsEnum.default] })
+  filterType?: ApplicationsFilterEnum;
```
ludtkemorgan, line 23: This isn't a problem because of this exact line, just commenting here as it is relevant to it. With the frontend passing the filtering type we still have the small problem of the page reloading whenever you click a different status on the left nav bar. This causes a flash of all of the numbers and it's not as smooth of an experience.

Would it be possible to always return all applications and do the filtering on the frontend? Especially now that the returned type from the backend is a lot smaller we don't have too much of a concern of there being too much data. Or in other words we only call the backend on initial load of the page and then when we go to the different tabs it filters for the correct applications. I do remember you talking about there being limitations on the UI component and accessibility so would be curious to get your thoughts on this

ludtkemorgan (changes requested):

api/src/services/application.service.ts
```diff
@@ -349,6 +358,74 @@ export class ApplicationService {
...
+    this will the most recent application the user has submitted
+  */
+  async publicAppsView(
+    params: PublicAppsViewQueryParams,
+    req: ExpressRequest,
+  ): Promise<PublicAppsViewResponse> {
+    const user = mapTo(User, req['user']);
+    if (!user) {
+      throw new ForbiddenException();
+    }
```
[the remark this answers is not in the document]
ColinBuyck: Just added the authorize action calls so that should be covered now! As for the second question, I wondered that too but ended up following existing patterns like the most-recent-application-query-param passes in the userId. I have no problem changing it but is there context behind that previous decision @YazeedLoonat?

api/src/dtos/applications/public-apps-view-params.dto.ts
ColinBuyck: Totally agree on it not being super smooth. I think if the long term plan was to stick with the Tabs component, then I would go for the FE filtering experience but since it is a temporary solution until the menu component is ready (which relies on links) and Jesse is advising that separate pages are the a11y best practice, I'm inclined to keep this approach as is. I did also have some conversations with @emilyjablonski about this decision if there's any thoughts she'd like to add here.

emilyjablonski: Yeah we've had a few chats w design about filtering on the same page vs having separate pages, and the separate pages was requested from design for the reasons Colin mentioned above

emilyjablonski: Agreed though that the filtering would be a better UX

Issue #4054, opened 2024-04-28
Lotteries - Public lottery view (FE)


emilyjablonski:

emilyjablonski:
Because the learn more links go to the help center in Doorway, those buttons will not exist in core

emilyjablonski:
QA Notes
When logged in as an admin, you can visit [this link](https://deploy-preview-4255--bloom-exygy-dev.netlify.app/account/application/af9e08a3-df18-411b-ad24-59f34fba03c4/lottery-results) in dev to QA. You could alternatively do the full flow manually, by creating a listing opted into the system lottery with a digital application, sign in as a public user and submit an application to that listing, and then go through the lottery run, release, and publish flow in partners.

It will just have mock data and will not reflect the listing or the application.

Known differences from the [Figma](https://www.figma.com/design/PAeCVANDmGdTz2Y53ouErM/Lotteries?node-id=3684-14239&t=YRyjWjDk5k5biian-0) due to core vs Doorway theming:
* Font weight of main header
* Vertical and horizontal padding in the card sections
* Primary color
* Font family

Known differences from the Figma because of Seeds details:
* Exclamation icon on the message is centered instead of floating to the top
* Heading section of the Card does not have more horizontal padding than the content

sarahlazarich:
This looks good! Only think I noticed is that the buttons are linking to https://www.exygy.com/ here, which is all good, we just want to be sure to link to the Help Center / Get Started page for Doorway

PR #4255, opened 2024-08-15
feat: public account lottery view

emilyjablonski:
This PR addresses #4054

## Description

Adds the lottery results page for the public site for an individual application.

When logged in as admin@example.com, you can visit [this link](https://deploy-preview-4255--bloom-exygy-dev.netlify.app/account/application/af9e08a3-df18-411b-ad24-59f34fba03c4/lottery-results) in dev to QA. You could alternatively do the full flow manually, by creating a listing opted into the system lottery with a digital application, sign in as a public user and submit an application to that listing, and then go through the lottery run, release, and publish flow in partners, and re-log in as the public user and view your application.

It will just have mock data and will not reflect the listing or the application.

I also did a small refactor to reduce duplication that pulled out some of the account application page blocks.

Known differences from the [Figma](https://www.figma.com/design/PAeCVANDmGdTz2Y53ouErM/Lotteries?node-id=3684-14239&t=YRyjWjDk5k5biian-0) due to core vs Doorway theming:
* Font weight of main header
* Vertical and horizontal padding in the card sections
* Primary color
* Font family

Known differences from the Figma because of Seeds details:
* Exclamation icon on the message is centered instead of floating to the top
* Heading section of the Card does not have more horizontal padding than the content

## Author Checklist:

- [ ] Added tests covering the changes <-- I think integration tests here will become relevant w the backend changes

## Review Process:


sites/public/src/pages/account/application/[id]/lottery-results.tsx
```diff
@@ -0,0 +1,169 @@
...
+        .then((app) => {
+          setApplication(app)
+          listingsService
+            ?.retrieve({ id: app.listings.id })
+            .then((retrievedListing) => {
+              setListing(retrievedListing)
+            })
+            .catch((err) => {
+              console.error(`Error fetching listing: ${err}`)
+            })
```
Comment (ludtkemorgan) [inline sites/public/src/pages/account/application/[id]/lottery-results.tsx:35]: Taking a look at the network calls, it appears like the listing name and id are returned in the application retrieve call. So this second call is unnecessary unless you need more than just those two fields

ludtkemorgan (approved):
This looks good, just one comment about the network call. But that could also be handled when connecting up the backend data

sites/public/src/pages/account/application/[id]/lottery-results.tsx
Comment (emilyjablonski) [reply sites/public/src/pages/account/application/[id]/lottery-results.tsx:35]: Ah yeah this was copy paste from the application page - we will also need the number of units as part of the BE ticket here.

sites/public/src/pages/account/application/[id]/lottery-results.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    <p className={styles["raw-rank"]}>57</p> {/* TODO: Plug in BE data */}
+                  </Card.Section>
+                  <Card.Section divider={"flush"}>
+                    <div>
+                      <p>{t("account.application.lottery.rawRank")}</p>
+                    </div>
+                    <div>
+                      <Button
+                        className={styles["section-button"]}
+                        href={"https://www.exygy.com"}
```
Comment (mcgarrye) [inline sites/public/src/pages/account/application/[id]/lottery-results.tsx:115]: Where should this be pointing?

mcgarrye (approved):
Looks good! Only wondering about the redirects, if we don't have the url yet can we add in some TODO's?

sites/public/src/pages/account/application/[id]/lottery-results.tsx
Comment (ludtkemorgan) [reply sites/public/src/pages/account/application/[id]/lottery-results.tsx:35]: Good call out. I'm cool with refactoring this when the backend endpoint is flushed out!

sites/public/src/pages/account/application/[id]/lottery-results.tsx
Comment (emilyjablonski) [reply sites/public/src/pages/account/application/[id]/lottery-results.tsx:115]: This value will be dependent on the jurisdiction, so we'll update it in Doorway like we have for the strings that say "Admins" vs "Doorway admins"

emilyjablonski:
Tysm, added comments!

Issue #3654, opened 2023-09-25
Save & Continue Flow Change


sarahlazarich:
https://www.notion.so/exygy/Preparing-an-accessiblity-plan-63733ba5f272471084773828d5b5aaa4

sarahlazarich:
Listing name and jurisdiction need to be filled out to save as a draft, but that shouldn't be an issue 
-  If someone has a published listing, and they make a change, that would fail a published listings validating. Published listings have way more validation.
-  Should we even allow this flow on a published listing? What do we do about that?
Detecting unsaved changes could be tricky 
The listing name is in the header, so if that's changed it will need to update there as well

sarahlazarich:
Per @eajensenwa  - the design recommendation is to not allow users to save in place on a published listing.  If we get feedback in the future that users need to do so for whatever reason, we can revisit.

sarahlazarich:
Per @eajensenwa  - the design recommendation is to not allow users to save in place on a published listing.  If we get feedback in the future that users need to do so for whatever reason, we can revisit.

sarahlazarich:
This should be in every state besides a closed listing

sarahlazarich:
lgtm!

PR #4256, opened 2024-08-16
feat: save and continue

jaredcwhite:
This PR addresses #3654 

## Description

This PR changes the "Save & Exit" button so it just reads "Save", and clicking it will save the listing and remain in the edit state. In addition, the "Cancel" button has been renamed "Exit" and it displays a prompt to the user to either save or discard changes.

Try saving or exiting listings in various stages (new, unpublished, published, etc.) and verify the behavior matches the AC/Figma.

## Author Checklist:

## Review Process:


sites/partners/cypress/e2e/default/03-listing.spec.ts
```diff
@@ -49,6 +49,26 @@ describe("Listing Management Tests", () => {
...
+    cy.getByID("listingsExitButton").click()
+    cy.getByID("listing-save-before-exit-dialog-content").contains(
+      "Do you want to save your changes before you exit?"
+    )
+    cy.getByID("saveBeforeExitDiscard").click()
+    cy.contains("Test - error messaging")
+    cy.getByID("listingEditButton").contains("Edit").click()
+    cy.getByID("name").clear()
+    cy.getByID("name").type("Test - error messaging DISCARD")
+    cy.getByID("listingsExitButton").click()
```
emilyjablonski, line 65: Could we use getByText for selectors like this where possible?

sites/partners/src/components/listings/ListingFormActions.tsx
```diff
@@ -74,14 +78,19 @@ const ListingFormActions = ({
   }, [listing])
 
   const actions = useMemo(() => {
+    // TODO: should be Exit, show Save confirmation dialog
```
emilyjablonski, line 81: Is this TODO valid?

sites/partners/src/components/listings/ListingFormActions.tsx
```diff
@@ -314,8 +324,9 @@ const ListingFormActions = ({
           type="button"
           className="w-full"
           onClick={() => {
-            // TODO throw a modal
-            submitFormWithStatus(true, ListingsStatusEnum.active)
+            // NOTE: confused by the following todo. Doesn't this show a modal? -JW
+            // TODO: throw a modal -EJ
```
emilyjablonski, line 328: Another q on the TODO here

sites/partners/src/components/listings/PaperListingForm/dialogs/CloseListingDialog.tsx
```diff
@@ -0,0 +1,54 @@
+import React from "react"
```
emilyjablonski, line 1: I like these living in this separate folder!

emilyjablonski:
Screenshot 2024-08-22 at 11 34 47 AM
When I get a backend error to throw I'm still getting the success toast

emilyjablonski:
Screenshot 2024-08-22 at 11 40 15 AM
I feel like the behavior here is preferred, but if a listing is already live and I click save, I get this modal and then get redirected to the edit page instead of this new save flow.

emilyjablonski (changes requested):
Would love to fix the listing name bug in particular but the rest are more questions / thoughts!

jaredcwhite:
Thanks for the review @emilyjablonski! Working through the various issues/questions now.

sites/partners/src/components/listings/ListingFormActions.tsx
jaredcwhite: will delete!

sites/partners/cypress/e2e/default/03-listing.spec.ts
jaredcwhite: Hmm, well for a simple string like "Exit" I'm not sure if we want to assume there's only one "Exit" on the page…?

jaredcwhite:
Just pushed up fixes for the listing name in header not changing and the error alert not disappearing.

jaredcwhite:
> Screenshot 2024-08-22 at 11 40 15 AM
> I feel like the behavior here is preferred, but if a listing is already live and I click save, I get this modal and then get redirected to the edit page instead of this new save flow - but I also think that makes sense, I just don't see it covered in the AC.

I think the AC was unspecified with regard to confirmation modals. I think this makes sense too, but we could ask for design QA to verify.

emilyjablonski (approved):
🛹

PR #4262, opened 2024-08-20
fix: remove db dependency from unit tests

ludtkemorgan:
There are several backend unit tests that happen to be working because we have a database running while the tests are running. If this were to not be the case these tests would start failing and there is even a failing test semi-related to this issue.

This PR mocks out all db calls that are needed in tests that are not already mocked out. 

To test:
- These are already passing in circleci, but to test it locally to make sure all are caught you can change your .env file to point to a non-existent DB. I did this by changing the port on the DB url


ColinBuyck (approved):
LGTM 🍵

Issue #4231, opened 2024-08-01
Allow a due date for all listing types


sarahlazarich:
Currently, on listings with opportunity type of open waitlist, the due date is removed. Partners have been doing a workaround to add one, but if that's a need we shouldn't hide this field. 

AC: 
- A Partner user is able to add a due date for opportunity type of open waitlist. It should be available but  not be required.

KrissDrawing:
QA notes: question about due date is removed. Due date input is always enabled.

sarahlazarich:
@KrissDrawing need to have some more conversation with Design on this before we release.

PR #4264, opened 2024-08-21
fix: allow a due date for open waitlist

KrissDrawing:
This PR addresses #4231 

## Description

Adds additional condition to display `Is there an application due date?` question when open waitlist selected.
Before as a workaround we could just switch to `available units` set due date, and switch to `open waitlist`. Now it can be added from open waitlist

On partners switch `What is the listing availability?` to open waitlist. Now in `Rankings & Results` we should see `Is there an application due date?` question

## Author Checklist:

## Review Process:


mcgarrye (approved):
LGTM

emilyjablonski:
The ticket is titled `Allow a due date for all listing types` so I'm curious if there should be no if check to show the field at all?

mcgarrye:
Was the newest change verified with product?

KrissDrawing:
@mcgarrye Yes i asked Sarah, it should be visible all the time

sites/partners/src/components/listings/PaperListingForm/sections/RankingsAndResults.tsx
```diff
@@ -273,6 +249,28 @@ const RankingsAndResults = ({ listing, isAdmin }: RankingsAndResultsProps) => {
             </Grid.Row>
           </>
         )}
+        <Grid.Row columns={2} className={"flex items-center"}>
```
mcgarrye, line 252: I think this logic should be moved to `sites/partners/src/components/listings/PaperListingForm/sections/ApplicationDates.tsx`. This question gets hidden when Lottery is selected and is no longer tied to FCFS so should be with the actual date and time fields at the bottom

emilyjablonski: @mcgarrye I think that would be a design change and Em has been checking out the UX of the toggled on fields already, I think this is a great point that would be part of separate work

mcgarrye: Very fair. Still should appear when lottery is selected though if we aren't moving it

emilyjablonski: Very sorry for the back and forth on the AC here @KrissDrawing! After chatting w Sarah and Em, the desired behavior is to remove the `Is there an application due date?` question and always have the due date field be visible and enabled.

KrissDrawing: Not a problem, if checks pass, it should be ready to review

emilyjablonski (approved):
thank you!

Issue #4242, opened 2024-08-08
Listing Cannot Be Edited After Lottery Result PDF


mcgarrye:
**Describe the bug**
A listing that has had a lottery result pdf uploaded can no longer be edited after that.
This is because we do not delete and recreate the asset in the listing service update for the listingEvents.

**To Reproduce**
1. Go to a closed lottery listing that is opted out of lottery service
2. Upload lottery results pdf
3. Attempt to make an edit to the listing

**Expected behavior**
Edit should succeed.

**Additional context**

**QA NOTES**
will be available in: https://partners.bloom.exygy.dev/sign-in 
Create or edit a listing such that: 
- it is closed
- it is set to lottery
- the lottery not run through the partner site (you will manually upload lottery results)
- manually uploaded lottery results

go back in to edit the listing -> attempt to save -> it should now save

sarahlazarich:
This is a Prisma issue - add an extra step to delete the PDF and re-add it or make it so it doesn't create a new PDF every time. 

This is maybe a misunderstanding of how jurisdictions are opening/closing listings? Why would they re-open that? Went over why

sarahlazarich:
this looks good!

PR #4277, opened 2024-08-27
fix: saving listing with uploaded lottery

YazeedLoonat:
This PR addresses #4242 

## Description
This fixes a bug where updating a listing with lottery results uploaded was not working

Create or edit a listing such that: 
- it is closed
- it is set to lottery
- the lottery not run through the partner site (you will manually upload lottery results)
- manually uploaded lottery results

go back in to edit the listing -> attempt to save -> it should now save

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/listing.service.ts
+++ b/api/src/services/listing.service.ts
@@ -1086,17 +1086,21 @@ export class ListingService implements OnModuleInit {
         });
       }),
     );
-    await Promise.all(
-      assetIds.map(async (assetData) => {
-        if (assetData.fileId) {
-          await this.prisma.assets.delete({
-            where: {
-              id: assetData.fileId,
-            },
-          });
-        }
-      }),
-    );
+    const fileIds = assetIds.reduce((accum, curr) => {
+      if (curr.fileId) {
+        accum.push(curr.fileId);
+      }
+      return accum;
+    }, []);
+    if (fileIds.length) {
+      await this.prisma.assets.deleteMany({
+        where: {
+          id: {
+            in: fileIds,
+          },
+        },
+      });
+    }
   }
 
   /*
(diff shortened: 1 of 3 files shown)


api/test/unit/services/listing.service.spec.ts
```diff
@@ -2391,6 +2399,15 @@ describe('Testing listing service', () => {
         id: 'example id',
         name: 'example name',
       });
+      prisma.listingEvents.findMany = jest.fn().mockResolvedValue([]);
+      prisma.listingEvents.update = jest.fn().mockResolvedValue({
+        id: 'example id',
+        name: 'example name',
+      });
+      prisma.assets.delete = jest.fn().mockResolvedValue({
```
emilyjablonski, line 2407: Could we add a test to this PR? Maybe just expecting this to be called?

api/src/services/listing.service.ts
```diff
@@ -1062,6 +1058,47 @@ export class ListingService implements OnModuleInit {
...
+    await Promise.all(
+      assetIds.map(async (assetData) => {
+        if (assetData.fileId) {
+          await this.prisma.assets.delete({
+            where: {
+              id: assetData.fileId,
+            },
+          });
+        }
+      }),
```
ludtkemorgan, line 1098: Could this be done as a `deleteMany` where id is in assetId list?

api/src/services/listing.service.ts
```diff
@@ -1207,22 +1246,25 @@ export class ListingService implements OnModuleInit {
...
-                  assets: event.assets
-                    ? {
-                        create: {
-                          ...event.assets,
-                        },
-                      }
-                    : undefined,
-                })),
+                create: dto.listingEvents.map((event) => {
+                  return {
```
ludtkemorgan, line 1250: question: Was the syntax without the return causing problems? Or is this just a preferred format?

api/src/services/listing.service.ts
```diff
@@ -1134,6 +1171,8 @@ export class ListingService implements OnModuleInit {
       dto,
       'listingsBuildingAddress',
     );
+    // Delete all assets tied to listing events before creating new ones
+    await this.updateListingEvents(dto.id);
```
ludtkemorgan, line 1175: This is the way we are handling other sub tables like addresses and paper applications, so I am cool with just deleting and recreating every time. But did you look into only creating if it doesn't exist in the db? Similar to how we do listingsBuildingSelectionCriteriaFile? The logic can get pretty gnarly that way though

ludtkemorgan (commented):
Mostly just questions

api/test/unit/services/listing.service.spec.ts
YazeedLoonat: I ended up writing a separate set of tests for this function, lemme know if that is okay or if you would prefer the tests being part of this

api/src/services/listing.service.ts
YazeedLoonat: nice call out! updated!

api/src/services/listing.service.ts
YazeedLoonat: nah I was trying to figure out what was going on with some console logs and didn't put things back properly

api/src/services/listing.service.ts
YazeedLoonat: yeah I think we should address that separately, this is something I think I mentioned but with prisma extensions ( https://www.prisma.io/docs/orm/prisma-client/client-extensions )

we could write a custom query that handles that for us, so we can sort simplify that into helper functions basically 

but I think we should do that in a separate ticketo with a spike to test out extensions a bit

emilyjablonski (approved):
➰

PR #4275, opened 2024-08-26
feat: public lottery view, rank data

emilyjablonski:
This PR addresses #4224

- [ ] Addresses the issue in full
- [x] Addresses only certain aspects of the issue

## Description

Pulls some of the backend data into the public lottery view screen. This just tackles showing the raw rank and preferences ranks - getting the total number of applicants and number of applicants per preference I'll handle in a separate PR.

## How Can This Be Tested/Reviewed?

You need a bit of data to test this out. I submitted two applications while signed in to a listing with a lottery and preferences. After running the lottery, when you view the lottery results, the correct ranks for raw rank and preference ranks if the application claimed it should show.

Additionally, we will also see the number of available units.

Example:

Screenshot 2024-08-26 at 3 40 11 PM

The new additions in this PR are `1 unit`, the raw rank, and the preference ranks. The above screenshot is from using a seeded listing locally which had seeded applications which was helpful for testing.

## Author Checklist:

- [ ] Added QA notes to the issue with applicable URLs - will wait to move to QA for the full AC
- [x] Reviewed in a desktop view
- [x] Reviewed in a mobile view
- [ ] Reviewed considering accessibility
- [x] Added tests covering the changes
- [ ] Made corresponding changes to the documentation
- [ ] Ran `yarn generate:client` and/or created a migration when required

## Review Process:

- Read and understand the issue
- Ensure the author has added QA notes
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Either (1) explicitly ask a clarifying question, (2) request changes, or (3) approve the PR, even if there are very small remaining changes, if you don't need to re-review after the updates

Issue #4224, opened 2024-07-24
Lotteries - Public lottery view (applicant data)


emilyjablonski:
Related to #4054

Includes "x applications were submitted for y units" and "out of x applicants on this list" as these may require backend changes to fetch that data in a performant way

We also need a way to fetch an application's rank and its preferences ranks, this could be a new lottery service call

sarahlazarich:
Extra Complexity is that in order to know those numbers, we need to get a count of all of them 
If we need to do that every time someone loads that page, it could be a lot 
the "Blue Box" number is easier, but the other ones we'll need to know the total list count 

App number - will not include duplicates, # of applications included in the raw rank lottery 
Will be the new BE endpoint need to figure out how to do that 
Could we just store it when we do the lottery? We could! In a json or something

emilyjablonski:
QA Notes
Submit an application to a listing and claim a preference. Run and release the lottery. Ensure that the rankings and the total number of applications and number of applications that claimed that preference are correct.

sarahlazarich:
done!

PR #4279, opened 2024-08-27
feat: public lottery view, totals data

emilyjablonski:
Related to merge of: https://github.com/bloom-housing/bloom/pull/4275

This PR addresses #4224

## Description

Adds in the total number of applicants and number of applicants per preference into the public site lottery results page.

I created a new table that stores the totals in a very similar way to how we store lottery positions. I had performance concerns over querying the lottery positions table every time a user needs to see their results page as it will have 1000s of rows after just one lottery is run.

There's also some misc. test coverage.

Submit an application to a listing and claim a preference. Run and release the lottery. Ensure that the total number of applications and number of applications that claimed that preference are correct.

Handling listings without preferences or a user who doesn't claim any is a separate ticket.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/controllers/lottery.controller.ts
+++ b/api/src/controllers/lottery.controller.ts
@@ -36,6 +36,7 @@ import { PermissionAction } from '../../src/decorators/permission-action.decorat
 import { permissionActions } from '../../src/enums/permissions/permission-actions-enum';
 import { AdminOrJurisdictionalAdminGuard } from '../../src/guards/admin-or-jurisdiction-admin.guard';
 import { PublicLotteryResult } from '../../src/dtos/lottery/lottery-public-result.dto';
+import { ApplicationLotteryTotal } from '../../src/dtos/applications/application-lottery-total.dto';
 
 @Controller('lottery')
 @ApiTags('lottery')
@@ -162,5 +163,21 @@ export class LotteryController {
       id,
       mapTo(User, req['user']),
     );
+  }
+
+  @Get(`lotteryTotals/:id`)
+  @ApiOkResponse({
+    type: ApplicationLotteryTotal,
+    isArray: true,
+  })
+  @ApiOperation({
+    summary: 'Get lottery totals by listing id',
+    operationId: 'lotteryTotals',
+  })
+  async lotteryTotals(
+    @Request() req: ExpressRequest,
+    @Param('id', new ParseUUIDPipe({ version: '4' })) id: string,
+  ): Promise<ApplicationLotteryTotal[]> {
+    return this.lotteryService.lotteryTotals(id, mapTo(User, req['user']));
   }
 }
--- a/api/src/controllers/lottery.controller.ts
+++ b/api/src/controllers/lottery.controller.ts
@@ -36,8 +36,7 @@ import { PermissionAction } from '../../src/decorators/permission-action.decorat
 import { permissionActions } from '../../src/enums/permissions/permission-actions-enum';
 import { AdminOrJurisdictionalAdminGuard } from '../../src/guards/admin-or-jurisdiction-admin.guard';
 import { PublicLotteryResult } from '../../src/dtos/lottery/lottery-public-result.dto';
-import { ApplicationLotteryTotal } from '../../src/dtos/applications/application-lottery-total.dto';
-import { PublicLotteryTotal } from 'src/dtos/lottery/lottery-public-total.dto';
+import { PublicLotteryTotal } from '../../src/dtos/lottery/lottery-public-total.dto';
 
 @Controller('lottery')
 @ApiTags('lottery')
--- a/api/src/controllers/lottery.controller.ts
+++ b/api/src/controllers/lottery.controller.ts
@@ -37,6 +37,7 @@ import { permissionActions } from '../../src/enums/permissions/permission-action
 import { AdminOrJurisdictionalAdminGuard } from '../../src/guards/admin-or-jurisdiction-admin.guard';
 import { PublicLotteryResult } from '../../src/dtos/lottery/lottery-public-result.dto';
 import { ApplicationLotteryTotal } from '../../src/dtos/applications/application-lottery-total.dto';
+import { PublicLotteryTotal } from 'src/dtos/lottery/lottery-public-total.dto';
 
 @Controller('lottery')
 @ApiTags('lottery')
@@ -167,7 +168,7 @@ export class LotteryController {
 
   @Get(`lotteryTotals/:id`)
   @ApiOkResponse({
-    type: ApplicationLotteryTotal,
+    type: PublicLotteryTotal,
     isArray: true,
   })
   @ApiOperation({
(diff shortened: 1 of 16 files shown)


api/src/services/lottery.service.ts
```diff
@@ -1113,31 +1132,71 @@ export class LotteryService {
       throw new ForbiddenException();
     }
 
-    const applicationUserId = await this.prisma.applications.findFirstOrThrow({
+    if (!user.userRoles.isAdmin) {
```
ludtkemorgan, line 1135: Public users don't have userRoles attached to them. So when I go through the flow with a public user I created this throws a null pointer exception

sites/public/src/pages/account/application/[id]/lottery-results.tsx
```diff
@@ -39,6 +41,14 @@ export default () => {
                 .publicLotteryResults({ id: applicationId })
                 .then((results) => {
                   setResults(results)
+                  lotteryService
```
Comment (ludtkemorgan) [inline sites/public/src/pages/account/application/[id]/lottery-results.tsx:44]: This network call chain has become pretty intense. Since these calls don't actually rely on the previous data in order to run do we have to chain it like this? Some of these calls could be done simultaneously. For example we have the listing id after fetching the application and don't need to wait for the lottery results to do that call.

api/src/services/lottery.service.ts
```diff
@@ -1113,31 +1132,71 @@ export class LotteryService {
...
+   */
+  public async lotteryTotals(
+    listingId: string,
+    user: User,
+  ): Promise<ApplicationLotteryTotal[]> {
+    if (!user) {
+      throw new ForbiddenException();
+    }
+
+    if (!user.userRoles.isAdmin) {
```
ludtkemorgan, line 1182: Same here about public users

sites/public/src/pages/account/application/[id]/lottery-results.tsx
Comment (ludtkemorgan) [reply sites/public/src/pages/account/application/[id]/lottery-results.tsx:44]: Also, we are retrieving more data than we need to. Unless I'm missing something it looks like we are just getting the application in order to get the listing id. I'm wondering if there is some consolidation we can do with the 4 endpoints. Or in other words would it be a viable solution to create one endpoint just for this page? I don't remember where the conversations ended on that discussion

Thompson1985 (approved):
Thanks guys

Thompson1985 (commented):
4224/applicant-number-data

api/src/controllers/lottery.controller.ts
```diff
@@ -163,4 +164,20 @@ export class LotteryController {
       mapTo(User, req['user']),
     );
   }
+
+  @Get(`lotteryTotals/:id`)
```
YazeedLoonat, line 168: nit pick: I'm not a fan of passing id params like this and instead think we should be passing these in as DTOs sorta how the listing controller's `getPaginatedSet` call works 

I dunno if you should change this now but that's the standard I'd really like us to move to so we don't have to remember to add the param validation on a bunch of endpoints

api/src/dtos/applications/application-lottery-total.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty()
+  listingId: string;
+
+  @Expose()
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @IsUUID(4, { groups: [ValidationsGroupsEnum.default] })
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty()
+  multiselectQuestionId: string;
```
YazeedLoonat, line 19: since this is optional in the prisma schema, should this be optional here?

api/src/services/lottery.service.ts
```diff
@@ -1113,31 +1132,71 @@ export class LotteryService {
...
+      await this.prisma.applications.findFirstOrThrow({
```
YazeedLoonat, line 1183: prisma errors are very hard to parse, we shouldn't rely on their `orThrow()` because troubleshooting them in production will be pretty hard. 

instead I recommend doing what we do in other services where we attempt to grab the record and then do a null check and throw an error we can understand if its missing

YazeedLoonat (changes requested):
looks good! along with what morgan said, had a few throughts as well

api/src/controllers/lottery.controller.ts
Thompson1985: Okay definitely makes sense

api/src/dtos/applications/application-lottery-total.dto.ts
Thompson1985: Looking into now

sites/public/src/pages/account/application/[id]/lottery-results.tsx
Comment (emilyjablonski) [reply sites/public/src/pages/account/application/[id]/lottery-results.tsx:44]: Agreed on the nesting and the endpoint consolidation, that just felt like it could turn into a bigger refactor to undertake here. Having one endpoint here I think also sounds ideal! If that feels blocking for this PR happy to do it now.

api/src/dtos/applications/application-lottery-total.dto.ts
emilyjablonski: I think so, I was duplicating the behavior with `ApplicationLotteryPosition` - should I change `multiselectQuestionId` to be optional there as well?

api/src/controllers/lottery.controller.ts
emilyjablonski: Sounds good! If that feels blocking here I am happy to update that now.

api/src/services/lottery.service.ts
emilyjablonski: Gotcha! I was copying patterns from elsewhere, particularly `lotteryStatus` - we have ~30 of instances of `findOrThrow`, should we make a ticket to remove them all?

api/src/dtos/applications/application-lottery-total.dto.ts
YazeedLoonat: ooh yeah can we change that to be optional as well please!

api/src/services/lottery.service.ts
YazeedLoonat: If I'm not mistaken all instances of `findOrThrow` are custom functions to encapsulate the logic

if you peep `ami-chart.service.ts ~line 137` you'll notice its a custom function that does a findUnique() and then checks to see

so those should be okay. Its more an issue with the inbuilt orThrow functions in prisma

emilyjablonski: Will take a look! If you have space let's peek at this in a huddle

api/src/dtos/applications/application-lottery-total.dto.ts
emilyjablonski: Can do!

sites/partners/src/lib/listings/formTypes.ts
```diff
@@ -167,6 +167,7 @@ export const formDefaults: FormListing = {
   //     rows: [],
   //   },
   // },
+  applicationLotteryTotals: [],
```
ludtkemorgan, line 170: nit: unrelated to this PR and I think I introduced it, but can we delete the commented out code above this

sites/public/src/pages/account/application/[id]/lottery-results.tsx
Comment (ludtkemorgan) [reply sites/public/src/pages/account/application/[id]/lottery-results.tsx:44]: This is not blocking, but we should probably create a ticket to address it at some point

ludtkemorgan (approved):

YazeedLoonat (approved):
Looks good! 2 small changes to DTOs

PR #4286, opened 2024-08-29
feat: lottery public list view

ColinBuyck:
This PR addresses #4053 

## Description

This PR adds the final touches to the applications lottery view of the public side. This mainly includes custom tags and dates. It also includes some string twekas (@eajensenwa approved the "View application" vs "View your application" tweak)

This can be tested by signing into the public site and applying to a few different listings. Then, update the listings to show as open without a due date, open with a due date, closed without a lottery date, closed with a lottery date, and lottery ran.

Additionally ensure all tests have passed below.
## Author Checklist:

## Review Process:


api/test/unit/services/application.service.spec.ts
```diff
@@ -768,6 +773,108 @@ describe('Testing application service', () => {
...
+        lotteryStatus: LotteryStatusEnum.publishedToPublic,
+      },
+    ];
+    const mockedValuesWithListing = mockedValues.map((mockedValue, idx) => {
+      return {
+        ...mockedValue,
+        listings: listingStatuses[idx],
+      };
+    });
+    prisma.applications.findMany = jest.fn().mockResolvedValue(
```
ColinBuyck, line 801: Is it worth mocking this value to this detail if I am only checking the length? I considered matching with more detail as well but felt that would be more complexity than value since the function isn'y transforming the response? Thoughts?

sites/public/src/components/account/StatusItem.tsx
```diff
@@ -3,39 +3,62 @@ import { t } from "@bloom-housing/ui-components"
...
-  applicationUpdatedAt: string
   confirmationNumber?: string
   listingName: string
   listingURL: string
+  listingStatus: ListingsStatusEnum
+  lotteryStartDate?: string
+  lotteryPublishedDate?: string
   lotteryResults?: boolean
   lotteryURL?: string
   strings?: {
```
ColinBuyck, line 19: Is this still worth building out? Considered adding custom string options for open closed, lottery run, applications close, lottery date, and lottery published at strings but not really sure how this is applicable now if the component isn't shared. Any thoughts @emilyjablonski

api/test/unit/services/application.service.spec.ts
```diff
@@ -768,6 +773,108 @@ describe('Testing application service', () => {
     });
   });
```
ColinBuyck, line 775: Love to get some advice on the best approach to unit testing here. Worried this one doesn't cover anything beyond the e2e but equally worried of too much mocking and also not getting much coverage.

api/test/unit/services/application.service.spec.ts
ColinBuyck: Also any tips to reduce the duplicate mocking code here?

sites/public/src/components/account/StatusItem.tsx
emilyjablonski: Yeah we used to export this, since we're just using it internally now we don't need this

sites/public/src/components/account/StatusItemWrapper.tsx
```diff
@@ -16,16 +16,21 @@ interface StatusItemWrapperProps {
...
     <StatusItem
       applicationDueDate={applicationDueDate && dayjs(applicationDueDate).format("MMMM D, YYYY")}
       applicationURL={`/account/application/${props.application?.id}`}
-      applicationUpdatedAt={dayjs(props.application?.updatedAt).format("MMMM D, YYYY")}
       confirmationNumber={props.application?.confirmationCode || props.application?.id}
       listingName={props.application?.listings?.name}
       listingURL={`/listing/${props.application?.listings?.id}`}
+      listingStatus={props.application.listings.status}
       key={props.application?.id}
+      lotteryStartDate={lotteryStartDate && dayjs(lotteryStartDate).format("MMMM D, YYYY")}
```
emilyjablonski, line 30: The design has the months here as `Jan` or `Dec` which would be `MMM` (docs: https://day.js.org/docs/en/display/format)

emilyjablonski:
The design has the text `Status:` in front of the status pills which is missing here. TBH I prefer the behavior on this PR, the text seems redundant, but just wanted to check in if you had passed that by Em.

api/test/unit/services/application.service.spec.ts
emilyjablonski: You can define all the static-ish variables, like the userid, requesting user, mocked values, statuses, and even the response under the describe block and shared them. If you want to overwrite specific parts of it, you can just spread ie { ...mainMock, customField: 'binky' }

sites/public/src/components/account/StatusItem.tsx
```diff
@@ -55,10 +78,10 @@ const StatusItem = (props: StatusItemProps) => {
           </div>
 
           <div className={styles["status-item__action"]}>
-            {props.applicationDueDate && (
+            {dueDate && (
               <p className={styles["status-item__due"]}>
-                {props.strings?.applicationsDeadline ?? t("listings.applicationDeadline")}:{" "}
-                <span className={styles["status-item__due-date"]}>{props.applicationDueDate}</span>
+                {`${deadlineText} `}
```
emilyjablonski, line 83: Nit: This can just be `{deadlineText}`

emilyjablonski (approved):
🐈‍⬛

sites/public/__tests__/components/applications/StatusItem.test.tsx
```diff
@@ -2,38 +2,117 @@ import React from "react"
...
         listingName={"Listing Name"}
+        listingStatus={ListingsStatusEnum.active}
         listingURL={"/listing/abcd1234/listing-name"}
       />
     )
 
     expect(getByText("Listing Name")).not.toBeNull()
-    expect(getByText(t("listings.applicationDeadline"), { exact: false })).not.toBeNull()
+    expect(getByText(t("account.openApplications"), { exact: false })).not.toBeNull()
+    expect(getByText(t("account.applicationsClose"), { exact: false })).not.toBeNull()
```
ludtkemorgan, line 24: Is there a reason these all have to have the `{exact: false}` portion? Looking at the code it looks like the string is a one to one in a span.

nit: `.toBeInTheDocument()` is slightly better than `.not.toBeNull()` https://testing-library.com/docs/guide-disappearance/#nottobeinthedocument

ludtkemorgan (approved):

sites/public/src/components/account/StatusItemWrapper.tsx
ColinBuyck: Nice catch 🐐

sites/public/__tests__/components/applications/StatusItem.test.tsx
ColinBuyck: Good point, I cleaned that up and switched to toBeInTheDocument()

api/test/unit/services/application.service.spec.ts
ColinBuyck: Just cleaned things up with some shared variables! Thanks!

ColinBuyck:
Thanks for the reviews! Responded to each thread and got confirmation from design that excluding the status text works @emilyjablonski. Will merge once new tests pass.

Issue #4181, opened 2024-07-08
Toast Infinite Re-Render


ColinBuyck:
If a user is able to navigate to a draft/closed listing application, then they will see a stack of toasts and they won't be redirected to the listings detail page. As part of https://github.com/bloom-housing/bloom/pull/4179, the addToasts call in choose-language.tsx and summary.tsx was commented out so the redirect occurs correctly but without any message.

Ideally, the toast should show once on the listing detail page you're redirected to. 

![Screenshot 2024-07-08 at 11 28 34 AM](https://github.com/bloom-housing/bloom/assets/53269332/b8c61784-3184-4a98-ba52-ad96e3b13aa7)

sarahlazarich:
Hard to saw what's going on, but could be a react affect getting stuck in a loop 
Could be missing some kind of dependency value (3ish value)

sarahlazarich:
When there's a redirect on the toast, infinite re-renders. 
The case in the ticket is (temp - just commented it out) fixed, but this is still relevant when a user is signed out 
Going to happen in the component itself 
Toast use affect logic has been a headache in the past 
Going to require research and implementation

sarahlazarich:
looks good

PR #4302, opened 2024-09-04
fix: restore toasts in a way which prevents infinite rerenders

jaredcwhite:
This PR addresses #4181

## Description

~~The Next.js `router.push` method can return a promise which is resolved when the new route has loaded. Instead of swallowing that (as `void`), we can provide a callback and use that for adding the toast to the toast stack. That resolves the infinite rendering issue. (I'm not quite sure why we haven't needed this solution _everywhere_, but at least now we know about it).~~

Turns out there _was_ a very specific bug which that callback fix worked around but it didn't really address it. I've since updated this PR. Here's what was going on:

Apparently, functions pulled out of React Contexts aren't stable (aka `addToast`), meaning even if you put them in the dependency array of a `useEffect` hook, it won't prevent that hook to re-execute. The fix was to write a custom hook that wraps the context in a stable ref. That way, you can put the ref itself in the dependency array, then pull `addToast` out of the ref's current value within `useEffect` for use, thus preventing unnecessary re-execution (aka infinite rerenders).

So that's good, but now there's two different methods within the codebase of accessing `addToast`—the original way which is normally fine, and the new way just for `useEffect`. I'm not thrilled about that, but I also don't want to mandate having to use the ref everywhere now which is a bit more cumbersome. Happy to entertain feedback about that in either direction.

Try applying to a listing which the backend has just switched off applications, or the due date is past, etc. You should see a single toast notification that the listing is now closed.

## Author Checklist:

## Review Process:


emilyjablonski:
Screenshot 2024-09-04 at 6 52 50 PM
We'll need this fix in more places, potentially everywhere we are wanting to show a toast on a different page than where we start. For example, the RequireLogin component, which kicks in when you try and visit `account/dashboard` on the public site when you're logged out.

jaredcwhite:
@emilyjablonski Oof…this might require a whole lot of refactoring in various places. I'm starting to wonder if there's an alternative solution, like maybe I can add a de-dup check inside of `addToast` somehow.

emilyjablonski:
@jaredcwhite Is this ready for re-review?

jaredcwhite:
@emilyjablonski It is now…I had to make a change based on how I actually determined was the root cause of the rerenders. See my updated PR description for the explanation: https://github.com/bloom-housing/bloom/pull/4302#issue-2506373944

emilyjablonski (approved):
This looks great! I would be in favor of using this new version everywhere since it works in both cases and so we don't have to remember which is which, but I don't think we necessarily need to do that conversion here.

ColinBuyck (commented):
I tried hitting http://localhost:3000/applications/start/choose-language?listingId={insert-closed-listing-id} and got two toasts. Specifically, using Valley Heights Senior Community after a reseed. Are you seeing this behavior as well?

Screenshot 2024-09-18 at 10 38 58 AM

jaredcwhite:
@ColinBuyck Hey, I tried to replicate this and I couldn't…I'm just seeing the single toast. Any other tips you could provide to trigger this?

ColinBuyck:
@jaredcwhite sorry for not following up to this sooner! I was just retesting and was able to reproduce but found a case where only one shows. 

When I tried to hit that url with the listing id of a closed listing **without** the common app enabled, only one showed. For example, Valley Heights Senior Community after reseeding

But using the listing id of a closed listing **with** the common app enabled, two shows. For example, I reseeded, closed Blue Sky Apartments, and used that listing id within the link above and saw two toasts. Let me know if that helps you repro!

jaredcwhite:
@ColinBuyck wow, so strange. Thanks for the info, I will investigate further!

jaredcwhite:
@ColinBuyck Alas…I still can't repo. I tried a few times and in a couple different browsers. Could be a dev/env specific issue of some sort?

I might recommend we do a test in a staging environment post-merge, unless you can trigger this in the deploy preview?

ColinBuyck:
@jaredcwhite Hmm I also can't seem to repro on the deploy preview. I wonder if there's some sort of race condition going on but I am okay with merging and keeping an eye on it! Thanks for the discussion on this ⏱️

ColinBuyck (approved):
LGTM 🥼

PR #4307, opened 2024-09-05
feat: listing duplication endpoint

mcgarrye:
This PR addresses [#(799)](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/metrotranscom/doorway/799)

## Description

Allows user to call `listings/duplicate` endpoint to create a copy listing from an existing listing.

Run partner portal and api and login as a admin user.
Have the listing id for the listing you would like to duplicate.
Fill out the payload for `listings/duplicate`. Set storedListing.id to the above.
Cases:

1. Set name to a unique name, set includeUnits to true. Should create a new draft listing with the name given which is otherwise identical to the original listing, including the units.
2. Set name to a unique name, set includeUnits to false. Should create a new draft listing with the name given which is otherwise identical to the original listing, excluding the units.
3. Set name to a the same name as the original listing. Should error with 400 BadRequest and a message 'New listing name must be unique'.

You can view the newly created in the Partners Portal

## Author Checklist:

## Review Process:


api/src/services/listing.service.ts
```diff
@@ -965,6 +967,68 @@ export class ListingService implements OnModuleInit {
...
+      delete mappedListing['units'];
+    }
+
+    const newListingData: ListingCreate = {
+      ...mappedListing,
+      name: dto.name,
+      status: ListingsStatusEnum.pending,
+      listingEvents: listingEvents,
+      listingMultiselectQuestions:
+        storedListing.listingMultiselectQuestions?.map((question) => {
```
ColinBuyck, line 1018: Nit: Could we use that =>({}) syntax brought up in yesterdays huddle?

api/src/services/listing.service.ts
```diff
@@ -965,6 +967,68 @@ export class ListingService implements OnModuleInit {
...
+          return {
+            id: question.multiselectQuestionId,
+            ordinal: question.ordinal,
+          };
+        }),
+      listingImages: listingImages,
+      lotteryLastRunAt: undefined,
```
ColinBuyck, line 1025: Just merged in a PR adding lotteryLastPublishedAt field which should be set to undefined here too

api/test/integration/listing.e2e-spec.ts
```diff
@@ -661,6 +661,57 @@ describe('Listing Controller Tests', () => {
     });
   });
 
+  describe('duplicate endpoint', () => {
```
ColinBuyck, line 664: Nit: Can these be separated to improve readability on if both duplication calls are broken or if it is dependent on the units piece?

ColinBuyck (approved):
LGTM 🫕 Just a few nits

api/.env.template
```diff
@@ -44,6 +44,8 @@ LOTTERY_PUBLISH_PROCESSING_CRON_STRING=58 23 * * *
 LOTTERY_PROCESSING_CRON_STRING=0 * * * *
 # how many days till lottery data expires
 LOTTERY_DAYS_TILL_EXPIRY=45
+# allow partners and jadmins to create duplicate listings
+ALLOW_PARTNERS_TO_CREATE_DUPLICATES=FALSE
```
YazeedLoonat, line 48: can we change this to `ALLOW_PARTNERS_TO_DUPE_LISTINGS` or something like that, to make it clear that this is about duplicating listings

api/src/services/listing.service.ts
```diff
@@ -965,6 +967,105 @@ export class ListingService implements OnModuleInit {
...
+    await this.permissionService.canOrThrow(
+      { ...requestingUser, userRoles: userRoles },
+      'listing',
+      permissionActions.create,
+      {
+        jurisdictionId: storedListing.jurisdictions.id,
+      },
+    );
+
+    if (dto.name.trim() === storedListing.name) {
```
YazeedLoonat, line 998: can you move this to right after we get the stored listing, so we error early and don't do more work than we need to

YazeedLoonat (changes requested):
looks good, some changes requested!

ColinBuyck (approved):
Yahoooo 🪨

PR #4309, opened 2024-09-05
fix: listing duplicate fe

ColinBuyck:
This PR addresses [#773](https://github.com/metrotranscom/doorway/issues/773)

## Description

This PR adds the FE component to the listing duplication feature based on the designs in the attached ticket. Additionally, this PR adds the created date and application due date to the listings table on the partners side.

Update: Additionally, this PR updates the BE logic to support a duplicate listing permissioning array which mirrors the pattern of the listing approval array. Lastly, there is a new feature flag to determine how limited the options on a closed listing should be (See AC updates at the bottom of the linked ticket.)

This can be reviewed by running this locally, clicking into any listing, clicking copy, and ensure that the form requires an original name and the includeUnits checkbox is functional. 

Additionally, try turning the feature flag LIMIT_CLOSED_LISTING_ACTIONS on/off or updating the duplicateListingPermissions array on the jurisdictions table.
 
## Author Checklist:

## Review Process:


api/src/services/listing.service.ts
```diff
@@ -1016,6 +1016,7 @@ export class ListingService implements OnModuleInit {
     const newListingData: ListingCreate = {
       ...mappedListing,
       name: dto.name,
+      assets: [],
```
ColinBuyck, line 1019: Discussed a BE error I was running into with Eric and since it only required this one tweak, we agreed to include that here to not block this PR(cc @mcgarrye)

sites/partners/src/components/listings/ListingFormActions.tsx
```diff
@@ -364,7 +380,8 @@ const ListingFormActions = ({
             elements.push(editFromDetailButton)
         }
 
-        // all users can preview
+        // all users can copy or preview
+        elements.push(copyButton)
```
mcgarrye, line 384: This should only be visible for everyone on Doorway, HBA and Core should only have this allowed for Admin

ColinBuyck:
On hold per discussions with @sarahlazarich around if this should be feature flagged or not 🥇

sarahlazarich:
@ColinBuyck @mcgarrye caught up with Izzie - if the following is true this doesn't need to be feature flagged

Doorway: 
- All user types should be able to copy a listing 

Core/HBA: 
- Only Admins should be able to copy a listing

api/prisma/seed-dev.ts
```diff
@@ -3,6 +3,7 @@ import {
   ListingsStatusEnum,
   MultiselectQuestionsApplicationSectionEnum,
   PrismaClient,
+  UserRoleEnum,
```
emilyjablonski, line 6: Looks like an unused import

emilyjablonski:
Screenshot 2024-10-01 at 3 33 21 PM
I'm still seeing Reopen / Unpublish, is this maybe a Doorway only change?

api/prisma/seed-dev.ts
ColinBuyck: 😢 🎃 🍋

ColinBuyck:
@emilyjablonski Good catch! Just talked to Sarah about this change and hba/dwy's needs and things seem to be back on hold until that piece is ironed out. Thanks for pointing this out 🧼

api/src/services/listing.service.ts
```diff
@@ -935,15 +936,40 @@ export class ListingService implements OnModuleInit {
...
+    //manually check for juris mismatch since logic above is forcing admin permissioning
+    if (
+      (requestingUser?.userRoles?.isJurisdictionalAdmin ||
+        requestingUser?.userRoles?.isPartner) &&
+      !requestingUser?.jurisdictions?.some(
+        (juris) => juris.id === storedListing.jurisdictionId,
+      )
+    ) {
+      throw new ForbiddenException();
+    }
```
ColinBuyck, line 972: @mcgarrye This PR is waiting on #4403 but wanted to flag this section to  you to get your thoughts since its connected to the tests we discussed. From what I can tell with the isAdmin overriding, we can't just rely on the permissionService.can to cover when a jurisdictional admin has the permissions to copy but attempts to do it in the wrong jurisdiction.

api/src/services/listing.service.ts
```diff
@@ -954,6 +970,20 @@ export class ListingService implements OnModuleInit {
...
+        !requestingUser?.jurisdictions?.some(
+          (juris) => juris.id === storedListing.jurisdictionId,
+        )) ||
+      (requestingUser?.userRoles?.isPartner &&
+        !requestingUser?.listings?.some(
+          (listing) => listing.id === storedListing.id,
+        ))
+    ) {
+      throw new ForbiddenException();
+    }
```
ColinBuyck, line 985: @mcgarrye This PR is waiting on https://github.com/bloom-housing/bloom/pull/4403 but wanted to flag this section above to you to get your thoughts since its connected to the tests we discussed. From what I can tell with the isAdmin overriding, we can't just rely on the permissionService.can to cover when a jurisdictional admin has the permissions to copy but attempts to do it in the wrong jurisdiction or partners on a forbidden listing since we'll pass isAdmin is true and avoid the other parts of the permissioning logic.

YazeedLoonat (changes requested):
looks good, but I think there is an issue with the DTO set up creating unnecessary types in the swagger doc

shared-helpers/src/types/backend-swagger.ts
```diff
@@ -6072,18 +6081,36 @@ export enum EnumJurisdictionCreateListingApprovalPermissions {
...
+  "admin" = "admin",
+  "jurisdictionAdmin" = "jurisdictionAdmin",
+}
 export enum EnumJurisdictionUpdateListingApprovalPermissions {
   "user" = "user",
   "partner" = "partner",
   "admin" = "admin",
   "jurisdictionAdmin" = "jurisdictionAdmin",
 }
+export enum EnumJurisdictionUpdateDuplicateListingPermissions {
```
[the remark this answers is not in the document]
ColinBuyck: Hmmm I can look into this but I was generally following the pattern of the listingApprovalPermissions which has its own enum just above the highlighted line. Are you thinking that both have issues or that its expected to be different from listingApprovalPermissions?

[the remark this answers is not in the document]
YazeedLoonat: Oh interesting, they likely both have the same problem. Enums should only get added once in the swagger doc and then that enum gets used everywhere so this should all use that UserRole enum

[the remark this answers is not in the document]
ColinBuyck: Adding the enumName field @ApiProperty seemed to fix it! I updated the listingApprovalPermissions as well 🦆 Thanks for the catch!

YazeedLoonat (approved):
LGTM

api/prisma/migrations/17_duplicate_listing_permissions/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "jurisdictions" ADD COLUMN "duplicate_listing_permissions" "user_role_enum"[];
```
mcgarrye, line 2: nit: missing EOF newline and there is already a migration 17

mcgarrye:
Is there a follow-up ticket to create a script to update jurisdiction duplication permissions? Otherwise how will this be turned on?

api/prisma/migrations/17_duplicate_listing_permissions/migration.sql
ColinBuyck: Just addressed this! Good catch 🥫

ColinBuyck:
@mcgarrye I was planning on just hitting the db directly to turn it on as if it was a boolean flag on the jurisdiction table. If you have a strong preference for a script though, I can totally make a follow-up!

mcgarrye:
> @mcgarrye Thanks for checking this out! I was planning on just hitting the db directly to turn it on as if it was a boolean flag on the jurisdiction table. If you have a strong preference for a script though, I can totally make a follow-up!

I think a script with documentation on how to run it in notion would be more sustainable as we add more jurisdictions and can allow us to change permissions dynamically as needed without hitting the db directly

emilyjablonski:
@mcgarrye I'm not 100% sure we should have a process that we use the script runner for the jurisdiction table - ideally we want to be able to turn those values on / off more easily, but I'm also not sure we should be hitting the DB directly. @ludtkemorgan Has this come up at all in the conversations about a new Detroit toggle system?

ColinBuyck:
@emilyjablonski @mcgarrye @ludtkemorgan (cc @YazeedLoonat) Thought more about this and I'm wondering if we can go forward with turning this feature on via the db directly and document Eric's concerns in a ticket. I agree that it is not optimal and want to leave room for building a new pattern that is safer but given how this ticket has been delayed, I'm hoping to get it into a testable environment as soon as possible. Thoughts?

ludtkemorgan:
@ColinBuyck @emilyjablonski @mcgarrye 
Hopefully once we have the new feature flag system in place we will no longer need to update the DB manually or via the script runner. However, since it's not in place yet doing it via a manually DB update is ok I think. We will also be migrating all of the existing feature flags to this new system so we don't need to worry about having them there at the moment

Issue #4300, opened 2024-09-04
Publish lottery data - Partner State Card


sarahlazarich:
Bug: 
- The date and time on the publish screen for Partners are not pulling through. 

<img width="1278" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWF2Qmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2f4f0c1fac790b647b075b8a47755c820c2a9c95/image.png" alt="image.png" />

Expected behavior: 
- The date/timestamp here should be the date the lottery was released to partners. 

QA: https://partners.bloom.exygy.dev/
This can be tested by creating a lottery listing as an admin and releasing that lottery. Then as a partner user see the date and time reflected in the lottery publish card. You can also test retracting and releasing the lottery to see the time updated.

sarahlazarich:
lgtm

PR #4315, opened 2024-09-10
fix: add releaseDate to partner card

ColinBuyck:
This PR addresses #4300 

## Description

This PR completes an outstanding todo to include the releaseDate in the partners messaging seen in the issue description. It also handles the case of releasing and retracting by finding the most recent release date.

This can be tested by creating a lottery listing as an admin and releasing that lottery. Then as a partner user see the date and time reflected in the lottery publish card. You can also test retracting and releasing the lottery to see the time updated.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/pages/listings/[id]/lottery.tsx
+++ b/sites/partners/src/pages/listings/[id]/lottery.tsx
@@ -198,6 +198,9 @@ const Lottery = (props: { listing: Listing | undefined }) => {
         (event) => event.type === ListingEventsTypeEnum.publicLottery
       )
       if (listing.lotteryStatus === LotteryStatusEnum.releasedToPartners) {
+        const lotteryReleaseDate = lotteryActivityLogData?.findLast(
+          (logItem) => logItem.status === LotteryStatusEnum.releasedToPartners
+        )?.logDate
         return (
           <CardSection>
             <Icon size="xl">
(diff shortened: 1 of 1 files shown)


sites/partners/src/pages/listings/[id]/lottery.tsx
```diff
@@ -198,6 +198,10 @@ const Lottery = (props: { listing: Listing | undefined }) => {
         (event) => event.type === ListingEventsTypeEnum.publicLottery
       )
       if (listing.lotteryStatus === LotteryStatusEnum.releasedToPartners) {
```
Comment (ColinBuyck) [inline sites/partners/src/pages/listings/[id]/lottery.tsx:200]: Alternative to using .findLast since it was causing build errors. Seems like it was only added recently so our versions may not be quite updated for it https://github.com/microsoft/TypeScript/issues/48829

sites/partners/src/pages/listings/[id]/lottery.tsx
```diff
@@ -198,6 +198,10 @@ const Lottery = (props: { listing: Listing | undefined }) => {
...
+        // reverse array to find most recent release date
+        const lotteryReleaseDate = lotteryActivityLogData
+          ?.reverse()
+          ?.find((logItem) => logItem.status === LotteryStatusEnum.releasedToPartners)?.logDate
```
Comment (ColinBuyck) [inline sites/partners/src/pages/listings/[id]/lottery.tsx:204]: Alternative to using .findLast since it was causing build errors. Seems like it was only added recently so our versions may not be quite updated for it https://github.com/microsoft/TypeScript/issues/48829

Comment (emilyjablonski) [reply sites/partners/src/pages/listings/[id]/lottery.tsx:204]: Ope we need typescript v 5 😮

emilyjablonski (approved):
🎰

sites/partners/src/pages/listings/[id]/lottery.tsx
Comment (ludtkemorgan) [reply sites/partners/src/pages/listings/[id]/lottery.tsx:204]: This is a great solution. `findLast` is definitely the best option, an alternative could be to filter and grab the last one. I don't know which is more performant, but either way it would be a minuscule difference
```
      const filtered = lotteryActivityLogData?.filter(
        (logItem) => logItem.status === LotteryStatusEnum.releasedToPartners
      )
      const lotteryReleaseDate = filtered[filtered.length - 1]
 ```

PR #4331, opened 2024-09-12
feat: improve client-side field validation on Partners listings

jaredcwhite:
This PR addresses metrotranscom/doorway#771

## Description

This PR sets up the Date/Time fields from UIC so that they can forbid non-numerical input. It also added a client-side validation trigger to the Listings field, which I tested a lot so it *shouldn't* add any weird edge cases to how saving a Listing works, but that should definitely have a bit of QA attention.

Add or edit a listing, and try to make sure you can't type letters and also nonsense numbers (like the year 400 or 23 for a month) for Date & Time fields like Application Due Date, Open House Dates, submitted time on Applications, etc.

## Author Checklist:

## Review Process:


sites/partners/src/components/listings/PaperListingForm/sections/RankingsAndResults.tsx
```diff
@@ -211,6 +213,7 @@ const RankingsAndResults = ({ listing, isAdmin }: RankingsAndResultsProps) => {
                   name={"lotteryEndTime"}
                   id={"lotteryEndTime"}
                   register={register}
+                  setValue={setValue}
```
mcgarrye, line 216: Previously if these fields were blank on publish/save, they would all be highlighted red with a warning. Now if they are left blank there is no red warning but an error still appears at the top. Not sure why only the date field is still correctly showing that error

mcgarrye (commented):
Overall things are looking good, if you could just fix the above comment and check out why that unit test is failing this should be good to go

jaredcwhite:
@mcgarrye Thanks for the review! I'll look into the issue you mentioned.

sites/partners/src/components/listings/PaperListingForm/sections/RankingsAndResults.tsx
jaredcwhite: @mcgarrye Can you take one more look at this? I'm not entirely sure why it previously worked in main, but as those date & time fields are required, I wouldn't expect them to validate entry on save without there being a `required` prop. So I've added those and it seems to be acting OK on the PR to me now.

mcgarrye: Screenshot 2024-10-18 at 12 23 56 PM
I'm still not seeing the red highlight on the fields :/

jaredcwhite: @mcgarrye Ah, so it _is_ working on the date field, but not on the time fields if one of those is blank. Hmm, I will investigate!

jaredcwhite:
@mcgarrye Third time's a charm?

mcgarrye (approved):
Looks good! Once you get the test passing lgtm

Issue #4338, opened 2024-09-17
Remove Dwy Specifc Lottery Export Terms Copy


ColinBuyck:
After speaking with product, the export terms that come up when a partner attempts to export the lottery is only relevant to Dwy. This ticket is to reduce the modal copy to only include the generalizable lottery information and not the terms of use section.

QA
Add a partner user to a lottery listing and attempt to export the lottery. You should see the reduced copy in the dialog.

sarahlazarich:
looks good

ludtkemorgan:
@ColinBuyck can this ticket be closed or are we waiting to do something with HBA?

ColinBuyck:
Good to close @ludtkemorgan !

PR #4339, opened 2024-09-17
fix: remove dwy copy

ColinBuyck:
This PR addresses #4338 

## Description

This change removes dwy specific lottery copy in the partner's lottery export terms (as approved by product). I left the Terms of Use string translations since Sarah said we would be using it soon for HBA's implementation.

Add a partner user to a lottery listing and attempt to export the lottery. You should see a reduced modal.

## Author Checklist:

## Review Process:


emilyjablonski:
To me the ticket reads as adjusting the terms to make them more general, as opposed to removing them - esp since we have a ticket to make the terms required by adding a checkbox, I'm wondering if we add the general terms into core?

ColinBuyck:
@emilyjablonski I should've also included this context but I talked to Sarah about the checkbox work and that is meant to be only done in Doorway. I verified with her that this is the desired core experience (and then wrote the ticket after 😔 ). Does that help clarify things?

emilyjablonski (approved):
Makes sense!

Issue #4343, opened 2024-09-18
Remove backend/core from codebase


ludtkemorgan:
With the new prisma backend (`api`) fully rolled out we can now remove the backend/core directory and all dependencies related to it.

This will help with the following:
- Clean up the code
- Reduce the yarn packages getting pulled in when building the front-end apps
- Fix some security issues by removing the unused outdated dependencies

PR #4344, opened 2024-09-18
fix: remove backend core directory

ludtkemorgan:
This PR addresses #4343

## Description

This PR removes the the entire backend/core package from the code base as well as the last remaining references to it in the rest of the codebase.

Notes:
- Several packages needed to be explicitly added to either the public or partner sites as they were getting it for free from the backend
- For some reason with this change MSW (mock service worker) was running into issues with useSWR. I was unable to figure out why. However updating MSW a major version seems to have resolved the issue. Note we are still a full major version behind, but that is a bigger change and we will need to upgrade other packages in order to get that to work.
- Hopefully this will remove some of our snyk and dependabot flagged packages as they were just from backend/core

## Author Checklist:

## Review Process:


YazeedLoonat (approved):
LGTM

this is HUGE

Issue #4312, opened 2024-09-10
Lottery Duplicates - email to valid app only


sarahlazarich:
As an applicant on Doorway who has submitted multiple applications, 
Only applications marked as valid should receive an email about my lottery results (& see lottery results in my account), 
So that I do not get confused by an email that says results are ready for an application that was opted out of the lottery. 

AC: 
- If I submit duplicate applications, I should only recieve a notification email for my valid applications.

emilyjablonski:
QA Notes

Submit two applications while logged in under different email accounts, but use same name / dob, as a partner mark one as a duplicate, and after publishing the lottery ensure only the valid email received a notification.

sarahlazarich:
lgtm!

PR #4346, opened 2024-09-18
fix: dont send lottery emails to duplicate applications

emilyjablonski:
This PR addresses #4312 

## Description

A change to not send lottery published emails to applications that have been marked as a duplicate. Also changes the email we are sending the lottery results to to the email associated with the account that applied, as opposed to the email field on the application, since the account email has been verified. Chatted with Sarah about what this means for HBA and lottery without them having mandated accounts - before we would release lottery in HBA we may have to consider turning mandated accounts on for the public side experience of receiving results - and that we should still use the account email here so folks know when they receive an email they need to login with that email to see the results.

Submit two applications while logged in under different email accounts, but use same name / dob, as a partner mark one as a duplicate, and after publishing the lottery ensure only the valid email received a notification.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/listing.service.ts
+++ b/api/src/services/listing.service.ts
@@ -278,53 +278,6 @@ export class ListingService implements OnModuleInit {
     return { emails: userEmails, publicUrl };
   }
 
-  public async getPublicUserEmailInfo(
-    listingId?: string,
-  ): Promise<{ [key: string]: string[] }> {
-    const userResults = await this.prisma.applications.findMany({
-      select: {
-        userAccounts: {
-          select: {
-            email: true,
-          },
-        },
-        language: true,
-      },
-      where: {
-        listingId,
-        markedAsDuplicate: {
-          not: true,
-        },
-      },
-    });
-
-    const emailUsers = userResults.filter((user) => !!user.userAccounts?.email);
-
-    const result = {};
-    Object.keys(LanguagesEnum).forEach((languageKey) => {
-      const applications = emailUsers
-        .filter((user) => user.language === languageKey)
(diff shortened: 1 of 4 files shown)


api/src/services/listing.service.ts
```diff
@@ -283,36 +283,36 @@ export class ListingService implements OnModuleInit {
...
-          emailAddress: {
-            not: null,
-          },
+        markedAsDuplicate: {
+          not: true,
         },
       },
     });
 
+    const emailUsers = userResults.filter((user) => !!user.userAccounts?.email);
```
emilyjablonski, line 301: With mandated accounts on, this is never needed, but as a interim in HBA

api/src/services/listing.service.ts
```diff
@@ -283,36 +283,36 @@ export class ListingService implements OnModuleInit {
   ): Promise<{ [key: string]: string[] }> {
     const userResults = await this.prisma.applications.findMany({
       select: {
-        language: true,
```
ludtkemorgan, line 286: This is not related to these specific changes, but I'm curious why this function is in the listing service? It's only being used by the lottery service and it never touches the listing db table. Would this make more sense in either the lottery or application services?

ludtkemorgan (approved):

api/src/services/listing.service.ts
emilyjablonski: Good q I'm not super sure! My guess is because it's related to the function above `getUserEmailInfo`

emilyjablonski: Moved it into lottery 👍

Issue #4316, opened 2024-09-10
Lottery email - publication notice to partners/admins


sarahlazarich:
Bug: 
- When testing automatic publication, the "results were published to applicant account" went out to some, but not all, of the admins on core.

Expected Behavior: 
- All partners and Admins should get an email when lottery is published 

<img width="719" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzNGQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--535b73429821a88c845481ebaa263f740224cc59/image.png" alt="image.png" />

emilyjablonski:
@sarahlazarich Do you have more details on how this was tested / what environment this was in?

sarahlazarich:
Yeah - I believe @mcgarrye and I were testing at https://partners.bloom.exygy.dev/sign-in

PR #4352, opened 2024-09-25
fix: dont attempt to send email to no users

emilyjablonski:
This PR addresses #4316

## Description

After debugging, I was unable to repro, and after reaching out to product, the email had actually come through successfully. The error is coming from the fact that the initial listing this was run on had no applications, and was being thrown while we were trying to send the similar lottery results email to public applicants. Right now we are in multiple situations passing an empty array to the SendGrid email function, and the error 'no personalizations' is what gets thrown when the 'to' field is empty. This change just doesn't call it if we pass the function no recipients.

The end result is just not throwing the error. A little hard to test, but you can use a seeded listing with seeded applications and publish lottery results. You shouldn't get any errors in the console (this error was also happening previously with seeded applications, and is throwing an error on main, bc we are using an account's email rather than the application email to send the results emails, and the applications we seed aren't tied to real accounts, so that was coming back empty as well).

## Author Checklist:

## Review Process:


mcgarrye (approved):
LGTM!

Issue #4357, opened 2024-09-30
Application export fails with null programs


emilyjablonski:
`/app/dist/src/services/application-csv-export.service.js:190                                       const program = programs.find((preference) => preference.key === curr);`

We expected programs to always be at minimum the empty array `[]` 
if programs is coming across as `null` it causes breakages we didn't account for, the fix here will be to change this so that if programs is `null` we skip over this work so we no longer error

emilyjablonski:
@YazeedLoonat Do you wanna add some more details here?

YazeedLoonat:
added!

PR #4362, opened 2024-09-30
fix: if programs is null no longer errors

YazeedLoonat:
This PR addresses #4357 

## Description
When an application's program data was null, it would cause the a crash in the api when trying to do an export. This fixes that

- create an application that has `null` from program data
- attempt to export the application data for the listing the application was added to
- the app should no longer error and should return an export

## Author Checklist:

## Review Process:


emilyjablonski:
How are you creating an application with null program data? Are you submitting an application to a listing without programs? Or not claiming any?

YazeedLoonat:
@emilyjablonski thats a good question, the error comes from the `program` column being `null` on the application table's application record

I forced the error to happen by creating an application then setting the program value to null in the db via the db

emilyjablonski:
@YazeedLoonat Do we know how this might have happening in staging?

YazeedLoonat:
@emilyjablonski not really sure on how it happened. Both the API and the db support null programs and when digging into the export code I found that the spreadsheet export already had corrected for this

so my assumption was that there is a way to get this data into the db from the front end I just carried over the fix from the spreadsheet export to the csv export

emilyjablonski:
Okay that all makes sense! Is there an easy way to add a test for this?

ColinBuyck (approved):
Vroom vroom 🛵

PR #4366, opened 2024-10-03
fix: tighten the rules for which afs job

ludtkemorgan:
This PR addresses #4208 

## Description

When the `DUPLICATES_CLOSE_DATE` and `DUPLICATES_PROCESSING_CRON_STRING` environment variables are set 
the new AFS cron job should run for all listings that meet either of the following scenarios:
- Listing is closed after `DUPLICATES_CLOSE_DATE`
- Listing is not yet closed

There is an issue with the second scenario for some listings because we currently don't remove the `closed_at` field when a listing is reopened. This means that the old job will still get triggered if the previously closed time was before the duplicate close date

1. Close a listing
2. Reopen the listing
3. Set `DUPLICATES_CLOSE_DATE` to now (or some time after the time stamp of when you first closed the listing) and restart backend
4. add applications. Preferably multiple and at least one duplication
5. Set both `AFS_PROCESSING_CRON_STRING` and `DUPLICATES_PROCESSING_CRON_STRING` with times that will happen in the next few minutes (makes it so you don't have to wait around forever). It's easier if they are at different times and  `AFS_PROCESSING_CRON_STRING` should be before `DUPLICATES_PROCESSING_CRON_STRING`
6. Restart backend and wait for the cron jobs to go off. Verify that the listing wasn't triggered for the old job, but was for the new one.

## Author Checklist:

- [n/a] Added QA notes to the issue with applicable URLs
- [n/a] Reviewed in a desktop view
- [n/a] Reviewed in a mobile view

## Review Process:

### DIFF
--- a/api/src/services/application-flagged-set.service.ts
+++ b/api/src/services/application-flagged-set.service.ts
@@ -513,7 +513,7 @@ export class ApplicationFlaggedSetService implements OnModuleInit {
             OR: [
               // Only run this job on listings that were closed after the DUPLICATES_CLOSE_DATE
               { closedAt: { gte: duplicatesCloseDate.toDate() } },
-              { closedAt: null },
+              { status: { not: 'closed' } },
             ],
           },
           // Allow the ability to force process even if application flag set has run more recent than last application update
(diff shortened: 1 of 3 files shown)


emilyjablonski:
I don't think we have this case, but what would happen for a re-opened listing that had applications previously run on the old job? If that's not a use case then nbd!

ludtkemorgan:
> I don't think we have this case, but what would happen for a re-opened listing that had applications previously run on the old job? If that's not a use case then nbd!

@emilyjablonski If that was the case the new job would run. This would cause a fresh cleanse of the application flagged sets in the system for that listing. However if any application was marked as a duplicate it would still be, just potentially in a different flagged set than what it used to be in and that flagged set would have to be manually resolved.

ColinBuyck (approved):
LGTM 🪜

PR #4375, opened 2024-10-08
feat: record of listing copy

mcgarrye:
This PR addresses [#(806)](https://github.com/metrotranscom/doorway/issues/806)

## Description

Adds `copyOfId` field and `copyOf` relationship to the Listing table. This stores the id of the original listing into the duplicate.
This is surfaced in the listing export by adding two new columns: `Copy or Original` and `Copied From`.
Per product, `Copied From` should show the id.

Run partner portal and api and login as a admin user.
Have the listing id for the listing you would like to duplicate.
Fill out the payload for `listings/duplicate`. Set storedListing.id to the above.
Click the listing export button in the parter portal.
Open the listing csv and view/verify the new columns.

## Author Checklist:

## Review Process:


api/prisma/migrations/18_copy_of_reference/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
+ALTER TABLE "listings" ADD COLUMN     "copy_of_id" UUID;
+
+-- AddForeignKey
+ALTER TABLE "listings" ADD CONSTRAINT "listings_copy_of_id_fkey" FOREIGN KEY ("copy_of_id") REFERENCES "listings"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```
YazeedLoonat, line 5: should this `ON UPDATE CASCADE` be part of this? I don't think updates to one of the listing should affect the other?

api/src/services/listing-csv-export.service.ts
```diff
@@ -348,6 +348,16 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
         format: (val: string): string =>
           formatLocalDate(val, this.dateFormat, this.timeZone),
       },
+      {
+        path: 'copyOfId',
+        label: 'Copy or Original',
+        format: (val: string): string => (val ? 'Copy' : 'Original'),
+      },
+      {
```
YazeedLoonat, line 356: I think since both this and the below have the same `path` key->value pair it will likely cause problems with the way the spreadsheet export wants to work (for when we get there)

api/src/services/listing.service.ts
```diff
@@ -897,6 +895,13 @@ export class ListingService implements OnModuleInit {
           : undefined,
         requestedChangesUser: undefined,
         contentUpdatedAt: new Date(),
+        copyOf: copyOfId
```
YazeedLoonat, line 898: can/do we have a unit test on this piece?

YazeedLoonat (commented):
Looks good, got a few questions

YazeedLoonat (approved):
LGTM!

Issue #4350, opened 2024-09-25
move log-in code to subject line 


sarahlazarich:
As a user trying to log in with an email code, 
I want the code to appear in the subject line of my email, 
So that the emails are not threaded in my account, making it difficult to find. 

AC: 
- log in code should be in the subject line of the email
QA captured in #4349

Issue #4349, opened 2024-09-25
Log-in code improvements


sarahlazarich:
As a user trying to log into my applicant account using a code, 
I want the timeout window for the code to be extended to 10 minutes,
So that I have more time to enter the code before it expires. 

As a user trying to log into my applicant account using a code, 
I want to receive the same code for the duration of the log-in window, 
So that I do not request multiple codes, repeatedly try to enter the wrong one, and be unable to access my account. 

AC: 
- Timeout length of each single use code should be extended to 10 minutes
-  For the duration of the timeout length, if a user clicks the “send the code” button, it sends the same code instead of a new one. 

QA: https://staging.housingbayarea.mtc.ca.gov/ (since not on in core)
1. This can be tested by making a new account on the partners portal that uses mfa verification. Sign in and notice the updated email copy. Then before entering the mfa code, press the back button and sign in again. You'll notice a new email with the same verification code will be sent.
2. Enable single sign on for the partners portal and create a new account. Notice that all references to the single use code expiration now reference 10 minutes rather than 5. Sign in with single use code and then click resend before entering the code. Like above, this should show with the updated email copy but the same code.

eajensenwa:
New subject copy: [######] is your secure [platform name] sign in code

sarahlazarich:
lgtm

PR #4376, opened 2024-10-08
fix: updated email and logic for single use code

ColinBuyck:
This PR addresses #4349 and #4350

## Description

This PR updates the single use code logic so if a user requests a new code within the valid window, we send the same code rather than generating a new one. I verified with Yazeed and Morgan that the udpatedAt time should be refreshed to avoid the confusing experience where a user only has a few seconds to use a code. 

This PR also updates some single use code copy and related emails to show the requested ten minute window (will be updated via env when released) and the code in the subject line.

1. This can be tested by making a new account on the partners portal that uses mfa verification. Sign in and notice the updated email copy. Then before entering the mfa code, press the back button and sign in again. You'll notice a new email with the same verification code will be sent.
2. Enable single sign on for the partners portal and create a new account. Notice that all references to the single use code expiration now reference 10 minutes rather than 5. Sign in with single use code and then click resend before entering the code. Like above, this should show with the updated email copy but the same code.

**NOTE**
3. If you want to test the script runner, it requires a few steps. Setup the db using main branch, switch to this branch, and now run the script via the open api. Because of how our seeding works the single use code email translations exists both in the default and in the jurisdiction-specific translations. As a result, the email will still show 5 minutes because the script doesn't target the jurisdiction specific translations (and jurisdiction overrides default) but if you query the translations table you will see the updated duration under the row where jurisdiction id is null. I have verified on hba and doorway that their email translations exist as the default there too.

## Author Checklist:

## Review Process:


api/src/utilities/passport-validator-utilities.ts
```diff
@@ -58,7 +58,7 @@ export function singleUseCodePresent(
  * @param storedSingleUseCode single use code stored on the user
  * @returns
  */
-export function singleUseCodeValid(
+export function singleUseCodeInvalid(
```
ColinBuyck, line 61: Update naming to better reflect functionality.

api/src/services/email.service.ts
ColinBuyck, line 1: Verified with design that extending this pattern across all code-related emails is preferred

api/prisma/migrations/18_single_use_code_expiration_translations/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- Update single-use code expiration translations
+UPDATE translations
```
emilyjablonski, line 2: Curious why we went the migration route vs the script runner?

ColinBuyck: Because I forgot about the script runner 😞 I'll work on that change now!

ColinBuyck: Updated with new approach and testing instructions if you're curious 🐦‍⬛

ludtkemorgan (approved):
Tested this on both email and SMS and it looks good. Only one small comment so approving it

api/test/unit/utilities/get-single-use-code.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { getSingleUseCode } from '../../../src/utilities/get-single-use-code';
+
+describe('getSingleUseCode', () => {
+  it('should return same mfa code if still valid', () => {
+    expect(getSingleUseCode(5, '00000', new Date(), 60000)).toEqual('00000');
+  });
+  it('should return new mfa code of the specified length if no longer valid', () => {
```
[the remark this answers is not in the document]
ColinBuyck: Ooo that's really helpful, thanks!

YazeedLoonat (approved):
LGTM

Issue #4297, opened 2024-09-03
Incorrect Edit link for Household Members on Review Summary


jaredcwhite:
When filling out an online application, upon arriving at the review step, the Edit link for the Household Members section is going to the same place as Household Details (the next section), and doesn't actually show the members editing step.

I believe this code needs to be adjusted:
https://github.com/bloom-housing/bloom/blob/6dd11ac50819494d032e0b7882fec4c92ff516d7/sites/public/src/components/shared/FormSummaryDetails.tsx#L360

sarahlazarich:
looks good!

PR #4385, opened 2024-10-09
fix: household members link on application review

jaredcwhite:
This PR addresses #4297 

## Description

This PR fixes the Review step of Common Application so that the Household Members section does indeed say Household Members and the edit link goes back to that step. It also adds reusable layout footer to the household members page — which means there's now a Save and Return to Review link (before there wasn't).

Fill out an Application with multiple household members, then when on the Review step, make sure you can go back to edit the members, then save and return to the Review step.

## Author Checklist:

## Review Process:


emilyjablonski (approved):
🛢️

Issue #4361, opened 2024-09-30
Flaky backend tests


emilyjablonski:
1. `Application flagged set Controller Tests › test process duplicates endpoint › should create multiple flag sets with chaining of flags` ([link](https://github.com/metrotranscom/doorway/actions/runs/11110284360/job/30867624906?pr=857))
2. `User Controller Tests › should fail to verify token when token mismatch` ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12163/workflows/dc290db7-f54a-4b35-895c-3ecd958dccaa/jobs/96873)) ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12136/workflows/d6aea776-45c9-4cf2-be8e-2c476c5766e5/jobs/96578)) ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12099/workflows/a8cfc3b3-ca35-4562-9d46-5812ea6a7802/jobs/96143))
3. `Lottery Controller Tests › publicLotteryResults › should return a raw rank` ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12062/workflows/5ec97f0c-2663-404a-bf87-10fb9df20ac0/jobs/95731)) ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12036/workflows/690fb219-3c17-406c-be78-13381fe70a75/jobs/95444)) ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12018/workflows/f0576921-ef56-4339-8f8a-9777c4925509/jobs/95252))
4. Generally seeing failures on creation for unique constraints, looks like the names in our factories sometimes overlap ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12018/workflows/f0576921-ef56-4339-8f8a-9777c4925509/jobs/95252)) ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12142/workflows/47904c75-dec9-4df2-87c6-850af6e3a7ca/jobs/96659)) ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12041/workflows/888bc972-6f5a-4cac-92b9-48ec94a17274/jobs/95503)) ([link](https://app.circleci.com/pipelines/github/bloom-housing/bloom/12041/workflows/00558830-db20-419b-9779-1a4257db240e/jobs/95493))
[...]

PR #4389, opened 2024-10-10
fix: flaky backend tests

mcgarrye:
This PR addresses [#(4361)](https://github.com/bloom-housing/bloom/issues/4361)

## Description

`User Controller Tests › should fail to verify token when token mismatch` & `Testing Permissioning of endpoints as Jurisdictional Admin in the correct jurisdiction › Testing user endpoints › should succeed for verify token endpoint` -> jwt tokens can be larger than 256 in length, upped dto to 542 based on math.

`Lottery Controller Tests › publicLotteryResults › should return a raw rank` & `Generally seeing failures on creation for unique constraints, looks like the names in our factories sometimes overlap` -> moved jurisdiction and reservedCommunityType creation to beforeAlls where applicable to reduce risk of collision and make sure reservedCommunityType is created.

`Listing Controller Tests › list endpoint › should get listings from list endpoint when params are sent` -> skipped as other tests in the describe block are because of db cannot be cleaned

Run e2e tests locally.

## Author Checklist:

## Review Process:


api/test/integration/listing.e2e-spec.ts
```diff
@@ -435,7 +435,8 @@ describe('Listing Controller Tests', () => {
       });
     });
 
-    it('should get listings from list endpoint when params are sent', async () => {
+    // without clearing the db between runs this test is flaky
+    it.skip('should get listings from list endpoint when params are sent', async () => {
```
ludtkemorgan, line 439: Rather than skipping this test could we explicitly set the listing names for listing1 and listing2? The names can be unique enough that no other test will have names like that

Something like 
```
      const listing1 = await listingFactory(jurisdictionAId, prisma, {
        listing: { name: 'filterListing1' } as Prisma.ListingsCreateInput,
      });
      const listing1Created = await prisma.listings.create({
        data: listing1,
      });

      const listing2 = await listingFactory(jurisdictionAId, prisma, {
        listing: { name: 'filterListing2' } as Prisma.ListingsCreateInput,
      });

ludtkemorgan (approved):
🎻

PR #4390, opened 2024-10-11
fix: moving express from dev dependencies

YazeedLoonat:
express was added as a dev dependency, but it should be a required dependency


ludtkemorgan (approved):
Interesting. Looks like express wasn't removed from dependencies and added to devDependencies, but rather added to package.json for the first time. I wonder if we were getting it for free somehow and with the upgrades we had to explicitly add it? Or maybe when you put it in devDependencies even if you are getting it for free from dependencies it takes precendent?

Either way, this looks good!

ludtkemorgan:
@emilyjablonski for awareness. We should probably do a mini team bash in core before we move all of these dependencies to HBA/Doorway

PR #4404, opened 2024-10-17
fix: cleanup lint warnings

ludtkemorgan:
This doesn't address any specific issue, but I noticed we had a decent amount of unused variables/imports. This PR cleans up the current warnings and also tightens the max number of warnings to our current level.

I did not fix any of the "any" types or the missing dependencies in hooks. These will require more testing/knowledge of that area of code

All tests should still work as well building and starting the app

## Author Checklist:

## Review Process:


emilyjablonski (approved):
LGTM! Curious why there were lock file changes when it doesn't look like any packages changed?

ludtkemorgan:
> LGTM! Curious why there were lock file changes when it doesn't look like any packages changed?

@emilyjablonski I tried upgrading eslint to see if it would solve the weird issue where two imports they think they aren't being used even though they are. But the major upgrade required a full rewrite of our eslint config file so I reverted it. However since we use the `^` for most dependencies when I reverted it yarn pulled in newer minor versions

Issue #4473, opened 2024-11-13
Failed Login Count Not Resetting Correctly


ColinBuyck:
1) A user runs through their allowed failed password attempts. 
2) They wait the 30 minutes and fail their password again. 
3) Instead of showing a fresh count of the allowed password remaining value they will have unlimited password attempts and see a negative failed attempts remaining count. They are still able to log in and after they do, it resets to zero but it does explain why we're seeing huge failed attempts counts and reports of confusing experiences from users.

lattaai9:
Hello, I tried to solve the issue.

This is what I did:

Modify the AuthService to properly handle failed password attempts, implement a lockout period, and reset the counter after the timeout. This includes adding new methods for tracking failed attempts and checking if the account is locked.

You can review changes in this commit: https://github.com/lattaai9/bloom-housing-bloom-4473/commit/13b6cc0a4f43b1b9a63945ef58a949e393fae8e6.


> [!CAUTION]
> Disclaimer: The commit was created by Latta AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---
This issue was tried to solve for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

Issue #4284, opened 2024-08-28
Failed Log In Missing Feedback


ColinBuyck:
When testing locally with passwordless turned on but logging in with an email and incorrect password, I am not getting any error message despite the network call showing as a failure

![Screenshot 2024-08-28 at 1 05 01 PM](https://github.com/user-attachments/assets/96bbadaa-1faf-4a2d-855a-0705fb129ef2)

emilyjablonski:
Fixed

PR #4474, opened 2024-11-13
fix: return updated failed login attempts

ColinBuyck:
This PR addresses #4473 

## Description

This PR changes the function isUserLockedOut to return the verified failedLoginAttempts count explicitly. Before, it was failing to reset after the lockout period was over because of an assumption that it was pass by reference rather than value.

Note that there is an existing bug in core: https://github.com/bloom-housing/bloom/issues/4284

This can be tested locally by locking out a public user via 5 failed password attempts (you'll have to verify in network tab). Then, go into the DB and set that user's last login at date to be last year. Lastly, try to log in again with an incorrect password and see that the count has been reset (via the network tab).

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/utilities/passport-validator-utilities.ts
+++ b/api/src/utilities/passport-validator-utilities.ts
@@ -6,14 +6,14 @@ import { HttpException, HttpStatus } from '@nestjs/common';
  * @param failedLoginAttemptsCount the number of times the user failed to log in (stored in db)
  * @param maxAttempts the maximum number of attempts before user is considered locked out (env variable)
  *
- * @returns throws error if user is already locked out
+ * @returns updated value of failed login attempts
  */
-export function isUserLockedOut(
+export function checkFailedLoginCount(
   lastLoginAt: Date,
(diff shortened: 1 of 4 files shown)


api/src/utilities/passport-validator-utilities.ts
```diff
@@ -6,14 +6,14 @@ import { HttpException, HttpStatus } from '@nestjs/common';
  * @param failedLoginAttemptsCount the number of times the user failed to log in (stored in db)
  * @param maxAttempts the maximum number of attempts before user is considered locked out (env variable)
  *
- * @returns throws error if user is already locked out
+ * @returns updated value of failed login attempts
  */
-export function isUserLockedOut(
+export function checkUserLockout(
```
ColinBuyck, line 11: Really struggled with naming this function since it's handling both the erroring case and the reset failed attempts case in one. I considered separating them out but it seemed most of logic would be duplicated... any thoughts?

YazeedLoonat (approved):
LGTM

Issue #4432, opened 2024-10-25
Backend - Home type


ludtkemorgan:
- Connect db fields up to DTOs and all appropriate endpoints (listing edit, create, etc)
- add to CSV export
- Should be behind a jurisdiction level feature flag

sarahlazarich:
the home type shows up in the export, so I believe this passes

Issue #4431, opened 2024-10-25
Partners - Home type


ludtkemorgan:
Add home type selection dropdown to listing edit page
- Use home type enums from backend
- Behind the jurisdiction level feature flag
![image](https://github.com/user-attachments/assets/c3bf3116-36dd-44b6-8c4a-cfef30429936)

ludtkemorgan:
QA notes:
- This functionality has been turned on in [core staging](https://partners.bloom.exygy.dev/) for the Bloomington jurisdiction
Steps for testing:
- Go to partner site and create a new listing
- Select Bloomington Jurisdiction and add a home type to the listing
- Save listing and the home type should appear on the listing
- Edit listing and change the home type. After save it should appear
- Export the Listing CSV and the "Home Type" column should be there with the new listing's home type in the appropriate location

ludtkemorgan:
@sarahlazarich sorry for the delay this one is ready to test in core

sarahlazarich:
@ludtkemorgan I don't see the home type reflected on the listing here

https://bloom.exygy.dev/listing/faa77899-fefa-44a3-b515-eb9634361bf1/test_home_type_4014_n_main_street_warren_mi

sarahlazarich:
nvm - talked in stand, just for the listing view in partners

PR #4484, opened 2024-11-20
fix: add home type feature to backend and partners

ludtkemorgan:
This PR addresses #4432, #4431 

## Description

Adds the "Home Type" select option to the listing create/edit page as well as connecting HomeType to the listing create endpoint, edit endpoint, and the CSV export.

This is also the first feature using the new feature flag table so some extra logic to utilize it has been added

NOTES:
- I did not get approval from product/design on the location of this field in Partners. The Detroit location is slightly different because it is within the "Unit Group" section which doesn't exist in core. I will get confirmation on the location before merging of this PR. 

- Reseed the database with staging data. This will add the new feature flag to the "Bloomington" jurisdiction
- Go to partner site and create a new listing
- Select Bloomington Jurisdiction and add a home type to the listing
- Save listing and the home type should appear on the listing
- Edit listing and change the home type. After save it should appear
- Export the Listing CSV and the "Home Type" column should be there with the new listing's home type in the appropriate location

Other things you can test:
- Verify the Home Type field doesn't appear on the UI if you have selected a different jurisdiction
- If no jurisdiction has the home type feature flag turned on the CSV export should not have that column

## Author Checklist:

## Review Process:


shared-helpers/src/locales/es.json
```diff
@@ -557,6 +557,10 @@
   "footer.disclaimer": "Exención de responsabilidades",
   "footer.forGeneralQuestions": "Si requiere información general sobre el programa, nos puede llamar al 000-000-0000.",
   "footer.giveFeedback": "Proporcione sus comentarios",
+  "homeType.apartment": "Apartamento",
```
ludtkemorgan, line 560: All of these translations will need to be added to Chinese, Vietnamese, and Tagalag when the public site features are pulled over. but those languages don't exist in Detroit so we will need get those

sites/partners/src/components/listings/PaperListingForm/sections/Units.tsx
```diff
@@ -190,6 +218,24 @@ const FormUnits = ({ units, setUnits, disableUnitsAccordion }: UnitProps) => {
             />
           </FieldValue>
         </Grid.Row>
+
+        {homeTypeEnabled && (
+          <Grid.Row columns={2}>
+            <FieldValue label={t("listings.homeType")}>
+              {homeTypes && (
+                <Select
```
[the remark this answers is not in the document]
ludtkemorgan: @emilyjablonski I've seen this linter flag this before. I'm assuming I can ignore it since we explicitly add a sr label?

sites/partners/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "target": "es2017",
```
ludtkemorgan, line 4: This was needed to use the `Set` functionality.

api/src/services/listing-csv-export.service.ts
```diff
@@ -316,7 +318,18 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
     return fieldValue;
   };
 
-  async getCsvHeaders(): Promise<CsvHeader[]> {
+  doAnyJurisdictionHaveFeatureFlagSet = (
+    jurisdictions: Jurisdiction[],
+    featureFlagName: string,
+  ) => {
+    return !!jurisdictions.find((juris) => {
```
YazeedLoonat, line 325: this can be `some` instead of `find` calls

YazeedLoonat (commented):
Looks good! 1 small nit

emilyjablonski (changes requested):
Haven't peeked at the code yet, but a few functionality notes!

Screenshot 2024-12-02 at 1 31 31 PM
Screenshot 2024-12-02 at 1 31 41 PM

And then when I select a value and save it, it shows as saved on the detail page, but if I reopen the edit page, it doesn't populate with the value.

sites/partners/cypress/e2e/default/03-listing.spec.ts
```diff
@@ -160,6 +160,9 @@ describe("Listing Management Tests", () => {
     cy.getByID("reservedCommunityDescription").type(listing["reservedCommunityDescription"])
     cy.getByTestId("unit-types").check()
     cy.getByTestId("listingAvailability.availableUnits").check()
+    if (listing["homeType"]) {
```
emilyjablonski, line 163: ❤️

shared-helpers/src/locales/general.json
```diff
@@ -791,6 +796,7 @@
   "progressNav.srHeading": "Progress",
   "region.name": "Bloomington",
   "states.AL": "Alabama",
+  "states.AK": "Alaska",
```
emilyjablonski, line 799: 😮

sites/partners/src/components/listings/PaperListingForm/index.tsx
```diff
@@ -166,7 +168,21 @@ const ListingForm = ({ listing, editMode, setListingName }: ListingFormProps) =>
   }, [listing?.units, listing?.listingEvents, setUnits, setOpenHouseEvents])
 
   // eslint-disable-next-line @typescript-eslint/unbound-method
-  const { getValues, setError, clearErrors, reset } = formMethods
+  const { getValues, setError, clearErrors, reset, watch } = formMethods
+
+  const selectedJurisdiction = watch("jurisdictions.id")
+
+  useEffect(() => {
+    const newFeatureFlags = profile.jurisdictions?.reduce((featureFlags, juris) => {
```
emilyjablonski, line 176: Just a nit, wondering if this makes sense as a helper that lives somewhere else

sites/partners/src/components/listings/PaperListingForm/sections/Units.tsx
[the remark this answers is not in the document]
emilyjablonski: Yep! I think this is a false positive

sites/partners/tsconfig.json
jaredcwhite: We might want to spike on frontend compiler targets generally speaking, as I'm not sure if we're currently shipping a bunch of polyfilles or other questionable aspects of how JS code gets compiled based on browser versions we wouldn't need to support. (e.g., even es2017 is quite old by now, we might want to aim for at least es2020)

ludtkemorgan: That's a good call out, I just chose 2017 because it is referenced in shared-helpers in the hopes that it would reduce the number of targets in our system. However, it doesn't make a big difference since we are already using several different targets. I'll bump it up to es2020 for this PR and make sure the build still works as expected

api/src/services/listing-csv-export.service.ts
ludtkemorgan: nice! fixed!

sites/partners/src/components/listings/PaperListingDetails/sections/DetailUnits.tsx
```diff
@@ -60,6 +60,13 @@ const DetailUnits = ({ setUnitDrawer }: DetailUnitsProps) => {
 
   return (
     <SectionWithGrid heading={t("listings.units")} inset>
+      {listing.homeType && (
```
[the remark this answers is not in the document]
ludtkemorgan: I don't want to show the home type if jurisdiction doesn't have it enabled. I have updated this so that it checks if the feature flag is turned on, if so it then will show "none" if no home type is selected.

ludtkemorgan:
@emilyjablonski Thanks for the review. Looks like I accidentally introduced the issue where the home type doesn't persist when editing with the useEffect. This has been fixed and you should be able to see the selected home type now.

I'll need to reach out to design about where to place the home type field as it doesn't really make a lot of sense in either of the spots. For now I updated the field to be in the same place for both edit and detail view 

image
image

emilyjablonski (approved):
LGTM!

ludtkemorgan:
I got confirmation from Em that the placement of the field is fine. We can revisit it if any non-detroit jurisdiction ever wants it

sites/partners/src/components/listings/PaperListingForm/index.tsx
ludtkemorgan: As we get more use cases I will refactor this to know the best place to put it

Issue #4459, opened 2024-10-30
Feature flag consumption


ludtkemorgan:
In order to use this new feature flag system we will need to make the following changes so that we can access the flags as well as updating/creating them.

- Create new featureFlag controller/service (could potentially combine with jurisdiction controller)
  - Endpoint to create new feature flags
  - Endpoint to get all feature flags
  - Endpoint to add/remove feature flags to a jurisdiction
  - Endpoint to add/remove feature flags to all jurisdictions
- Update existing endpoints:
  - Jurisdiction get endpoints should return the attached feature flags

mcgarrye:
Not QA-able by Product

PR #4489, opened 2024-11-20
feat: feature flag consumption

mcgarrye:
This PR addresses [#(4459)](https://github.com/bloom-housing/bloom/issues/4459)

## Description

Creates the module, controller and service files for the new Feature Flag table. Allows for standard CRUD methods, as well as a list endpoint and an endpoint to add/remove jurisdiction associations with the flag (`associateJurisdictions`). Permissions are limited to admin only. Flags associated with jurisdictions will appear when jurisdiction data is returned.

Start up the backend. Login as an admin, either through the API or partners portal. Test each endpoint:
- Create a feature flag
- Get the feature flag using the id returned ^^
- Update the feature flag using the id
- Get the feature flag to verify update
- Associate jurisdiction(s) with the flag
- Get the feature flag to verify association
- Get the jurisdiction to verify association
- Call the list endpoint
- Remove the jurisdiction(s) using associateJurisdictions
- Get the feature flag to verify removal
- Get the jurisdiction to verify removal
- Delete feature flag
- Call list endpoint to verify

## Author Checklist:

## Review Process:


ludtkemorgan (approved):
Tested out all of the endpoints and they work great! Just the one question about if we can change the type on the associate endpoint. But otherwise good, so approving!

api/src/dtos/feature-flags/feature-flag-associate.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+  @IsUUID(4, { groups: [ValidationsGroupsEnum.default] })
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty()
+  id: string;
+
+  @Expose()
+  @Type(() => IdDTO)
+  @IsArray({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty({ type: IdDTO, isArray: true })
+  associate: IdDTO[];
```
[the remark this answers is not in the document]
mcgarrye: I used IdDTO because I felt we will likely need the id AND name when using this endpoint instead of just sending the id. Ordinal is unnecessary but didn't see a better way to send both with the existing DTOs and this the pattern used elsewhere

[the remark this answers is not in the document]
mcgarrye: I wanted the validations on the id field which would be missing with a `string[]` being passed in. This is the patten I could find elsewhere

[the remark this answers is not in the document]
ludtkemorgan: Is the validation just that it is a UUID? Are you not able to do the validation on string because it's an array?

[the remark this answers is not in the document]
mcgarrye: Yes, just that it is a UUID. I couldn't find anywhere where we do uuid validation on an array. Can try to find a way to do nested validation

api/src/dtos/feature-flags/feature-flag-associate.dto.ts
```diff
@@ -0,0 +1,31 @@
...
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty()
+  id: string;
+
+  @Expose()
+  @IsArray({ groups: [ValidationsGroupsEnum.default] })
+  @IsUUID(4, {
+    groups: [ValidationsGroupsEnum.default],
+    each: true,
+  })
```
ludtkemorgan, line 19: Awesome! Thanks for making this change

api/src/controllers/feature-flag.controller.ts
```diff
@@ -0,0 +1,115 @@
...
+  @ApiOperation({
+    summary: 'Get a feature flag by id',
+    operationId: 'retrieve',
+  })
+  @ApiOkResponse({ type: FeatureFlag })
+  @UseGuards(ApiKeyGuard)
+  async retrieve(
+    @Param('featureFlagId', new ParseUUIDPipe({ version: '4' }))
+    featureFlagId: string,
+  ): Promise<FeatureFlag | null> {
```
YazeedLoonat, line 112: do you need the `| null`

api/src/controllers/feature-flag.controller.ts
```diff
@@ -0,0 +1,115 @@
...
+    return await this.featureFlagService.list();
+  }
+
+  @Post()
+  @ApiOperation({
+    summary: 'Create a feature flag',
+    operationId: 'create',
+  })
+  @ApiOkResponse({ type: FeatureFlag })
+  @UseGuards(ApiKeyGuard)
```
YazeedLoonat, line 60: since this requires an admin to making these changes wondering if the whole controller should have the ApiKeyGuard applied instead of on basically all the endpoints

api/src/services/feature-flag.service.ts
```diff
@@ -0,0 +1,207 @@
...
+export class FeatureFlagService {
+  constructor(
+    private prisma: PrismaService,
+    private jurisdictionService: JurisdictionService,
+  ) {}
+
+  /**
+        this will get a set of feature flags
+      */
+  async list(): Promise<FeatureFlag[]> {
```
YazeedLoonat, line 29: should this be jurisdiction specific? right now it would return all flags across all jurisdictions instead of all flags on a jurisdiction

although maybe I'm wrong on how this will be used

YazeedLoonat (approved):
Looks good, just got a few questions

api/src/services/feature-flag.service.ts
mcgarrye: This does not need to be jurisdiction specific. Use case for this is we want to associate a jurisdiction with flags, we call all flags to see which we want to associate with

api/src/controllers/feature-flag.controller.ts
mcgarrye: Huh guess not. I followed the jurisdiction controller when making most of this, but looking at other controllers it doesn't seem we do this. Will remove

Issue #4440, opened 2024-10-28
Accessibility Features verification


ludtkemorgan:
The Accessibility Features section is already in the core codebase but it hasn't been tested in a while.

Verify that the partner section appears when the `enableAccessibilityFeatures` is set to true for the jurisdiction.
Verify that when editing and creating listing the values are properly stored
Verify that the accessibility features section appears on public when a listing has saved values

![image](https://github.com/user-attachments/assets/9bac8870-4d04-4ed1-94c5-b10b0200a5ee)

mcgarrye:
Add backend unit tests and frontend integration tests

ludtkemorgan:
QA notes:
- This has been enabled in core https://partners.bloom.exygy.dev/ for the Bloomington jurisdiction
- Go to [add listing page](https://partners.bloom.exygy.dev/listings/add) and select Bloomington jurisdiction
  - The "Accessibility Features" section should be displayed
  - You should be able to save the listing after selecting accessibility features and is should save correctly (should display on listing detail page on partners after save)
  - The "Accessibility Features" section should not be displayed if you select a different jurisdiction and save with that jurisdiction
- Go to listing edit page and the same functionality should exist

sarahlazarich:
@ludtkemorgan this looks good and is responsive to the changing jurisdiction! 

I have the same question here about the listings export, but I think that may be a follow-up?

ludtkemorgan:
@sarahlazarich shoot also didn't realize this was part of the export. I'll create a follow-up ticket that will have both of them added to the CSV.

sarahlazarich:
Handling CSV in a follow-up, good to release

Issue #4439, opened 2024-10-28
Utilities Included verification


ludtkemorgan:
The Utilities Included section is already in the core codebase but it hasn't been tested in a while. 

- Verify that the partner section appears when the `enableUtilitiesIncluded` is set to true for the jurisdiction. 
- Verify that when editing and creating listing the values are properly stored
- Verify that the utilities included section appears on public when a listing has saved values
![image](https://github.com/user-attachments/assets/0788ccd6-bc97-4a19-b47c-4910b10dc5da)

mcgarrye:
Add backend unit tests and frontend integration tests

ludtkemorgan:
QA notes:
- This has been enabled in core https://partners.bloom.exygy.dev/ for the Bloomington jurisdiction
- Go to [add listing page](https://partners.bloom.exygy.dev/listings/add) and select Bloomington jurisdiction
 - The "Utilities Included" section should be displayed
 - You should be able to save the listing after selecting utilities included and is should save correctly (should display on listing detail page on partners after save)
  - The "Utilities Included" section should not be displayed if you select a different jurisdiction and save with that jurisdiction
- Go to listing edit page and the same functionality should exist

sarahlazarich:
@ludtkemorgan I still see this section when I switch the jurisdiction to "crestview" and save. Also wondering if I should also see this in the listings export?  

See [here](https://partners.bloom.exygy.dev/listings/a193792b-3d16-44d5-856e-54bd0731def0)

<img width="633" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd0tHQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--00b99f90ebf25ce30a2f4f80162f4ab60c98f790/image.png" alt="image.png" />

<img width="653" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd09HQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--025e62b150542e5d0ec1e5109542db67ed8c0408/image.png" alt="image.png" />

ludtkemorgan:
@sarahlazarich shoot, I didn't realize that this was part of the export in Detroit. I'll create a separate ticket for adding this and the accessibility features to the CSV.

[...]

PR #4503, opened 2024-12-05
fix: enable utilities and accessibility features

ludtkemorgan:
This PR addresses #4440 #4439 

## Description

Listing utilities and accessibility features functionality already exist. However it was not working correctly. This PR fixes all of the bugs as well as connecting it to the new feature flag functionality

NOTE: the jurisdiction table fields have also been removed since we can now use the feature flag table

Both feature flags are added to the staging seeded data for "Bloomington" jurisdiction so the following can be tested with a new reseed

- Create a new listing in partners for the "Bloomington" jurisdiction.
  - Verify that the Utilities checkbox section and the accessibility features checkbox section appear
  - Verify that you can select specific utilities and accessibility features. On save make sure they are saved correctly
- Edit the listing and change the utilities/accessibility features
  - Verify that the fields are correctly updated on save
- Create a new listing not for "Bloomington" jurisdiction and verify that the two sections don't appear
- Go to the public site and verify that the first listing created shows the correct utilities and accessibility features on the listing details page

## Author Checklist:

## Review Process:


sites/partners/src/lib/listings/AdditionalMetadataFormatter.ts
```diff
@@ -71,19 +71,21 @@ export default class AdditionalMetadataFormatter extends Formatter {
       this.data.reviewOrderType = ReviewOrderTypeEnum.waitlist
     }
 
-    if (this.data.listingFeatures) {
+    if (this.data.accessibilityFeatures) {
```
ludtkemorgan, line 74: This was the main reason the feature was not working. We need to send an object to the backend, but the frontend form is looking for an array. By using a different field to store the values we can convert it to the correct format on submission

api/prisma/migrations/23_remove_jurisdiction_enable_fields/migration.sql
```diff
@@ -0,0 +1,11 @@
+/*
+ Warnings:
```
YazeedLoonat, line 2: **polish:** I think its best if we remove these warnings in the migrations as long as we did our due diligence to make sure the end result is "safe"

YazeedLoonat (approved):
looks good, one small 🧹

api/prisma/migrations/22_add_bengali_and_arabic/migration.sql
```diff
@@ -6,8 +6,8 @@
...
 ALTER TYPE "languages_enum"
 ADD
-    VALUE 'bn';
+    VALUE IF NOT EXISTS 'bn';
 
 ALTER TYPE "languages_enum"
 ADD
-    VALUE 'ar';
\ No newline at end of file
+    VALUE IF NOT EXISTS 'ar';
```
mcgarrye, line 13: **nitpick:** EOF missing

api/prisma/migrations/23_remove_jurisdiction_enable_fields/migration.sql
```diff
@@ -0,0 +1,4 @@
+-- AlterTable
+ALTER TABLE
+  "jurisdictions" DROP COLUMN "enable_accessibility_features",
+  DROP COLUMN "enable_utilities_included";
```
mcgarrye, line 4: **nitpick (non-blocking):** EOF missing

mcgarrye (commented):
issue: signed in as an admin, I can still see the sections for a listing for another jurisdiction on creation

https://github.com/user-attachments/assets/6ef9a62a-f8a9-4c51-8efc-4b77739ecfcc

api/prisma/migrations/22_add_bengali_and_arabic/migration.sql
ludtkemorgan: fixed. For some reason on save the autoformatting keeps removing this line.

ludtkemorgan:
@mcgarrye good catch. I was missing an `s` and it now works correctly so is ready for re-review

emilyjablonski:
This feels like a very niche use case, but if I take a listing that has both utilities and a11y features, and then change the jurisdiction to be one that doesn't include them, the listing keeps the data.

Screenshot 2025-01-07 at 4 43 27 PM
I also think the spacing above the utilities here is too big, and the text size is small enough to be considered inaccessible - could we increase it a bit? Is there a design for what this should look like in core?

ludtkemorgan:
@emilyjablonski There is no design for this as these fields already existed in both public and partners, there was just some bugs with the saving of the values. So the design is the same as when it was implemented 3 years ago. Looks like that styling is coming from a ui-components component `AdditionalFees`.

sites/partners/src/components/listings/PaperListingDetails/sections/DetailUnits.tsx
```diff
@@ -60,10 +60,7 @@ const DetailUnits = ({ setUnitDrawer }: DetailUnitsProps) => {
     return t("t.none")
   }, [listing])
 
-  const enableHomeType =
-    profile?.jurisdictions
-      ?.find((j) => j.id === listing.jurisdictions.id)
-      ?.featureFlags?.find((flag) => flag.name === "homeType")?.active || false
+  const enableHomeType = doJurisdictionsHaveFeatureFlagOn("homeType", listing.jurisdictions.id)
```
ludtkemorgan, line 63: @YazeedLoonat I have added the util function for home type as mentioned https://github.com/housingbayarea/bloom/pull/830/files/93e3e7390be46afca3fb242d5ae09cd8b28fe90a#r1915735487

emilyjablonski:
@ludtkemorgan I'm about to rewrite that section, so it's no worries to me!

emilyjablonski (approved):
🐧

PR #4510, opened 2024-12-16
fix: household member delete + order

ColinBuyck:
This PR addresses [#977 ](https://github.com/metrotranscom/doorway/issues/977)

## Description

This PR updates the deleteModal open logic to handle an orderId=0. When a user applies online, the first household member is set to an OrderId of 0 but when added via the partner's side, it adds it with an orderId of 1. This PR also updates the ordering on the partner's side so that a user sees the household members in the order they added them (public side is already demonstrating this behavior). However, I stopped short of refactoring the orderId to be used more consistently to avoid any risks in conflicting with existing data. Instead, that work is captured here: https://github.com/bloom-housing/bloom/issues/4511

Note: I had originally assumed the orderId was null causing this issue but querying production showed no household members with null orderIds so this should resolve the issue.

This can be tested a few ways. On the public side, sign in and apply to a listing while adding household members, editing and deleting household members. Then submit, and do the same on an autofill application. Ensure that the household members behave as expected and maintain the order at which they were added.

Then on the partner's side, check these same applications and click delete to ensure the modal opens for each of them. Also, check that the order updates and saves correctly when deleting a household member.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/components/applications/PaperApplicationForm/sections/FormHouseholdMembers.tsx
+++ b/sites/partners/src/components/applications/PaperApplicationForm/sections/FormHouseholdMembers.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo, useCallback, useEffect } from "react"
+import React, { useState, useMemo, useCallback } from "react"
 import {
   HouseholdMember,
   HouseholdMemberUpdate,
@@ -84,7 +84,6 @@ const FormHouseholdMembers = ({
       const { birthMonth, birthDay, birthYear } = member
       const sameResidence = member.sameAddress
       const workInRegion = member.workInRegion
-      console.log(member.orderId)
 
       return {
         name: {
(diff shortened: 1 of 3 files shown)


sites/partners/src/components/applications/PaperApplicationForm/sections/FormHouseholdMembers.tsx
```diff
@@ -170,7 +170,7 @@ const FormHouseholdMembers = ({
       </Drawer>
 
       <Dialog
-        isOpen={!!membersDeleteModal}
```
ColinBuyck, line 173: **note:** When the modal number was set to zero (when app submitted on public side) this would be false

emilyjablonski:
Seeing this in the issue: `This appears to be related to the script Doorway uses to upload paper apps. The specific app they had an issue with was missing a specific field which wouldn't happen if creating the application from the UI. However, we should use the id rather than the ordinal number here here to avoid this happening in the future.` Just want to make sure that was accommodated?

ColinBuyck:
Good call out @emilyjablonski! I think there was an issue with my query checking the production db because as discussed in workshop... there are certainly household members with null order ids. I just pushed a new change that updates the order ids when they are pulled into the Paper Application Form and sorts them so they display in order that the user added them. While I don't love it, the reason I took this path rather than using the id is because a freshly added household member will not have an id until the paper app has been submitted so users would be unable to edit or delete household members while working on the paper application form.

ludtkemorgan (approved):
🦓

PR #4512, opened 2024-12-17
fix: set optOut watch default value

ColinBuyck:
This PR addresses https://github.com/metrotranscom/doorway/issues/1004

## Description

This PR updates the watch function to include a default value. The issue was that we were defaulting the canYouOptOutQuestion to true on all preferences that didn't explicitly set to no but depending on the watch value to be Yes in order to show the optOutText field. As a result, the watch value was undefined on load and wouldn't set to Yes until the user directly interacted with the radio leading to the confusing experience described in ticket. This one line change handles that case. 

This can be tested by editing existing preferences and adding new ones. In both cases, if the canYouOptOutQuestion shows yes, the optOutText field should be visible

## Author Checklist:

## Review Process:


emilyjablonski (approved):
🐩

PR #4513, opened 2024-12-17
fix: handle empty other race options

ColinBuyck:
This PR addresses https://github.com/metrotranscom/doorway/issues/1009

The ticket also mentions cleaning up the instances where doorway entered a space to get the "other" field to save which this PR doesn't cover.

## Description

This PR includes two different empty string checks in the formKeys file which prevents the selection of other without entering further information from being filtered out or being stored with a colon and space. 

This can be tested by 
a) submitting an online app with a mix of demographic data (including other selections without any text inputted) and going to the partners portal to see that it was stored correctly
b) submit a **paper** app to the same listing with a unique mix of demographic data also including empty other selections and then edit this paper app multiple times to ensure that the selections are saved
c) export the applications to that listing to see that they print in a legible way.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/utilities/application-export-helpers.ts
+++ b/api/src/utilities/application-export-helpers.ts
@@ -542,12 +542,12 @@ export const getHouseholdCsvHeaders = (
 export const convertDemographicRaceToReadable = (type: string): string => {
   const [rootKey, customValue = ''] = type.split(':');
   //only show colon if user entered a custom value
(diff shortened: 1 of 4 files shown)


ludtkemorgan (approved):
This looks good and solves the issue.

I did notice that if you use a `:` within the free text everything after the colon disappears. for example `human:alien` is saved as `human`. Looks like that issue also exists in main.

Issue #4517, opened 2024-12-18
Clean-up Unused Email Logic


mcgarrye:
Unused email template, email service function and user service logic exist in regards to the portalAccountUpdate email. The user service logic became unreachable after the prisma rewrite and since no translations could be found in the db, it is likely this email was never sent at all. Can be removed from the following places:
https://github.com/metrotranscom/doorway/blob/main/api/src/views/portal-account-update.hbs
https://github.com/metrotranscom/doorway/blob/main/api/src/services/email.service.ts#L238-L255
https://github.com/metrotranscom/doorway/blob/main/api/src/services/user.service.ts#L724-L740

mcgarrye:
QA: Test that existing account creation for public and partners still works.

sarahlazarich:
lgtm

PR #4526, opened 2025-01-07
chore: remove unused email logic

mcgarrye:
This PR addresses [#(4517)](https://github.com/bloom-housing/bloom/issues/4517)

## Description

Removes unreachable and unused email code that was lost in the prisma rewrite.

Test that existing account creation for public and partners still works.

## Author Checklist:

## Review Process:


api/src/services/user.service.ts
```diff
@@ -718,25 +718,6 @@ export class UserService {
...
-      // Partner user that is given access to an additional jurisdiction
-    } else if (
-      forPartners &&
-      existingUser &&
-      'userRoles' in dto &&
-      existingUser?.userRoles?.isPartner &&
-      dto?.userRoles?.isPartner &&
-      this.jurisdictionMismatch(dto.jurisdictions, existingUser.jurisdictions)
-    ) {
-      const newJurisdictions = this.getMismatchedJurisdictions(
```
ludtkemorgan, line 731: suggestion(non-blocking): The function `getMismatchedJurisdictions` can also be cleaned up. There is one other instance of it being used but in that case we just need to know if there is a mismatch. So we could delete this function and re-implement the appropriate logic in jurisdictionMismatch

ludtkemorgan (approved):
Looks good. Just one small suggestion

PR #4538, opened 2025-01-14
docs: update readme

emilyjablonski:

## Description

## Author Checklist:

## Review Process:


README.md
```diff
@@ -78,6 +78,8 @@ Running `yarn dev:all` from root runs 3 processes for both apps and the backend
 
 You can also run each process individually from separate terminals with the following command in each directory: `yarn dev`.
 
+We have a number of default users seeded for local development, the most basic of which being (email: `admin@example.com`, password: `abcdef`) which will login to both the public and partners sites, but you can view other default seeded users and their permissions by checking out the user section of the [seed file](https://github.com/bloom-housing/bloom/blob/aed77bf06525be359ef9205044fabbea2ab2576d/api/prisma/seed-staging.ts#L67).
```
mcgarrye, line 81: **thought:** I have a concern about adding the admin info directly to the readme as it does work in core and staging? environments

emilyjablonski: Yeah I hear that - all of our seed data is still publicly accessible though - I think if we don't want that data public, then we also need to remove it from the seed. We get this question a lot from folks onboarding - would it be better to just link to the seed file here?

mcgarrye: I think so. We should get this in with the new folks joining the team

mcgarrye (approved):
LGTM

Issue #2173, opened 2021-11-13
Applicant should not be able to user browser back button to re-submit an application


slowbot:
**Describe the bug**
Currently an applicant and use the browser back button post submission to navigate back and re-submit their applicant creating a new application id

**To Reproduce**
1. Go to any listing and submit an application
2. Use the browser back button to move to the terms page, Recheck terms and confirm application
3. You should get a new confirmation number a duplicate application is created

**Expected behavior**
You should get a browser warning when you try and navigate back from the confirmation screen. Your session should be reset and you should not be able to re-submit

**Desktop (please complete the following information):**
 - OS: Big Sur
 - Browser Chrome
 - Version 95.0.4638.69

sarahlazarich:
this looks good!

PR #4550, opened 2025-01-15
fix: app submission edge cases

ColinBuyck:
This PR addresses #2173 

## Description

This PR protects two duplicate submission paths. The first being spam clicking the submission button which was caused by the button loadingMessage prop not sufficiently blocking the submission call which is captured here (https://github.com/bloom-housing/ui-seeds/issues/98) and is now handled with a !submitting check at the top of the submit function. The second path was clicking back on the confirmation screen and clicking submit again. This is handled by redirecting the user if they already have a confirmation code since that is only saved on submitted applications.

Note that the ticket also includes reference to a "warning" but after talking to produce these two fixes are the desired behavior for handling this ticket.
 
This can be tested on the public side by logging in, fill out an application, and attempt to spam click the submit button. Then, click the back button on the confirmation page and notice the toast and redirect. You should end up on the my applications page where you will only see one submission for your attempted spam click.

Then, repeat this process as a signed out user and you should be redirected to the listings page and only see one submission in the applications table (via db).

## Author Checklist:

## Review Process:


sites/public/src/pages/applications/review/terms.tsx
```diff
@@ -42,47 +44,50 @@ const ApplicationTerms = () => {
   // eslint-disable-next-line @typescript-eslint/unbound-method
   const { register, handleSubmit, errors } = useForm()
   const onSubmit = (data) => {
-    setSubmitting(true)
-    const acceptedTerms = data.agree === "agree"
-    conductor.currentStep.save({ acceptedTerms })
-    application.acceptedTerms = acceptedTerms
-    application.completedSections = 6
+    // blocks multiple clicks and previously submitted applications
+    if (!submitting && !application.confirmationCode) {
```
ColinBuyck, line 48: **note:** This is the only line that's actually changed

jaredcwhite:
@ColinBuyck This is cool! I did notice in testing that you can still rapid-fire go back to one of the previous steps, where it doesn't load in any listing and we start seeing Next.js errors like `TypeError: Cannot read properties of undefined (reading 'save')`. But protecting all of the form routes might be its own effort (not sure if we have any existing ticket about that though).

ColinBuyck:
@jaredcwhite Thanks for checking this out! I think protecting all of the Application form routes will be its own effort but do you mind writing up that ticket since I am not certain what flow you tested?

jaredcwhite:
@ColinBuyck agreed, I wrote up a new ticket for those additional checks. #4587

Issue #4370, opened 2024-10-03
Zip the application csv export


ludtkemorgan:
We currently are unable to download very large application csv exports in Heroku.

We should do the following thing to help address this issue:
- In order to reduce the file size of the application export for listings, it should be zipped.
  - We have examples of doing this for both lottery export and the listing export

sarahlazarich:
Break into two tickets: 
- zip file: 2  
- look into netlify: 2

mcgarrye:
QA: 
With a listing that has applications, go to the applications tab.
Export the applications.
Confirm export was a zip file.
Open zip file, confirm csv is created.
Open CSV file and check contents.

Edit listing to be a lottery.
Run lottery.
Export lottery.
Confirm export was a zip file.
Open zip file, confirm XSLX is created.
Open XSLX file and check contents.

PR #4588, opened 2025-01-21
chore: refactor and extend zip code

mcgarrye:
This PR addresses [#(4370)](https://github.com/bloom-housing/bloom/issues/4370)

## Description

Generalized zip exporting code to allow for applications export to be zipped.

With a listing that has applications, go to the applications tab.
Export the applications.
Confirm export was a zip file.
Open zip file, confirm csv is created.
Open CSV file and check contents.

Edit listing to be a lottery.
Run lottery.
Export lottery.
Confirm export was a zip file.
Open zip file, confirm XSLX is created.
Open XSLX file and check contents.

Bring down the app.
Change `APPLICATION_EXPORT_AS_SPREADSHEET` to TRUE in the sites/partners/.env file.
With a listing that has applications, go to the applications tab.
Export the applications.
Confirm export was a zip file.
Open zip file, confirm XSLX is created.
Open XSLX file and check contents.

## Author Checklist:

## Review Process:


api/src/controllers/application.controller.ts
```diff
@@ -115,15 +116,28 @@ export class ApplicationController {
...
     queryParams: ApplicationCsvQueryParams,
   ): Promise<StreamableFile> {
-    return await this.applicationExportService.csvExport(req, res, queryParams);
+    const user = mapTo(User, req['user']);
+    const readStream = await this.applicationExportService.csvExport(
+      req,
+      res,
+      queryParams,
+    );
+    return await zipExport(
```
YazeedLoonat, line 133: **suggestion:** I think in general having logic in the controller just makes our life harder in the long run. Is there a way to change this so that calling the csvExport() returns the zipExport() on its own?

api/src/controllers/application.controller.ts
```diff
@@ -139,12 +153,21 @@ export class ApplicationController {
...
   ): Promise<StreamableFile> {
-    return await this.applicationExportService.spreadsheetExport(
+    const user = mapTo(User, req['user']);
+    const readStream = await this.applicationExportService.spreadsheetExport(
       req,
       res,
       queryParams,
       false,
     );
+    return await zipExport(
```
YazeedLoonat, line 163: **suggestion:** same as above

api/src/utilities/zip-export.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ *
+ * @param readStream
+ * @param zipFilename
+ * @param filename
+ * @param isSpreadsheet
+ * @returns a promise containing a streamable file
+ */
+export const zipExport = (
```
YazeedLoonat, line 14: **praise:** love this change!

sites/partners/src/lib/hooks.ts
```diff
@@ -555,10 +529,15 @@ export const useSpreadsheetExport = (
             { id: listingId, includeDemographics },
```
YazeedLoonat, line 529: **suggestion:** this may be a me thing but I think its easier to parse this code if we change it to a if/else if/else block. If you disagree feel free to ignore this suggestion

YazeedLoonat (changes requested):
Looks good, a few code changes I think will make this easier to maintain in the long run

api/src/controllers/application.controller.ts
mcgarrye: I did this to keep the csvExport() function testable, because if we return the zip file then we cannot test the result. I think I can make a new function to wrap all this into one though.

YazeedLoonat (approved):
LGTM

Issue #4543, opened 2025-01-15
Partners - Put the close button behind feature flag


ludtkemorgan:
Detroit currently doesn't support the ability "Close" an open listing. Instead they just "unpublish" a listing when they no longer want it visible on the public site. In order to accomodate this we will need to put the button behind a feature flag
![image](https://github.com/user-attachments/assets/04f3f1bd-1dad-46e3-a334-161ff3705d54) ![image](https://github.com/user-attachments/assets/9cdc496d-1e1c-4a92-9bff-b3594bb695d7)

Tasks needed:
- Add a feature flag for not displaying the "Close" button on listing edit. Something like `showCloseButton`
  - Note: The feature flag is to remove the button, not to add the button. This is because the main use case is to have the button shown
- Wrap the button in display in `ListingFormActions` so that if the feature flag is on to not display the button
  - Note: there is three separate button logics in that file based on the user's permissions

KrissDrawing:
QA Notes: It was core change, and it should change nothing in core it still should show 'close' button on partners edit open listing (when you are logged as admin).

sarahlazarich:
FF isn't turned on in Core, we should set up a core jurisdiction that is detroit-like for testing 
Still trying to figure out a way to keep the new detroit repo in sync

ludtkemorgan:
@sarahlazarich this feature has been turned on in the [new Detroit repo](https://bloom-detroit-partners.netlify.app/). You should no longer see the "close" button on an open listing

sarahlazarich:
great, looks good

PR #4598, opened 2025-01-22
fix: add feature flag to show close button

KrissDrawing:
This PR addresses #4543 

## Description

Uses `showCloseListingButton` to display `close` button on edit listing page. It is made only to hide that button, so when not present it will show that close button.

1. Without any seeding check edit open listing page. Close button will be visible.
2. Then either reseed staging db or add `showCloseListingButton` flag with `true` value it should still show `close` button.
3. Switch flag to `false`. It should hide `close` button

## Author Checklist:

## Review Process:


mcgarrye (approved):
LGTM

sites/partners/src/components/listings/ListingFormActions.tsx
```diff
@@ -77,6 +77,10 @@ const ListingFormActions = ({
     (jurisdiction) => jurisdiction.id === listing?.jurisdictions?.id
   )
 
+  const showCloseButton = listingJurisdiction?.featureFlags?.find(
+    (flag) => flag.name === "showCloseListingButton"
+  )?.active
```
ludtkemorgan, line 82: suggestion: There was a helper function created to handle this called "doJurisdictionsHaveFeatureFlagOn". It is part of the AuthContext. If you think it makes sense we could use it here

api/prisma/seed-staging.ts
```diff
@@ -79,6 +79,14 @@ export const stagingSeed = async (
...
     ),
   });
+  await prismaClient.featureFlags.create({
+    data: featureFlagFactory(
+      'showCloseListingButton',
+      true,
+      'When true, close button is displayed on the listing edit form',
+      [jurisdiction.id],
+    ),
+  });
```
ludtkemorgan, line 89: I think this is on me for having a bad description in the task. We want the close button showing as the default so if the feature flag doesn't exist on a jurisdiction the button will appear. 

In other words the feature flag should be "hideCloseListingButton" where if it is turned on for a jurisdiction we hide the button. So the inverse will have to happen on ListingFormActions. 

Sorry about the confusion!

ludtkemorgan (changes requested):
Sorry Kris for making this ticket confusing, but I think we'll need to to the inverse of what was done. Let me know if there are still some questions on what I asked

sites/partners/__tests__/components/listings/ListingFormActions.test.tsx
```diff
@@ -85,6 +85,10 @@ let partnerUser: User = {
   userRoles: { isPartner: true },
 }
 
+const doJurisdictionsHaveFeatureFlagOn = () => {
```
ludtkemorgan, line 88: You could add one test to this file to test the case where this returns true and the close button is not there

ludtkemorgan (approved):
Thanks for making those changes. Just one suggestion

Issue #4515, opened 2024-12-18
Convert listing list endpoints to post to stay in sync with Doorway


emilyjablonski:

YazeedLoonat:
we need to make sure we update doorway to consume the listing list endpoint as a post endpoint instead of a get endpoint

emilyjablonski:
Fixed

PR #4608, opened 2025-01-24
feat: filterable listing list post endpoint

mcgarrye:
This PR addresses [#(4515)](https://github.com/bloom-housing/bloom/issues/4515)

## Description

Creates new listings list endpoint with the same functionality but uses a POST instead of a GET.

Using the API, call the new endpoint `listings/list` using POST. Should have the same results as the GET `listings?` endpoint.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/test/integration/listing.e2e-spec.ts
+++ b/api/test/integration/listing.e2e-spec.ts
@@ -392,7 +392,7 @@ describe('Listing Controller Tests', () => {
       });
 
       const res = await request(app.getHttpServer())
-        .post('/listings/list')
+        .post('/listings')
         .set({ passkey: process.env.API_PASS_KEY || '' })
         .expect(200);
 
@@ -541,9 +541,9 @@ describe('Listing Controller Tests', () => {
       });
 
       const res = await request(app.getHttpServer())
-        .post('/listings/list')
+        .post(`/listings/list`)
         .set({ passkey: process.env.API_PASS_KEY || '' })
-        .expect(200);
+        .expect(201);
 
       expect(res.body.meta.currentPage).toEqual(1);
       expect(res.body.meta.itemCount).toBeGreaterThanOrEqual(2);
--- a/api/test/integration/listing.e2e-spec.ts
+++ b/api/test/integration/listing.e2e-spec.ts
@@ -380,20 +380,19 @@ describe('Listing Controller Tests', () => {
       });
     });
 
-    // without clearing the db between runs this test is flaky
-    it.skip('should get listings from list endpoint when no params are sent', async () => {
+    it('should get listings from list endpoint when no params are sent', async () => {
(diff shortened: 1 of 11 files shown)


shared-helpers/src/types/backend-swagger.ts
```diff
@@ -235,6 +235,28 @@ export class ListingsService {
       axios(configs, resolve, reject)
     })
   }
+  /**
+   * Get a paginated set of listings
+   */
+  filterableList(
```
ludtkemorgan, line 241: The swagger UI is working correctly, but the sample body is not valid parameters. For example, filter should be an array but it is just the first object in the sample.
```
{
  "page": 1,
  "limit": 10,
  "filter": {
    "$comparison": "=",
    "status": "active"
  },
  "view": "fundamentals",
  "orderBy": "mostRecentlyUpdated",
  "orderDir": "[\"desc\"]",
  "search": "search"
}
```
I wonder if there are things we could change in the `listings-query-params.dto.ts` file to make it work correctly

ludtkemorgan: Looks like the values are taken directly from `example` above each value. If those are updated it will fix the problem

api/test/integration/listing.e2e-spec.ts
```diff
@@ -531,6 +531,180 @@ describe('Listing Controller Tests', () => {
...
+          currentPage: 1,
+          itemCount: 0,
+          itemsPerPage: 1,
+          totalItems: 0,
+          totalPages: 0,
+        },
+      });
+    });
+
+    it('should get listings from list endpoint when params are sent', async () => {
```
ludtkemorgan, line 624: Suggestion: This tests that filtering by name works correctly. I don't think we have tests for some of the other filtering options such as `jurisdiction` or `zipCode`, could we add some more tests to have better coverage there?

ludtkemorgan (approved):
This change looks good functionally. Just some test and swagger suggestions. 
Do we have a ticket to switch the frontends to use this new endpoint?

api/test/integration/listing.e2e-spec.ts
mcgarrye: I will expand testing of the filter options in my next ticket since I'm adding a bunch of new filters.

api/test/integration/listing.e2e-spec.ts
```diff
@@ -531,6 +531,180 @@ describe('Listing Controller Tests', () => {
     });
   });
 
+  describe('filterableList endpoint', () => {
+    // without clearing the db between runs this test is flaky
+    it.skip('should not get listings from list endpoint when no params are sent', async () => {
```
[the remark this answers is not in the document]
mcgarrye: Able to get one working. The test to see it return no results cannot be done with our current testing infra

PR #4611, opened 2025-01-24
fix: set studio + sro to zero bedrooms

ColinBuyck:
This PR addresses https://github.com/metrotranscom/doorway/issues/1059 and https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4220

## Description

This PR saves Studios and SROs to have zero bedrooms rather than null bedrooms. This doesn't change much in core but it is critical for filtering in doorway which is a feature core will likely get soon.

After merge, I will be running the following query in Core, HBA, and Doorway since querying existing data showed that all null bedroom units were Studios or SROs
`Update units set num_bedrooms=0 where num_bedrooms is null;`

This can be tested by creating SROs and Studios on the partner side and then checking the units table to ensure they are saved as zero rather than null number of bedrooms.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/lib/listings/UnitsFormatter.ts
+++ b/sites/partners/src/lib/listings/UnitsFormatter.ts
@@ -25,6 +25,10 @@ export default class UnitsFormatter extends Formatter {
         case "oneBdrm":
           unit.numBedrooms = 1
           break
+        case "studio":
+        case "SRO":
+          unit.numBedrooms = 0
+          break
         default:
           unit.numBedrooms = null
       }
--- a/sites/partners/src/lib/listings/UnitsFormatter.ts
+++ b/sites/partners/src/lib/listings/UnitsFormatter.ts
@@ -25,12 +25,8 @@ export default class UnitsFormatter extends Formatter {
         case "oneBdrm":
           unit.numBedrooms = 1
           break
-        case "studio":
-        case "SRO":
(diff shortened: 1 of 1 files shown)


sites/partners/src/lib/listings/UnitsFormatter.ts
```diff
@@ -25,6 +25,10 @@ export default class UnitsFormatter extends Formatter {
         case "oneBdrm":
           unit.numBedrooms = 1
           break
+        case "studio":
+        case "SRO":
+          unit.numBedrooms = 0
+          break
```
ludtkemorgan, line 31: question(non-blocking): We don't have a use case for null number of bedrooms, so I wonder if we could make the default `0` instead. But that might be beyond this ticket

ludtkemorgan (approved):
Works as expected!

sites/partners/src/lib/listings/UnitsFormatter.ts
ColinBuyck: I'd be fine to make that change, just waiting for an answer from product in the original support thread. Will get that answered and this merged in Monday's sync if I don't hear back!

ColinBuyck: Got confirmation from product so updated that default!

Issue #4599, opened 2025-01-22
Backend - utilities included and accessibility features added to listing export


ludtkemorgan:
Both the "utilities included" and the "accessibility features" features should be added to the listing export CSV.

![Image](https://github.com/user-attachments/assets/8ec20145-5432-4b4e-b57a-2f3aca103837)

![Image](https://github.com/user-attachments/assets/f42f0a04-98d1-465d-8544-a5c3e0a6acd4)

Steps needed:
- Update the listing CSV export to have a column for utilities included
  - Should be a comma separated list of all utilities
  - Only show the column if at least one jurisdiction has this feature enabled
- Update the listing CSV export to have a column for accessibility features
  - Should be a comma separate list of all features
  - Only show the column if at least one jurisdiction has this feature enabled

ludtkemorgan:
QA Notes: 
- In the [core partner site](https://partners.bloom.exygy.dev/) click on the "Export to CSV"
- "Utilities Included" and "Property Amenities" columns should appear with appropriate data in each row

sarahlazarich:
lgtm

PR #4612, opened 2025-01-27
fix: add utilities included and accessibility to csv

ludtkemorgan:
This PR addresses #4599 

## Description

The "utilities included" and "accessibility features" feature need to be added to the listing CSV export when the features are enabled.

Also tackled the re-naming of the Home Type feature flag to keep it in line with the naming conventions for feature flags. Discussion mentioned [here](https://github.com/metrotranscom/doorway/pull/1055#issuecomment-2607626124)

Lastly, as the list of feature flags is growing it would be nice to have a place where all feature flags in the code base existed in one place. This was done by creating an enum that can be used in both the backend and the two frontends

Note:
- The names of the CSV columns come from the names used in the Detroit CSV export

Steps to test:
- Reseed the backend with staging data
  - This will give a variety of utilities included and accessibility features. Also one listing has neither
- Log into the [partner site](http://localhost:3001/) and click "Export to CSV"
- Verify that the "Property Amenities" and "Utilities Included" columns appear in the `listings.csv` with appropriate data 

There are no tests for this part of the code. I would like to add some tests, but want to get this PR out so will do it as a separate PR. 

## Author Checklist:

## Review Process:


api/src/dtos/feature-flags/feature-flag.dto.ts
```diff
@@ -10,14 +10,19 @@ import { ApiProperty } from '@nestjs/swagger';
...
   @MaxLength(256, { groups: [ValidationsGroupsEnum.default] })
   @IsDefined({ groups: [ValidationsGroupsEnum.default] })
-  @ApiProperty()
-  name: string;
+  @ApiProperty({
+    enum: FeatureFlagEnum,
+    enumName: 'FeatureFlagEnum',
+    example: 'sampleFeatureFlag',
+  })
+  name: FeatureFlagEnum | string;
```
emilyjablonski, line 25: **note:** Should this just be `FeatureFlagEnum`?

api/src/services/listing-csv-export.service.ts
```diff
@@ -418,13 +421,64 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
     ];
 
     if (
-      this.doAnyJurisdictionHaveFeatureFlagSet(user.jurisdictions, 'homeType')
+      this.doAnyJurisdictionHaveFeatureFlagSet(
```
emilyjablonski, line 424: **question:** Curious why we would use this here? I read this as "If the requesting user has any jurisdictions where this flag is enabled, add this to the listing csv export" when I may have expected "If the jurisdiction attached to this listing has it enabled" but could be misunderstanding!

ludtkemorgan: This export is for all listings the user has access to. So the headers of the CSV export need to support the case that at least one of the jurisdictions has `enableHomeType` turned on even if there is just a single listing on the export for that jurisdiction. Within each row we can look at what the listing's jurisdiction has set, but not at the header level.

api/src/dtos/feature-flags/feature-flag.dto.ts
ludtkemorgan: We need to accept all strings since it is not a requirement that the feature flags need to be from the Enum. But having the enum here exposes it in the swagger doc which allows the front-end to know about it

Issue #4560, opened 2025-01-15
Frontend Integration Tests - public/src/pages/sign-in


emilyjablonski:

adriencyberspace:
making sure the right fields are on the page, that password is required and throws an error, that it redirects you if you're already logged in, etc

ludtkemorgan:
This has been merged. Discussed offline that we would be splitting the ticket up to two. The tests that weren't handled as part of this ticket were added as "todos" in the test file

PR #4623, opened 2025-01-30
test: sign-in page integration test 

adriencyberspace:
This PR addresses #4560 

- [x] Addresses only certain aspects of the issue (explained below)

## Description

This PR adds unit tests for:
- All page elements on page
- Field validation
- If user is not logged in

This PR does not add a test for:
- If user is logged in, they should be redirected to /account/dashboard. This currently is not happening, so I created a separate ticket to address this and handle the test after fixing.

- Run `yarn test:unit __tests__/pages/sign-in.test.tsx` in `public`
- LMK if i'm missing anything!

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/src/locales/general.json
+++ b/shared-helpers/src/locales/general.json
@@ -463,7 +463,7 @@
   "authentication.signIn.mfaError": "This is a partner account, which for security reasons cannot login to the public site.",
   "authentication.signIn.passwordError": "Please enter a valid password",
   "authentication.signIn.passwordOutdated": "Your password has expired. Please reset your password.",
-  "authentication.signIn.pwdless.createAccountCopy": "Sign up quicky with no need to remember any passwords.",
+  "authentication.signIn.pwdless.createAccountCopy": "Sign up quickly with no need to remember any passwords.",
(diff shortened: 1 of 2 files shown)


ludtkemorgan (changes requested):
These are great tests. However, I think there are a lot more parts of the SignIn Component that will need to be tested in order to have full coverage. 

Here are some example of things that should be covered 
- There are two different forms within the SignIn component, FormSignInPasswordless and FormSignInDefault.
  - We should check that the correct one renders with the proper env set and make sure there is a test for the content within each one
- There is a "mandated account" flow that has a lot of additional content if it is turned on
  - There is also slightly different content if on desktop vs mobile
- We should add tests for all of the scenarios after successfully typing in password
  - regular vs passwordless
  - Make sure toast appears
  - "User is not confirmed" flow as well as "Resend Confirmation" flow
  
Let me know if you have any questions from a product perspective about these or need some pointers on the best way to test them

ludtkemorgan (approved):
@adriencyberspace I've added some "todos" into this file for future tests as discussed offline, but I think this PR is good to merge and we will follow-up with additional tests

Issue #4542, opened 2025-01-15
Backend: add all required filters to listing endpoint


ludtkemorgan:
The current listings endpoint allows for filtering but cannot filter on all of the fields we need it to to support both Detroit and Doorway.

The following aspects need to be filterable in the `/listings` endpoint so that we can add filters to the public site
| Name | Location | Notes |
| ------ | ---- | ------ |
| Number of Bedrooms | units | This exists currently but doesn't work properly |
| Community Type | reservedCommunityType | The frontend will only know the name of the types so filtering will be done on reservedCommunityType.name |
| Availability (open waitlist, available units, etc) | |  |
| Accessibility Features | listingFeatures | |
| Monthly rent | units | UI will have min rent and max rent. Which can be done by passing this filter twice (once with ">=" and once with "<=") |
| Home Type | homeType | |
| Region | region | |
| Confirmed Listings | isVerified | This functionality is not setup yet in partners or public |
| Section 8 Accepted | section8Acceptance | |
| Counties | buildingAddress.county | This is only needed for Doorway |
| Number of bathrooms | units | This is only needed for Doorway |

Things needed for this ticket:
- Add the above options to `listings-filter-params.dto.ts`
- Update `buildWhereClause` in listing.service.ts to handle all of the above filters

sarahlazarich:
also add tests

sarahlazarich:
would be nice to do this after https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4515

mcgarrye:
QA Notes: New filters can only be tested through the API. Test existing  happy paths on listings pages.

Issue #4616, opened 2025-01-29
BE - Availability Filtering 


mcgarrye:
Availability Filtering as it exists in Detroit can not be implemented and tested until after [#4523](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4523) is complete.

Additionally, Doorway also uses a filter called `availability` but with units instead of unit groups. The implementation should be compatible with both.

Options:
- Rename the fields to something like `unitAvailability` and `unitGroupAvailability`
- Find a way to utilize the unit group feature flag
- Other?

sarahlazarich:
There was a question on unit groups PR about this, was left out of Kris' PR, would want to tackle as a part of this 
There's a thing called "units available" that we don't use in Doorway 

This will need more research, will touch more than just the filtering

mcgarrye:
https://exygy.slack.com/archives/C01Q4QG5R8Q/p1738092399405379

PR #4628, opened 2025-02-04
feat: expand listing filters

mcgarrye:
This PR addresses [#(4542)](https://github.com/bloom-housing/bloom/issues/4542)

## Description

Adds the filters described in the ticket, as well as any missing filters in [Doorway](https://github.com/metrotranscom/doorway/blob/main/api/src/dtos/listings/listings-filter-params.dto.ts).
Logic has been alphabetized for read-ability. Unit tests were added for all filters. Should have no impact on existing functionality in Core, HBA or Doorway.

Start up the backend API.
Using the OpenAPI, call the POST `listings/list` endpoint.
Adjust the `filters` to test each filter.
Note1: the `search` field at the bottom of the sample payload will prevent anything from being found if not removed or changed.
NOTE2: we do not have any guardrails with which comparisons are used for each case. So using `=` on a list or `<=` on a string or `IN` on a number may have unpredictable results or even 500 errors.

Happy path testing for public listings page and partners listing page.
For partners listing page, try the search and signing in as jadmin and partner.

## Author Checklist:

## Review Process:


api/src/dtos/shared/base-filter.dto.ts
```diff
@@ -10,6 +10,7 @@ export enum Compare {
   'IN' = 'IN',
   '>=' = '>=',
   '<=' = '<=',
+  'LIKE' = 'LIKE',
```
ludtkemorgan, line 13: Thought: I know you mentioned this in the PR description and we don't have to do it as part of this PR since it is already fairly large. It seems like the filters more often than not only accept specific comparisons and if we pass in the wrong one it throws 500s on the backend. We could add additional validation using a [ValidationPipe](https://docs.nestjs.com/pipes#the-built-in-validationpipe). We use those in a few spots for more complicated validation

api/src/dtos/listings/listings-filter-params.dto.ts
```diff
@@ -1,59 +1,174 @@
...
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  [ListingFilterKeys.leasingAgent]?: string;
+
+  @Expose()
+  @ApiPropertyOptional({
+    type: Array,
+    example: ['elevator'],
+  })
+  @IsArray({ groups: [ValidationsGroupsEnum.default] })
+  [ListingFilterKeys.listingFeatures]?: string[];
```
ludtkemorgan, line 108: suggestion: we could make this of an enum type instead of string to add extra validation on the input

ludtkemorgan (commented):
I'm halfway done with the review and will finish tomorrow!

api/src/services/listing.service.ts
```diff
@@ -357,6 +357,200 @@ export class ListingService implements OnModuleInit {
 
     if (params?.length) {
       params.forEach((filter) => {
+        // TODO: handling availibility for units and unitGroups
```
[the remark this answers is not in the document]
mcgarrye: Separate ticket, as previously discussed #4616

api/src/dtos/shared/base-filter.dto.ts
mcgarrye: Not sure if this was discussed last bloom workshop but imo the team should talk about this and create a separate ticket to rework the functionality if decided on. Wouldn't want to balloon this ticket scope.

ludtkemorgan: We decided to not talk about it in the workshop. Sounds good to make it a separate ticket!

api/src/dtos/listings/listings-filter-params.dto.ts
mcgarrye: We would need to maintain both the db and enum to be in sync. Do we envision there being changes or additions to the features in the future?

api/test/unit/services/listing.service.spec.ts
```diff
@@ -1078,6 +1074,480 @@ describe('Testing listing service', () => {
     });
   });
 
+  describe('Test buildWhereClause helper', () => {
```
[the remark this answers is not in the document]
mcgarrye: Added!

api/prisma/seed-helpers/address-factory.ts
```diff
@@ -17,6 +17,7 @@ export const addressFactory =
 export const yellowstoneAddress = {
   placeName: 'Yellowstone National Park',
   city: 'Yellowstone National Park',
+  county: 'Teton County',
```
YazeedLoonat, line 20: **suggestion:** Since county isn't a required field I think some of our addresses should be missing county. Maybe add a few without county? or possibly just modify so that some of these don't have county baked in

shared-helpers/src/types/backend-swagger.ts
```diff
@@ -160,7 +160,7 @@ export class ListingsService {
       /**  */
       limit?: number | "all"
       /**  */
-      filter?: ListingFilterParams[]
+      filter?: any | null[]
```
YazeedLoonat, line 163: **issue:** I'm pretty sure this should still be a ListingFilterParam[] instead of having been changed to a really generic type. 

I think this means theres an issue with the DTO and we would need to peep to figure out what exactly went wrong

shared-helpers/src/types/backend-swagger.ts
```diff
@@ -2693,7 +2693,7 @@ export interface ListingsQueryParams {
   limit?: number | "all"
 
   /**  */
-  filter?: string[]
+  filter?: []
```
YazeedLoonat, line 2696: **issue:** same as above I think

YazeedLoonat (changes requested):
Looks good, I think changes to the DTO caused some loss in typing in the swagger file that needs addressing though

api/test/unit/services/listing.service.spec.ts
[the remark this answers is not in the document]
ludtkemorgan: The new tests look great!

YazeedLoonat (approved):
LGTM

api/src/dtos/listings/listings-filter-params.dto.ts
ludtkemorgan: There could potentially be changes in the future. I also think that an enum could be useful when grabbing the values in the frontend. However, that can be a future ticket

ludtkemorgan (approved):

Issue #4438, opened 2024-10-28
BE - Neighborhood Amenities


ludtkemorgan:
The "Neighborhood Amenities" feature needs to be ported over from Detroit to Core

Connect db fields up to the DTOs and all appropriate endpoints (listing create, edit, etc)

Notes: 
- The db fields (listing_neighborhood_amenities table) are already in core but it is not connected to anything
- The endpoints for this feature are in the Detroit codebase but will need to be updated for core to use Prisma
- Update endpoint will require deletion and insertion of external table data
- Frontend integration done in #4436 and #4437

sarahlazarich:
Table exists, we'd just need to update edit and create + properly save

sarahlazarich:
This is just free text so not sure if sql injection is a concern but should think about it

sarahlazarich:
make sure GET endpoint returns this field as well

PR #4629, opened 2025-02-04
feat: add neighborhood amenities to the backend

jaredcwhite:
This PR addresses #4438

## Description

This PR updates entities, seed data, listing service, and CSV export service to include neighborhood amenities.

This is just the backend portion, so the unit tests should cover the relevant flows.

## Author Checklist:

## Review Process:


api/src/services/listing-csv-export.service.ts
```diff
@@ -580,6 +580,30 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
           path: 'servicesOffered',
           label: 'Services Offered',
         },
+        {
```
ludtkemorgan, line 583: issue: I just realized this wasn't added to the ticket, but the neighborhood amenities will only appear if a feature flag is turned on. Because of that we only want to show these columns in the export if the feature flag is turned on. 

You can reference the "homeType" column further up in this file. You'll also need to create the feature flag in the DB to see it so I'd recommend creating it in "seed-staging" like what we do for `enableUtilitiesIncluded`. The new feature flag should be called something like `enableNeighborhoodAmenities`

api/prisma/seed-staging.ts
```diff
@@ -40,6 +40,7 @@ import { ValidationMethod } from '../src/enums/multiselect-questions/validation-
 import { randomNoun } from './seed-helpers/word-generator';
 import { householdMemberFactorySingle } from './seed-helpers/household-member-factory';
 import { featureFlagFactory } from './seed-helpers/feature-flag-factory';
+import { ListingNeighborhoodAmenities } from 'src/dtos/listings/listing-neighborhood-amenities.dto';
```
ludtkemorgan, line 43: chore: This import is not being used, it can be cleaned up.

ludtkemorgan (changes requested):
Looks good, we just need to put the one part behind a feature flag. Sorry I missed that in the story creation!

jaredcwhite:
@ludtkemorgan Ah I was just about to ask this after looking at the frontend task…yeah I'll work on moving the CSV export behind a feature flag.

ludtkemorgan:
> @ludtkemorgan Ah I was just about to ask this after looking at the frontend task…yeah I'll work on moving the CSV export behind a feature flag.

@jaredcwhite I also just merged a PR that adds an enum for feature flags that you can add the new feature flag to so we don't need to hardcode the name on both the frontend and the backnend

jaredcwhite:
@ludtkemorgan Cool, I'll take a look!

api/src/services/listing-csv-export.service.ts
ludtkemorgan: @jaredcwhite, looks like you forgot to remove these after adding them behind the feature flag above

jaredcwhite:
I did find a couple BE issues here I'm working through in the partners FE piece, so depending on the timing it may make more sense to address that PR with the BE included and close this PR…

jaredcwhite:
Closing in lieu of #4647

Issue #4631, opened 2025-02-05
NaN Summarizing Units Without Rent Information


ColinBuyck:
A listing in san jose had multiple units without the rent or rent as percentage income specified. This led to the experience captured down below. Further testing showed that a single unit without rent information was handled by showing n/a but multiple was leading to this confusing output. Ideally, the multiple units case should also display as n/a

![Image](https://github.com/user-attachments/assets/f9f563e6-e4af-4585-a552-4dc53475b2c8)

QA: https://bloom.exygy.dev/ and https://partners.bloom.exygy.dev/sign-in
Add multiple units of the same unit type to a listing that have no rent or rent as a percentage of income information. Ensure that the unit summary displays with n/a rather than NaN per month.

PR #4632, opened 2025-02-05
fix: null checks in minMax process

ColinBuyck:
This PR addresses #4631

## Description

The units summarization process was failing to handle multiple units without rent information. The first unit was correctly handled with "t.n/a" in the getDefaultSummary function but then from there the inputs to a minMax function were t.n/a and null leading to the NaN. This fix avoids the minMax comparison if the unit's associated field is null.

Add multiple units of the same unit type to a listing that have no rent or rent as a percentage of income information. Ensure that the unit summary displays with n/a rather than NaN per month.

## Author Checklist:

## Review Process:


emilyjablonski (approved):
🦩

Issue #4523, opened 2025-01-03
Add unit groups to core backend


ludtkemorgan:
We need to accommodate unit groups in Core so that Detroit can continue to work. 

Here is the full plan for this functionality: https://docs.google.com/document/d/1ijqTJhWZRwQAt0_UFPMJvxFrU6-b5-hxUAuTuB08ATY/edit?usp=sharing. 

This ticket will cover the following parts of the above document:
- Update database tables:
  - add `created_at` and `updated_at` columns to the `unit_group_ami_levels` table.
    - These should be automatically updated on create and update respectively
  - add `created_at` and `updated_at` columns to the `unit_group table.
    - These should be automatically updated on create and update respectively
- Create a new feature flag using the feature flag table:
  - Determine the name of the feature flag to be used in the below logic
  - staging seeded data should create this flag
- Update the create listing endpoint:
  - Add `UnitGroups` to the Listing DTO as an optional field
  - Properly save the unit group data to the DB if the DTO contains `UnitGroups`
    - Note: the data will span both `unit_group` and `unit_group_ami_levels`
  - Add additional verification that the request only has unit groups if the jurisdiction has the feature flag turned on. Similarly verify that it only has units if the feature flag is set to false.
    - In other words the request should not have both `units` and `unitGroups` and if it has one of them make sure it matches the feature flag set for that jurisdiction.
    - This can be done in the listing service immediately after checking the permissions of the user
- Update the update listing endpoint:
  - Properly save the unit group data to the DB if the DTO contains `UnitGroups`
    - Note: before update all existing unit groups associated to the listing need to be deleted. This is the same pattern we do for units
  - Add additional verification that the request only has unit groups if the jurisdiction has the feature flag turned on. Similarly verify that it only has units if the feature flag is set to false.
    - In other words the request should not have both `units` and `unitGroups` and if it has one of them make sure it 
- Update the get/list listing endpoints:
  - Update the returned view so that unit groups are returned with the data
  - unitsummaries logic will need to be updated to return correctly for unit groups
- Update CSV export:
  - Add an additional CSV to the listing CSV export that is for unit groups. 
    - Only need to generate this file if unit groups exist for the jurisdiction/listing

sarahlazarich:
BE steps laid out here, should be behind a FF based on jurisdiction 
Do we want to use new FF system? YES!!

sarahlazarich:
no product QA until FE

PR #4634, opened 2025-02-06
fix: add unit groups to core backend

KrissDrawing:
This PR addresses #4523

## Description

Adds unit groups from detroit.
there is feature flag `enableUnitGroups` it will turn on unitGroups instead of units.
it should also work for csv export.
It does not include `units summary`.

For testing i used something like this code below. For edit and create (for create it requires `id` just for unitTypes and amiChart). For edit it requires all id to be valid from your db.
There is check to validate if listing contains `units` and `unitGroups`, so it might require to manually remove them.

```
    const mockUnitGroup = {
      id: '4098205d-c43d-4779-aa53-87587c06d25e',
      createdAt: undefined,
      updatedAt: undefined,
      maxOccupancy: 2,
      minOccupancy: 1,
      floorMin: 1,
      floorMax: 5,
      totalCount: 10,
      totalAvailable: 8,
      bathroomMin: 1,
      bathroomMax: 1,
      openWaitlist: true,
      sqFeetMin: 400,
      sqFeetMax: 500,
      unitTypes: [
        {
          id: 'a7078959-0488-41ca-b5ed-c063275adddf',
        },
      ],
      unitAccessibilityPriorityTypes: {
        id: '1606f784-7560-4766-a1e1-1e7d667f0d2a',
      },
      unitGroupAmiLevels: [
        {
          id: undefined,
          amiPercentage: 30,
          monthlyRentDeterminationType:
            MonthlyRentDeterminationTypeEnum.percentageOfIncome,
          percentageOfIncomeValue: 30,
          flatRentValue: 2000,
          amiChart: {
            id: '3c717c11-7d9f-4e32-926e-034eb0ae34f5',
          },
        },
      ],
    };

    if (enableUnitGroups) {
      dto.units = undefined;
      dto.unitGroups = [mockUnitGroup];
    }
```
for `list()` and `findOne()` will return both `units` and `unitGroups` (but one or other should not be there, so it will return empty array for at least one).

## Author Checklist:

## Review Process:


api/src/services/listing.service.ts
```diff
@@ -195,11 +195,34 @@ export class ListingService implements OnModuleInit {
       }
     }
 
+    const viewInclude = views[params.view ?? 'full'];
+
+    const finalInclude = params.enableUnitGroups
```
ludtkemorgan, line 200: question: do we need to know if the unit groups are enabled? can we just always grab them and it won't find any if there aren't any unit groups? If you think it makes more sense to explicitly grab either or, can we move this logic within the `view` creation (roughly line 86 in this file)

api/src/services/listing.service.ts
```diff
@@ -654,10 +682,29 @@ export class ListingService implements OnModuleInit {
...
+      where: {
+        id: dto.jurisdictions.id,
+      },
+      include: {
+        featureFlags: true,
+      },
+    });
+
+    const enableUnitGroups = rawJurisdiction?.featureFlags?.find(
+      (featureFlag) => featureFlag.name === 'enableUnitGroups',
```
ludtkemorgan, line 699: Chore: My PR was merged this morning after you created this PR so this is mostly just a note. I created a new enum called `FeatureFlagEnum` where we can add the name of each feature flag. Once you rebase you could add this flag name there and replace all places of `enableUnitGroups` with the enum.

api/src/dtos/listings/listings-query-params.dto.ts
```diff
@@ -5,6 +5,7 @@ import { ListingFilterParams } from './listings-filter-params.dto';
 import {
   ArrayMaxSize,
   IsArray,
+  IsBoolean,
```
ludtkemorgan, line 8: Chore: `IsBoolean` isn't used anywhere in this file so it can be 🧹

ludtkemorgan (commented):
I'm still making my way through the PR, but in case I don't get to it again before you start I wanted to get my initial PR comments down

KrissDrawing:
Aside of above changes, for some reason i was not using `IdDTO` types for stuff that is just there to connect, so i changed that (and updated sample data in description).

api/src/services/listing.service.ts
```diff
@@ -655,9 +689,32 @@ export class ListingService implements OnModuleInit {
...
+    )?.active;
+
     dto.unitsAvailable =
-      dto.reviewOrderType !== ReviewOrderTypeEnum.waitlist && dto.units
-        ? dto.units.length
+      dto.reviewOrderType !== ReviewOrderTypeEnum.waitlist &&
+      (enableUnitGroups ? dto?.unitGroups?.length : dto?.units?.length)
+        ? enableUnitGroups
+          ? dto?.unitGroups?.length
+          : dto?.units?.length
```
KrissDrawing, line 717: I'm quite sure it's not right 😞 . I don't even know if we're using that in Detroit right now. in core it seem to be the one in partners listing table. So i don't know if i should just make it `undefined` for `enableUnitGroups`, or sum `totalAvailable` from each `unitGroup` 🤔
@ludtkemorgan

api/src/decorators/validate-units-required.decorator.ts
```diff
@@ -0,0 +1,42 @@
+import {
+  registerDecorator,
+  ValidationOptions,
+  ValidationArguments,
+} from 'class-validator';
+import { PrismaService } from '../services/prisma.service';
+import { ListingPublishedUpdate } from '../dtos/listings/listing-published-update.dto';
+import { FeatureFlagEnum } from '../enums/feature-flags/feature-flags-enum';
+
+export function ValidateUnitsRequired(validationOptions?: ValidationOptions) {
```
ludtkemorgan, line 10: 💯 love this decorator
nit: Can you add a comment above this so future people know why we have this function and why it's only required if `enableUnitGroups` is false

api/src/services/listing-csv-export.service.ts
```diff
@@ -222,19 +258,30 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
...
-    );
-    // TODO: the below is essentially the same as above in this.createCsv
+  async createUnitCsv(
+    filename: string,
+    listings: Listing[],
+    enableUnitGroups?: boolean,
+  ): Promise<void> {
+    const csvHeaders = enableUnitGroups
+      ? this.getUnitGroupCsvHeaders()
+      : this.getUnitCsvHeaders();
```
ludtkemorgan, line 268: Issue: This doesn't work for the case that both "unit group" jurisdictions as well as "unit" jurisdictions exist in partners. It will only output the unit group version. Instead in that case the zip should have both the `unit.csv` and `unitGroups.csv`

api/src/services/listing-csv-export.service.ts
```diff
@@ -431,6 +478,24 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
...
+      headers.push({
+        path: 'unitGroups.length',
+        label: 'Number of Unit Groups',
+      });
+    } else {
+      headers.push({
+        path: 'units.length',
+        label: 'Number of Units',
+      });
+    }
```
ludtkemorgan, line 497: Issue: same here. If there is a case where a unitGroup jurisdiction exists along with a unit jurisdiction we will want both columns to appear. I don't know if in production we'll ever see that, but we should handle it just in case

api/src/services/listing.service.ts
```diff
@@ -1303,12 +1451,29 @@ export class ListingService implements OnModuleInit {
...
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const transactions = await this.prisma.$transaction([
-      // delete all connected units before recreating in update
-      this.prisma.units.deleteMany({
-        where: {
-          listingId: storedListing.id,
-        },
-      }),
+      // Conditionally delete either units or unitGroups based on feature flag
+      ...(enableUnitGroups
```
ludtkemorgan, line 1455: nit: Once again, I don't think this case will ever come up in production. But if a listing changes from a jurisdiction that has units and then requires unitGroups the original units wouldn't get deleted. We could potentially just always do all three of these deletes and not put it behind the `enableUnitGroups` flag. But I'd be curious what others think.

api/src/services/listing.service.ts
```diff
@@ -1507,63 +1672,112 @@ export class ListingService implements OnModuleInit {
...
-                    ? {
-                        create: {
-                          items: unit.unitAmiChartOverrides.items,
-                        },
-                      }
-                    : undefined,
-                  unitAccessibilityPriorityTypes:
-                    unit.unitAccessibilityPriorityTypes
+          units:
+            !enableUnitGroups && dto.units
```
ludtkemorgan, line 1676: question: I wonder if it would make sense to replace some of these checks for `enableUnitGroups` with throwing a `BadRequestException` at the beginning of this function if units are submitted on a jurisdiction that should have `unitGroups`

api/src/utilities/unit-utilities.ts
```diff
@@ -542,3 +544,49 @@ export const summarizeUnits = (
   );
   return data;
 };
+
+export const convertToTitleCase = (value: string): string => {
```
ludtkemorgan, line 548: This and the functions would be a good functions to add unit tests for in `unit-utilities.spec.ts`

ludtkemorgan (changes requested):
This is looking great. Haven't found any major issue. Just some thoughts and minor issues found so far

api/src/services/listing.service.ts
KrissDrawing: I updated that, but after changing condition for preventing submitting both units and unitGroups (line: 1583) it should throw an error when such case happen, before transaction.

api/src/services/listing-csv-export.service.ts
```diff
@@ -37,6 +37,10 @@ import { ListingMultiselectQuestion } from '../dtos/listings/listing-multiselect
 import { Jurisdiction } from '../dtos/jurisdictions/jurisdiction.dto';
 import { ListingUtilities } from '../dtos/listings/listing-utility.dto';
 import { ListingFeatures } from '../dtos/listings/listing-feature.dto';
+import { UnitType } from 'src/dtos/unit-types/unit-type.dto';
+import { UnitGroupAmiLevel } from 'src/dtos/unit-groups/unit-group-ami-level.dto';
```
mcgarrye, line 41: **nitpick (blocking):** these imports should be updated to `../`

api/src/services/listing-csv-export.service.ts
```diff
@@ -119,33 +127,82 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
...
+        listings as unknown as Listing[],
+        true,
+      );
+    }
+
+    await this.createUnitCsv(
+      unitFilePath,
+      listings as unknown as Listing[],
+      false,
+    );
```
mcgarrye, line 185: **question (blocking):** Should this be wrapped in an `if (hasUnits)`?

api/src/services/listing-csv-export.service.ts
```diff
@@ -119,33 +127,82 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
...
     const unitCsv = createReadStream(unitFilePath);
```
mcgarrye, line 186: **suggestion (non-blocking):** You can move this line to the if-then block at line 200

api/src/services/listing-csv-export.service.ts
```diff
@@ -330,6 +398,17 @@ export class ListingCsvExporterService implements CsvExporterServiceInterface {
...
+    jurisdictions: Jurisdiction[],
+    featureFlagName: string,
+  ) => {
+    return jurisdictions.some((juris) => {
+      return !juris.featureFlags.some(
+        (flag) => flag.name === featureFlagName && flag.active,
+      );
+    });
+  };
+
```
mcgarrye, line 411: **suggestion (non-blocking):** Can we move this and the `doAnyJurisdictionHaveFeatureFlagSet` functions into a utilities helper file? Would be cleaner to have them separated and will likely need them in other services going forward

api/src/services/listing.service.ts
```diff
@@ -855,6 +889,32 @@ export class ListingService implements OnModuleInit {
...
+        id: dto.jurisdictions.id,
+      },
+      include: {
+        featureFlags: true,
+      },
+    });
+
+    const enableUnitGroups = rawJurisdiction?.featureFlags?.find(
+      (featureFlag) => featureFlag.name === FeatureFlagEnum.enableUnitGroups,
+    )?.active;
```
mcgarrye, line 903: **thought:** Could be a helper function

api/src/utilities/unit-utilities.ts
```diff
@@ -10,6 +10,8 @@ import { UnitType } from '../dtos/unit-types/unit-type.dto';
 import { UnitAccessibilityPriorityType } from '../dtos/unit-accessibility-priority-types/unit-accessibility-priority-type.dto';
 import { AmiChartItem } from '../dtos/units/ami-chart-item.dto';
 import { UnitAmiChartOverride } from '../dtos/units/ami-chart-override.dto';
+import { isEmpty } from 'class-validator';
+import UnitGroupAmiLevel from 'src/dtos/unit-groups/unit-group-ami-level.dto';
```
mcgarrye, line 14: **nitpick:** `../` for import

ludtkemorgan (approved):
This looks good. Thanks for making the changes!

PR #4635, opened 2025-02-06
4570/Partners listing page integration tests

matzduniuk:
This PR addresses #4570

## Description

* Extensive suite of integration tests for the partners' site single listing page
* Small code cleanup by removing unnecessary type assertions
* Adding missing translations to the general.json translation files

Remaining Issues:
  * The ` <FieldValue>` shared component uses a `<p>` as a wrapper for the children component, which raises console errors when some elements are passed (e.g.: `<div>`, `<ul>`)>. This error does not interfere with the application's operation. Still, it could be fixed by either replacing the `<p>` element with `<span>` or adding a new optional prop for selecting a type of wrapper component.
 
* The `amiCharts` API call for the unit details drawer "Eligibility" section proved to be difficult to capture for mocking therefore, this field has been tested only for properly displaying the field label and the missing data placeholder. 

* Newly added label strings are missing translation for languages outside of English  

Run the full suite of test using command:
```
yarn jest --verbose -t "listing data"
```

## Author Checklist:

## Review Process:


shared-helpers/src/locales/general.json
```diff
@@ -750,6 +750,12 @@
   "listings.waitlist.submitAnApplication": "Once ranked applicants fill all available units, the remaining ranked applicants will be placed on a waitlist for those same units.",
   "listings.waitlist.submitForWaitlist": "Submit an application for an open slot on the waitlist.",
   "listings.waitlist.unitsAndWaitlist": "Available Units and Waitlist",
+  "listings.applicationType.onlineApplication": "Online Applications",
```
emilyjablonski, line 753: **note:** We don't have this documented well, but we split translations for the public and partners sites into two separate files, because we don't translate any of the partners strings. Thank you for catching these! We just need to move them to `sites/partners/page_content/locale_overrides/general.json`

emilyjablonski (approved):
These look awesome! And the coverage reports look great. Just need to move the strings over to the other file.

Issue #4596, opened 2025-01-22
Partners - FE/BE - Region


emilyjablonski:

emilyjablonski:
@ludtkemorgan Noticed this one was missing - want to make sure we do intend to bring it over! If not I can close

ludtkemorgan:
@emilyjablonski nice catch. Looks like I missed this. I think we need to have a mini spike to see how we can handle the enums as the long term solution they shouldn't be hardcoded. I'll create a subtask for that

sarahlazarich:
@emilyjablonski do we need designs for this, or would this be ready to pull in?

sarahlazarich:
Question for @emilyjablonski and @eajensenwa : 
- Will we ever need to show the application date into in the right rail?

emilyjablonski:
@sarahlazarich Yes we do this already (if I'm understanding correctly) like in this listing: https://housingbayarea.mtc.ca.gov/listing/ec69bcb8-8334-4c3b-9756-46cd466dcfd8/riviera_family_apartments_1515_riviera_ave_walnut_creek_ca

ludtkemorgan:
This is ready to be QA'd now at https://bloom-detroit-partners.netlify.app/ 

![Image](https://github.com/user-attachments/assets/0c29950d-91ab-425f-8e9d-bd0fd2d2e0da)

sarahlazarich:
Noticed one thing - 

was looking at the export and saw there was this "Hollywood" option in the neighborhood field certain listings, https://bloom-detroit-partners.netlify.app/listings/af2dc778-5a7c-4d36-811a-71d5850f0b68 for example, but I don't think that's an option. 

<img width="240" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzBtQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2216c1d4167ea5f0bf6d04ac3c10e5b47e1bbf37/image.png" alt="image.png" />



<img width="223" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3dtQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5102cd0f5a59cbdce8c32f382d16ac8673d6c200/image.png" alt="image.png" />


[...]

PR #4643, opened 2025-02-14
4596/partners region and neighborhood

matzduniuk:
This PR addresses #4596 

## Description

* Add new `enableRegions` jurisdictions feature flag
* Update listing DTOs' to include the region field
* Update the Building Details section to show the region field if the feature flag is set to true
* Update Neighborhood and Region edit page input fields to be either text or controlled select input types

* Turn on the enable `enableRegions`
* Open any listing page and view the `Building Details` section (if not set the Fields return `n/a`)
* Enter the listing edit page, and select a given region (this will result in the `Neighborhood` field reset)
  * At this point you can select Neighborhoods from different regions to test the auto-changing fields.
* Save the listing and re-verify the `Building Details` section

## Author Checklist:

## Review Process:


sites/partners/src/components/listings/PaperListingForm/sections/BuildingDetails.tsx
```diff
@@ -144,6 +152,20 @@ const BuildingDetails = ({
       : defaultMessage
   }
 
+  const enableRegions = doJurisdictionsHaveFeatureFlagOn(
+    FeatureFlagEnum.enableRegions,
+    listing?.jurisdictions?.id
+  )
```
ludtkemorgan, line 158: Issue: This isn't working as expected. I have the "Bloomington" jurisdiction set to true and another jurisdiction where it is not turned on. No matter what jurisdiction you have selected on a new listing the enableRegions is showing. The only way I can get it to disappear is to save it with the other jurisdiction and then open the listing to edit it.

Potential solution would be to switch out `listing?.jurisdictions?.id` with the jurisdiction id from the form. This can be grabbed with `const jurisdiction = watch("jurisdictions.id")`. `AdditionalFees.tsx` is a good example of it

sites/partners/src/components/listings/PaperListingForm/sections/BuildingDetails.tsx
```diff
@@ -277,16 +318,52 @@ const BuildingDetails = ({
...
+                    className: "w-full",
+                    onChange: () => setValue("neighborhood", undefined),
+                  }}
+                  options={Object.keys(RegionEnum).map((entry) => ({
+                    value: entry,
+                    label: entry.toString().replace("_", " "),
+                  }))}
+                />
+              </FieldValue>
+            ) : (
```
ludtkemorgan, line 338: Issue:
I think this year built field should display on the left column if the region field exists rather than in the third column. Could we just always display the field and it would end up in the left column if the region also exists?

image

sites/partners/src/components/listings/PaperListingForm/sections/BuildingDetails.tsx
```diff
@@ -277,16 +318,52 @@ const BuildingDetails = ({
...
+                placeholder={t("listings.yearBuilt")}
+                type={"number"}
+                register={register}
+              />
+            )}
           </Grid.Cell>
         </Grid.Row>
+        <GridRow columns={3}>
+          <GridCell>
+            <p className="field-sub-note">{t("listings.requiredToPublish")}</p>
```
ludtkemorgan, line 352: Issue: The "required to publish" is floating pretty far below the fields 

This PR:
image

In Main:
image

ludtkemorgan (changes requested):

sites/partners/src/components/listings/PaperListingForm/sections/BuildingDetails.tsx
matzduniuk: If all the fields are placed in the same row, the size of each input is too small for good UX, especially the `City` field, as most names will be cut off.  
Nevertheless I implemented the new design so feel free to test it yourself

mcgarrye:
Seems you have some merge conflicts that need to be addressed

mcgarrye:
Non-blocking: Would love for us to consider improving the layout of the listings details here
Screenshot 2025-02-25 at 2 34 05 PM
Don't want to hold this ticket up but can we review this? @ludtkemorgan @emilyjablonski @eajensenwa

mcgarrye:
question: should the neighborhood drop down options change if a region is already selected? If we can narrow the options that may be helpful. Opinions: @ludtkemorgan ?

mcgarrye (approved):
Looks good! Approving but hoping questions can be addressed before merge.

Also want to call out that in the unlikely case listings have neighborhoods and then the flag is turned on, they will be overwritten in subsequent edits. And if the flag is turned off, the neighborhood can be altered in the text box. This is an edge case and unlikely to occur.

sites/partners/src/components/listings/PaperListingForm/sections/BuildingDetails.tsx
ludtkemorgan: Hey @matzduniuk. I was picturing something more like this. Where year built is on a new line but in the left column.
image

I realize that my original comment isn't very valid as in order to accommodate this design it will require having "yearBuilt" be two places in the code depending on if `enableRegions` is set or not

ludtkemorgan:
@mcgarrye It was decided that we should keep the behavior as close to what it is in Detroit right now. So always display all of the neighborhoods but if they select a neighborhood adjust the region appropriately. I think in the future we can think through possible iterations

ludtkemorgan (approved):
This looks great. Thanks for making the changes!

PR #4654, opened 2025-02-19
fix: minor filter adjustments for doorway

mcgarrye:
This PR addresses #(4542)

## Description

Doorway frontend (as well as Detroit) send back strings for bedrooms and bathrooms. Filter being a number without casting the string to a number caused 500 errors and the map to break. Adding @Type casts the string to a number and allows for numbers to be sent as well. Still catches bad strings that are not numbers.

Other fixes caught:
- isArray accidentally removed from @ApiPropertyOptional for OrderBy and OrderDir fields, caused complaints in FE code, no errors
- Adjust listing-factory to alleviate issue in Doorway e2e tests

Spin up backend.
Test bedroom and bathroom filters. Use both numbers, numbers as strings and random strings (success, success, failure respectively)

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/dtos/listings/listings-filter-params.dto.ts
+++ b/api/src/dtos/listings/listings-filter-params.dto.ts
@@ -1,6 +1,6 @@
 import { ApiPropertyOptional } from '@nestjs/swagger';
 import { HomeTypeEnum, ListingsStatusEnum, RegionEnum } from '@prisma/client';
-import { Expose } from 'class-transformer';
+import { Expose, Type } from 'class-transformer';
 import {
   IsArray,
   IsBoolean,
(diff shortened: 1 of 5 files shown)


api/src/dtos/listings/listings-filter-params.dto.ts
```diff
@@ -29,16 +29,18 @@ export class ListingFilterParams extends BaseFilter {
 
   @Expose()
   @ApiPropertyOptional({
-    example: 2,
+    example: '2',
   })
   @IsNumber({}, { groups: [ValidationsGroupsEnum.default] })
```
ludtkemorgan, line 34: thought: The `IsNumber` class validator allows decimals by default. Since we wouldn't probably ever want a fraction of number you could restrict this more
```
@IsNumber({{maxDecimalPlaces: 0 }, { groups: [ValidationsGroupsEnum.default] })
```
However, if you do pass in a decimal number the functionality still works (e.g pass in 3.5 for bedrooms and 4 bedroom+ are returned). So we don't need to restrict it if you think it is overkill

ludtkemorgan (approved):
Looks great!

api/src/dtos/listings/listings-filter-params.dto.ts
mcgarrye: May want to leave this if we need to support something like half bath or half/flex bedroom

Issue #4639, opened 2025-02-10
Public - Section 8


sarahlazarich:
As a user of the public site in Detroit,
I want to be able to see the section 8 disclaimer on my listing detail page
So that I know if my voucher will apply for that listing. 

On listing detail page: 

<img width="1450" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejdOQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d2ea392edb41568de9b3920c1dda5859154a9da2/image.png" alt="image.png" />

<img width="1368" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBei9OQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e9115ccffe529fbe3e519c28b57015e018d117b5/image.png" alt="image.png" />

AC: 
- If a listing accepts section 8, the section 8 disclaimer should show up on the public site in the listing detail page under the unit types table in RentSummary
- If a listing accepts section 8, the section 8 disclaimer should show up on the public site in the listing detail page in the eligibility section
- You should be able to filter based on this in the filter drawer and in rental finder - if that's not present already, we'll handle that in a separate ticket

Note: Blocked by https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/4638

sarahlazarich:
looks good!

PR #4660, opened 2025-02-24
4639/public section 8

matzduniuk:
This PR addresses #4639 

## Description

* Add a new translation string for the public site section 8 info text
* Render the section 8 info text to the public listing details view page

* Make sure that the `section8Acceptance` flag is turned on for a given listing
* Enter the given listing details page

## Author Checklist:

## Review Process:


ColinBuyck (commented):
This looks good! Only thing is that I am thinking it should be included in the seeds listing view as well? Since that's what Detroit will be getting if I am understanding the plan correctly? Any thoughts here @emilyjablonski

emilyjablonski:
@ColinBuyck Yes good call! We're building new frontend public designs, specifically to be able to support Detroit. These are by default toggled off but can be toggled on with `SHOW_NEW_SEEDS_DESIGNS=TRUE` in the .env in public. We'll need to add new Detroit features to these new files - for the listing view, it's `ListingViewSeeds`

sites/public/src/components/listing/ListingViewSeeds.tsx
```diff
@@ -174,9 +175,19 @@ export const ListingViewSeeds = ({ jurisdiction, listing, preview }: ListingProp
             reviewOrderType={listing.reviewOrderType}
             unitsSummarized={listing.unitsSummarized}
           />
+          {listing.section8Acceptance && (
+            <div className="my-2">
```
emilyjablonski, line 179: **issue:** We're moving off of Tailwind for CSS and toward our internal Seeds component library tokens & classes, so instead of the Tailwind class here, we'll be able to use the tokens/classes specified [here](https://github.com/bloom-housing/ui-seeds/blob/main/src/global/spacing.scss) for spacing. I know it was impossible for you to know this because this particular ticket didn't have a design asset - but we'll also want to use the text spacing token instead of the `2` value, and only above this section not also below - so this will look like just changing the class to `seeds-m-bs-text`, m for margin, bs for block-start, and text for our text-specific spacing.

This is also not appearing correctly on mobile by not having enough block margin. To fix this, we can move this section into the RentSummary component above it at the bottom, and it will automatically get the right spacing!

Lastly, the text size is too small, which is coming from the `custom-counter__subtitle` class on the Markdown component - we can just remove that and we should be good to go!

sites/public/src/components/listing/listing_sections/Eligibility.tsx
```diff
@@ -4,17 +4,30 @@ import { CollapsibleSection } from "../../../patterns/CollapsibleSection"
...
   if (!eligibilitySections.length) return
   return (
     <CollapsibleSection
       title={t("listings.sections.eligibilityTitle")}
-      subtitle={t("listings.sections.eligibilitySubtitle")}
+      subtitle={
+        <div>
+          {t("listings.sections.eligibilitySubtitle")}
+          {section8Acceptance && (
+            <>
```
emilyjablonski, line 23: **issue:** Let's rely on the same token spacings here instead of the `<br />` elements - this would look like this!

```suggestion
             {section8Acceptance && (
            <div className="seeds-m-bs-content">
              <Markdown>{t("listings.section8VoucherInfo")}</Markdown>
            </div>
          )}
```

sites/public/src/components/listing/ListingViewSeeds.tsx
emilyjablonski: Screenshot 2025-02-26 at 11 08 16 AM
I think we still need to move this into `RentSummary` so that it appears correctly on mobile @matzduniuk

emilyjablonski (approved):

Issue #4605, opened 2025-01-22
Public - New Designs - Content page


emilyjablonski:

sarahlazarich:
Core, then pull over
No change to links out in doorway (all external) - don't need to toggle, assuming that the content in FAQ won't change 
Test this with multiple styles per last paragraph

sarahlazarich:
list every page that this touches in the ticket

PR #4672, opened 2025-03-05
fix: inverse styling support

ColinBuyck:
This PR is a follow up to #4605 

## Description

As part of the Doorway uptake, I remembered that the core base layout is actually the inverse and supporting Doorway without pattern divergence will require this inverse prop.

This should have no visual changes on disclaimer or privacy policy page.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/src/patterns/PageHeaderLayout.tsx
+++ b/sites/public/src/patterns/PageHeaderLayout.tsx
@@ -4,7 +4,7 @@ import styles from "./PageHeaderLayout.module.scss"
 
 export interface PageHeaderProps {
   /** Heading at the top of the page header */
-  heading: string
+  heading: React.ReactNode
   /** Description below the heading */
   subheading?: React.ReactNode
 }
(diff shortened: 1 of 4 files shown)


sites/public/src/patterns/PageHeaderLayout.tsx
```diff
@@ -4,14 +4,16 @@ import styles from "./PageHeaderLayout.module.scss"
 
 export interface PageHeaderProps {
   /** Heading at the top of the page header */
-  heading: string
+  heading: React.ReactNode
```
ColinBuyck, line 7: **polish:** For consistency with subheading and component typing

emilyjablonski (approved):
Thank you for making this change! I think this could spark good conversation in our next check-in w design about how to set a core=inverse, doorway=primary change like this at a global theme level, vs for example, what we will have to do which is making a change directly in the content pages in Doorway to remove the inverse flag. This looks great tho!

PR #4696, opened 2025-03-11
Feat: public create account integration tests

matzduniuk:

## Description

* Add the event testing library package to the public sites module
* Add `es2020` target to public module TS configuration
* Implement password reset page integration tests following best practices

Run the command below in the main repo directory or run general tests using CircleCI:
```bash
cd sites/public && yarn jest -w --verbose -t "Create Account Page"
```

## Author Checklist:

## Review Process:


sites/public/src/pages/create-account.tsx
```diff
@@ -116,9 +116,9 @@ export default () => {
                   divider={"inset"}
                   className={accountCardStyles["account-card-settings-section"]}
                 >
-                  <label className={styles["create-account-header"]} htmlFor="firstName">
+                  <legend className={styles["create-account-header"]}>
```
ludtkemorgan, line 119: question: switching to "legend" breaks some of the screen reader flow. It no longer reads "your name". @emilyjablonski this might be a question for you.

I do notice that it currently only reads it for the "first name" field, so maybe it's incorrect already

matzduniuk: @ludtkemorgan 
Just as a note here, I made this change to match the legend in the `<DOBField>` component. The `<legend>` element there does not seem to have an explicitly set ARIA values.

Also, I ran the macOS voice over and it seemed to work fine, can You give me an example where this breaks so I can also test those in the future.

emilyjablonski: @ludtkemorgan The screen reader experiences are the same to me in between core and this PR preview - do you mind sharing a video or more details of what you're experiencing?

This looks good to me! Legend is great for a set of related fields. If we're going to use legend though, let's just also wrap the three fields with `<fieldset>` like in Example 3 in this [accessibility documentation](https://www.w3.org/WAI/tutorials/forms/grouping/).

ludtkemorgan: @matzduniuk I touched base with Emily and changing this to legend is the correct path to go here. I'm good approving this PR assuming you can tackle the `<fieldset>` request @emilyjablonski suggested above.

ludtkemorgan (approved):
This test looks great! Just the one suggested accessibility change and then this can be merged

emilyjablonski (approved):

Issue #4450, opened 2024-10-29
Public - listing favoriting


ludtkemorgan:

emilyjablonski:
@eajensenwa We will likely need at least wireframes to confirm what should happen with favorites. When favorites is on in core, where should the icon/text show on the detail/directory page? For example for the directory page, where might it fit in here and what would it look like?
<img width="600" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMm1HQ0E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8f51aa5f29343a91ab8251c86fc96a10e79f4431/Screenshot%202025-01-21%20at%207.07.25%E2%80%AFPM.png" alt="Screenshot 2025-01-21 at 7.07.25 PM.png" />
I also feel like our empty state for an account favorites page looks a little off at the moment. We can re-center it, but would want to confirm the copy there as well.

eajensenwa:
@emilyjablonski here are mockups for the favorites feature, I included screens for both Core and Detroit (simplified designs)
https://www.figma.com/design/9NTevgQX4FGLccXTC9VA7y/detroit-platform-updates?node-id=2048-6470

emilyjablonski:
These look awesome!! Thank you!

ludtkemorgan:
@emilyjablonski @sarahlazarich this ticket probably shouldn't be done until [listing search page](https://github.com/bloom-housing/bloom/issues/4649) is done as that is used for both favoriting and the "favorites" page. We could split this ticket up since it is fairly large. However, the only portion that would be unblocked is the favoriting portion on the listing details page

sarahlazarich:
fine to not put these in seeds off the bat, it will be easy to fix

sarahlazarich:
would like integration test coverage and a cypress test 
ML warned that there are many potential edge cases

jaredcwhite:
@sarahlazarich I'm wondering what translations we should support for the new language (Favorites, My Favorites, etc.). I'm copying over Spanish, Bengali, and Arabic from the old Detroit repo, but in core we also have Vietnamese, Chinese, and Tagalog.

sarahlazarich:
@jaredcwhite eventually -  We can use google translate for what we're missing in core for now!

sarahlazarich:
[...]

PR #4703, opened 2025-03-13
feat: add a Favorite button and My Favorites view to the Public app

jaredcwhite:
This PR addresses #4450 

## Description

This PR adds a feature for saving listings as favorites and viewing saved favorites from the dashboard. It is scoped as a feature flag.

Provided the feature flag is set (this is true for the seed data), you should be able to view the Listings page or a single listing, and favorite a listing. Then using the nav menu or the dashboard card, you can navigate to the My Favorites page to view those favorited listings. You can also un-favorite a listing from any of those various views.

**Note:** see below for thoughts on unit/e2e testing, and also some engineering notes.

## Author Checklist:

## Review Process:


jaredcwhite:
I stubbed out a new Cypress test for the My Favorites page, but otherwise this PR doesn't include frontend tests specifically for this feature. I'm not sure the best way to do this (unit tests vs. Cypress tests), and given time constraints I'm wondering if we should do manual QA on this for now and file a ticket to add FE tests separately.

jaredcwhite:
Regarding how this was built, I had to jump through some hoops to manage data loading/state for the favorited listings, which resulted in some React code I feel is a bit gnarly but I can't think of a more elegant way to do it. Part of the problem is we have to load the user profile again (we can't use what's cached in the browser) which is a client-side call to the backend, and then based on those favorite IDs we can either match up with listings we've already loaded or load just those listings (on the My Favorites page). It results in a small visual delay between when pages initially render and when we have the data to update state for favorited listings, which I don't think is ideal.

Also as I mentioned in the recent workshop, I had to come up with a localStorage-based mechanism for checking the feature flag in order to show My Favorites in the nav bar across all pages. We don't have a global (aka DRY) way of loading a jurisdiction across all pages and making global layout changes accordingly. My technique may be OK "for now" but I'd love to think of coming up with a better strategy long-term.

api/src/passports/jwt.strategy.ts
```diff
@@ -35,6 +35,7 @@ export class JwtStrategy extends PassportStrategy(Strategy, 'jwt') {
       include: {
         listings: true,
         userRoles: true,
+        favoriteListings: true,
```
YazeedLoonat, line 38: **suggestion:** 
If my understanding is correct this is here so that the get user profile endpoint includes favorite listing data. 

If thats the case then I think it would be better to change how that endpoint works instead of adding that here. This is generated and used with every request on the backend and if we don't need this data on endpoints we should probably alter the get profile endpoint to get user data itself and return that 

this is more of a performance thing since getting and handing around extra data on the api that we don't need, I think should be cut down wherever possible

sites/public/src/lib/helpers.tsx
```diff
@@ -273,5 +275,25 @@ export const downloadExternalPDF = async (fileURL: string, fileName: string) =>
...
+) => {
+  await userService.modifyFavoriteListings({
+    body: {
+      id: listingId,
+      action: favorited ? ModificationEnum.add : ModificationEnum.remove,
+    },
+  })
+}
+
+export const fetchFavoriteListingIds = async (userService: UserService) => {
```
YazeedLoonat, line 294: **suggestion:** Instead of refetching the profile like this I think you could leverage AuthContext's `loadProfile` 

this refetchs the profile for the logged in user, so you could add that in after doing the `saveListingFavorite` call instead then leverage the profile in context 

this would cut down on extra api/db calls when they may or may not be needed

sites/public/src/layouts/application.tsx
```diff
@@ -12,6 +12,13 @@ const Layout = (props) => {
   const { profile, signOut } = useContext(AuthContext)
   const { toastMessagesRef, addToast } = useContext(MessageContext)
   const router = useRouter()
+  const [showFavorites, setShowFavorites] = useState(false)
+
+  useEffect(() => {
```
YazeedLoonat, line 17: **suggestion:** If we are refetching profile on favorite/unfavorite then the profile in AuthContext would be up to date. 

That should mean we don't need to store this in localStorage and can instead rely on that AuthContext Profile

@emilyjablonski may check me on if that is more performant or not though so localStorage could be better I'm unsure for whats better on the UI

sites/public/src/pages/account/dashboard.tsx
```diff
@@ -38,7 +45,14 @@ function Dashboard(props: DashboardProps) {
         : props.router.query.alert
       setAlertMessage(alert)
     }
-  }, [props.router, profile])
+
+    window.localStorage.setItem(
```
YazeedLoonat, line 49: **suggestion:** same note as above for context vs localStorage

sites/public/src/pages/sign-in.tsx
```diff
@@ -69,7 +78,14 @@ const SignIn = () => {
       pageTitle: "Sign In",
       status: UserStatus.NotLoggedIn,
     })
-  }, [])
+
+    window.localStorage.setItem(
```
YazeedLoonat, line 82: **suggestion:** same note as above for context vs localStorage

YazeedLoonat (changes requested):
Looks good, some feedback and questions

sites/public/src/components/account/FavoritesView.module.scss
```diff
@@ -0,0 +1,7 @@
+.favorites-none-layout {
+  --page-header-layout-content-bg-color: var(--seeds-bg-color-canvas);
+}
+
+.favorites-listings {
+  margin-inline: 0 !important; /* override standard browse layout */
```
emilyjablonski, line 6: **note:** Is it possible to make this an overridable variable?

shared-helpers/src/locales/general.json
```diff
@@ -32,7 +32,10 @@
   "account.lotteryRun": "Lottery run",
   "account.myApplications": "My Applications",
   "account.myApplicationsSubtitle": "See lottery dates and listings for properties for which you've applied",
+  "account.myFavorites": "My Favorites",
```
emilyjablonski, line 35: **note:** If this is a new string let's use sentence casing (`My favorites`)

sites/public/src/components/listing/ListingViewSeeds.tsx
```diff
@@ -125,9 +150,7 @@ export const ListingViewSeeds = ({ jurisdiction, listing, preview }: ListingProp
         depositMax={listing.depositMax}
         depositMin={listing.depositMin}
         utilitiesIncluded={
-          jurisdiction?.featureFlags?.some(
-            (flag) => flag.name === "enableUtilitiesIncluded" && flag.active
-          )
+          isFeatureFlagOn(jurisdiction, FeatureFlagEnum.enableUtilitiesIncluded)
```
emilyjablonski, line 153: **note:** ❤️

sites/public/src/lib/hooks.ts
```diff
@@ -42,6 +44,42 @@ export const useFormConductor = (stepName: string) => {
...
+        if (listingIds.length > 0) {
+          listingsService
+            .list({
+              filter: [
+                {
+                  $comparison: EnumListingFilterParamsComparison.IN,
+                  ids: listingIds,
+                },
+              ],
+              limit: "all",
```
emilyjablonski, line 64: **note:** Once we have pagination which is also in review, I imagine we will paginate this

emilyjablonski (commented):
The active button state is persisting after clicking the button, so it looks like nothing is happening. In the second half here, I start clicking off in between. May be a Seeds bug.
https://github.com/user-attachments/assets/868e76e9-0f6b-4190-a213-bb526abf686f


https://github.com/user-attachments/assets/8aedd82c-7f9c-4ce7-b935-95c0d77f6ea7


https://github.com/user-attachments/assets/3310adcc-de58-424b-a5c6-bab23055f061

Screenshot 2025-03-19 at 12 19 08 PM
Screenshot 2025-03-19 at 12 24 51 PM

sites/public/src/layouts/application.tsx
emilyjablonski: If this seems like an equally easy implementation, using the context sounds great

sites/public/cypress/e2e/pages/my-favorites.spec.ts
```diff
@@ -0,0 +1,10 @@
+describe("My favorites page", function () {
+  it("renders the my favorites page", function () {
```
emilyjablonski, line 2: **note:** I would love to have this test sign in as a user, favorite something, make sure its in the favorites view, unfavorite it, and refresh and make sure it's not there

sites/public/src/layouts/application.tsx
jaredcwhite: Hmm, I'm confused by your question @YazeedLoonat, I'm not storing any favorites in localStorage. The only thing I'm using localStorage for is saving the jurisdiction feature flag for the favorites feature. And I don't think we can save that flag in AuthContext because the context may be restarted in some navigations and we'd lose the flag state.

api/src/passports/jwt.strategy.ts
jaredcwhite: @YazeedLoonat makes sense., I'll look into how we might set up a specific call just to get favorite IDs based on a user ID.

jaredcwhite:
Thanks @YazeedLoonat and @emilyjablonski for your feedback! I'm digging in and will keep you posted…already responded to a couple of notes.

jaredcwhite:
All right I think I've addressed most of the feedback now. Some outstanding items:

* Still need to fill out the Cypress e2e test, but that shouldn't hold up another review pass I think.
* The issue with the button hover/pressed state, particularly on mobile, does seem like a general Seeds button issue. I'd like to address that holistically rather than try to monkeypatch a solution in here.
* In a similar vein, with regard to error states/toasts/etc. in the case of a network failure there are a variety of bespoke solutions for this across the public app and in some cases nothing happens except a console log (and this is an architectural issue I've raised in the past), so this sounds like a broader conversation and I hesitate to come up with another bespoke solution. (I did add a try/catch where there wasn't before for the save action in order to do that console lot.)

If I missed anything else, let me know!

jaredcwhite:
Hmm, I made improvements to the Cypress e2e test, but it fails here because it can't find `.seeds-card` …and I'm wondering if that's because CI isn't opted-into the new design for listings browse which I have running locally. 🤔

ludtkemorgan:
@jaredcwhite I hope you don't mind, but I pushed a small commit to this branch that sets the circle-ci configuration to have the seeds flag turned on. It solves the favorite test, but now the apply for a listing test is failing. I think it would be an easy fix, but curious if you want me to tackle it?

Also, I can assist with the merge conflicts if you need help. I refactored the seed file pretty significantly.

emilyjablonski:
@ludtkemorgan I think we need two different Cypress suites like we had for listings approval - if we turn the seeds toggle on for Cypress suites, what we're testing is not what's going out to production.

emilyjablonski:
The icon fill looks great now! I do think before release but maybe not before merge, we need to:
* Confirm the loading state design with Jesse, I'm happy to post it in the design channel
* Have some way that handles an error and throws a toast - there are definitely places where we aren't handling this well, but I'd love to do so for all new features, even if that improvement creates inconsistency, happy to post in the design channel as well!
* Fix the Seeds button bug - it makes it look like it's not working at all on mobile (where we have the most users), even though it is

emilyjablonski:
@jaredcwhite If you get the Cypress suite working locally with the toggle on, I think you can merge this PR with the test skipped, and we can pull in this ticket and then turn it back on: https://github.com/bloom-housing/bloom/issues/4762

ludtkemorgan:
@emilyjablonski @jaredcwhite Oh duh! Yeah, I like Emily's suggestion of skipping the test for now. Feel free to revert my change to the `config.yml` file

jaredcwhite:
Ah yeah, there's a difference in the menu IDs if favorites item is there vs. not, so Cypress is still having a hiccup here which I can fix.

I'll work on a toast for favoriting not working, and maybe think of some ways it can be pattern-ized for the future.

emilyjablonski:
Sounds great! For Cypress hopefully we just have to [revert Morgan's change](https://github.com/bloom-housing/bloom/pull/4703/commits/56d41945eda61d4d9981e8f6cc4761af4c318636) vs update the menu IDs

jaredcwhite:
All right, one more round on this! I added toast handling for when favoriting/unfavoriting a listing fails, and I tried to come up a with a pattern using an exception thrown so there's separation from the API utility code and the parent environment of the React page. I still think I could DRY up some of the duplication on the page side, but again that feels like yet another pattern I wish could be more codified down the road. I also fixed up the height of the loading state just a bit, thought I know we'd like to come up with a better Seeds-ified loader.

The partners Cypress failure seems to be contagion from main…I'm pretty sure all relevant tests for this PR are passing (or skipped) now.

emilyjablonski (approved):
🥳 looks awesome!!

emilyjablonski:
We can merge over the Cypress partners test failure btw, there's a fix coming in a different PR

jaredcwhite:
It looks like there are some non-trivial merge conflicts now I need to work through, some of the layout structure & styles for listings view got changed since I last touched those files.

YazeedLoonat (changes requested):
Hey @jaredcwhite code looks good but I'm no longer seeing the favorites button. I set the env variable in the jurisdiction and it seems to have come through on the local storage but I dont see the favorites button anywhere: 
image


weirdly it does show the favorites section: 
image

but I can only navigate to there from the dashboard, it doesn't show in the dropdown at all

emilyjablonski:
@YazeedLoonat You need to have Seeds turned on with `SHOW_NEW_SEEDS_DESIGNS=TRUE`

YazeedLoonat (changes requested):
okay! After issues on my side got resolved almost everything looks great @jaredcwhite 
only issue I found was 
image


getting dupe favorite buttons on the detail page

emilyjablonski:
^ this is my fault from the rebase btw!

emilyjablonski:
so sorry didnt meant to close!!

YazeedLoonat (approved):
LGTM

PR #4723, opened 2025-03-19
chore: upgrade axe-core/react

emilyjablonski:
## Description

Related to: https://github.com/metrotranscom/doorway/pull/1129

I'm removing this dependency based on the guidance [here](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/react/README.md#react-support) that it will no longer be supported in React versions and they recommend using their other newer tools instead. We just had this in case we wanted to uncomment the piece in `_app` to debug locally, which I don't think we were using.

## Review Process:


mcgarrye (approved):
yerrrr

Issue #4740, opened 2025-03-21
Update core jurisdictions to better represent reality


ludtkemorgan:
We would like to have the core environment better represent our various jurisdictions by doing the following:

Have 3 jurisdictions in both our core deployed environment and local seeded data.
* Main jurisdiction (Bloomington) be all of our features enabled
  * Few exceptions include: units instead of unit groups, still show close listing button
* Second jurisdiction be as close to Detroit configuration as possible
* Third jurisdiction be as close to HBA/Doorway as possible

This means entails the following steps:
* Update core deployed environment to have the above three jurisdictions with appropriate feature flags turned on and values set on jurisdiction table
* Create two new core frontend deploys
* Update seeded data to have the three jurisdictions with appropriate feature flags
  * Also add listing data to the other two jurisdictions
* Create one place in the code that lists all of the feature flags with descriptions. Have these all be created when seeded.

Along with this document our feature flags and map them to what jurisdictions are using them.

ludtkemorgan:
@sarahlazarich the documentation of feature flags has been recorded here: https://www.notion.so/exygy/Feature-Flags-1c33433f03d080fd8d04ecc8aa163d5a. 

Note: this should be a living document and there are some gaps in the documentation while we build things out

PR #4741, opened 2025-03-21
fix: add better seeded jurisdictions

ludtkemorgan:
This PR addresses #4740 

## Description
Note: this work is getting paired with updating the core environment to add these jurisdictions as well as adding documentation for what each feature flag does/where it is configured

This PR tackles the following two things:
* Make adding feature flags to instances easier
* Have locally seeded data better represent the jurisdictions we support

### New Jurisdictions
The following jurisdictions are created when seeding the stage data. You can see what feature flags are in each by looking at `seed-staging.ts`
#### Bloomington
This has as many feature flags enabled as possible. An example of flags that aren't enabled is "enableUnitGroup" and "hideCloseButton". Both of those would conflict with other features.
#### Lakeview
This is similar to our Detroit configuration
#### Bridge Bay
This is similar to HBA or Doorway configurations
#### Nada Hill
This has no feature flags set

### Feature flag creation and maintenance
We currently don't have one place in the code that holds all of the feature flags including the name and description. This PR adds the full list of feature flags to `feature-flags-enum.ts` via a list with the feature flags name and description.

Also a new endpoint was created that adds all feature flags to the DB. This should be run after every deployment where a new flag was added to the system. Note: it will only add feature flags that aren't in the DB yet and won't associate any jurisdictions

This should be tested by running the stage seeded data locally and verify the following:

1. Three jurisdictions are created
2. The appropriate feature flags are tied to each jurisdiction
3. Partner users tied to each seeded listing exist

The new endpoint should also be tested via the following:
1. Seed data
2. Add a feature flag to the `featureFlagMap` 
3. Hit the `/featureFlags/addAllNew` endpoint and verify that the new feature flag was added to the DB

## Author Checklist:

## Review Process:


emilyjablonski:
What do you think about adding a very basic set of themed overrides to one of the jurisdictions, probably Lakeview? It has been a little tricky to test in advance what Detroit's themed overrides are going to look like in core. It could be really nice to be able to set the local core jurisdiction to Lakeview and see it theme. We would probably have to dynamically add a single class if the jurisdiction is Lakeview which is maybe a little funky, and that class would have these overrides: https://github.com/bloom-housing/bloom-detroit/pull/19/files Curious for your thoughts!

api/prisma/seed-staging.ts
```diff
@@ -1282,44 +1279,174 @@ export const stagingSeed = async (
...
+        depositMax: '0',
+        depositHelperText:
+          "or one month's rent may be higher for lower credit scores",
+        disableUnitsAccordion: false,
+        leasingAgentEmail: 'bloom@exygy.com',
+        leasingAgentName: 'Bloom Bloomington',
+        leasingAgentOfficeHours: null,
+        leasingAgentPhone: '(555) 555-5555',
+        leasingAgentTitle: null,
+        name: 'Lakeview villa',
```
emilyjablonski, line 1326: **nitpick:** Could this be `Lakeview Villa` just to match our other names?

api/prisma/seed-staging.ts
```diff
@@ -37,115 +38,80 @@ import {
...
+    }),
   });
-  await prismaClient.featureFlags.create({
-    data: featureFlagFactory(
-      FeatureFlagEnum.enableMarketingStatus,
-      false,
-      "When true, the 'marketing status' sub-section is displayed in listing creation/edit and the public listing view",
-      [jurisdiction.id],
-    ),
+  // "Core" features enabled jurisdiction
```
emilyjablonski, line 95: **nitpick:** I don't think I would consider this the "Core" set if Bloomington is going to have them al

api/prisma/seed-staging.ts
```diff
@@ -37,115 +38,80 @@ import {
...
-  await prismaClient.featureFlags.create({
-    data: featureFlagFactory(
-      FeatureFlagEnum.enableUnitGroups,
-      false,
-      'When true, uses unit groups instead of units',
-      [jurisdiction.id],
-    ),
+  // unit group jurisdiction
+  const lakeviewJurisdiction = await prismaClient.jurisdictions.create({
+    data: jurisdictionFactory('Lakeview', {
```
emilyjablonski, line 74: **praise:** I love calling them Lakeview/Detroit and Bridge Bay/HBA :)

emilyjablonski:
Only issue is that if I set my local jurisdiction to Lakeview, I get an error when I try to open that listing on the public site.

api/prisma/seed-staging.ts
ludtkemorgan: fixed!

ludtkemorgan:
@emilyjablonski thanks for the review. It's ready to re-review
* Looks like the frontend issue with Lakeview was related to the StackedTable changes which I believe will be fixed with Chris's changes. However, I did add an additional null check so the page will render in the meantime! 
* The theming suggestion is a great idea but I think it should be done separately. I want the core environment to avoid using language specific to our current jurisdictions. But it does make sense to have the ability to show different color themes as that will also be useful to demo Bloom's capabilities. I'll create another ticket for that.

api/prisma/seed-staging.ts
emilyjablonski: Sounds great!

emilyjablonski (approved):

PR #4785, opened 2025-04-07
fix: seeds a11y misc improvements

emilyjablonski:
This includes three ad-hoc seeds a11y fixes.

## Description

1. `aria-controls` cannot accept spaces, so I removed spaces from these IDs
2. `<ol>` must contain `<li>` as direct children, so I moved dividers between our list elements into the `<li>` elements themselves
3. The subtext on our listing tables wasn't meeting AAA color contrast

You can ensure dividers between eligibility sections still appear, and check out the now subtly darker subtext (`per month`) on the browse page.

## Author Checklist:

## Review Process:


sites/public/src/patterns/CollapsibleSection.tsx
```diff
@@ -39,7 +41,7 @@ export const CollapsibleSection = (props: CollapsibleSectionProps) => {
         onClick={() => setCollapsed(!collapsed)}
         aria-label={!collapsed ? t("t.collapseSection") : t("t.expandSection")}
         aria-expanded={!collapsed}
-        aria-controls={props.title}
+        aria-controls={controlsId}
```
emilyjablonski, line 44: **note:** The `aria-controls` value cannot include spaces - this was affecting our `Additional Information` section

sites/public/src/components/listing/listing_sections/Eligibility.tsx
```diff
@@ -32,17 +32,16 @@ export const Eligibility = ({ eligibilitySections, section8Acceptance }: Eligibi
...
-                order={index + 1}
-                title={section.header}
-                subtitle={section.subheader}
-                note={section.note}
-              >
-                {section.content}
-              </OrderedSection>
-              {index < eligibilitySections.length - 1 && <hr />}
-            </div>
+            <OrderedSection
```
emilyjablonski, line 35: **note:** The `<li>` element should be the first direct child. In order to remove the div, I had to move the divider into the OrderedSection component itself wrapped within the list element.

ColinBuyck (approved):
LGTM 🪭 Did want to flag this discrepancy in the use of  stacked table subtext between the units section and hmi to see if you had any thoughts. Specifically around the "per month"
Screenshot 2025-04-08 at 10 43 30 AM

emilyjablonski:
@ColinBuyck I feel ya on this one ty for the callout! My ideal would be just relying on the `/ month` in the header and remove the text altogether because it's so repetitive, but when we have multiple ami levels the `/ month` is dropped, like in [this listing](https://bloom.exygy.dev/listing/ca935750-769d-4062-a0b7-d52005c3aa74/testing_design_emily_64_grinnell_dr_west_glacier_mt). In the meantime we could drop the per month to the subtext, but I think if we're going to make an effort to adjust the table data, would love to find a way to drop it all together. I'll connect w Jesse and post updates in the design system channel 🫡

PR #4796, opened 2025-04-09
fix: allow favorites to work in deployed environments

emilyjablonski:
## Description

Currently, in deployed environments when the favorites toggle is on (like [here](https://bloom-detroit.netlify.app/)), when you log in you can see the favorites button on the browse and detail pages, but you can't see the link in the header or the card on the dashboard. This is likely because we set a flag in local storage that should turn this on, which is in a use effect that runs when the jurisdiction changes, but the jurisdiction is set in static props - so it doesn't update on a built version of the application and we never set the flag to true.

It's more difficult to test this in core - but there is a PR up in Detroit with exactly the same changeset, and you can test [on the deploy preview there](https://deploy-preview-30--bloom-detroit.netlify.app/) because db/app has the correct configuration! When you log in, you should see the favorites card on the dashboard and the favorites link under my account.

## Author Checklist:

## Review Process:


ColinBuyck (approved):
LGTM 📁 but do you mind explaining why it is called in both useEffects again... not fully following the description

emilyjablonski:
@ColinBuyck Yea! I ✨ think ✨  what's happening is because [getStaticProps](https://nextjs.org/docs/pages/building-your-application/data-fetching/get-static-props) runs at build time / essentially caches, that `props.jurisdiction` value never actually changes and the use effect never runs. So one handles a local scenario, and the other a deployed scenario. However, I think how it was before is  _supposed_ to work in a deployed environment - but still looking at docs tbh to see why!

Issue #4761, opened 2025-03-27
BE - Update listing filters to use unit groups


mcgarrye:
**What is this feature or what feature is this part of?**
Now that unit groups have been implemented, the listing filters for `bathrooms`, `bedrooms` and `monthly rent` need to be updated to work with unit groups.

**Describe the solution you'd like**
Follow the code used for existing filters. Base new filters on unit group fields. Add testing. 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

**What is the acceptance criteria/definition of done?**
Filtering based on `bathrooms`, `bedrooms` and `monthly rent` works with unit groups.

**QA Review Instructions**
Can only be tested through the API until the UI is complete for unit groups and filtering.

sarahlazarich:
we do need to accommodate units and unit groups, shouldn't need a feature flag

PR #4799, opened 2025-04-10
feat: add filters for unitGroups

KrissDrawing:
This PR addresses #4761 

## Description

Adds missing filters for unit groups related fields. It doesn't include number of bathroom, as it's not present in detroit (and it's not really applicable).
For monthly rent it should work fine for single filters, but if it got double filters like greater than 2400 and less than 2600 it will return all listings. Not sure if we got any ways of handling that so i left it for now.
Also `monthlyRent` for `units` seem to be comparing strings not numbers so for example `9` is bigger than `1000`, again not sure if i'm missing something so left it.

seed data with unitGroups (it takes bedrooms from unitTypes, and monthlyRent from unitGroupAmiLevels.flatRentValue, also it will return that listing when percentageOfIncomeValue is not null)
I used request body in swagger UI like this (not all at once usually)

```
{
  "page": 1,
  "limit": 10,
  "filter": [
    {
      "$comparison": ">=",
      "monthlyRent": "2498"
    },
   {
      "$comparison": "<=",
      "monthlyRent": "2499"
    },
   {
      "$comparison": "=",
      "bedrooms": "5"
    }
  ],
  "view": "fundamentals"
}
```

## Author Checklist:

## Review Process:


mcgarrye (commented):
Looking good so far. Can you add in integration tests here:
https://github.com/bloom-housing/bloom/blob/main/api/test/integration/listing.e2e-spec.ts#L918
https://github.com/bloom-housing/bloom/blob/main/api/test/integration/listing.e2e-spec.ts#L1201
And remove this todo:
https://github.com/bloom-housing/bloom/blob/main/api/test/integration/listing.e2e-spec.ts#L888

mcgarrye (commented):
I am seeing strange responses when filtering bedrooms `=` to `0`. Everything seems to be included

KrissDrawing:
@mcgarrye it was parsing `0` to false in `if` condition. It was also the case for bathrooms.

api/src/services/listing.service.ts
```diff
@@ -670,13 +680,29 @@ export class ListingService implements OnModuleInit {
...
-            })),
+              })),
+              ...builtFilter.map((filt) => ({
+                unitGroups: {
+                  some: {
+                    unitGroupAmiLevels: {
+                      some: {
+                        OR: [
+                          { flatRentValue: filt },
+                          { percentageOfIncomeValue: { not: null } },
```
mcgarrye, line 698: **question:** should we be returning listings where there isn't flatRentValue when searching on monthly rent? We don't return listings with `units` that only have `monthlyRentAsPercentOfIncome` set.

api/src/services/listing.service.ts
```diff
@@ -670,13 +680,29 @@ export class ListingService implements OnModuleInit {
...
-              },
-            })),
+              })),
+              ...builtFilter.map((filt) => ({
+                unitGroups: {
+                  some: {
+                    unitGroupAmiLevels: {
+                      some: {
+                        OR: [
+                          { flatRentValue: filt },
```
mcgarrye, line 697: **issue:** seeing issues when using `>=` and `<=`. returning incorrect listings

KrissDrawing: Ok it should now work for `unitGroups` range. I believe that it won't work for `units` as long as they are stored as `string`, but not sure if there is workaround for that 🤔 couldn't find anything that will solve it.

api/src/services/listing.service.ts
mcgarrye: Alright. I'm going to raise this question to product but not a blocker for this PR

Issue #4731, opened 2025-03-19
Listing approval flow:


sarahlazarich:

ludtkemorgan:
QA Notes:
* This can be tested on the [core environment](https://partners.bloom.exygy.dev/)

To see the scenario with listing approval turned off
* Create a Lakeview partner user that has access to at least one Lakeview listing
* Login as this new user
* Go to edit the listing when it is in draft and you should see "Publish" instead of "Submit"

To see the scenario with listing approval turned on but not for partners
* Create a Bloomington partner user that has access to at least one Bloomington listing
* Login as this new user
* Go to edit the listing when it's in draft and you should see "Submit" instead of "Publish"

sarahlazarich:
looks great! As does the doc

PR #4803, opened 2025-04-11
fix: allow listing approval to be turned off for partners

ludtkemorgan:
This PR addresses #4731 

## Description

There was a bug where if you turn off listing approval roles (have no roles set on the jurisdiction) Partners are still seeing the "Submit" button instead of the "Publish" button

This fixes that and adds tests to cover that case.

A couple of notes of the test changes:
- Switch from `ToBeTruthy` to `ToBeInTheDocument`. This is essentially the same, but [ToBeInTheDocument](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tobeinthedocument) more accurately tests the dom
- Switch to use `screen`. Same result, but [more modern practices](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen)
- Use `getByRole` instead of `getByText`. If at all possible we should be using `getByRole` because it takes accessibility aspects into account. Here is the [order](https://testing-library.com/docs/queries/about/#priority) of queries that should take priority

If you do the staging seeding you'll be able to test the differences between "Lakeview" and "Bloomington"

To see the scenario with listing approval turned off
* Log in as a lakeview partner user (partner-user-lakeview-villa@example.com)
* unpublish the listing you have access to
* Go to edit the listing and you should see "Publish" instead of "Submit"

To see the scenario with listing approval turned on but not for partners
* Log in as a bloomington partner user (partner-user-little-village-apartments@example.com)
* Go to edit the listing and you should see "Submit" instead of "Publish"

## Author Checklist:

## Review Process:


api/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
-    "target": "es2017",
+    "target": "es2020",
```
emilyjablonski, line 9: **question:** Curious about this change?

ludtkemorgan: This is required to use the `replaceAll` function.

Ideally we would have the same target for all of our packages, so I set it to the same one that is in partners and public. Hopefully we can update them all to a more modern target

mcgarrye (approved):
LGTM

Issue #4751, opened 2025-03-25
Partners - Copy listing feature validation


sarahlazarich:

ludtkemorgan:
Findings:
* Listings with unit groups were copied correctly including the unit groups
* If you unselect the "Include Units" checkbox the unit groups are still copied over
  * This is just a one line change

Due to the easy fix, I will be creating a PR directly against this ticket instead of spinning off a new ticket.

ludtkemorgan:
QA notes:
* Both the copying of a unit group listing with the "Include Units" checkbox selected and unselected should work as expected.
* This can be tested on [core](https://partners.bloom.exygy.dev/) when copying a "Lakeview" listing

sarahlazarich:
lgtm!

PR #4823, opened 2025-04-17
fix: allow duplication to not copy unit groups

ludtkemorgan:
This PR addresses #4751 

## Description

The duplication (copy) of listings already works for listings that use unit groups instead of units. However, if you don't want to copy the units a small change was needed.

* Note: open question. Do we require a language change from "include units" to "include unit groups" when it is a unit group jurisdiction

To Copy with units groups
* Seed the staging data
* open up one of the "lakeview" listings in the partner site
* Click "copy" for the listing
* Enter a new listing name and submit
* Verify the listing along with the unit groups were transferred.

To copy without unit groups
* Seed the staging data
* open up one of the "lakeview" listings in the partner site
* Click "copy" for the listing
* uncheck the "include units" checkbox
* Click submit
* Verify the listing was copied, but the unit groups were not

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/listing.service.ts
+++ b/api/src/services/listing.service.ts
@@ -1317,9 +1317,11 @@ export class ListingService implements OnModuleInit {
                       level.monthlyRentDeterminationType,
                     percentageOfIncomeValue: level.percentageOfIncomeValue,
                     flatRentValue: level.flatRentValue,
-                    amiChart: {
-                      connect: { id: level.amiChart.id },
-                    },
+                    amiChart: level.amiChart?.id
+                      ? {
+                          connect: { id: level.amiChart.id },
+                        }
+                      : undefined,
                   })),
                 },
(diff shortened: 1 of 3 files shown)


api/src/services/listing.service.ts
```diff
@@ -1428,6 +1430,7 @@ export class ListingService implements OnModuleInit {
 
     if (!dto.includeUnits) {
       delete mappedListing['units'];
+      delete mappedListing['unitGroups'];
```
emilyjablonski, line 1433: **question:** Curious why this was removed?

mcgarrye (approved):
LGTM!

api/src/services/listing.service.ts
mcgarrye: This is when you unselect including the units when copying

Issue #4698, opened 2025-03-11
Containerization of Bloom


ludtkemorgan:
To build and deploy Bloom the code does not have to be containerized. The main reason for this is that it is not a requirement for both local development as well as deploying to Netlify and Heroku.

Not having the system containerized has led to actual issues before, during, and after deployments. Here are a few examples:
* Building the backend on Heroku installs slightly different package versions and a package was accidentally upgraded
* Netlify specific packages stop working and only notice when trying to deploy there
* Issue happening in a deployed environment but not locally

TODO: add more details

PR #4827, opened 2025-04-22
fix: local docker updates and type fixes

seanmalbert:
Address #4698 specifically #4998

## Description

This PR adds the ability to run Bloom Housing via docker with docker compose. While setting this up, I updated the prisma version, which allowed me to encounter two bugs that were earlier surfaced by a JTC fellow: a type errors in api/src/services/auth.service.ts and api/src/utilities/model-fields.ts. 

Instructions for using this docker locally can be found in `docker.md`

This was only tested on my computer which is an M2 mac. Follow the instructions in `docker.md` to run both just the postgres instance and all applications.

For people that run into the same mac issues I had take note of the env variable changes required:
- `DATABASE_URL=postgresql://postgres:prisma@docker.for.mac.host.internal:5433/bloom_prisma` in the api .env file
- `BACKEND_API_BASE=http://docker.for.mac.host.internal:3100` in the public and partners .env files

## Author Checklist:

## Review Process:


emaliewebb (approved):

api/test/unit/utilities/model-fields.spec.ts
```diff
@@ -1,29 +1,74 @@
+import { randomUUID } from 'node:crypto';
```
mcgarrye, line 1: **nitpick:** everywhere else we import, it's just `from 'crypto'`

mcgarrye:
Does this docker setup all for the docker files to pickup live development changes?

ludtkemorgan:
@YazeedLoonat 
1. When building it for the first time is does take a long time. On my machine it is not 30 minutes, but probably like 10 minutes. Subsequent builds don't need to pull down all of the images and do all of the steps so it is relatively fast. When it is building my computer is mostly unusable as it uses the majority of the machine's CPU.
2. I am able to hit the api at `http://localhost:3100/api`. I have to update the db url to be `DATABASE_URL=postgresql://postgres:prisma@docker.for.mac.host.internal:5433/bloom_prisma` (note the 5433). And the public .env should have BACKEND_API_BASE=http://docker.for.mac.host.internal:3100. Let me know if that doesn't work for you

@mcgarrye Local changes should get picked up and not require a re-build. Let me know if you aren't seeing that

docker.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Full bloom application
+
+The entire Bloom application can also be run via a docker compose script. This starts up a database, the api, the partner site, and the public site all within docker. The application is still in dev mode so any changes will be picked up without needing to restart. The backend also re-seeds the database on startup.
+
+This setup still uses the `.env` files from the respective packages. Any changes to those files will need a restart of the application to be picked up. Take note of the required changes for the database url in the above section.
+
+#### Command to start:
+
+`docker compose -f docker-compose-full.yml up --build -d`
```
mcgarrye, line 30: **suggestion:** I recommend we update this to the following to remove unused containers and images after build
```suggestion
`docker compose -f docker-compose-full.yml up --build -d --remove-orphans && docker image prune -a -f`
```

docker.md
```diff
@@ -0,0 +1,35 @@
...
+
```
mcgarrye, line 31: **suggestion:** Can we add a section here to recommend using `docker compose stop` and `docker compose start` when needing to stop containers instead of using up and down every time when there are no changes made that require rebuilding

ludtkemorgan:
Closing in favor of #5170

Issue #4772, opened 2025-04-01
Public - FE - Listing Verification


sarahlazarich:
We need to add a tag for when a listing is confirmed by the property in Detroit. This is tied to the `isVerified` field.

Listing Directory Page: 
<img width="1096" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMkdBQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--46990f1cc4bbfb1c22e51d9514194424d4227b04/image.png" alt="image.png" />

Listing Detail Page: 
<img width="760" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMldBQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--da122c844d2896a9bb921fd244c17d9b19fe878d/image.png" alt="image.png" />

Notes: 
- We are moving this tag off of the image card - it can sit with the other listing tags
- It should come first
- It should show on both the search and detail pages for listings
- Here is a link to the [design](https://www.figma.com/design/9NTevgQX4FGLccXTC9VA7y/detroit-platform-updates?node-id=2569-16510&t=YoTf6ElBDotem0vO-1)! It should be the warn color

sarahlazarich:
@slowbot - do we want to keep this tag where it is or follow a different pattern?

sarahlazarich:
Looks great!

PR #4832, opened 2025-04-23
feat: public fe listing verification tag

KrissDrawing:
This PR addresses #4772 

## Description

Adds tag that will be displayed when `isVerified` is set to true.

Set `isVerified` for listing. Then check if its visible for `/listings` and `/listings/[:id]` pages.

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/src/locales/general.json
+++ b/shared-helpers/src/locales/general.json
@@ -772,6 +772,7 @@
   "listings.vacantUnit": "Vacant Unit",
   "listings.vacantUnits": "Vacant Units",
   "listings.vacantUnitsAvailable": "Vacant Units Available",
+  "listings.verifiedListing": "Confirmed by Property",
   "listings.waitlist.closed": "Waitlist Closed",
   "listings.waitlist.currentSize": "Current Waitlist Size",
   "listings.waitlist.finalSize": "Final Waitlist Size",
--- a/sites/public/src/components/listing/listing_sections/MainDetails.module.scss
+++ b/sites/public/src/components/listing/listing_sections/MainDetails.module.scss
@@ -28,10 +28,13 @@
 
 .listing-tags {
   margin-bottom: calc(-1 * var(--seeds-s1));
-  span {
-    margin-bottom: var(--seeds-s1);
+  .tag {
+    margin-block-end: var(--seeds-s1);
     &:not(:last-child) {
       margin-inline-end: var(--seeds-s2);
+    }
+    svg {
+      margin-inline-end: var(--seeds-s1);
     }
   }
 }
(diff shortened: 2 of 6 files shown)


sites/public/src/components/listing/listing_sections/MainDetails.module.scss
```diff
@@ -28,10 +28,13 @@
 
 .listing-tags {
   margin-bottom: calc(-1 * var(--seeds-s1));
-  span {
-    margin-bottom: var(--seeds-s1);
+  .tag {
+    margin-block-end: var(--seeds-s1);
```
emilyjablonski, line 32: **praise:** ❤️

shared-helpers/src/locales/general.json
```diff
@@ -772,6 +772,7 @@
   "listings.vacantUnit": "Vacant Unit",
   "listings.vacantUnits": "Vacant Units",
   "listings.vacantUnitsAvailable": "Vacant Units Available",
+  "listings.verifiedListing": "Confirmed by Property",
```
emilyjablonski, line 775: **issue:** Let's keep property lowercase, so `Confirmed by property`

sites/public/src/components/listing/listing_sections/MainDetails.tsx
```diff
@@ -17,6 +17,7 @@ import FavoriteButton from "../../shared/FavoriteButton"
 import { Availability } from "./Availability"
 import listingStyles from "../ListingViewSeeds.module.scss"
 import styles from "./MainDetails.module.scss"
+import { CheckIcon } from "@heroicons/react/20/solid"
```
emilyjablonski, line 20: **note:** If you use the 16 import `@heroicons/react/16/solid` it'll be smaller and warp less

sites/public/src/components/listing/listing_sections/MainDetails.tsx
```diff
@@ -39,6 +41,14 @@ export const getListingTags = (
 ): ListingTag[] => {
   const listingTags: ListingTag[] = []
 
+  if (listing.isVerified) {
```
emilyjablonski, line 44: **issue:** Let's check the feature flag here, just in case a jurisdiction had this feature flag on and then turned it off

like:

```
  const enableIsVerified = doJurisdictionsHaveFeatureFlagOn(
    "enableIsVerified",
    listing.jurisdictions.id
  )
```

emilyjablonski (changes requested):
A few small requests! Otherwise looks good 👍

sites/public/src/components/listing/listing_sections/MainDetails.tsx
KrissDrawing: hmm, for some reason `doJurisdictionsHaveFeatureFlagOn` doesn't work here. It seem like it works only when we're logged in (so probably for partners site). But i used different approach and it should work.

emilyjablonski (approved):

Issue #4817, opened 2025-04-16
Partners Listings Table View - Missing/Repeat Listings


mcgarrye:
- [ ] General
- [x] Backend
- [ ] Public
- [x] Partners

## Background / Issue
Listings table view on the Partners site is showing repeat listings across pages which causes other listings never to show when sorting by Status.
## Technical Context
The Partner site is sending back a request to OrderBy Status and OrderDir ASC, with Limit 8. The `list` function in ListingService is calling the db, but the returned listings are not sorted/paginated? correctly. Further investigation needed.
## Acceptance Criteria
Listing table view does not have any repeats and shows all listings when sorting by Status.
## QA Notes

** Please provide as much information as possible (screenshots, documentation, etc) **

mcgarrye:
QA: Play around with sorting on the partners listing view. Have more than 8 listings with a single status to make sure issue is resolved.

PR #4835, opened 2025-04-23
fix: listing status sort bug

mcgarrye:
This PR addresses [#(4817)](https://github.com/bloom-housing/bloom/issues/4817)

## Description

When sorting and paginating on listings with OrderBy set to `status`, Prisma may return repeats on different pages since `status` is not unique and can include a large number of listings, causing some listings not to appear at all.

The solution is to include a secondary sort by name. This way it will sort by status and then subsort by name.

Play around with sorting on the partners listing view. Have more than 8 listings with a single status to make sure issue is resolved.

## Author Checklist:

## Review Process:


api/test/unit/services/listing.service.spec.ts
```diff
@@ -837,6 +840,9 @@ describe('Testing listing service', () => {
           {
             name: 'asc',
           },
+          {
```
emilyjablonski, line 843: **question:** J curious why this is duplicated?

emilyjablonski:
Do you have a repeatable set of repro steps? I can't repro this on main, and want to make sure I can compare the same steps! I have 11 open listings locally.

api/test/unit/services/listing.service.spec.ts
mcgarrye: The code doesn't check if it is already attempting to sort by name so it appends the second name sort regardless.

mcgarrye:
> Do you have a repeatable set of repro steps? I can't repro this on main, and want to make sure I can compare the same steps! I have 11 open listings locally.

It's tricky to repo. The more listings with the same status the more likely it is to appear. Sometimes bring the app down and restarting it, and hard reloading the page will give a different result but no promises

emilyjablonski:
@mcgarrye I don't know that we'll be able to confidently know we've fixed it unless we get repeatable steps - otherwise it's just a best guess, which is fine if that's all we have! Is it possible there's a specific number of listings / default sorting that generates the issue?

mcgarrye:
> https://deploy-preview-4835--partners-bloom-dev.netlify.app

The issue is currently present on Core for open listings. One could download a copy of the core db locally and see if it resolves the issue

emilyjablonski:
Sounds good! @mcgarrye Can I confirm if that's a step you took in testing? It was looking good to me locally, but just want to confirm that's been done at least once :) Then lgtm!

mcgarrye:
> Sounds good! @mcgarrye Can I confirm if that's a step you took in testing? It was looking good to me locally, but just want to confirm that's been done at least once :) Then lgtm!

I sure did!

emilyjablonski (approved):
Fabulous ✨

PR #4844, opened 2025-04-30
fix: add coverage thresholds

ludtkemorgan:

## Description

## Author Checklist:

## Review Process:


sites/partners/jest.config.js
```diff
@@ -21,10 +21,10 @@ module.exports = {
   coverageDirectory: "test-coverage",
   coverageThreshold: {
     global: {
-      branches: 0,
-      functions: 0,
-      lines: 0,
-      statements: 0,
+      branches: 41,
```
emilyjablonski, line 24: **question:** Just curious how these numbers were chosen?

ludtkemorgan: I ran the coverage reports and chose the closest whole number. These numbers need to be close to what we currently have so that if new tests aren't added it will trigger these threshold numbers

ludtkemorgan: https://docs.google.com/spreadsheets/d/1d55fjCK5g-9DdfPcxYQz3rlymyV3y90Gff6T7zCXVEA/edit?gid=1873339964#gid=1873339964

YazeedLoonat: I believe you would need to get enough coverage on new things to keep the coverage thresholds at least at where they are 

I think we can probably give ourselves some wiggle room though, the api always had wiggle room (even though that allowed our coverage % to gradually drop off 😭 )

ludtkemorgan: I can lower the thresholds for the ones that are super close. However, more tests have been written since I created these thresholds so there probably is more buffer now

YazeedLoonat (approved):
@ludtkemorgan this looks good, and I'm gonna approve but I do have a comment in one of the comment threads asking for a lil bit of wiggle room to be built into these coverage thresholds

PR #4852, opened 2025-05-05
fix: reduce fetch on individual application

emilyjablonski:
This PR addresses #4853

## Description

The individual application view page was calling the application and listing each twice - and it was fetching the `full` view for a listing when it only needed the `base`.

Can use the [deploy preview](https://deploy-preview-4852--bloom-exygy-dev.netlify.app/) and login as `admin@example.com` which will already have apps.

Otherwise, apply for an application while logged in. While viewing the application summary, check out the network tab and ensure there is only one call for the application + listing. I also pulled in the new Seeds loading state. 

## Author Checklist:

## Review Process:


ludtkemorgan (approved):
The change looks good. This is probably too much for this specific PR, but we should prioritize the integration test around this page so that this case could be tested. https://github.com/bloom-housing/bloom/issues/4545

Issue #4794, opened 2025-04-09
Filter Pt. 3 - Support filtering by URL


ColinBuyck:
AC
The url should now have a decoding step which breaks down the url and shows the existing selections in the filter drawer

The fields should be toggle-able based on feature flags

QA: This can be tested at https://bloom-public-seeds.netlify.app/listings The filters should all be functional with the exception of the monthly rent numeric inputs (that can be tested with unit groups through the lakeview jurisdiction). The listings displayed should match the entered filters and the pagination totals should change based on the results. Additionally, the filter drawer should always reflect the filtered state.

Issue #4793, opened 2025-04-09
Filter Pt. 2 - Filter listings from filter page


ColinBuyck:
Handle onSubmit and URL encodings

AC
A user should be able to select options, click submit and see the corresponding filtered results

NOT AC
The filter should not reflect existing selections of the filtered view

QA: See Part 3

Issue #4441, opened 2024-10-28
FE Implementation - Filter Drawer


ludtkemorgan:
Design: 
Design has recommended starting with this pattern to build the filter Drawer: https://www.figma.com/design/xqyamlgdsASkLTYlcifBwq/core-patterns?node-id=1686-9745&m=dev 

It's not a 1-1 asset, but will establish general core filtering patterns. We will need to add all the questions with this pattern  for each type of question ourselves. We will not move forward yet with the `Any` checkbox pattern, as that is new and not yet fully fleshed out.

We don't currently have a version of the Seeds Drawer that is the width shown in the design, so we will start with the default Seeds drawer without any customizations and go from there! If we can rely on existing Drawer tokens to change the width without any Seeds changes, we are able to do that. We may need to check in with design on what the width token should be for this smaller width.

Some remaining questions for design:

Notes:
- Inputs for current filter in detroit [here](https://www.notion.so/exygy/Filter-Drawer-Inputs-1a43433f03d080f3a983e60c0e1ea7d6)
- Core vs Detroit filter inputs [here](https://www.notion.so/exygy/Core-vs-Detroit-Filter-1d13433f03d080959201fd684032fe90)
- This will only be in core, leave doorway as-is 
- Designs are not 1:1, will need to extrapolate a bit - Eng that picks this up will need to coordinate with @slowbot 

AC: 
- Test the filter drawer on the replatformed Detroit public site 
- Filtering returns accurate listings

mcgarrye:
Time box technical portion to 2 days

sarahlazarich:
Start with designs - output should be an asset

emilyjablonski:
@eajensenwa I'm 50/50 on this being a wireframe or a high fidelity asset. If it's easier, I think we could start with a wireframe of what we want the filtering experience to look like in core.

emilyjablonski:
Design has recommended starting with this pattern: https://www.figma.com/design/xqyamlgdsASkLTYlcifBwq/core-patterns?node-id=1686-9745&m=dev It's not a 1-1 asset, but will establish general core filtering patterns. We will not move forward yet with the `Any` checkbox pattern as that is new and not yet fully fleshed out. Some remaining questions for design:
* Having the actionable button say "Done" isn't a pattern we have anywhere else - the filters won't take effect until the user clicks the button. Doorway says "Show matching listings" - what do you think about something more action-sounding?
[...]

PR #4868, opened 2025-05-13
feat: filtering

ColinBuyck:
This PR addresses partially addresses #4441 
Focusing on https://github.com/bloom-housing/bloom/issues/4793 and https://github.com/bloom-housing/bloom/issues/4794

## Description

This PR builds the functionality for filtering by url through encoding and decoding the input from the filter drawer. As requested, this functionality is covered by a feature flag and utilizes seeds components as much as possible. 

Ongoing Work:
- Unit tests for each of the new helper functions, expanded integration tests for filter drawer and listing browse
- The way it is currently set up special characters are only encoded after pagination but not on initial filter submit. This is a nit since its still functional but I'd like to be consistent on this

Product Questions
- There should be an indication when filters are applied to the user (maybe of the filter button). Two, there should a unique message for when a filter results in no listings and when filtering failed. Think the map work is a nice example of more directive messaging. 
- How do we want to handle SROs? Should coming soon be under construction?

This can be tested locally ! The filters should all be functional with the exception of the monthly rent numeric inputs (that can be tested with unit groups through the lakeview jurisdiction). The listings displayed should match the entered filters and the pagination totals should change based on the results. Additionally, the filter drawer should always reflect the filtered state.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/src/components/browse/FilterDrawerHelpers.tsx
+++ b/sites/public/src/components/browse/FilterDrawerHelpers.tsx
@@ -77,6 +77,10 @@ const availabilityOrdering = {
   [FilterAvailabilityEnum.comingSoon]: { ordinal: 3 },
 }
 
+/**
+ *
+ * @returns an array of availability keys in the displayed order
+ */
 export const getAvailabilityValues = () => {
   // TODO: https://github.com/metrotranscom/doorway/issues/1278
   const availabilityFiltered = Object.keys(FilterAvailabilityEnum).filter(
@@ -114,10 +118,18 @@ export const unitTypesSorted = Object.keys(UnitTypeEnum).sort(
   (a, b) => unitTypeMapping[a].ordinal - unitTypeMapping[b].ordinal
 )
 
+/**
+ * Generates field data to mantain consistency across all filter fields 
+ * 
+ * @param filterType key to connect individual options to filtering type
+ * @param labelInfo string if all translations have consistent translation key base in json
+   array for strings to be passed in manually for special cases
+ * @param keyArr array of individual filter selection keys
+ * @param existingData filter data to inform defaultChecked field
+ * @returns array of formatted field data which will be passed to each individual filter field
+ */
 export const buildDefaultFilterFields = (
   filterType: ListingFilterKeys,
-  // string if all translations have consistent translation key base in json
-  // array for strings to be passed in manually for special cases
   labelInfo: string | string[],
   keyArr: string[],
   existingData: FilterData
@@ -208,7 +220,13 @@ export const RentSection = (props: RentSectionProps) => (
   </fieldset>
 )
 
-// built to support filtering by url decoding and filtering by state directly from form data
+/**
+ * Transforms filter data to backend filter formatting
+ * Note: built to support filtering by url decoding and filtering by state directly from form data
+ *
+ * @param data object containing form selections or url params decoded
+ * @returns array of formatted backend filters
+ */
 export const encodeFilterDataToBackendFilters = (data: FilterData = {}): ListingFilterParams[] => {
   const filters: ListingFilterParams[] = []
   Object.entries(data).forEach(([filterType, userSelections]) => {
@@ -272,15 +290,33 @@ export const encodeFilterDataToBackendFilters = (data: FilterData = {}): Listing
   return filters
 }
 
+/**
+ * Checks if url query contains filtering information
+ *
+ * @param contextQuery ParsedUrlQuery from context.query in getServerSideProps call
+ * @returns boolean if any filter params are passed
+ */
 export const isFiltered = (contextQuery: ParsedUrlQuery) => {
   return Object.keys(contextQuery).some((param) => Object.keys(ListingFilterKeys).includes(param))
 }
 
+/**
+ * Removes pagination information from url to isolate filter query
+ *
+ * @param contextQuery ParsedUrlQuery from context.query in getServerSideProps call
+ * @returns string of url params related to filtering
+ */
 export const getFilterQueryFromURL = (url: ParsedUrlQuery) => {
   delete url["page"]
   return encode(url)
 }
 
+/**
+ * Transforms data from filter submission into url query
+ *
+ * @param data form data capturing user selections
+ * @returns string of user's selections in url param format
+ */
 export const encodeFilterDataToQuery = (data: FilterData): string => {
   const queryArr = []
   const cleanedFilterData = removeUnselectedFilterData(data)
@@ -304,6 +340,12 @@ export const encodeFilterDataToQuery = (data: FilterData): string => {
   return queryArr.join("&")
 }
 
+/**
+ * Transforms url query into filter data to repopulate filter form with current selections
+ *
+ * @param data form data capturing user selections
+ * @returns FilterData of selections reflected in url params
+ */
 export const decodeQueryToFilterData = (parsedQuery: ParsedUrlQuery): FilterData => {
   const filterData = {}
   Object.entries(parsedQuery).forEach(([filterType, userSelections]) => {
@@ -327,6 +369,12 @@ export const decodeQueryToFilterData = (parsedQuery: ParsedUrlQuery): FilterData
   return filterData
 }
 
+/**
+ * Isolates selected fields from full filter data
+ *
+ * @param data form data capturing all fields
+ * @returns FilterData including only the user's selections
+ */
 export const removeUnselectedFilterData = (data: FilterData): FilterData => {
   const cleanedFilterData: FilterData = {}
   Object.entries(data).forEach(([filterType, userSelections]) => {
--- a/sites/public/__tests__/components/browse/FilterDrawer.test.tsx
+++ b/sites/public/__tests__/components/browse/FilterDrawer.test.tsx
@@ -1,0 +1,3 @@
+describe("FilterDrawer", () => {
+  it.todo("should return all filter fields correctly")
+})
--- a/sites/public/__tests__/components/browse/FilterDrawerHelpers.test.tsx
+++ b/sites/public/__tests__/components/browse/FilterDrawerHelpers.test.tsx
@@ -13,8 +13,87 @@ describe("filter drawer helpers", () => {
     })
   })
   describe("buildDefaultFilterFields", () => {
-    // const existingFilterData
-    // it("should return correct fields when labelInfo passed as string", () => {})
-    // it("should return correct fields when labelInfo passed as string array", () => {})
+    it.todo("should return correct fields when labelInfo passed as string")
+    it.todo("should return correct fields when labelInfo passed as string array")
+    it.todo("should return correct fields when existindData is empty")
+    it.todo("should return correct fields when existindData is partially filled")
+    it.todo("should return correct fields when existindData is the full form object")
+  })
+
+  describe("encodeFilterToBackendFiltering", () => {
+    it.todo("should return correct BE filters for single IN comparisons")
+    it.todo("should return correct BE filters for multiple IN comparisons")
+    it.todo("should return correct BE filters for single = comparison")
+    it.todo("should return correct BE filters for multiple = comparisons")
+    it.todo("should return correct BE filters for >= comparison")
+    it.todo("should return correct BE filters for <= comparison")
+    it.todo("should return correct BE filters for full rent information")
+    it.todo("should returm correct BE filters for all comparison types combined")
+  })
+
+  describe("isFiltered", () => {
+    it.todo("should return true if filter params present in context.query")
+    it.todo("should return false if filter params not present in context.query")
+  })
+
+  describe("getFilterQueryFromURL", () => {
+    it.todo("should return filter query without page when page is present in url")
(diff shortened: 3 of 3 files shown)


sites/public/src/components/browse/ListingBrowse.tsx
```diff
@@ -103,6 +119,35 @@ export const ListingBrowse = (props: ListingBrowseProps) => {
     }
   }
 
+  const onFilterSubmit = async (data: FilterData) => {
```
ColinBuyck, line 122: **note:** Go through the unnecessary steps of encoding and decoding to demonstrate pattern that will be used once listings fetch is completed

sites/public/src/components/browse/ListingBrowse.tsx
```diff
@@ -103,6 +122,17 @@ export const ListingBrowse = (props: ListingBrowseProps) => {
     }
   }
 
+  const onFilterSubmit = (data: FilterData) => {
+    const updatedFilterQuery = encodeFilterDataToQuery(data)
+    setIsFilterDrawerOpen(false)
+    if (updatedFilterQuery != filterQuery) {
```
emilyjablonski, line 128: **nitpick:** 
```suggestion
    if (updatedFilterQuery !== filterQuery) {
```

sites/public/src/components/browse/ListingBrowse.tsx
```diff
@@ -103,6 +122,17 @@ export const ListingBrowse = (props: ListingBrowseProps) => {
...
+      setIsLoading(true)
+      router.pathname.includes("listings-closed")
+        ? void router.push(`/listings-closed?${updatedFilterQuery}`)
```
emilyjablonski, line 131: **nitpick:** 
```suggestion
        ? void router.push(`/listings${router.pathname.includes("listings-closed") ? "-closed" : "}?${updatedFilterQuery}`)
```

sites/public/src/components/listing/FilterDrawer.module.scss
```diff
@@ -0,0 +1,24 @@
+.filter-drawer {
+  --drawer-width: var(--seeds-width-xl);
+  font-family: var(--seeds-font-alt-sans);
+
+  .filter-section {
+    padding-block-end: var(--seeds-spacer-content);
+    border-bottom-width: var(--seeds-border-1);
```
emilyjablonski, line 7: **nitpick:** 
```suggestion
    border-block-end-width: var(--seeds-border-1);
```

sites/public/src/components/listing/FilterDrawer.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { isTrue } from "../../lib/helpers"
+
+export interface FilterDrawerProps {
+  filterState: FilterData
+  isOpen: boolean
+  onClose: () => void
+  onSubmit: (data: FilterData) => void
+}
+
+// remove doorway specific enum references
```
emilyjablonski, line 28: **suggestion:** Let's not have jurisdictional references in core. If this is temporary NBD.

sites/public/src/components/listing/FilterDrawer.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { useForm } from "react-hook-form"
+import { listingFeatures } from "@bloom-housing/shared-helpers"
+import {
+  FilterAvailabilityEnum,
+  RegionEnum,
+  UnitTypeEnum,
+  HomeTypeEnum,
+  ListingFilterKeys,
+} from "@bloom-housing/shared-helpers/src/types/backend-swagger"
+import styles from "./FilterDrawer.module.scss"
```
emilyjablonski, line 12: **nitpick:**  Can we order these up so the imports go from external --> internal relative to this file

sites/public/src/components/listing/FilterDrawerHelper.tsx
```diff
@@ -0,0 +1,313 @@
...
+  key: string
+  label: string
+  defaultChecked: boolean
+}
+
+export interface CheckboxGroupProps {
+  groupLabel: string
+  fields: FilterField[]
+  register: UseFormMethods["register"]
+  customRowNumber?: number
```
emilyjablonski, line 39: **question:** What is `customRowNumber`?

sites/public/src/components/listing/FilterDrawerHelper.tsx
```diff
@@ -0,0 +1,313 @@
...
+            id={`${ListingFilterKeys.monthlyRent}.maxRent`}
+            name={`${ListingFilterKeys.monthlyRent}.maxRent`}
+            label={t("listings.maxRent")}
+            type="currency"
+            prepend="$"
+            register={props.register}
+            getValues={props.getValues}
+            setValue={props.setValue}
+            defaultValue={props.filterState?.[ListingFilterKeys.monthlyRent]?.maxRent}
+          ></Field>
```
emilyjablonski, line 149: **nitpick:** 
```suggestion
          <Field
            id={`${ListingFilterKeys.monthlyRent}.maxRent`}
            name={`${ListingFilterKeys.monthlyRent}.maxRent`}
            label={t("listings.maxRent")}
            type="currency"
            prepend="$"
            register={props.register}
            getValues={props.getValues}
            setValue={props.setValue}
            defaultValue={props.filterState?.[ListingFilterKeys.monthlyRent]?.maxRent}
          />
```

sites/public/src/components/listing/FilterDrawerHelper.tsx
```diff
@@ -0,0 +1,313 @@
...
+        </Grid.Cell>
+      </Grid.Row>
+      <Grid.Row key="0">
```
emilyjablonski, line 152: **question:** Why does this have `key=0`?

sites/public/src/components/listing/FilterDrawerHelper.tsx
```diff
@@ -0,0 +1,313 @@
...
+  ListingFilterKeys.availabilities,
+]
+
+const booleanFilters: ListingFilterKeys[] = [
+  ListingFilterKeys.isVerified,
+  ListingFilterKeys.section8Acceptance,
+]
+
+// two filters below have yet to be implemented, captured in part 4
+const indvidualFilters: ListingFilterKeys[] = [
```
emilyjablonski, line 65: **nitpick:** 
```suggestion
const individualFilters: ListingFilterKeys[] = [
```

emilyjablonski (commented):
This is looking great!! A few qs:
* It looks like a11y features are or vs and, and they should be and - not an issue for this PR but just want to make sure that is captured
* It looks like we're missing listing name and community types which to my understanding is expected and not intended to be completed here, just want to make sure that is captured
* Does this include incorporating the a11y feedback? No need for it to, also just want to make sure that is captured
* The order of the availability checkboxes doesn't match current Detroit production, so would love to ensure they do!

sites/public/src/components/listing/FilterDrawerHelper.tsx
```diff
@@ -0,0 +1,313 @@
...
+  [UnitTypeEnum.fiveBdrm]: 5,
+}
+
+export const buildDefaultFilterFields = (
+  filterType: ListingFilterKeys,
+  stringBase: string,
+  keyArr: string[],
+  existingData: FilterData
+): FilterField[] =>
+  keyArr.map((key) => {
```
ludtkemorgan, line 85: issue: This might be covered in a different ticket, but noticed that these aren't getting translated. Same goes for "Home Type"

ludtkemorgan (commented):
Still making my way though, but won't have time to finish for a while so giving my initial comments now

sites/public/src/components/listing/FilterDrawerHelper.tsx
ColinBuyck: Must've slipped through from some a11y testing, removed it!

sites/public/src/components/browse/ListingBrowse.tsx
ColinBuyck: My preference is to stick to the more readable so I kept things as is if that's alright

sites/public/src/components/listing/FilterDrawer.tsx
emilyjablonski: @ludtkemorgan Question for you about this piece! What would it look like to move this to the backend / not have it be hardcoded?

I think the issue here comes down to (Colin walked me through this) that Doorway uses `waitlistOpen` and everywhere else uses `openWaitlist` so the enum has both values and I think the long term fix is to migrate Doorway to `openWaitlist` - wdyt?

ColinBuyck: Sounds like a plan! Made the issue here https://github.com/metrotranscom/doorway/issues/1278 and updated the comment

emilyjablonski:
More of a UX question - as the browser size decreases, the number of columns goes from 3 -> 2 -> 1, but the size of the container isn't actually changing. We should ideally be at one size until it takes up the full screen, and then a smaller size from there. Given the length of the accessibility features my initial thought is that 1 column on mobile and 2 columns above might feel better than 3, but that's just one perspective so curious what feels right to you as well / if this was discussed w design!

https://github.com/user-attachments/assets/9419e88f-b8cc-45c3-a357-aec41cc31345

ColinBuyck:
@emilyjablonski Interesting! I hadn't considered that piece and agree with the shift to two columns until one column on mobile but would want to get designs thoughts before changing anything. @eajensenwa What do you think? Always down to huddle if more context/questions are needed 🥅

emilyjablonski:
@ColinBuyck Would love to handle that in a follow-up! I can make a small ticket & we can coordinate w design

api/src/enums/feature-flags/feature-flags-enum.ts
```diff
@@ -12,6 +12,7 @@ export enum FeatureFlagEnum {
   enableIsVerified = 'enableIsVerified',
   enableListingFavoriting = 'enableListingFavoriting',
   enableListingOpportunity = 'enableListingOpportunity',
+  enableListingFiltering = 'enableListingFiltering',
```
mcgarrye, line 15: **nitpick:** please keep alphabetical :)

sites/public/src/components/browse/FilterDrawerHelpers.tsx
```diff
@@ -0,0 +1,351 @@
...
+        const filter = {
+          $comparison: EnumListingFilterParamsComparison["IN"],
+        }
+        filter[filterType] = selectedFields
+        filters.push(filter)
+      }
+    } else if (
+      booleanFilters.includes(ListingFilterKeys[filterType]) &&
+      // filter data direct from form is boolean, decoded from url is string
+      (userSelections == true || userSelections === "true")
```
ludtkemorgan, line 245: nit: `userSelections === true`

Or you could do `[true, "true"].includes(userSelections)` but that might be less readable

sites/public/src/components/browse/FilterDrawerHelpers.tsx
```diff
@@ -0,0 +1,351 @@
...
+    } else if (filterType === ListingFilterKeys.monthlyRent && typeof userSelections === "string") {
+      //custom separator to avoid conflicts with higher values with commas
+      const rentArr = userSelections.split("-")
+      filterData[filterType] = { minRent: rentArr[0], maxRent: rentArr[1] }
+    }
+  })
+  return filterData
+}
+
+export const removeUnselectedFilterData = (data: FilterData): FilterData => {
```
ludtkemorgan, line 328: nit: some of these functions aren't the most intuitive on what they are doing. Can you add a function JSDoc for this and the other function in this file. Also lets make sure tests will be added for this so can you add `it.todo`s for this function in `FilterDrawerHelpers.test.tsx`

ludtkemorgan (approved):
Mostly minor comments. I'm ok with this going in without a full test suite as we know we will quickly adding them, but can you add the todos for all of the tests so we know what still needs to be added?

api/src/enums/feature-flags/feature-flags-enum.ts
ColinBuyck: You've got a deal

sites/public/src/components/browse/FilterDrawerHelpers.tsx
```diff
@@ -0,0 +1,351 @@
...
+export interface FilterData {
+  availability?: Record<FilterAvailabilityEnum, boolean | "true" | "false">
+  bedroomTypes?: Record<UnitTypeEnum, boolean | "true" | "false">
+  homeType?: Record<HomeTypeEnum, boolean | "true" | "false">
+  isVerified?: boolean | "true" | "false"
+  listingFeatures?: Record<keyof ListingFeatures, boolean | "true" | "false">
+  monthlyRent?: Record<"maxRent" | "minRent", string>
+  regions?: Record<RegionEnum, boolean | "true" | "false">
+  section8Acceptance?: boolean | "true" | "false"
+}
```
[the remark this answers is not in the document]
ColinBuyck: Niceeeee

PR #1262, opened 2021-05-24
Fix/cypress wait

dominikx96:
# Pull Request Template

## Issue

#1257 

Addresses # (issue)
- [x] This change addresses the issue in full
- [ ] This change addresses only certain aspects of the issue
- [ ] This change is a dependency for another issue
- [ ] This change has a dependency from another issue

## Description

I tried to follow all CircleCI steps in my local setup and noticed in the **build** section in the config.yml  command: `yarn build:app:public ` - BUT we just set up backend without running it. It means - the build doesn’t contain all the necessary pages which come from the listing so this build is not correct.

I moved the build command into sites/public `package.json`, then `yarn start` command should run after build.

Also fixed broken `MyApplications` test - it requires to wait for route change after successful login.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Prototype/POC (not to merge)
- [ ] This change is a refactor/address technical debt
- [ ] This change requires a documentation update
- [ ] This change requires a SQL Script

## How Can This Be Tested/Reviewed?

Please describe the tests that you ran to verify your changes. Provide instructions so we can review. Please also list any relevant details for your test configuration
- [x] Desktop View
- [ ] Mobile View
- [ ] Test A
- [ ] Test B

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have reviewed the changes in a desktop view
- [ ] I have reviewed the changes in a mobile view
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have assigned reviewers
- [x] I have updated the changelog to include a description of my changes

PR #4872, opened 2025-05-15
fix: add additional logging to afs job (#1262)

ludtkemorgan:
This PR addresses [#1261](https://github.com/metrotranscom/doorway/issues/1261)

## Description

NOTE: this has already been merged into Doorway, but we would like to keep it in sync with core

The AFS cron job is occasionally failing in Doorway starting the other day. In order to figure out why it is failing we will need to know what listing it fails on as well as the rule and rule key it complains about.

This PR does two things:
- Adds a try catch around the creation of the flag set so that it doesn't fail the entire job, but rather just that one duplicate set.
- Adds additional logging. More info on what listings the job is getting run for as well as if it fails what did it fail on

Running the AFS job locally still works.

1. make sure you have the following env variables set
```
DUPLICATES_PROCESSING_CRON_STRING=0,5,15,25,30,35 * * * *
DUPLICATES_CLOSE_DATE=2024-10-08 00:00 -08:00
```
2. Update the cron string to be something in the near future
3. Reseed the database
4. Start up the backend and watch the logs. When the cron job is triggered it should output the listing ids that it is running against.

## Author Checklist:

- [n/a] Reviewed in a desktop view
- [n/a] Reviewed in a mobile view

## Review Process:

### DIFF
--- a/api/src/services/application-flagged-set.service.ts
+++ b/api/src/services/application-flagged-set.service.ts
@@ -537,7 +537,11 @@ export class ApplicationFlaggedSetService implements OnModuleInit {
       },
     });
     this.logger.warn(
-      `running duplicates check on ${outOfDateListings.length} listings`,
+      `running duplicates check on ${
+        outOfDateListings.length
+      } listings. Listings - ${outOfDateListings
+        ?.map((listing) => listing.id)
+        .toString()}`,
     );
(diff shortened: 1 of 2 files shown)


mcgarrye (approved):
LGTM

Issue #4818, opened 2025-04-16
Partners - User Search not working with spaces


mcgarrye:
- [ ] General
- [ ] Backend
- [ ] Public
- [x] Partners

## Background / Issue
Searching for users with a space in their name results in no results returned. Works for listings with spaces in the value.
## Technical Context
Investigation needed
## Acceptance Criteria
Searching for users by their name with a space should return the users that match.
## QA Notes

** Please provide as much information as possible (screenshots, documentation, etc) **

PR #4876, opened 2025-05-16
feat: support searching first + last name in users admin

jaredcwhite:
This PR addresses #4818

## Description

This adds a little more smarts to the users query so that we can search for first name + last name in the search query.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/utilities/build-user-where.ts
+++ b/api/src/utilities/build-user-where.ts
@@ -12,20 +12,37 @@ export const buildWhereClause = (
   const filters: Prisma.UserAccountsWhereInput[] = [];
 
   if (params.search) {
+    const namesClause = [];
+
+    /*
+        We'll add a first/last name OR pair for each word, in order to support searching for
+        first name + last name in the query
+      */
+    params.search
+      .split(' ')
+      .filter((str) => str !== '')
+      .forEach((word) => {
+        namesClause.push({
+          OR: [
+            {
+              firstName: {
+                contains: word,
+                mode: Prisma.QueryMode.insensitive,
+              },
+            },
+            {
+              lastName: {
+                contains: word,
+                mode: Prisma.QueryMode.insensitive,
+              },
+            },
+          ],
+        });
+      });
+
     filters.push({
       OR: [
-        {
-          firstName: {
-            contains: params.search,
-            mode: Prisma.QueryMode.insensitive,
-          },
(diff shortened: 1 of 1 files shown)


api/src/utilities/build-user-where.ts
```diff
@@ -12,20 +12,37 @@ export const buildWhereClause = (
   const filters: Prisma.UserAccountsWhereInput[] = [];
 
   if (params.search) {
+    const namesClause = [];
+
+    /*
+        We'll add a first/last name OR pair for each word, in order to support searching for
+        first name + last name in the query
+      */
+    params.search
```
ludtkemorgan, line 21: question:
Your example of "Eric Jablonski" finding both "Eric McGarry and Emily Jablonski" I think makes sense. However, if you type something that would catch more names it could be a bit confusing. Here is a screenshot of an example search of "e villas" which basically finds all users since any user with "e" in either first or last name is returned.
What are people's thoughts on that. @mcgarrye do you have an opinion?
image

mcgarrye: I feel we may be over-engineering this solution. Pretty sure this used to work. I believe we can switch to matching first AND last names exactly (not contains) we have the solution we need. We don't need to have a fuzzy search or return more than we need to. Usually the admin will know exactly who they are looking for

jaredcwhite: I do also like having a "fuzzier" search of sorts…I just added one check to require a word for name search to be at least two characters, and I think the frontend doesn't even start the filtering until at least three characters so that should alleviate the issue Morgan found.

ludtkemorgan: I think the fuzzy search makes sense. And I like the requirement that it has to have at least 2 characters which alleviates most of my concern.
What I was most concerned by was something happening like the following example
* User is trying to look up a user where they know the last name and the first initial of the first name
* They type that in and it basically doesn't filter
* They wouldn't know that it would be better to remove that first initial than include it in the search.

Overall though I like this change!

ludtkemorgan (approved):

jaredcwhite:
@ludtkemorgan Yeah, I can see that about wanting to support initials, but it's probably hard to do without a lot of additional logicking. Anyway, I'll go ahead and merge this!

PR #4882, opened 2025-05-21
feat: adds ui load tester & getStaticProps from getServerSideProps

YazeedLoonat:
This PR addresses https://github.com/metrotranscom/doorway/issues/1271

## Description
This adds a ui load tester script

more importantly, this changes how we get listing data for the public listing detail page from `getServerSideProps` to `getStaticProps` this should mean that instead of the server generating the HTML for the public listing detail page each time it generates it once then holds onto it for 30 seconds before rebuilding it

Stand up the public site locally and verify you can hit the listing detail page
Aim the load tester at your listing detail page and make sure it stands up to most load tests (it'll break down at higher ends still) 

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/src/pages/listing/[id]/[slug].tsx
+++ b/sites/public/src/pages/listing/[id]/[slug].tsx
@@ -102,5 +102,8 @@ export const getStaticProps: GetStaticProps = async (context: {
   }
   const jurisdiction = fetchJurisdictionByName()
 
-  return { props: { listing: response.data, jurisdiction: await jurisdiction }, revalidate: 30 }
+  return {
+    props: { listing: response.data, jurisdiction: await jurisdiction },
+    revalidate: Number(process.env.cacheRevalidate),
+  }
 }
--- a/sites/public/src/pages/listing/[id]/[slug].tsx
+++ b/sites/public/src/pages/listing/[id]/[slug].tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useContext } from "react"
 import Head from "next/head"
+import { GetStaticPaths, GetStaticProps } from "next"
 import axios from "axios"
 import { t } from "@bloom-housing/ui-components"
 import {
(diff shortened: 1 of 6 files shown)


ui-load-tester.rb
```diff
@@ -0,0 +1,29 @@
+# What is this:
+#   This is a ruby script to load test any of our front end urls
+
+# Why is this written in ruby?
+#   Ruby supports multithreading much better natively than node/js does
+
+# How to use:
+#   check out how `ab` works. On mac you can run `man ab` in a terminal to get the man page for ab
```
emilyjablonski, line 8: **question:** What is `ab`?

ui-load-tester.rb
```diff
@@ -0,0 +1,29 @@
...
+#   you can change the n arguments on these calls to increase the total number of requests
+#   you can change the c arguments on these calls to increase the throughput basically (number of requests made concurrently)
+#   you can change the url arguments on these calls to direct which url will recieve the traffic
```
emilyjablonski, line 11: **praise:** 
```suggestion
#   you can change the url arguments on these calls to direct which url will receive the traffic
```

ui-load-tester.rb
```diff
@@ -0,0 +1,29 @@
...
+#   to execute once your changes are made from the root directory run `ruby tester.rb`
+#   In order to test mem usage you can attach the node debugger to either the public or partner sites and track mem usage that way
```
emilyjablonski, line 13: **question:** What does attaching the debugger look like?

ui-load-tester.rb
```diff
@@ -0,0 +1,29 @@
...
+#   check out how `ab` works. On mac you can run `man ab` in a terminal to get the man page for ab
+#   you can change the n arguments on these calls to increase the total number of requests
+#   you can change the c arguments on these calls to increase the throughput basically (number of requests made concurrently)
+#   you can change the url arguments on these calls to direct which url will recieve the traffic
+#   to execute once your changes are made from the root directory run `ruby tester.rb`
+#   In order to test mem usage you can attach the node debugger to either the public or partner sites and track mem usage that way 
+#   You can also track cpu usage, or track it through a cpu activity monitor
+
+def main()
+  # call main page english
```
emilyjablonski, line 17: **question:** What's the purpose of `english` here?

ui-load-tester.rb
ludtkemorgan: +1 I've never worked with ruby before and not exactly sure how to run this script to test it out

ui-load-tester.rb
```diff
@@ -0,0 +1,29 @@
+# What is this:
```
ludtkemorgan, line 1: suggestion: I wonder if we should find a better location for this script. I'm cool with it living in this repo, but could we create a directory for things like scripts? This being the first item there

sites/public/src/pages/listing/[id]/[slug].tsx
```diff
@@ -79,68 +80,27 @@ export default function ListingPage(props: ListingProps) {
...
       },
     })
   } catch (e) {
     return { notFound: true }
   }
-  const jurisdiction = fetchJurisdictionByName(context.req)
+  const jurisdiction = fetchJurisdictionByName()
 
-  return { props: { listing: response.data, jurisdiction: await jurisdiction } }
+  return { props: { listing: response.data, jurisdiction: await jurisdiction }, revalidate: 30 }
```
Comment (ludtkemorgan) [inline sites/public/src/pages/listing/[id]/[slug].tsx:105]: suggestion: When a listing goes live or closes we want to make sure that this page is updated in a relatively quick amount of time so I think 30 seconds makes sense. But I could also see this being an environment variable that we could default to 30 seconds and override it in case we need to adjust.

Also, this doesn't have to be part of this PR, but Next just support [On-Demand ISR](https://nextjs.org/docs/14/pages/building-your-application/data-fetching/incremental-static-regeneration#on-demand-revalidation). So we could theoretically have this be a much larger "revalidate" time and have the backend refresh the page when the listing updates. That might be something we want to look at for long term sustainability

ludtkemorgan (approved):
This works as expected 🏎️ . Can you update the script documentation a bit but otherwise good to merge so I'm approving it now

ui-load-tester.rb
YazeedLoonat: Hey @emilyjablonski `ab` is a utility that is baked into unix https://httpd.apache.org/docs/2.4/programs/ab.html

if you run `man ab` in your terminal it'll give you the manual page for the utility 
I'm not sure how much more explanation of ab we should keep in our documentation since its a built in utility with a man page baked in does that help/make sense?


@ludtkemorgan I do have the following "from the root directory run `ruby tester.rb`" as part of the comment ~line 12 

I can also put that into the package.json as a script so you could call it with something like `yarn ui-load-test` or something like that?

ui-load-tester.rb
YazeedLoonat: Hey @emilyjablonski we have set it up so we can attach a debugger to the public/partner site already https://github.com/bloom-housing/bloom/blob/f40f838efc7c0499815616cb85bc5ff58f48dcca/sites/public/package.json#L14

There are a bunch of aways to attach a UI to actually have insight into the debugger outputs:
https://nodejs.org/en/learn/getting-started/debugging
https://code.visualstudio.com/docs/nodejs/nodejs-debugging
https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27

I used the chrome tools one but found it not the most helpful when tracking CPU usage and much better at tracking mem usage (for nextjs anyway)

ui-load-tester.rb
YazeedLoonat: good question: When trouble shooting where cpu usage was going one of the pastabilities was that translated pages were the ones eating up cpu usage so at one point this had a pretty broad number of pages hit as I tried to narrow down what exactly was causing the problem

I left english here to be pretty precise that this call is hitting the english page as someone else may want to hit the spanish page for example

I can remove it if you would like

sites/public/src/pages/listing/[id]/[slug].tsx
Comment (YazeedLoonat) [reply sites/public/src/pages/listing/[id]/[slug].tsx:105]: I do like the idea of this eventually going to on-demand ISR!

emilyjablonski:
@YazeedLoonat Do you mind just adding a bit more documentation on running the script in the areas we had questions about pre-merge?

Issue #4682, opened 2025-03-06
Public - FE - Rental Finder - All questions


emilyjablonski:
A subtask of #4451 which has some more context.

This should be the second subtask. We should build off of the first page and include all of the questions in the [finder](https://homeconnect.detroitmi.gov/finder). When the form submits, it should redirect to the listings page with those filters applied. We will be able to reuse the helper functions built during the filtering work in order to construct an appropriate URL that will filter the browse page of listings.

If we the backend work to be able to fetch programs (aka community types) isn't finished yet by the time we pick this up, we can just mock that page visually with the header, and will make a subtask attached to this same parent to make that page actually call for community types.

sarahlazarich:
Blocked by initial question format 
Won't be able to actually submit + filter as a part of this - just UI 
@emilyjablonski to add a subtask for that part

PR #4884, opened 2025-05-23
feat: rental finder all questions

matzduniuk:
This PR addresses #4682 

## Description

* Add new translation strings for Rental Finder question based on the existing Detroit Finder
* Update the existing Finder initial page to include all required questions 
* Add new finder component for the disclaimer section
* Integrate form into finder with persistence between _'pages'_ a clearing on `skip`
* Handle rental form submission with query string formatting

1. Go to the finder page `/finder`
2. Go through the finder questionnaire, selecting at least one option (if no options are selected, no query string will be added on routing)
3. Verify the navigation URL query against previously selected options
4. _[Additionally]_ Verify that selecting no options or skipping the form results in no query string on route

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/__tests__/pages/finder.test.tsx
+++ b/sites/public/__tests__/pages/finder.test.tsx
@@ -532,7 +532,7 @@ describe("<RentalsFinder>", () => {
 
       await waitFor(() => {
         expect(pushMock).toBeCalledWith(
-          "/listings?bedrooms=studio,oneBdrm&regions=Greater_Downtown,Westside&monthlyRent=1500-3000&section8Acceptance=true&listingFeatures=wheelchairRamp,elevator,serviceAnimalsAllowed&reservedCommunityTypes=veterans"
+          "/listings?bedroomTypes=studio,oneBdrm&regions=Greater_Downtown,Westside&monthlyRent=1,500.00-3,000.00&section8Acceptance=true&listingFeatures=wheelchairRamp,elevator,serviceAnimalsAllowed&reservedCommunityTypes=veterans"
         )
       })
     })


sites/public/src/components/finder/RentalsFinder.tsx
```diff
@@ -113,7 +182,36 @@ export default function RentalsFinder() {
...
+    setSectionIndex(rentalFinderSections.length - 1)
+    setStepIndex(0)
+  }, [rentalFinderSections.length])
+
+  const onSubmit = useCallback(() => {
+    const urlQueryElements = []
+
+    Object.entries(formData).forEach((entry) => {
+      const [key, value] = entry
+      if (Array.isArray(value) && value.length) {
```
emilyjablonski, line 198: **note:** Once @ColinBuyck 's filter work is merged, we can likely pull this from a shared helper

shared-helpers/src/locales/general.json
```diff
@@ -573,6 +573,23 @@
...
+  "finder.building.subtitle": "Some affordable housing rental properties are dedicated to specific populations, like seniors. We will use your selections to find you rentals that may match your housing needs.",
+  "finder.building.withDisabilities": "Rentals for Residents with Disabilities",
+  "finder.building.senior55": "Rentals for Seniors 55+",
+  "finder.building.senior62": "Rentals for Seniors 62+",
+  "finder.building.homeless": "Supportive Housing for the Homeless",
+  "finder.building.veterans": "Rentals for Veterans",
+  "finder.disclaimer.question": "Before you see the results",
+  "finder.disclaimer.subtitle": "We have some final notes on understanding Regulated Affordable Housing.",
+  "finder.disclaimer.alert": "Disclaimer: The information in this personalized rental finder should be used for informational purposes only. Due to the changing nature of property information, the best way to see if you qualify for a property or if it has any availability is by contacting the property itself.",
+  "finder.disclaimer.info1": "Detroit Home Connect will show you regulated affordable rentals, meaning they are affordable because of governmental funding.",
```
emilyjablonski, line 588: **nitpick:** We can't have jurisdictional references in core. We will override this string in Detroit so that it has this exact copy, but in core it should be something generic like `This website will show you regulated affordable rentals, meaning they are affordable because of governmental funding.`

sites/public/src/components/finder/RentalsFinder.tsx
matzduniuk: The code required more updates, but I stitched the code to reduce repetition.

**note**: For the future, we could probably think about detecting and adding different messages to the `<ListingBrowse>` when there are _"no listings"_ vs _"no listing matching filers"_

emilyjablonski: Good call out, I think Colin is pulling that change into his next filter ticket

Issue #4683, opened 2025-03-06
Public - FE - Rental Finder - Toggle questions


emilyjablonski:

matzduniuk:
QA Notes: 
This PR allows for control of whether the `Regions` and `Accessibility` are available on the `finder` questionnaire based on their matching feature flags. 

To test this, different combinations of settings for the `enableRegions` and `enableAccessibilityFeatures` feature flags should be used, and verified if the matching section appears on the `/finder` page

sarahlazarich:
lgtm!

PR #4911, opened 2025-05-29
feat: add feature flag based finder question toggling

matzduniuk:
This PR addresses #4683 

## Description

* Add feature flag-based rental finder question section toggling

* Access Prisma database and update jurisdiction feature flags. Currently available:
   * `enableRegions`
   * `enableAccessibilityFeatures`
* Go to `/finder` and go through the finde

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/package.json
+++ b/shared-helpers/package.json
@@ -16,7 +16,7 @@
     "prettier": "prettier --write \"**/*.ts\""
   },
   "dependencies": {
-    "@bloom-housing/ui-components": "12.4.0",
+    "@bloom-housing/ui-components": "12.7.2",
     "@bloom-housing/ui-seeds": "1.22.1",
     "@heroicons/react": "^2.1.1",
     "axios-cookiejar-support": "^5.0.5",
--- a/sites/partners/package.json
+++ b/sites/partners/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@bloom-housing/shared-helpers": "^7.7.1",
-    "@bloom-housing/ui-components": "12.4.0",
+    "@bloom-housing/ui-components": "12.7.2",
     "@bloom-housing/ui-seeds": "1.22.1",
     "@heroicons/react": "^2.1.1",
     "@mapbox/mapbox-sdk": "^0.13.0",
--- a/sites/public/package.json
+++ b/sites/public/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@bloom-housing/shared-helpers": "^7.7.1",
-    "@bloom-housing/ui-components": "12.4.0",
+    "@bloom-housing/ui-components": "12.7.2",
     "@bloom-housing/ui-seeds": "1.22.1",
     "@heroicons/react": "^2.1.1",
     "@mapbox/mapbox-sdk": "^0.13.0",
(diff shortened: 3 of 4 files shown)


sites/public/src/components/finder/RentalsFinder.tsx
```diff
@@ -33,7 +34,11 @@ type FinderSection = {
   sectionSteps: FinderStep[]
 }
 
-export default function RentalsFinder() {
+export type RentalsFinderProps = {
+  activeFeatureFalgs: FeatureFlagEnum[]
```
emilyjablonski, line 38: **nitpick:** 
```suggestion
  activeFeatureFlags: FeatureFlagEnum[]
```

sites/public/src/pages/finder.tsx
```diff
@@ -1,10 +1,32 @@
+import {
+  FeatureFlagEnum,
+  Jurisdiction,
+} from "@bloom-housing/shared-helpers/src/types/backend-swagger"
 import RentalsFinder from "../components/finder/RentalsFinder"
 import Layout from "../layouts/application"
+import { fetchJurisdictionByName } from "../lib/hooks"
+
+export default function Finder({ jurisdiction }: { jurisdiction: Jurisdiction }) {
+  const activeFeatureFlags: FeatureFlagEnum[] =
```
emilyjablonski, line 10: **question:** Because we're getting the jurisdiction in static props, I think once we build and deploy this, we won't be able to dynamically turn these flags on or off because we're getting them from the static-ly fetched jurisdiction @YazeedLoonat does this sound right to you? I know we chatted about this recently and am wondering if we will need `getServerSideProps` for every page which could be affected by a toggle

sites/public/__tests__/pages/finder.test.tsx
```diff
@@ -59,8 +67,202 @@ describe("<RentalsFinder>", () => {
     expect(screen.queryByRole("button", { name: /back/i })).not.toBeInTheDocument()
   })
 
+  describe("should hide toggle sections based on feature flags", () => {
+    it("should hide regions section if not toggled on", async () => {
```
emilyjablonski, line 71: **suggestion:** Can we limit what these tests are testing? This could fail in a lot of places beyond checking if the regions section should hide if toggled off - is there a way to just test that? I think we're covered on what else the test is doing in other tests

emilyjablonski (commented):
Super minimal changes + a question!

sites/public/__tests__/pages/finder.test.tsx
matzduniuk: The problem is how to reliably test if the section is rendered or skipped. 

In case of the `Accessibility` section, it's a bit easier as I can check if the progress bar contains an item for this section. It also affects the header section, which details the total number of sections.

With `Regions`, it's a bit more complicated as this is a step in the section, therefore, there is no indication that the step is missing unless we test that its contents are not to be found where they are supposed to be. 

All other contents that are used in these tests are also being checked in other place,s therefore even if those checks don't fail here, they will somewhere else in this suite

sites/public/src/pages/finder.tsx
YazeedLoonat: yeah we should use the revalidate flag so that it refetches every so often

generally since jurisdictions change very infrequently I think we can use the env variable we use else where for the revalidate value

emilyjablonski (approved):

Issue #4776, opened 2025-04-02
Public - Additional Waitlist Section


sarahlazarich:
This is blocked by [this PR](https://github.com/bloom-housing/bloom/pull/4812) which adds the waitlist fields to partners.

Detroit has additional fields available relating to waitlist numbers.

[Here is the new design](https://www.figma.com/design/xqyamlgdsASkLTYlcifBwq/core-patterns?node-id=3465-4770&m=dev)

[And here is an example of what this component looks like now
](https://bloom-public-seeds.netlify.app/listing/cf908bad-5e0d-41ce-b688-f4be27408449/test_another_lottery_1233_w_main_chicago_il
)
The fields currently appear near the top in the right rail (current waitsize size, open waitlist slots, final waitlist size).
<img width="1507" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeUZSQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--60d7959586765d679896c24d9cf9337044ba076a/Screenshot%202025-03-18%20at%204.39.11%E2%80%AFPM.png" alt="Screenshot 2025-03-18 at 4.39.11 PM.png" />

sarahlazarich:
<img width="1882" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGlaQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5a82aa5f58787b59524ef709d1be1cbb157d9626/image.png" alt="image.png" />

<img width="1481" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNG1aQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9e4959b89f368af5b103e757ce7b1cb48bdcd4f9/image.png" alt="image.png" />

emilyjablonski:
@sarahlazarich I know we chatted about this offline, but confirming that seeing the date twice is for now expected

PR #4915, opened 2025-06-02
feat: add additional waitlist size on public

KrissDrawing:
This PR addresses #4776 

## Description

Adds open waitlist section on public listing view.
Also it modifies a bit lottery section to show sum of total units from unit Groups

Enable `enableWaitlistAdditionalFields` set `Do you want to show a waitlist size?` on partners site (edit listing) to yes and fill values. It should show section with set numbers

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/src/components/listing/listing_sections/Availability.tsx
+++ b/sites/public/src/components/listing/listing_sections/Availability.tsx
@@ -66,6 +66,10 @@ export const Availability = ({ listing, jurisdiction }: AvailabilityProps) => {
     jurisdiction,
     FeatureFlagEnum.swapCommunityTypeWithPrograms
   )
+  const showAdditionalWaitlistFields = isFeatureFlagOn(
+    jurisdiction,
+    FeatureFlagEnum.enableWaitlistAdditionalFields
+  )
 
   const statusMessage = getListingStatusMessageContent(
     listing.status,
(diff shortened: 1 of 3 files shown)


sites/public/src/components/listing/listing_sections/Availability.tsx
```diff
@@ -119,6 +131,40 @@ export const Availability = ({ listing, jurisdiction }: AvailabilityProps) => {
           {content && <p className={"seeds-m-bs-label"}>{content}</p>}
           {subheading && <p className={`seeds-m-bs-label`}>{subheading}</p>}
         </Card.Section>
+        {showAdditionalWaitlistFields && (
+          <Card.Section divider="flush">
+            <Heading size={"md"} priority={3}>
+              {t("listings.waitlist.open")}
+            </Heading>
+            <p className={styles["bold-subheader"]}>
+              {listing.isWaitlistOpen
```
emilyjablonski, line 140: **issue:** I don't think this field on the listing is updated correctly with unit groups, because whether or not the waitlist is open is not determined at the listing level. It results in the following:
Screenshot 2025-06-02 at 5 55 30 PM

emilyjablonski (changes requested):
Left a small note above on using one of the listing fields.
A visual update to make is that we should include these new fields at the bottom of the existing waitlist section. [Here is an example listing](https://bloom-public-seeds.netlify.app/listing/cf908bad-5e0d-41ce-b688-f4be27408449/test_another_lottery_1233_w_main_chicago_il) that shows how the section appears now. At the moment, we can have two sections appear that both indicate a waitlist.

KrissDrawing:
Ok, so i assumed that both section may appear only when `enableUnitGroups = true`.
But tbh it's more confusing than not 😄 I tried to test every possible variant, but there is a high chance i missed something

emilyjablonski:
Thanks for the updates! I'm reviewing this with design and I think it might be more complicated than initially thought. Will send some updates 👍

emilyjablonski:
Would love to get on eyes on this from someone else since we've both done work here! Can poke the team tmrw

YazeedLoonat (changes requested):
Looks good, issue I found was that the change is not represented on the `listing preview` page on the public site even though the "live" version of the listing does show the new changes

YazeedLoonat (approved):
after talking with Emily the preview page is not seedsified yet so this change wouldn't be visible there 
giving the approval

PR #4979, opened 2025-06-10
fix: update seed with due date

emilyjablonski:
Updates the seed to always add a due date. Otherwise, we're unable to save seeded listings until we add a due date.


ludtkemorgan (approved):
This change is good for now. Once the due date isn't required for all jurisdictions we'll want to update this again to have some listings not have application due date.

PR #4992, opened 2025-06-11
test: add unit testing for the Lottery Results page in Public

jaredcwhite:
This PR addresses #4546

## Description

I think this is pretty good coverage now of the feature (should cover the three cases suggested by Emily), but if you can think of additional ways to test LMK.

The unit test passed in CI.

## Author Checklist:

## Review Process:


ludtkemorgan (commented):
@jaredcwhite great test! I think the main place that should have additional testing is the chained API calls that are made on this page. There are a lot of error catching that we should add a test for to prove that the page still loads if not all of the data does.

jaredcwhite:
@ludtkemorgan Good idea, I'll look at adding that in.

ludtkemorgan (approved):
Thanks for making the new test. I'm glad that the new test caught a bug 🐛 . And yeah, I agree that the chained requests should be refactored as well as our error handling in general.

Looks great though!

Issue #5002, opened 2025-06-17
[Bug]: common digital application question not properly selecting in partners


ludtkemorgan:
### What is the issue?

The digital common app radio checkbox is not appearing checked for either yes or no on the partner site

### How can you reproduce the bug?

1. Log into the partner site and go into a listing that "digital application" is enabled
2. Look at the field "Are you using the common digital application?" and neither "Yes" or "No" is selected.
3. After save the same persists

### Is there any additional context?

It appears like this was introduced on this [PR](https://github.com/bloom-housing/bloom/pull/4935) when we updated UI-components

### What browsers are you seeing the problem on?

Chrome, Firefox, Safari

### Where are you seeing this issue?

Desktop

emilyjablonski:
QA Notes

Create or update a listing with a value to `Are you using the common digital application?` which appears after selecting `Yes` to `Is there a digital application?` and ensure the value re-populates after saving and reopening the form.

We can test this in [core](https://partners.bloom.exygy.dev/sign-in).

PR #5004, opened 2025-06-18
fix: common app selection

emilyjablonski:
This PR addresses #5002

## Description

The "Are you using the common digital application?" has stopped autofilling.

The commit that "caused" it had a UIC upgrade that ended up including 10 UIC commits - [this is the UIC commit](https://github.com/bloom-housing/ui-components/pull/169) that causes the issue but I still don't know why. It did include a few package upgrades that might have had some unexpected effects.

Create or update a listing with a value for `Are you using the common digital application?` which appears after selecting `Yes` to `Is there a digital application?` and ensure the value re-populates after saving and reopening the form.

I updated our listing Cypress test to also re-open an entered listing to make sure all of the fields populate as expected, as we've had lots of bugs around this.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/components/listings/PaperListingForm/sections/ApplicationTypes.tsx
+++ b/sites/partners/src/components/listings/PaperListingForm/sections/ApplicationTypes.tsx
@@ -36,6 +36,36 @@ const ApplicationTypes = ({ listing }: { listing: FormListing }) => {
   const { register, setValue, watch, errors, getValues } = useFormContext()
   const { doJurisdictionsHaveFeatureFlagOn, getJurisdictionLanguages } = useContext(AuthContext)
 
+  const getDefaultMethods = () => {
+    const temp: Methods = {
+      digital: null,
+      paper: null,
+      referral: null,
+    }
+    const applicationMethods =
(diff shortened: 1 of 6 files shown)


sites/partners/src/components/listings/PaperListingForm/sections/ApplicationTypes.tsx
```diff
@@ -36,6 +36,36 @@ const ApplicationTypes = ({ listing }: { listing: FormListing }) => {
   const { register, setValue, watch, errors, getValues } = useFormContext()
   const { doJurisdictionsHaveFeatureFlagOn, getJurisdictionLanguages } = useContext(AuthContext)
 
+  const getDefaultMethods = () => {
```
emilyjablonski, line 39: **note:** Moves this logic to set the default values out of the useEffect because it never needed to be in one, and this solved the issue. I think it was seeing a null value on page load, and then was being set too late, but I'm not sure what caused the issue.

ludtkemorgan (approved):
Thanks for making such a quick fix! Interesting bug

Issue #4874, opened 2025-05-16
Partners - Last Updated At


sarahlazarich:

emilyjablonski:
Do we want this in core, or do we want this feature toggled?

skeller-exy:
@jaredcwhite I am not seeing the `enableListingUpdatedAt` flag in https://bloom-detroit-partners.netlify.app/admin. Is this ready to test?

PR #5027, opened 2025-06-30
feat: add listing updated at to detail page, based on feature flag

jaredcwhite:
This PR addresses #4874

## Description

## Author Checklist:

## Review Process:

### DIFF
--- a/api/prisma/seed-staging.ts
+++ b/api/prisma/seed-staging.ts
@@ -84,6 +84,7 @@ export const stagingSeed = async (
         FeatureFlagEnum.enableListingFiltering,
         FeatureFlagEnum.enableListingOpportunity,
         FeatureFlagEnum.enableListingPagination,
+        FeatureFlagEnum.enableListingUpdatedAt,
         FeatureFlagEnum.enableMarketingStatus,
         FeatureFlagEnum.enableNeighborhoodAmenities,
         FeatureFlagEnum.enablePartnerDemographics,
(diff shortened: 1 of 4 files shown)


sites/public/__tests__/components/listing/ListingViewSeeds.test.tsx
```diff
@@ -14,4 +15,33 @@ describe("<ListingViewSeeds>", () => {
...
+              description: "",
+              active: true,
+              jurisdictions: [],
+              name: FeatureFlagEnum.enableListingUpdatedAt,
+            },
+          ],
+        }}
+      />
+    )
+    expect(view.getAllByText("Listing Updated: Dec 31, 2019")).toBeDefined()
```
ludtkemorgan, line 45: nit: If there is more than one I like to have an explicit expect for the length. So this would become
```
    expect(view.getByText("Listing Updated: Dec 31, 2019")).toHaveLength(2)
```

Also, I do find it weird that there is a second "Listing Updated" in the DOM. I see that when I render the page but the other instance is not visible by the user. Do you know why that is?

ludtkemorgan (approved):
This change looks good! Just one comment about related to the test

sites/public/__tests__/components/listing/ListingViewSeeds.test.tsx
jaredcwhite: @ludtkemorgan hmm, somehow I dismissed that as being weird because I assumed there was some duplicated DOM that was shown/hidden depending on viewport size, but I wonder if that's actually the case. I'll take a look at it!

jaredcwhite: Yeah, that was the reason. I've updated the test to check specifically for the length 2.

Issue #4886, opened 2025-05-28
Testing Bug - Community Types Tag


sarahlazarich:
Steps to Reproduce: 
- Go on to a listing that has community types added using the detroit version of community types (aka repurposed programs) in the partners portal
- Go on to a listing that has an "old" community type added (see example here) 

Current Behavior
- The listing with the "new" community type does not display a tag on the listing 
- The listing with community types added the "old" way does display a tag on the listing 

Expected Behavior
- The listing with the Detroit community types added should have one or multiple community type tags displaying depending on how many are in Partners 
- The "old" community type tags should not be displaying on listings 

Notes: 
- https://exygy.slack.com/archives/C0203460K6Y/p1747859403598999

sarahlazarich:
I don't see the community types / programs tag on listings, now, and I think that's what we'd expect -- 

<img width="545" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd21YQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c5839de39498662f46bbd0301d88fb9d9e64daea/image.png" alt="image.png" />

PR #5042, opened 2025-07-07
fix: display program tags instead of community types tag

KrissDrawing:
This PR addresses #4886 

## Description

It's fix for failed QA ticket (see comments on zenhub). It will show program tags when `swapCommunityTypesWithPrograms` flag is on. There is some auto-translation of programs for listing view which doesn't exist for `/listings` and `under construction` view (on home page). Thats why i used the solution that exist in detroit where i translate on frontend programs that are not translated.

having `swapCommunityTypesWithPrograms = true` add on partners listing edit page "programs" labeled as `Community Types`. On public listing view (`http://localhost:3000/listing/[id]`) it should display all of those programs as tags.
Same for listing view, and under construction view on home page (`enableUnderConstruction` flag to `true`, and marketing set to `under Construction`

## Author Checklist:

## Review Process:


sites/public/src/components/listing/listing_sections/MainDetails.tsx
```diff
@@ -42,7 +43,8 @@ export const getListingTags = (
   hideHomeTypeTag?: boolean,
   hideAccessibilityTag?: boolean,
   enableIsVerified?: boolean,
-  swapCommunityTypeWithPrograms?: boolean
+  swapCommunityTypeWithPrograms?: boolean,
+  translatePrograms?: boolean
```
emilyjablonski, line 47: **question:** I think using two different methods of translation for the browse page and the detail page could lead to two different translations. @ludtkemorgan Curious what you think about this because we don't have machine translations on the browse page? Does it look like using the strings on the detail page too?

ludtkemorgan: Yeah, I agree the two different translations can become tricky. 

The problem with the strings route is that If we ever need to add a new "program" then we will have to add the translation to file. And it could be hard to keep that up to date. Even the default seeded programs don't show up correctly

Image

The problem with the translation route is it might not look good in a tag if it translates different than what we expect.

For this PR I'd recommend we keep the same of using the strings for both the details page and the browse page. But the thing I'd change is that if possible the fallback would be the program text if the translation is not found via the json. I don't know if our `t` function handles that scenario though.

Long term I'd like us to fine tune the returned response from the API so that the fields we care about are actually returned. For example we return the entire multiselect question information on the browse page and all we care about is the title of it. If we can go down that route we could request the translated multiselect text. But we don't need to do that as part of this PR

emilyjablonski (approved):

Issue #4984, opened 2025-06-11
Public - Full-Time Student Common Application Question


ColinBuyck:
The Detroit common app has a set of adjustments made pertaining to if the applicant or if people associated to them are students.

This adjustment should occur in three places:

1. It should ask "Are you a full-time student?" below the "Do you work in county question" in the contact information section.
2. It should ask "Are they a full-time student?" at the end of the household member section following the "what is their relationship to you question
3. It should ask "Are all household members students?" as its one page at the end of the household members section. Note that this addition will be replacing the Household Includes Student or Member Nearing 18 information with this new updated question and description. See exact description in the link below!

For the official language and placement refer to the finalized copy document [here](https://docs.google.com/document/d/1KOWK_W1yxgOTCzRhaFmBgrL4fxfpHkqB9Me_apd4EOM/edit?tab=t.0)

Notes:
- Feature flagged based on #4987
- Will need to update the mapping to format the new application data into the database's accepted type

skeller-exy:
@KrissDrawing Thanks for the additional note. 🙏 My confusion is regarding what does autofill mean when those fields are hidden? I am thinking from the user's perspective, if the fields are not currently in the form what does autofilling them look like? Is there any "behind the scenes" behavior happening I should test? Thanks!

PR #5044, opened 2025-07-07
feat:  public - add Full-Time Student question to common app

KrissDrawing:
This PR addresses #4984 

## Description

Adds full-time student questions to common app.
Wasn't quite sure if it should be included in summary (as some fields are not displayed there), but i included it with `Full-Time Student` label (using naming approach from [this PR](https://github.com/bloom-housing/bloom/pull/5026))

Apply for listing with `enableFullTimeStudent = true`.
- it should show question in contact address `http://localhost:3000/applications/contact/address`
- Then select `Other people will live with me` and in household member it should be another full-time student question (`/applications/household/member`)
- for `/applications/household/student` it should have different copy
- in summary it should show all changes (unless it shouldn't then i will roll it back)
Make sure that application with `enableFullTimeStudent = false` works as before

## Author Checklist:

## Review Process:


ColinBuyck (commented):
This is looking really good just have one question around how we handle autofill for this. I'm noticing that we clear the selection of work in region in appAutofill.ts but right now the full-time student selection is carried over. That's probably more of a product question @sarahlazarich 
Screenshot 2025-07-08 at 5 12 13 PM

At the very least, we should make sure to set it as null if the flag is off in the autofill step to protect against an edge case of a user adding an application with the flag on and then autofilling with it off... this would mean the data would float with them despite not being asked about

ColinBuyck (approved):
LGTM 💵 Thanks for making that Partners update too!

sarahlazarich:
Caught up with Izzie, and we think that the Work in Region question was cleared from autofill because a person's job location could change regularly, so that's intentional.

I think people are more likely to be full-time students for years, so it makes sense to include that in autofill.

Issue #4838, opened 2025-04-24
Testing Bug - Listing Directory Search Order


sarahlazarich:
Steps to Reproduce 
- Go to Listings page on public site

Expected Behavior 
- Should order listings by “comingSoon” (aka “Under Construction”) first to mirror existing behavior 

Current Behavior 
- Sorting by "Last Updated"

sarahlazarich:
Will be 
- ordered by coming soon 
- ordered by name

jaredcwhite:
There are some rather specific ordering SQL code in the old Detroit repo and I'm not sure if/how to reproduce some of that. See: https://github.com/CityOfDetroit/bloom/blob/c098e93e237d7db7d3d9af1ba128b56b223e9b45/backend/core/src/listings/listings.service.ts#L432

So far what I've done is update `fetchOpenListings` in `sites/public/src/lib/hooks.ts` to include these:

```
orderBy: [ListingOrderByKeys.marketingType, ListingOrderByKeys.mostRecentlyPublished],
orderDir: [OrderByEnum.desc, OrderByEnum.desc],
````

but that feels simplistic and might not work as expected.
cc @YazeedLoonat

YazeedLoonat:
Hey @jaredcwhite super sorry it took me so long to get back to ya on this. 

I did a bit of digging and you're right just setting the sort order like that would change the way the functionality would work compared to what it did in typeorm. In that nulls would be part of the ordering instead of being forced to be last like in typeorm. 


There is an experimental feature flag we can turn on: https://github.com/prisma/prisma/releases/tag/4.1.0 if you search for `orderByNulls` in that release doc. I think we can turn that on and then apply the `:nulls: 'last'` property and basically recreate the same thing we had in typeORM! 

and with the added bonus of being able to use this in other places! 

hopefully that helps you out a bit but lemme know if that is causing you some problems!

jaredcwhite:
@YazeedLoonat That's an interesting feature, we might need that for something else down the road. But in working on this further I discovered that `marketingType` field is `NOT NULL` meaning we don't need to worry about nulls and where they might show up in the sort.

I'm more concerned about the other aspects of the old query, which maybe aren't relevant anymore but I'm not sure:

```js
        qb.addOrderBy(`to_char(listings.marketingDate, 'YYYY')`, "ASC")
        qb.addOrderBy(
[...]

PR #5054, opened 2025-07-07
feat: add marketing sort ordering

jaredcwhite:
This PR addresses #4838 

## Description

This PR adds additional sort ordering for public site listings to include the season and the year. While only some jurisdictions would use those fields, it's fine if they're null and thus the ordering will fall back to typical reverse-chron.

There's also a schema change and migration from the previous `marketingDate` field to the new `marketingYear` field (previously only the year portion of the date was used for anything, so it's simpler just to store year as an integer).

(I suspect this can't be tested in a preview deploy, since it requires updated database schema)

Update listing(s) with "coming soon" marketing type and a marketing season/year and verify the sort order for listings has changed.

## Author Checklist:

## Review Process:


api/prisma/migrations/20250703205631_change_marketing_date_to_marketing_year/migration.sql
```diff
@@ -0,0 +1,11 @@
+/*
+  Warnings:
+
+  - You are about to change column `marketing_date` on the `listings` table. All the data in the column will be transformed to year only.
+
+*/
+-- AlterTable
+ALTER TABLE "listings"
+ALTER COLUMN "marketing_date" TYPE INTEGER
+USING EXTRACT(YEAR FROM "marketing_date");
```
YazeedLoonat, line 10: **praise:** woah! I never knew this functionality existed! v cool!

api/prisma/migrations/20250703205631_change_marketing_date_to_marketing_year/migration.sql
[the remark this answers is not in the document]
YazeedLoonat: we shouldn't have to, this migration updates the data as it renames columns/re-types columns

YazeedLoonat (commented):
I couldn't test this locally rn but the migration and code looked good to me

will come back to this to test locally once I can reseed my local db

api/prisma/migrations/20250703205631_change_marketing_date_to_marketing_year/migration.sql
emilyjablonski: OH is this why we don't need a script?

jaredcwhite: @YazeedLoonat I didn't know about it either at first, took a bit of sleuthing to find the right syntax but yeah, it's cool!

emilyjablonski:
@YazeedLoonat Do you mind re-reviewing this one?

api/prisma/migrations/28_change_marketing_date_to_marketing_year/migration.sql
```diff
@@ -0,0 +1,11 @@
+/*
```
YazeedLoonat, line 1: **nitpick:** we can drop this warning from here

YazeedLoonat (approved):
LGTM just a small nitpick

Issue #5041, opened 2025-07-04
[Bug]: Detroit's Community Types Breaking Add Appliction


ColinBuyck:
### What is the issue?

While reproducible in https://bloom-detroit-partners.netlify.app/, this change will be made in core. 

Currently, clicking the add application button on a listing in detroit that has a community type added results in a page not found error. This is because of a missing null check on line 57 of the FormMultiselectQuestions.tsx file. This is an easy fix but want to check if there are other ways we want to support the community type program swap that is happening in Detroit.

### How can you reproduce the bug?

Click the add application button on a listing in bloom-detroit that has a community type added results in a page not found error

### Is there any additional context?

-This is coming from seeding based on dev data

### What browsers are you seeing the problem on?

Chrome

### Where are you seeing this issue?

Desktop

QA Notes: https://bloom-detroit-partners.netlify.app/sign-in
Add a paper application to a listing with a community type added. Note: There will still be no options since that is a DB change

Issue #5040, opened 2025-07-03
[Bug]: Application Flow Blocked when No Languages


ColinBuyck:
### What is the issue?

If a listing's jurisdiction has no languages or one language set in the db, the choose language page of the application will show no buttons to continue on to the application. These cases should be handled rather than blocking the flow entirely.

![Image](https://github.com/user-attachments/assets/cbbc5dd3-8bb0-4427-a96c-351d85316957)

### How can you reproduce the bug?

1. Set a jurisdiction's language field to {} or {es}
2. Create a listing connected to that jurisdiction and enable common app
3. Try to apply to the listing and see the view above

### Is there any additional context?

QA will be handled by @ColinBuyck since it will require db manipulation.

ColinBuyck:
QA completed while pairing with @skeller-exy and discussion led to the creation of #5082

Issue #4991, opened 2025-06-11
Implement Plan for Common App to Support Non-Required Fields


ColinBuyck:
Currently, Detroit has no required fields and that interrupts the expected application flow and the submission verification at the end of the application. This ticket is to identify the data expectations of the common application and make a plan for handling the minimum required fields for our current application.

Possible Outcomes:
- Only unit groups required for application so that the unit application page would be skipped
- Set of fields required and are easier handled on the application by skipping pages
- Set of fields required and are unique cases that would be easier handled by making them required on the listings submission

OUTCOME Revisions:
Per https://github.com/bloom-housing/bloom/pull/4942, updating the verify endpoint or dtos shouldn't be done custom in this work but instead leverage this PR. As a result, the focus should be on determining a list of fields required for the FE experience and check for value defaulting and ensuring that no misleading values are set.

Technical notes:
- Known case: The unit page of the application expects some unit groups to be filled out
- Maintain communication with product to keep alignment on what changes will be made to the application or listing flow such that things can be communicated back to the client

AC:

- Document required field findings
- Successful application submission no matter the listing information

QA Notes: https://bloom-detroit.netlify.app/
1. Fully submit a common application with a listing with unit groups and no unit groups and ensure there is no "income is too low" error on the income step

PR #5055, opened 2025-07-07
fix: add null checks

ColinBuyck:
This PR partially addresses #4991 by fixing #5041 and #5040 

## Description

This PR adds three different null checks based on common application flow testing.

1. Currently the community types (programs under the hood) from Detroit prod have no options set. This will need to be handled by updating the data but as a partial step, I added a null check so the paper application form can load even with these listings attached to them.
2. Detroit prod data had {en} set for the languages field on the jurisdiction table. This resulted in no options appearing on the choose-language.tsx page. This check allows for one button to show.
3. Common app flows with unit groups were failing on the income step because of the verify income logic. This PR checks for units before verifying the income. I also made a follow up ticket [here](https://github.com/bloom-housing/bloom/issues/5056) to explore adding income verification for unit groups but that is not easily done right now with the fields that unit groups collects.

NOTE: I am still testing other parts of the common application flow but these changes should help free up testing and allow for common app submissions.

This can be tested by 

1. Delete the options on a program, add it to a Lakeview listing, try to add a paper app
2. Set the languages field on the jurisdiction table to be {en} and then try to apply to a listing in that jurisdiction
3. Try to submit a common app to a listing with unit groups or no unit and no unit groups and see if you can get past the income step

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/components/applications/PaperApplicationForm/sections/FormMultiselectQuestions.tsx
+++ b/sites/partners/src/components/applications/PaperApplicationForm/sections/FormMultiselectQuestions.tsx
@@ -54,7 +54,7 @@ const FormMultiselectQuestions = ({
   const allOptionFieldNames = useMemo(() => {
     const keys = []
     questions?.forEach((listingQuestion) =>
-      listingQuestion?.multiselectQuestions.options.forEach((option) =>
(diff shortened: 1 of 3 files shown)


emilyjablonski (approved):
🐣

Issue #4822, opened 2025-04-16
Rent Range Validation


ColinBuyck:
- [x] Public

On the filter drawer, the min and max rent should validate **on blur** to ensure no negative rent is entered or the min rent is not greater than the max rent. The error messages are seen in the old detroit but should circle back with design for exact desired messaging.

** Please provide as much information as possible (screenshots, documentation, etc) **

This was attempted by matching the Detroit pattern but it is not working in the Detroit site as intended and it will require a broader change to the currency field to get it to work onBlur. The currency field already used onChange, onBlur, and onFocus functions in its default implementation.

Alternatively, we could only validate on submission if onBlur proves to be too difficult.

ColinBuyck:
Time box to 1-2 hours for investigation and full implementation

KrissDrawing:
QA Notes: On https://bloom-detroit.netlify.app/listings click `filter`.
- It should clear both errors on blur both when typing correct value or removing one of them.

Writing that QA Notes i realised that we have similar min/max rent fields in https://bloom-detroit.netlify.app/finder 
If that was part of this ticket i did not implemented it :)
@sarahlazarich @ColinBuyck

sarahlazarich:
Ah @KrissDrawing - This looks good to me in the drawer, but we should also fix in the rental finder. Can we do that as a part of this or do we need to be separate?

sarahlazarich:
this looks good

PR #5057, opened 2025-07-08
fix: trigger error on blur for min max rent

KrissDrawing:
This PR addresses #4822

## Description

Adds validation for rent fields. Task also mentioned negative rent values, but i couldn't input those, so just ignored error for that case. LMK if you find some tricky way to have negative values

on `/http://localhost:3000/listings` click `filter`. then check rent validation on field blur + clearing errors

## Author Checklist:

## Review Process:


sites/public/src/components/browse/FilterDrawerHelpers.tsx
```diff
@@ -181,56 +184,96 @@ export const CheckboxGroup = (props: CheckboxGroupProps) => {
...
-  </fieldset>
-)
+export const RentSection = (props: RentSectionProps) => {
+  const validateRentValues = () => {
+    const minValue = props.getValues(`${ListingFilterKeys.monthlyRent}.minRent`)
+    const maxValue = props.getValues(`${ListingFilterKeys.monthlyRent}.maxRent`)
+
+    if (!minValue || !maxValue) return
+
+    const numericMinValue = parseFloat(minValue.replaceAll(",", ""))
```
emilyjablonski, line 194: **question:** Do we also need to check for `.`?

ColinBuyck (changes requested):
It's looking good, just found a little bug around clearing errors. It looks like if the error is triggered deleting one of the fields doesn't clear the error on the other field and it forces you to change that value/or at least enter focus there to proceed.

https://github.com/user-attachments/assets/cf20b7bd-3a32-4ac3-843f-64e87f1ee291

ColinBuyck (approved):
LGTM ❄️ Thanks for the speedy update

Issue #5049, opened 2025-07-07
Testing Bug - HMI Section Not Displaying


sarahlazarich:
Steps to Reproduce: 
- Compare [this listing](https://bloom-detroit.netlify.app/listing/6e11d372-edc8-4570-be57-7485884e3ac9/marwood_townhomes_67_77_marston_detroit_mi) on our replatformed site to the one [in production](https://homeconnect.detroitmi.gov/listing/6e11d372-edc8-4570-be57-7485884e3ac9/marwood_townhomes_67_77_marston_detroit_mi) 

Expected Behavior: 
- The listing information for both should match 

Observed Behavior: 
- The production version does not show an HMI table but ours does 

Notes: 
- This seems like it is actually a bug in prod and we should be displaying the HMI 
- We should verify that there isn't any underlying logic that is hiding this or any reason why the HMI might be being shown dynamically.  
-  There is another example [here](https://bloom-detroit.netlify.app/listing/6da133f6-7fe0-434c-8b4d-e52620590546/shelborne_square_apartments_8931_mount_elliott_street_detroit_mi) where it does show on our version and [prod’s version](https://homeconnect.detroitmi.gov/listing/6da133f6-7fe0-434c-8b4d-e52620590546/shelborne_square_apartments_8931_mount_elliott_street_detroit_mi) but the data is just clearly bad, as it only shows 1 row and 1 person occupancy up to 3BR. Can we verify why the data is being pulled this way?

emilyjablonski:
QA Notes
In Lakeview, create a listing with a unit group with an AMI chart and AMI percentage, but do not fill in any min/max occupancy information. Ensure the HMI table does not appear on the public site.

skeller-exy:
A/C
- Neither [this](https://bloom-detroit.netlify.app/listing/6e11d372-edc8-4570-be57-7485884e3ac9/marwood_townhomes_67_77_marston_detroit_mi) listing on our replatformed site nor [this](https://homeconnect.detroitmi.gov/listing/6e11d372-edc8-4570-be57-7485884e3ac9/marwood_townhomes_67_77_marston_detroit_mi) listing in production display the hmi chart ✅
- The following listings with unit group combinations **do not display** the hmi table
    - unit group with ami chart + ami percentage and no min or max occupancy specified  ✅
    - unit group with min/max occupancy specified and no ami chart ✅
- The following listings with unit group combinations **do display** the hmi table
[...]

PR #5059, opened 2025-07-08
fix: show hmi ranges if multiple ami charts are used

emilyjablonski:
This PR addresses #5049 and #5070

## Description

Two related fixes in the HMI table:
* For #5049 we have a listing in our replatformed Detroit that is showing a single-row HMI table where Detroit isn't showing one at all. This is because the way that we are calculating min/max occupancy has a default value of 1 where in Detroit it is null. We end up with an HMI table with one row but it should be empty, because the listing did not enter any min/max occupancy numbers on its unit groups.
* For #5070 we have a bug both in our replatformed site and our production site where listings that have units or unit groups that use multiple AMI charts are supposed to show ranges in the HMI table but are always showing the maximum value

You will have to test locally due to the backend changes.

* For 5049, in Lakeview, create a listing with a unit group with an AMI chart and AMI percentage, but do not fill in any min/max occupancy information. Ensure the HMI table does not appear on the public site.
* For 5070: (1) In Bloomington, create a listing with 2 units that have two different ami charts and the same percentage, and ensure ranges appear on the public site (2) In Lakeview, create a listing with 2 unit groups that have two different ami charts and the same percentage and ensure ranges appear on the public site

New charts were added into the seed that should help testing this locally

*** This will also affect production with seeds off ***
Screenshot 2025-07-10 at 4 40 41 PM

## Author Checklist:

## Review Process:


api/src/utilities/unit-groups-transformations.ts
```diff
@@ -127,42 +147,39 @@ export const getHouseholdMaxIncomeSummary = (
...
-  const amiChartItems = getMaxIncomeAmiChartItems(amiCharts);
-
   let occupancyRange: MinMax;
   const amiPercentages = new Set<number>();
-  // aggregate household sizes across all groups based off of the occupancy range
+
   unitGroups.forEach((group) => {
     if (occupancyRange === undefined) {
       occupancyRange = {
-        min: group.minOccupancy || 1,
```
emilyjablonski, line 145: **note:** This was causing occupancy to always to have at least a min and max of 1, even if no occupancy data was entered. That meant the HMI table was always showing with at least 1 row for a household size of 1. That is not the behavior of current Detroit prod. In Detroit, if there is no occupancy information listed for unit groups, we don't show the HMI table at all.

api/src/utilities/unit-utilities.ts
```diff
@@ -305,19 +305,25 @@ export const generateHmiData = (
               max: usd.format(firstChartValue),
             } as MinMaxCurrency,
           );
+
+          const yearRange =
```
emilyjablonski, line 309: **note:** This way of sending this data with the special characters is so strange, but it would be a big rewrite to change it, and that's already on our radar. All this does is include our already-calculated ranges in the string.

api/prisma/seed-helpers/ami-chart-factory.ts
```diff
@@ -4,24 +4,28 @@ import { randomName } from './word-generator';
 export const amiChartFactory = (
   numberToCreate: number,
   jurisdictionId: string,
+  offset?: number,
```
emilyjablonski, line 7: **note:** This allows us to generate an ami chart with more than just one set of values

api/prisma/seed-staging.ts
```diff
@@ -833,7 +839,7 @@ export const stagingSeed = async (
               monthlyRentDeterminationType:
                 MonthlyRentDeterminationTypeEnum.flatRent,
               flatRentValue: 1400.0,
-              amiChart: { connect: { id: amiChart.id } },
+              amiChart: { connect: { id: lakeviewAmiChart.id } },
```
emilyjablonski, line 842: **note:** We weren't associating ami charts in lakeview with those in the right jurisdiction

sites/partners/__tests__/components/listings/PaperListingForm/UnitGroupForm.test.tsx
```diff
@@ -513,12 +505,7 @@ describe("<UnitGroupForm>", () => {
       await userEvent.selectOptions(maxFloorInput, "1")
     })
 
-    expect(
-      await screen.findByText(/minimum floor must be less than or equal to max floor/i)
-    ).toBeInTheDocument()
-    expect(
-      await screen.findByText(/max floor must be greater than or equal to minimum floor/i)
-    ).toBeInTheDocument()
+    expect(await screen.findByText(/floor must be greater than or equal to/i)).toBeInTheDocument()
```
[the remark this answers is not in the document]
emilyjablonski: This for some reason ended up being fairly flaky for me - but it is working in the browser, and it's just that one or the other isn't there for the assert. I tried a bunch of stuff around the async bit of selecting the value and waiting for the query, but I think if we see either error exists, bc the logic is the same we should be good to go, but I'm open to continuing to investigate. I would personally prefer that in a separate ticket just bc it was taking up more time than felt valuable in this PR context!

api/src/utilities/unit-groups-transformations.ts
```diff
@@ -97,18 +100,35 @@ export const getMaxIncomeAmiChartItems = (
...
+
     if (existingItemIndex === -1) {
+      currentItem.incomeString = usd.format(income);
       result.push(currentItem);
-    } else if (currentItem.income > result[existingItemIndex].income) {
+    } else {
+      const newIncome = result[existingItemIndex].income;
+
+      if (income > newIncome) {
+        currentItem.incomeString = `${usd.format(newIncome)} - ${usd.format(
```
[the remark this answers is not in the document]
emilyjablonski: Honestly I'm not sure! Could be a good question for design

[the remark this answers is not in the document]
emilyjablonski: I don't think we do localization elsewhere, so I might be part of a larger localization convo

Issue #4717, opened 2025-03-18
Toggle required fields for publishing listing


ludtkemorgan:
This is the implementation of the required field changes as outlined here in #4630 and written up [here](https://docs.google.com/document/d/1dqoYLVeT6YCaQTUiA601v0fA3DJ4y5UraI5stViZF0E/edit?tab=t.0) 

## Background
In order to accommodate Detroit we need to allow listings to go live with less fields required. The core repo has a very strict set of fields that it expects when publishing a listing. These fields align closely with the expected fields in HBA and Doorway jurisdictions which is still a list we would like to maintain. With our push to not make any backend changes in jurisdiction repos we will need to make the required list dynamic so that both the backend and the frontend can handle the different jurisdictions.

### backend
This will require a refactor of how we set up and use DTOs for listing creates/edits. Currently there are separate DTOs for all the scenarios (create, update, and publish).
#### database
* Add new field to the jurisdictions table called `requiredListingFields`
  * value of String[]
  * Note this field will need to be exposed to the partner site when retrieving jurisdictions.
#### New Decorator
We need to have a validator that bypasses the other validation if the field is not in the required field list, is not trying to be an "active" listing, and there isn't anything in the field. This can be done with the following.
* Create a new decorator that utilizes `ValidateIf` validator called `ValidateListingPublish`
* [Sample version of it](https://github.com/bloom-housing/bloom/blob/4630/validate-required-fields/api/src/decorators/validate-listing-publish.decorator.ts)
#### DTOs
* Update `listing.dto.ts` to have optional field `requiredFields` with value array of strings
* Move all additional validation for each field from `listing-published-update.dto.ts` to the equivalent field on both `listing-update.dto.ts` and `listing.dto.ts`
  * examples: MaxLength, EnforceLowerCase, IsDefined
* Delete `listing-published-update.dto.ts` and `listing-published-create.dto.ts`
* Add new decorator to every field that may be required to `listing-update.dto.ts` and `listing.dto.ts`
  * TODO: add details here
#### Update listing-create-update-pipe.ts
* [Sample update](https://github.com/bloom-housing/bloom/blob/4630/validate-required-fields/api/src/validation-pipes/listing-create-update-pipe.ts) (note just the "transform function" is required
* TODO: add additional details
#### Update edit/create listing service
* With the `requiredFields` on the listing dto it will need to be removed from the object before attempting to save it

The partner site changes are pulled out to a separate ticket (ticket goes here)
~### Partner site~
[...]

PR #5071, opened 2025-07-10
fix: application due date work correctly validates

ludtkemorgan:
This PR addresses #4717

## Description

While pulling the #4717 changes into HBA I noticed there was a missed validation around application due date.

There is conditional validation on the applicationDueDate field for listings. It only is validated as required if the listing is not of review order type "waitlist". This was done via a `ValidateIf` on the dto. We should have this carried over for the new dynamic required fields

This validation will only get triggered if the `applicationDueDate` is listed in `requiredListingFields` so the following scenarios should be tested

Jurisdiction with `applicationDueDate` in `requiredListingFields`
1. Save as a draft a listing with `applicationDueDate` filled out should save correctly
2. Save as a draft a listing with `applicationDueDate` not filled out should save correctly
3. publish a listing with `applicationDueDate` not filled and select "waitlist". Should save correctly (assuming you have all other requiredListingFields filled out)
4. publish a listing with `applicationDueDate` not filled and select "open units". should fail on validation of the applicationDueDate field

jurisdiction with `applicationDueDate` *not* in `requiredListingFields`
1. publish a listing with `applicationDueDate` not filled and select "waitlist".  should save correctly (assuming you have all other requiredListingFields filled out)
2. publish a listing with `applicationDueDate` not filled and select "open units". should save correctly (assuming you have all other requiredListingFields filled out)

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/dtos/jurisdictions/jurisdiction.dto.ts
+++ b/api/src/dtos/jurisdictions/jurisdiction.dto.ts
@@ -138,7 +138,7 @@ export class Jurisdiction extends AbstractDTO {
   featureFlags: FeatureFlag[];
 
   @Expose()
-  @ValidateNested({ groups: [ValidationsGroupsEnum.default] })
+  @IsString({ groups: [ValidationsGroupsEnum.default], each: true })
   @IsDefined({ groups: [ValidationsGroupsEnum.default] })
   @ApiProperty({ isArray: true })
   requiredListingFields: string[];
--- a/api/src/dtos/listings/listing.dto.ts
+++ b/api/src/dtos/listings/listing.dto.ts
@@ -210,6 +210,16 @@ class Listing extends AbstractDTO {
   @ValidateListingPublish('applicationDueDate', {
     groups: [ValidationsGroupsEnum.default],
   })
+  @ValidateIf(
+    (o) =>
+      !(
+        o.applicationDueDate == undefined &&
+        o.reviewOrderType == ReviewOrderTypeEnum.waitlist
+      ),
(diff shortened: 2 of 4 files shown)


emilyjablonski:
I think I have more of philosophical question around this - it seems like it's necessary, but feels somewhat odd to include a listing in required fields that is not actually required in all cases. I'm not sure how it's going to affect the frontend implementation of showing which fields are required - j curious if you have thoughts about that!

ludtkemorgan:
@emilyjablonski apparently my comment didn't save so sorry about the late response.

My thought here is that the backend needs to be dumb and just decide if something is required for a jurisdiction based on the one specific thing. So in our case the "requiredListingFields" field. And then any other field specific validation is done independently. Those specific validations are always separate so we can get more nuanced and they will be the same for every jurisdiction. We only offer the jurisdiction the ability to turn required on/off for the field and not add all of the additional logic specific for that jurisdiction. Maybe one day we will need to support that use case.

I also have some concerns about what this means for the frontend. Mostly in around how fields that have extra validation we need to add that to the frontend as well. For this field we will only display the required asterisk if it's in the required list and the user has selected open waitlist. There are probably a couple of other cases where some business logic will be on both the backend and frontend.

Lastly, this field is going to be required for Doorway and HBA and not Detroit so we want it to be toggleable.

ColinBuyck (approved):
LGTM 🥼 I read your response Morgan and that makes sense to me! Approving since I know people are swamped and reviews are a bit backed up atm

PR #5100, opened 2025-07-17
fix: multi image modal theme

emilyjablonski:

## Description

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/styles/overrides.scss
+++ b/sites/public/styles/overrides.scss
@@ -150,6 +150,26 @@
     }
   }
 
+  // These are temporary manual overrides for the UIC multi image modal to enable it to theme with Seeds while the new Seeds multi image modal is still in development
+  .image-card__gallery-modal {
+    --bloom-color-primary: var(--seeds-color-primary);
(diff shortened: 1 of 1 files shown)


sites/public/styles/overrides.scss
```diff
@@ -164,4 +164,31 @@
       --button-border-radius-md: var(--seeds-rounded-3xl);
     }
   }
+
+  // The multi-image card overrides below exist outside of the jurisdictional class. This allows it to still get the lakeview overrides - if the class does exist, then apply the same overrides to its sibling.
+  .lakeview + div {
+    .image-card__gallery-modal {
+      @extend .lakeview;
+    }
+  }
```
ludtkemorgan, line 173: question: Will we have to do this for all of our jurisdictional repos? Or is this just the case because we are adding lakeview overrides via a class?

ludtkemorgan (approved):
Looks good. Just a question

sites/public/styles/overrides.scss
emilyjablonski: This is just for the lakeview class, it works in Detroit without this

Issue #5038, opened 2025-07-03
Required for listing publishing fields on partner site


ludtkemorgan:
- [ ] General
- [ ] Backend
- [ ] Public
- [x] Partners

## Background / Issue
The backend logic to toggle required fields has been completed via #4717. This allows jurisdictions to have different fields as being required for publication of listings. For example HBA listings require images but Detroit does not. In order to fully communicate these differences to the end user the parter listing create/edit page needs to display what fields are required. This can be done via a sub-not "required to publish"

![Image](https://github.com/user-attachments/assets/c2fbd994-c07d-441c-bf72-22f2a31bc537)

## Technical Context
The backend changes now provide a new field on the jurisdiction called `required_listing_fields` that have a list of the fields that are required. We already get the jurisdiction data when editing a listing so if this field doesn't already get returned by the API we can add it to the call already being done. 

The items in the list match the name/id of each field. Each field should have a helper class to dynamically display the "required to publish" if the id of the field is in the `requiredListingFields` list.

Questions:
* If a jurisdiction doesn't explicitly set this field we use a default list setup in `listing-create-update-pipe.ts` meaning the frontend won't know what these fields are and a solution will need to be made
  * Option 1: also have this default list set on the frontend and if no jurisdiction is selected or the selected jurisdiction does not have the field just use that list
  * Option 2: update the backend to instead of using this list put a default list for the field in the database
* Are there any fields that have additional frontend validation? If so, how do we handle these dynamically?

## Acceptance Criteria
* After selecting a jurisdiction on listing create or edit all required fields should display a sub note "required to publish"
## QA Notes

** Please provide as much information as possible (screenshots, documentation, etc) **

emilyjablonski:
@ludtkemorgan I'm definitely happy to make changes so that radios get default values, but I'm wondering if that feels like a separate ticket

skeller-exy:
# Testing Notes

- No jurisdiction selected ✅
  - shows * to indicate required field ✅
  - validates missing data in required field ✅
- Lakeview jurisdiction selected ✅
  - shows * to indicate required fields (listing name, jurisdiction, street address, city, state, zip code) ✅
  - validates missing data in required fields ✅ (see styling issue below ⚠️)
  - can publish listing with only required fields filled out ✅
- Bloomington jurisdiction
  - shows * to indicate required fields ✅
  - validates missing data in required fields ✅ 
  - front end validation for sub-required fields (unit type, ami chart, etc) ✅
[...]

PR #5101, opened 2025-07-18
feat: frontend required fields

emilyjablonski:
This PR addresses #5038

## Description

This PR is the frontend for allowing for limited set of fields on the listing to be dynamically configured as required to publish.

Happy to do a PR walkthrough as this is a larger PR. I've separated out two pieces of this that I think could be better as followups to try to keep this PR smaller (backend cleanup of required field defaults, setting a default value on radios that don't have one).

This query will set as required all of the possible fields that can be configured with the `requiredFields` set:
`update jurisdictions set required_listing_fields = ARRAY['listingsBuildingAddress','name','developer','listingImages','leasingAgentEmail','leasingAgentName','leasingAgentPhone','jurisdictions','units','digitalApplication','paperApplication','referralOpportunity','rentalAssistance','neighborhood', 'yearBuilt', 'reservedCommunityTypes', 'reservedCommunityDescription', 'communityDisclaimerTitle', 'disableUnitsAccordion', 'homeType', 'applicationFee', 'depositMin', 'depositMax', 'depositHelperText', 'costsNotIncluded', 'amenities', 'accessibility', 'unitAmenities', 'smokingPolicy', 'petPolicy', 'servicesOffered', 'creditHistory', 'rentalHistory', 'criminalBackground', 'requiredDocuments', 'programRules', 'specialNotes', 'whatToExpect', 'leasingAgentTitle', 'managementWebsite', 'leasingAgentOfficeHours', 'listingsLeasingAgentAddress', 'additionalApplicationSubmissionNotes', 'applicationDueDate'] where name = 'Bloomington'`

Test the Lakeview jurisdiction, and the Bloomington jurisdiction after running the above query, to see the difference in which fields are required to publish on the listing form.

Why other fields are not included is documented [here](https://docs.google.com/spreadsheets/d/1QCrMPGHca6YLG-Oo0IUXsxAKQRM4Cm93vLa8B7zAbos/edit?gid=0#gid=0) and there are subtasks made under this ticket.

Other changes in this PR:
* Removing placeholder text (intentional, more context in this [loom](https://www.loom.com/share/38334c88c198482eb453a9d5df49a509?sid=8cdc827d-6fa8-48f1-a17f-b434a51cb19b))
* We had a variety of label sizes and colors - this standardizes those
* The `<FieldValue>` component was often wrapping fields instead of using the existing field prop. This means there were form fields that had no labels at all and instead used a `<p>` tag. I'm not clear why we were doing that and it's an a11y issue, so some FieldValue components were removed, to instead use the baked in input label.

## Author Checklist:

## Review Process:


sites/partners/src/components/listings/PaperListingForm/ListingForm.module.scss
emilyjablonski, line 1: **note:** This UIC PR will allow me to remove most of this https://github.com/bloom-housing/ui-components/pull/173

sites/partners/__tests__/components/listings/PaperListingForm/sections/CommunityType.test.tsx
```diff
@@ -154,14 +154,12 @@ describe("CommunityType", () => {
     await act(() => userEvent.click(screen.getByRole("radio", { name: "Yes" })))
 
     expect(
-      screen.getByRole("textbox", { name: "Reserved Community Disclaimer Title" })
+      screen.getByRole("textbox", { name: "Reserved Community Disclaimer Title *" })
```
ludtkemorgan, line 157: question: Is there an easy way to write a test that covers adding the asterisk to all of the fields? I know the listing form doesn't have full test coverage so it could be a lot of work.

ludtkemorgan (approved):
This is looking good. I haven't found issues so approving it. Just a couple of logistical questions:

Have you documented what fields are remaining for validation? If so can you add it to the description so other reviewers know which fields to not look at. And if not do you have a follow-up ticket created?

For the fields that only appear if some frontend only field is selected do you think we could just make those frontend validations? For example if you select "Is there a paper application?" to be "yes" shouldn't it always be required that a paper application is added?

Lastly, do you think this can be released without the remaining changes completed?

emilyjablonski:
@ludtkemorgan I think this can be released without have every question be configurable as required because this set covers the use cases we have in all our current jurisdictions. [Here is a spreadsheet](https://docs.google.com/spreadsheets/d/1QCrMPGHca6YLG-Oo0IUXsxAKQRM4Cm93vLa8B7zAbos/edit?gid=0#gid=0) that documents which fields are currently configurable, and which ones aren't and why. I added three sub-issues to the [parent](https://github.com/bloom-housing/bloom/issues/5038) that break out three of the separate cases where it was harder to make fields configurable.

mcgarrye:
"applicationDueDate" is included in the file of togglable fields and in the test file but not in the query in the description. Is there a reason for this?

Issue #5080, opened 2025-07-15
Detroit Bash 7/14


ColinBuyck:
- [x] General
- [ ] Backend
- [ ] Public
- [ ] Partners

### Bash PR
1. In Detroit, there is a default image that is shown if a listing image is not uploaded
2. Unit group validation is not clearing once the issue is corrected in Firefox. Since this issue is existing in the unit drawer (in addition to the unit group drawer and only in Firefox, I pulled it out [here](https://github.com/bloom-housing/bloom/issues/5112) to be handled at a later time)
<img width="1139" height="983" alt="Image" src="https://github.com/user-attachments/assets/253edd3f-567e-42fb-82ea-218f15c80c8b" />
3. Core text style in the "You" header on the application review page
<img width="535" height="372" alt="Image" src="https://github.com/user-attachments/assets/9eab4d03-cdb3-4b47-abe2-4e38fb88f6b3" />

4. On a listing with an external application and mandated accounts enabled, the apply online button shows as if its a common application, you sign in and then it redirects you back to the listing page where you see the external apply online button... you should not be forced to sign in if it is an external application

### Need Further Support from Product @sarahlazarich 
1. How did you hear options has reference to AC -consulting with product on this one
2. FCFS issues - text on what to expect for a listing, terms page, application card subtext... this is based off of core logic that looks at the review order (which is only fcfs / lottery / waitlist in core) but we don't use the review order in the same way for Detroit. While a waitlist status in core is set based on that one toggle, in Detroit it’s whether or not unit groups have a waitlist so we can't rely on the review order logic consistently (cc @emilyjablonski if there's more detail you'd like to add!) In order to be able to rely on the review order when unit groups are on, we could make a change to the way that field is saved on the backend, and set the status as waitlist if a listing has unit groups with a waitlist. Right now, the review order is FCFS on what should be waitlist listings if we are using unit groups.

Examples of FCFS references
<img width="525" height="692" alt="Image" src="https://github.com/user-attachments/assets/aa3f340e-be40-4de6-afc1-ebda1b1a1e83" />

**(note that the image above is also missing translations but that is related to @jaredcwhite work on #4601)**

<img width="527" height="582" alt="Image" src="https://github.com/user-attachments/assets/bceffa28-f09f-407c-ba42-5a45da2d765c" />
<img width="529" height="693" alt="Image" src="https://github.com/user-attachments/assets/0b4f0b19-258b-453f-9e60-7b1e43c130f3" />

## Technical Context
[...]

PR #5104, opened 2025-07-21
feat: bash fixes 1

ColinBuyck:
This PR addresses #5080 

## Description

This PR addresses the following points in the issue:

1. The fallback image is now utilized. There was a typo creating this bug.
3. The styling is fixed by removing the unnecessary tailwind which allows the overrides to apply correctly. NOTE that this slightly changes the visuals of the core experience but since anything seeds related should be using sans serif and I assume we'd want the h3 headers to align in styling rather than create a custom one off styling for the application, I made the update
You is a example of old styling, Household Details is an example of new styling
Screenshot 2025-07-21 at 8 21 34 AM

4. Mandated accounts should not force a user to sign in if the application is not a common app. This is achieved by pulling a doorway PR back to core (with slight tweaks) https://github.com/metrotranscom/doorway/pull/593

In Lakeview

Point 1. Create a listing with no listing image and ensure the fallback image shows on the listing browse page
Point 3. See that the h3 styling in the application summary view matches that of the other h3 headers across the site. (Personally tested in detroit and it worked as desired with the overrides)

Point 4. (with mandated accounts enabled) Try to apply to a listing with the common app (while signed out) and it should redirect, a listing with the common app (while signed in) and it should not redirect, and an external online application and it should not redirect

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/src/utilities/photos.ts
+++ b/shared-helpers/src/utilities/photos.ts
@@ -36,5 +36,5 @@ export const imageUrlFromListing = (listing: Listing, size = 400): string[] => {
         : asset.fileId
     })
 
-  return imageUrls?.length > 0 ? imageUrls : ["/images/fallback-listing.png"]
+  return imageUrls?.length > 0 ? imageUrls : ["/images/listing-fallback.png"]
 }
--- a/sites/public/__tests__/components/listing/listing_sections/Apply.test.tsx
+++ b/sites/public/__tests__/components/listing/listing_sections/Apply.test.tsx
@@ -1,5 +1,6 @@
 import React from "react"
 import dayjs from "dayjs"
+import { screen } from "@testing-library/dom"
 import { render, cleanup, fireEvent } from "@testing-library/react"
 import {
   ApplicationMethodsTypeEnum,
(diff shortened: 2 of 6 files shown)


shared-helpers/src/utilities/photos.ts
```diff
@@ -36,5 +36,5 @@ export const imageUrlFromListing = (listing: Listing, size = 400): string[] => {
         : asset.fileId
     })
 
-  return imageUrls?.length > 0 ? imageUrls : ["/images/fallback-listing.png"]
+  return imageUrls?.length > 0 ? imageUrls : ["/images/listing-fallback.png"]
```
emilyjablonski, line 39: **note:** oof

sites/public/__tests__/components/listing/listing_sections/Apply.test.tsx
emilyjablonski, line 1: **praise:** love it, stunning, fabulous

sites/public/src/components/listing/ListingView.tsx
```diff
@@ -373,15 +373,18 @@ export const ListingView = (props: ListingProps) => {
 
   const getOnlineApplicationURL = () => {
     let onlineApplicationURL
+    let isCommonApp
```
emilyjablonski, line 376: **nitpick:** you can do `let isCommonApp = false` and then you don't need 380

sites/public/src/components/listing/ListingView.tsx
emilyjablonski, line 1: **question:** this is in the deprecated file, just confirming that is that intended? i presume so bc it's a bug fix we're pulling over

emilyjablonski (approved):
just v small comments, lgtm!

sites/public/src/components/listing/ListingView.tsx
ColinBuyck: Just confirming this update but I removed 385 rather than 380... lmk if that wasn't a typo @emilyjablonski

sites/public/src/components/listing/ListingView.tsx
ColinBuyck: Exactly!

Issue #5089, opened 2025-07-16
Testing Bug - Filter Options


sarahlazarich:
Steps to Reproduce: 
- Compare the unit size options in the Unit Creation drawer on the partners site to the replatformed public site. 

Expected Behavior: 
- These two options should match 
- For SRO, this should be toggled in the filter drawer based on the enableUnitGroups flag. (Hide if unit groups is on)

Observed Behavior: 
- They do not match 

<img width="1411" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN1cvQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--166205223c26c920c381bf2f7766cfe8b2720b99/image.png" alt="image.png" />

<img width="1617" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2EvQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--98408c080081034f8adeac8dcc158265d34baac1/image.png" alt="image.png" />

Note: 

I think ideally, we should match existing production.  Did we decide there was a toggling limitation there?  
<img width="957" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2kvQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--28550c604b2b910434b3ddc407d5fb3c0cbd8958/image.png" alt="image.png" />

<img width="484" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN20vQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9c4145c00310c425bc030569143f5537d63320c7/image.png" alt="image.png" />

sarahlazarich:
also expect 5 br option to be removed per https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/5177

sarahlazarich:
lgtm

Issue #73, opened 2019-10-19
Add API tests for the listings service


bk3c:
Not for immediate work given the current amount of flux in the API, but in the future we should seek to have (I'm assuming) Jest tests of the service APIs directly, and then move the front-end tests to using fixed static inputs. Less value in the short term since we're using fixed static inputs for real, but as soon as we're not we'll need tests.

software-project:
Initial part of this one is done @bk3c so I think we can close it.

PR #5114, opened 2025-07-22
fix: testing bug filter options

KrissDrawing:
This PR addresses #5089 

## Description

Will use Detroit unit types for filters (and finder) on public site.
For partners we will be able only to add detroit related unit types (technically it populates unit types from backend so with production data it should work before for new detroit, but this one will adapt for switching jurisdictions with different `enableUnitGroups` flag)

For `enableUnitGroups` it should work as before.
With enabled on public (so for lakeview)
- filter drawer should show unit types from current detroit page
- same for `/finder`
- on partners for listing with unit groups it should have new unit types (in unit groups drawer) -> if we switch to jurisdiction with enableUnitGroups = false it should show current unit types for units

## Author Checklist:

## Review Process:


ludtkemorgan (changes requested):

ludtkemorgan:
@KrissDrawing the error is during the building step
```
5:17:32 AM: ../../shared-helpers/src/utilities/unitTypes.ts:29:52
5:17:32 AM: Type error: Argument of type 'string' is not assignable to parameter of type 'UnitTypeEnum'.
5:17:32 AM:   27 |   if (!units) return []
5:17:32 AM:   28 | 
5:17:32 AM: > 29 |   return units.sort((a, b) => UnitTypeSort.indexOf(a.name) - UnitTypeSort.indexOf(b.name))
5:17:32 AM:      |                                                    ^
5:17:32 AM:   30 | }
5:17:32 AM:   31 | 
5:17:32 AM:   32 | export const getUniqueUnitTypes = (units: Unit[]): GetUnitTypeNamesReturn[] => {
```

KrissDrawing:
@ludtkemorgan  Just curious if i should align this to https://github.com/bloom-housing/bloom-detroit/pull/73 .
I think it will work fine (but there will be additional `listings.unitTypes.expanded.fourPlusBdrm` key).

ludtkemorgan:
> @ludtkemorgan Just curious if i should align this to [bloom-housing#73](https://github.com/bloom-housing/bloom-detroit/pull/73) . I think it will work fine (but there will be additional `listings.unitTypes.expanded.fourPlusBdrm` key).

@mcgarrye can you address this question? Looks like it's tackling a similar thing that you just worked on

mcgarrye:
Was not made aware of this ticket before tackling the work in #73. You can remove all changes to copy and tests with regard to `4+` as it should only exist within the Detroit repo. The key will remain the same, only the copy will change

Issue #5075, opened 2025-07-14
[Bug]: No images on listing not displaying correctly


ludtkemorgan:
### What is the issue?

There are two issues related to the scenario when there are no images tied to a listing.

1. (This is on the seeds version of the browse page) When you don't add an image to a listing and publish the listing rather than showing the fallback image there is a blank space where the image should be
<img width="1038" height="300" alt="Image" src="https://github.com/user-attachments/assets/c3da8dfd-f454-42c3-91ed-8104e78179bd" />
2. If there are no `listingImages` or `assets` tied to the listing a 500 is returned and the entire listings browse and listing details page doesn't work

### How can you reproduce the bug?

1. Save and publish a listing in the [new Detroit partner site](https://bloom-detroit-partners.netlify.app/) without any images attached
2. Go to the listing browse page on the public site and you won't see anything where the image should be

To reproduce the second issue you will need to manually edit the data to not have data

### Is there any additional context?

Slack thread: https://exygy.slack.com/archives/C0203460K6Y/p1752505124468509

Error logs:
```
ERROR  TypeError: imageAssets?.filter is not a function
    at imageUrlFromListing (/var/task/sites/public/.next/server/chunks/79.js:4031:34)
    at ListingCard (/var/task/sites/public/.next/server/chunks/7931.js:693:137)
```

If there are no images in the `listingImages` we fall back to `assets` due to the old way we stored images. We should update the code to handle the case where assets also does not exist

### What browsers are you seeing the problem on?

Chrome, Safari, Firefox

### Where are you seeing this issue?

Desktop, Mobile

PR #5125, opened 2025-07-23
fix: use listing assets only when array, refactor of photo helper

jaredcwhite:
This PR addresses #5075

## Description

If a listing gets published with no image, there was old fallback code to look at assets directly attached to a listing. The problem was that JSON value could be `{}` instead of `[]`. I'm not sure why even after spending some time poking around, so I simply add this Array type check. If we don't like that approach, an alternative might be to stop look at listings assets (is that even still a requirement?).

Save a new published listing with no image, then verify you can view it on public and see a fallback image placeholder.

## Author Checklist:

## Review Process:


jaredcwhite:
I added that new followup ticket here: https://github.com/bloom-housing/bloom/issues/5165

Issue #5035, opened 2025-07-01
Occupancy rows duplicated in unit groups


emilyjablonski:
[Example listing](https://bloom-detroit.netlify.app/listing/f8ba76f0-d26d-4ee1-973e-76c403e233fe/conner_creek_apartments_4661_outer_dr_e_detroit_mi)

When there are multiple unit groups with the same unit type, the occupancy table is duplicating the rows. The occupancy table should have one row per unit type.

<img width="667" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBLzZTQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--12707db68780354731cc4366f52d652011d3087b/Screenshot%202025-07-01%20at%204.22.04%E2%80%AFPM.png" alt="Screenshot 2025-07-01 at 4.22.04 PM.png" />

skeller-exy:
Sending this back to "In Development" based on discussion in today's standup.

PR #5140, opened 2025-07-24
Fix: Fix occupancy formatting for multiple unit groups with only max or min values

matzduniuk:
This PR addresses #5035 

## Description

Fixes and issues when formatting multiple unit groups with only max or min occupancy set would result in a `n/a` label

* Make sure the `enableUnitGroups` is enabled for the currently configured jurisdiction
* Go to the partners site dashboard and select any listing for the configured jurisdiction 
* Edit the listing by adding multiple unit groups with shared unit types by only max or min occupancy set for a value (example below)
Screenshot 2025-07-24 at 11 13 09
* Save the updated listing
* Go to public sites `/listings` page
* Click on the updated listing and scroll down to the "Occupancy" section
* The occupancy table should now show `at least <x> people` or `no more than <x> people` (result for the example below):
Screenshot 2025-07-24 at 11 16 46

## Author Checklist:

## Review Process:


shared-helpers/__tests__/occupancyFormatting.test.tsx
```diff
@@ -309,6 +337,27 @@ describe("occupancy formatting helper stacked table", () => {
             name: UnitTypeEnum.threeBdrm,
             numBedrooms: 3,
           },
+        ],
+      },
+      {
+        ...unitGroup,
+        maxOccupancy: 4,
+        minOccupancy: undefined,
```
mcgarrye, line 345: **nitpick:** Can the order of the fields be the same everyone for easier readability?

mcgarrye:
Screenshot 2025-07-28 at 4 40 20 PM
This seems a bit weird what I was able to do with Studio. Is that expected?

matzduniuk:
> Screenshot 2025-07-28 at 4 40 20 PM This seems a bit weird what I was able to do with Studio. Is that expected?

Technically, this is correct behavior, as each entry in unit groups with a different set of unit types is considered a separate entity; therefore, their ranges are also calculated separately. I worked with the already existing underlying logic for aggregation, so this should probably be consulted with someone who knows the target behavior

mcgarrye (approved):
LGTM

Issue #4913, opened 2025-05-30
Rental Finder A11y


ColinBuyck:
The updated and seedsified version of the rental finder doesn't currently include some of the a11y customizations made in partnership with LH. Particularly, the focus management that is done when switching between pages so that a user is reminded of the page change, their progress, and when changing sections vs steps.

KrissDrawing:
QA Notes: on https://bloom.exygy.dev/finder having `screen reader` on when we click `next` it should focus either on title when we switch to next step of section, or it should focus on section name when we switch between sections. 
When focusing on `minimum rent` and `maximum rent` fields it should have better voice description

PR #5221, opened 2025-08-06
fix: a11y improvements for finder

KrissDrawing:
This PR addresses #4913 

## Description

Adds finder a11y improvements from detroit page. It will be:
1. focus on card title on step change.
2. focus on section "title" on section change
3. adds aria label for rent fields

turn on screen reader and walk through `/finder`.

## Author Checklist:

## Review Process:


ColinBuyck (changes requested):
I'm noticing that the focus stays on the next button when the user changes from one page to another (within the same section). My guess is that we'll have to use the ref approach for both rather than the getElementById. You can see in the recording that the focus reset works with the section change but not the step change. Lmk if you have any questions

https://github.com/user-attachments/assets/ff7d20b3-39f8-44d0-8c09-b7ae0da60ea5

PR #5226, opened 2025-08-06
fix: only require name and jurisdiction

emilyjablonski:
QA for #5038

## Description

The backend was adding a default set of required fields on listing create if the required fields was an empty array. However, the frontend relies on the required fields to know when to show asterisks. The bug this created was if you selected Nada Hill and filled out only the name and jurisdiction and tried to publish, errors would throw on fields that did not have the asterisk. 

This is the set of default fields being added:
'listingsBuildingAddress',
'name',
'developer',
'listingImages',
'leasingAgentEmail',
'leasingAgentName',
'leasingAgentPhone',
'jurisdictions',
'units'

However, this default set is more than even Detroit has, and it should be the minimal set.
This PR updates it to just name and jurisdiction, and allows for even address to be null and for the listing page to still render.

In partners, select Nada Hill, hit publish, and ensure the only error on the form is on the name field. Fill out the name, hit publish, and you should be able to publish successfully. Hit preview, and ensure the detail page renders without error.

## Author Checklist:

## Review Process:


ludtkemorgan:
@emilyjablonski Can this PR be closed?

Issue #4924, opened 2025-06-02
Testing Bug - Open Text Collapse


sarahlazarich:
Feature: Dynamic Text Fields
Partners or Public? Public

Steps to Reproduce
- On Partners, add a suuuupppper long entry into a text field (I used rental history, and I think it had no spaces), save and check that listing out on the public site

Expected Result
- The listing should appear normal but the field should have the option to “View More/less”

Actual Result
- The listing image blows up and the page breaks

jaredcwhite:
@sarahlazarich yes, that's correct. the fix was just to ensure long text strings will wrap and it doesn't blow out the layout.

sarahlazarich:
okay then all good.

PR #5245, opened 2025-08-12
fix: ensure that very long strings don't blow out the listings page layout

jaredcwhite:
This PR addresses #4924

## Description

Some of the longer text entry fields which then display on the listing page had the potential to blow out the layout because the strings wouldn't wrap. This is probably unlikely in "normal" usage as people aren't typing out passages like asjkdhflasjkdfh, but it's still a good to idea to guard against any user-supplied content influencing the layout like that.

I think I touched on all the relevant placed in my updates to the CSS, but would love verification of that.

Type long nonsense strings into different text fields for a listing, then see what it looks like in the new Seeds layout.

## Author Checklist:

## Review Process:


jaredcwhite:
OK, got some false test failures here but it should be alright (the Netlify seeds app did rebuild correctly, not sure why it's still showing as a failure, and nothing changed re: partners for Cypress). You can test that link above and see that it's displaying correctly now!

Issue #5247, opened 2025-08-12
Testing Bug - Clearing Filters


sarahlazarich:
Steps to reproduce: 
- Go to the filter drawer at https://bloom-detroit.netlify.app/listings
- Select filters 
- Click "show matching listings" 
- Return to the drawer 

Expected behavior: 
- You should be able to clear your selections 
- Existing prod shows desired behavior [here](https://homeconnect.detroitmi.gov/listings)

Current behavior: 
- There is no option to clear the results 
- User has to manually unselect or delete options

ColinBuyck:
@sarahlazarich Hmmm I think we'll have to get designs input here (cc @eajensenwa) since the original Figma only shows the "show matching" and "cancel" buttons... don't think we originally planned for the clear button

https://www.figma.com/design/xqyamlgdsASkLTYlcifBwq/core-patterns?node-id=1686-9745&m=dev

sarahlazarich:
This makes sense to me @eajensenwa , thanks for the fast response. 

@ColinBuyck @KrissDrawing lmk if that clears things up

ColinBuyck:
Some concerns raised by Jared in the review process @eajensenwa @sarahlazarich https://github.com/bloom-housing/bloom/pull/5258#issuecomment-3189619643

sarahlazarich:
this lgtm

PR #5258, opened 2025-08-14
fix: clear filters

KrissDrawing:
This PR addresses #5247

## Description

Adds clear filters button

Go to `/listings` open filter drawer, it should have now `clear` button instead of close.
male sure it clears filters:
- when no filters applied
- when filters applied

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/src/locales/es.json
+++ b/shared-helpers/src/locales/es.json
@@ -629,6 +629,7 @@
   "leasingAgent.contact": "Comuníquese con el agente de alquiler",
   "leasingAgent.dueToHighCallVolume": "Debido al alto volumen de llamadas, usted podría escuchar un mensaje.",
   "leasingAgent.officeHours": "Horario de oficina",
+  "listingFilters.clear": "Borrar",
   "listingFilters.program.Families": "Familias",
   "listingFilters.program.Residents with Disabilities": "Residentes con discapacidades",
   "listingFilters.program.Seniors 55+": "Adultos mayores de 55 años",
--- a/shared-helpers/src/locales/general.json
+++ b/shared-helpers/src/locales/general.json
@@ -620,6 +620,7 @@
   "leasingAgent.contact": "Contact leasing agent",
   "leasingAgent.dueToHighCallVolume": "Due to high call volume you may hear a message.",
   "leasingAgent.officeHours": "Office hours",
+  "listingFilters.clear": "Clear",
   "listingFilters.program.Families": "Families",
   "listingFilters.program.Residents with Disabilities": "Residents with disabilities",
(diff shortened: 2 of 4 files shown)


jaredcwhite:
I understand this makes that secondary button behave as in Detroit prod, but I was very confused by the UX:

- I clicked clear, then I clicked away from the Drawer to close it and the listings weren't reset
  - (yet the filters number up top did go away)
- I see there's a dedicated Reset button on prod next to the Filter, but we don't seem to have a corresponding Reset button (see example image below)
- The only way to reset is to clear the filters, then click "Show matching listings" which is confusing to me because there's nothing to "match", I just cleared all the filters. 😅

Maybe we could reset the listings if the filters get cleared and then the drawer is closed (via clicking away or clicking the X button)? Or we should add a dedicated Reset button? Here's what's currently on prod:

image

sarahlazarich:
Ugh. Good points @jaredcwhite . I have a few thoughts, but really interested in @ColinBuyck and @eajensenwa 's take too: 
- If when you click clear, then you click away from the drawer, the listings don't reset but the filter number does, that feels like a bug to me. That's misleading. I feel like the counter should only change if the listings actually change or reset. 
- Good catch. We should bring this in to bring the experiences to parity. It could be a p1 follow-up, though, I don't think that is a huge deal. If it's super easy to just do that as a part of this I'm open to that too, keeping in mind we'd want this in early next week. 
- How much of this has to do with the button text being changed? If we switch it from "Show matching listings" back to "Done" does that fix some of the confusion?

eajensenwa:
Oh, interesting! Expected behavior based on the designs I shared would be that "clear" and "reset" actions result in the same outcome, which would be to immediately update the listings page to remove all applied filters(I actually didn't even know that "reset" appeared as an action on listings page once filters were applied). Is it possible to remove applied filters on "clear", instead of on "show matching listings"? Do we need both buttons if they result in the same action? If we do end up keeping both, the labels should match ( both should read "Reset" or "Clear"...but not one of each).

KrissDrawing:
It should be updated to match behaviour from last comment above.
For it to work without flicker i got to remove `key` prop from `ListingBrowse`. Not sure if it got any other purpose than clearing loading state.

shared-helpers/src/locales/es.json
ColinBuyck, line 1: **issue:** Could you add translations in all languages?

ColinBuyck (approved):
LGTM 🌵 Once the translations are added

jaredcwhite (approved):
Looking good to me, UX feels a lot better. After Colin's comments are resolved, LGTM! 😁

PR #5273, opened 2025-08-21
fix: translation clean up

emilyjablonski:
Addresses feedback on this PR: https://github.com/bloom-housing/bloom-detroit/pull/91

## Description

1. Removes strings from non-English files that had no English key/string. The vast majority of these are in the Arabic / Bengali files, because we brought over those full files from Detroit, but Detroit (a) was missing the fix we did in core that moved partners-only strings to a different file and (b) included strings that were specific to the old Detroit implementation that are intentionally not included in the new Detroit designs. These could also be unused from historically having cleaned up an unused key in the English file, and just not also cleaning it up in our other files, so there are a few strings removed in other languages as well.
2. Fixes some machine-translated Arabic strings which had variables, where the percent character used in the variable was instead using the Arabic percent character (`٪` vs `%`), which doesn't work with the variables. These strings were showing `{variable}٪` in the UI. An example is on the second page of an application, where we are using the `firstName` variable in the header:
479634678-a7ad1f8a-c741-487c-a84e-35c0b9f53ef5
3. Two Spanish key captialization updates (`Revisar los términos de uso` --> `Revisar los Términos de uso` to be consistent with other cases of Terms of Use, and `Mi Tablero de control` --> `Mi tablero de control` for sentence casing)

4. Removes < br > at the end of one of the strings on the terms page - only some languages had it, but it's not needed and appears as < br > in the UI instead of a line break.
479634663-9c3cebf6-7aef-489f-917a-3131137fd71c

5. Adds missing translations for one string from a PR that went in recently ("whatToExpectAdditionalText.label")
6. Ensures community types in the filter drawer are translated

Running the `unused-foreign-keys` script in shared-helpers will now output no strings, whereas on main, there are 814.
Running the `missing-translations` script should also output no values.
[More detailed instructions for how to run the scripts](https://www.notion.so/exygy/How-use-our-translation-scripts-2473433f03d080218b6adff4bd523a6c?source=copy_link)
Community types in the filter drawer should translate

## Author Checklist:

## Review Process:


emilyjablonski:
Tysm for the thorough review! Here’s some notes!
1. Bolding missing for Arabic on vouchers page
    1. t’s more subtle than our other languages, but the text is bolded here. It actually likely should not be bolded - there are some languages where bolding creates readability issues. For now, I think we can sit on this one, it would be a larger design task to consider what the site would look like without any bolded text.
2. Currency input in RTL
    1. Follow-up bug ticket in UIC: https://github.com/bloom-housing/bloom-detroit/issues/95
3. Nationality only translates on page load
    1. We have this problem for other parts of the application as well - some translations only come through when you start the application in that language. I believe it is related to the way we store google translations.
4. RTL Arabic numbers issue
    1. We chatted in standup! https://github.com/bloom-housing/bloom/issues/5281
5. <br> issue
    1. I’m not sure why some but not all languages have the <br> at the end of this strings - I removed them from all of them
6. Updated `Mi Tablero de control` to be sentence case
7. Community types just needed to be passed to our translation fxn
9. First name Arabic variable issue
    1. There were more of these issues beyond this string - some of the machine translated Arabic strings were used the Arabic percentage character, which does not work and causes this issue
10. Privacy policy & terms
    1. We chatted in standup! https://github.com/bloom-housing/bloom/issues/5282

5, 6, 7, and 8 are fixed in this core PR! https://github.com/bloom-housing/bloom/pull/5273

ColinBuyck (approved):
LGTM 🎎 very tidy!

PR #5315, opened 2025-08-27
fix: display vacant units for available units in partners

KrissDrawing:

## Description

It will display sum of vacant units for `available units` table for `unit groups` listings.
It does not exist for current detroit, so it seem like best value to display

log in to partners. For `unit group` listings it will show now `available units` based on unit groups.

## Author Checklist:

## Review Process:


mcgarrye (commented):
The issue is not the formatting, it is that `unitsAvailable` is not being set in the database for existing Detroit listings. We should reconsider this approach

mcgarrye:
There is also a bug with this [logic](https://github.com/bloom-housing/bloom/blob/main/api/src/services/listing.service.ts#L2536-L2545) in the listing service. It is always defaulting to the the length of units as units is an empty list and is always evaluating to true. Can be fixed by changing the check to units?.length instead of units. Might as well still fix the logic for the future

Issue #5210, opened 2025-08-05
[Mobile Issue] Langauge options cut off in smaller screen sizes


skeller-exy:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Observed Behavior
On iphone se (2nd gen, screen size 4.7in) the English language option is cut off in bloom.exygy.dev. After clicking Español language option I am unable to select English again. See screenshot below.

<img width="450" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGZ6Q2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0cb91e872b7d4ea7321afedf765bbdef3bc25fa7/Screensho.jpg" alt="Screensho.jpg" />

## Expected Behavior
Language options should be available even on smaller screen sizes. 

## Steps to replicate
Issue is partially replicable using chrome dev tools if you don't have a smaller screen size phone. 
1. Navigate to bloom.exygy.dev in chrome or firefox
2. Use dev tools and select iphone se view. Language options will be partially cut off.

## QA Notes

** Please provide as much information as possible (screenshots, documentation, etc) **

ludtkemorgan:
QA notes:
* Go to https://bloom.exygy.dev/ and shrink the screen down to mobile size. All languages should be visible in the top nav bar

sarahlazarich:
lgtm

PR #5340, opened 2025-09-03
fix: display language nav with smaller/tighter font on mobile

jaredcwhite:
This PR addresses #5210 

## Description

The size and placements of the language options in the language nav (both pre-Seeds and Seeds versions) didn't change on smaller/mobile screen sizes, so with a lot of languages it resulted in some falling off the edge of the screen (and also weird wrapping within language names). This PR adds a responsive design change so the languages nav display is smaller and tighter so it fits on a small screen.

Check the Seeds and pre-Seeds public sites in a small screen and compare with the previous layout.

| image | image |
| --- | --- |
| image | image |

Note that this does mean that the touch targets are smaller now, but I think that's an acceptable tradeoff. If you think we need a review from design, let's discuss that in the comments. (There's also a valid question here of if this change is needed when there are fewer languages shown…Detroit for example only has four and I think fits alright as-is.)

review sites top-left to bottom-right:
* [seeds - this PR] https://deploy-preview-5340--bloom-public-seeds.netlify.app
* [seeds - main] https://bloom-public-seeds.netlify.app
* [pre-seeds - this PR] https://deploy-preview-5340--bloom-exygy-dev.netlify.app
* [pre-seeds - main] https://bloom-exygy-dev.netlify.app

## Author Checklist:

## Review Process:


emilyjablonski (approved):
Fwiw, this isn't a problem in any of our jurisdictions, only core, due to the number of languages. For LA, there's a plan to redesign this to be a dropdown because I think LA would run into this. Lgtm!

Issue #5310, opened 2025-08-26
Testing Bug - Improve Pill Padding on Multi-line Text


ColinBuyck:
### What is the issue?

Could go either way on if the padding should reduce based on the text wrapping but at the very least, the top left padding seems to tight
<img width="389" height="549" alt="Image" src="https://github.com/user-attachments/assets/e5b63bda-9cf7-4c6b-bf54-fa3450683bf8" />

### How can you reproduce the bug?

Seethe listing names _Every Field Filled Out_ here: https://bloom-detroit-launch--detroit-public-dev.netlify.app/

### What browsers are you seeing the problem on?

Chrome

### Where are you seeing this issue?

Desktop

PR #5341, opened 2025-09-03
chore: adjust padding of Seeds tag for multi-line text

jaredcwhite:
This PR addresses #5310

## Description

> [!WARNING]
> This is an adjustment to the metrics of the Seeds tag. We may want to make this adjustment in Seeds directly, or come up with a different approach with guidance from design.

With that caveat, this PR might be helpful at least to kick off discussion. I adjusted the padding slightly as well as tightened the line height for when the text wraps (which I don't think we ever accounted for previously). I think that looks better, but YMMV.

Before/after:

image

image

review app: https://deploy-preview-5341--bloom-lakeview.netlify.app

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/styles/overrides.scss
+++ b/sites/public/styles/overrides.scss
@@ -3,6 +3,12 @@
     -webkit-font-smoothing: antialiased; // restore macOS styling that had been unset
   }
 
+  .seeds-tag {
+    --tag-padding-inline: var(--seeds-s3);
+    --tag-padding-block: var(--seeds-s1_5);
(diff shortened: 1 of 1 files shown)


emilyjablonski:
The fix here makes sense to me, but I personally would greatly prefer we make this change in Seeds. Seems like this would benefit everyone! Best case scenario is we don't have any overrides in core that aren't 100p based on tokens.

jaredcwhite:
@emilyjablonski agreed, I'll give this a go over there and close this override one out.

Issue #5354, opened 2025-09-09
Public 404 Error and Partners Autofilling Error around What To Expect


mcgarrye:
- [ ] General
- [ ] Backend
- [x] Public
- [x] Partners

## Background / Issue
The new rich text editor for the "What To Expect" field is causing two bugs.

Public: When the "What To Expect" field is not set (aka null) on the partners site, attempting to open the Listing Details page on the public site throws a 404 not found error as the Markdown does not expect null.

Partners: If a user save a listing with the "What To Expect" field not set and then re-opens the edit view, the "What To Expect" field will be repopulated with the default text when it should remain empty. Can cause saving the default text when not intended.

## Technical Context
Check if there is a value in the "What To Expect" field on the Public site and if in edit mode for a Listing on the Partners site.

## Acceptance Criteria
Listing Details page on the Public site appears correctly whether or not the "What To Expect" field is set.
"What To Expect" field is not re-populated in the edit mode for a listing on the Partners site.

## QA Notes
Create/Edit a listing on the Partner site. Delete the default text from the "What To Expect" field. Save changes/publish listing.
Go to the above listing on the Public site. Make sure you can see the listing and that no "What To Expect" section is present.
Go back to the Partners site and edit the listing. Make sure the "What To Expect" field is still empty. Add some text into the field.
Back to Public site again. Make sure you can see the text that was added to the "What To Expect" section.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNUpVQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--52497298ecb18a8d00f951f3c2630eb37980897c/Screen%20Recording%202025-09-09%20at%2012.37.13%E2%80%AFPM.mov)

** Please provide as much information as possible (screenshots, documentation, etc) **

PR #5355, opened 2025-09-09
fix: public 404 error and partners autofilling error around what to expect

mcgarrye:
This PR addresses [#(5354)](https://github.com/bloom-housing/bloom/issues/5354)

## Description

The new rich text editor for the "What To Expect" field is causing two bugs.

Public: When the "What To Expect" field is not set (aka null) on the partners site, attempting to open the Listing Details page on the public site throws a 404 not found error as the Markdown does not expect null.

Partners: If a user save a listing with the "What To Expect" field not set and then re-opens the edit view, the "What To Expect" field will be repopulated with the default text when it should remain empty. Can cause saving the default text when not intended.

These are fixed by checking if there is a value in the "What To Expect" field on the Public site and if in edit mode for a Listing on the Partners site.

Create/Edit a listing on the Partner site. Delete the default text from the "What To Expect" field. Save changes/publish listing.
Go to the above listing on the Public site. Make sure you can see the listing and that no "What To Expect" section is present.
Go back to the Partners site and edit the listing. Make sure the "What To Expect" field is still empty. Add some text into the field.
Back to Public site again. Make sure you can see the text that was added to the "What To Expect" section.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/components/listings/PaperListingForm/index.tsx
+++ b/sites/partners/src/components/listings/PaperListingForm/index.tsx
@@ -187,13 +187,15 @@ const ListingForm = ({ listing, editMode, setListingName }: ListingFormProps) =>
         return
       }
 
-      if (!whatToExpectEditor?.storage.characterCount.characters()) {
-        whatToExpectEditor.commands.setContent(selectedJurisdictionData.whatToExpect)
-      }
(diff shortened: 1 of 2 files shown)


ColinBuyck (approved):
LGTM 🧥

Issue #5338, opened 2025-09-03
[Feature] - New support vendor role


ludtkemorgan:
- [ ] General
- [x] Backend
- [ ] Public
- [x] Partners

## Background / Issue
The Doorway team is taking on a new support vendor to help with applicant and partner support. This new vendor will need access to listings across Doorway, but they would like to restrict certain actions from this new support vendor permission type vs the BAHFA admin type. In order to accomplish this we will need a new role.

The new role should be called `supportAdmin` within the enum and `Admin (support)` as the display name in the partner site.

This should be done in core and only after #5345 is complete
## Technical Context
Full TDD can be found [here](https://docs.google.com/document/d/1jMopeqCgJNf1NgeKLVSWwqNAik1p8pb19zJ4r10dEts/edit?tab=t.0).
* Backend
  * Create feature flag that enables this role 
  * Update schema with new roles as well as creating a migration file
  * Update seed data to have users with the two roles as well as at least one jurisdiction with the feature flags set
  * DTO updates
  * Services updates - listing, listing-csv-export, user, lottery, permission, user-csv-export
    *Make sure the appropriate spots are updated with these new role guards
* Frontend
  * Using the [PRD](https://docs.google.com/document/d/16peItcCvGtR4lBs3d9j8T7icLatBdb8vY6sguPSJLbc/edit?tab=t.0#heading=h.z2we68wtdnff) as a reference update all of the locations that need this role as a guard.
    * Examples being ListingFormActions.tsx, FormUserManage.tsx, and pages/users/index.tsx
  * Use the feature flag to determine if someone can create this role of a user
## Acceptance Criteria
* A new role is created in core behind a feature flag that is able to do the actions outlined in the [PRD](https://docs.google.com/document/d/16peItcCvGtR4lBs3d9j8T7icLatBdb8vY6sguPSJLbc/edit?tab=t.0#heading=h.z2we68wtdnff)
* Admins are able to add new users to the partner portal via the UI with the role `Admin (support)` as the option
## QA Notes

** Please provide as much information as possible (screenshots, documentation, etc) **

PR #5369, opened 2025-09-15
feat: new support vendor role

millord:
This PR addresses #5338

## Description

Purpose: add a new support vendor role

The Doorway team is taking on a new support vendor to help with applicant and partner support. This new vendor will need access to listings across Doorway, but they would like to restrict certain actions from this new support vendor permission type vs the BAHFA admin type. In order to accomplish this we will need a new role.

The new role should be called supportAdmin within the enum and Admin (support) as the display name in the partner site.

Log in using the support-admin@example.com account.

use this doc to verify all the actions for New Support Admin align with the requirements 

doc: https://docs.google.com/document/d/16peItcCvGtR4lBs3d9j8T7icLatBdb8vY6sguPSJLbc/edit?tab=t.0#heading=h.z2we68wtdnff

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/components/listings/ListingFormActions.tsx
+++ b/sites/partners/src/components/listings/ListingFormActions.tsx
@@ -445,8 +445,7 @@ const ListingFormActions = ({
       if (isListingApprover && !profile?.userRoles.isPartner) {
         elements.push(approveAndPublishButton)
         elements.push(requestChangesButton)
-      }
-      if (profile?.userRoles.isSupportAdmin) {
+      } else if (profile?.userRoles.isSupportAdmin) {
         elements.push(requestChangesButton)
       }
       if (profile?.userRoles.isPartner) {


api/src/enums/feature-flags/feature-flags-enum.ts
```diff
@@ -51,6 +52,10 @@ export const featureFlagMap: { name: string; description: string }[] = [
     name: FeatureFlagEnum.disableJurisdictionalAdmin,
     description: 'When true, jurisdictional admins cannot be created',
   },
+  {
+    name: FeatureFlagEnum.disableSupportAdmin,
+    description: 'When true, support admins cannot be created',
+  },
```
mcgarrye, line 58: **issue:** As this is new functionality, we should have this (and as many other feature flags as possible) as an `enable`, not `disable`. Thus flipping the meaning. Otherwise the default is that it is already turned on everywhere which we don't want.

api/src/services/listing.service.ts
```diff
@@ -1627,8 +1631,17 @@ export class ListingService implements OnModuleInit {
 
     //manually check for juris/listing mismatch since logic above is forcing admin permissioning
     if (
-      ((requestingUser?.userRoles?.isJurisdictionalAdmin ||
-        requestingUser?.userRoles?.isLimitedJurisdictionalAdmin) &&
+      (requestingUser?.userRoles?.isSupportAdmin &&
+        !requestingUser?.jurisdictions?.some(
+          (juris) => juris.id === storedListing.jurisdictionId,
+        )) ||
```
mcgarrye, line 1637: **issue:** This check is not accurate. Support Admins will have access to all jurisdictions so this logic can be removed

api/src/services/listing.service.ts
```diff
@@ -1597,6 +1599,8 @@ export class ListingService implements OnModuleInit {
 
     const userRoles =
       requestingUser?.userRoles?.isAdmin ||
+      (requestingUser?.userRoles?.isSupportAdmin &&
+        duplicateListingPermissions?.includes(UserRoleEnum.supportAdmin)) ||
```
mcgarrye, line 1603: **issue:** This should follow `isAdmin` similar to above. The support admin will always have duplicating ability

api/src/services/permission.service.ts
```diff
@@ -79,6 +79,25 @@ export class PermissionService {
...
+            `(${permissionActions.read}|${permissionActions.create}|${permissionActions.update}|${permissionActions.delete})`,
+          );
+          await enforcer.addPermissionForUser(
+            user.id,
+            'listing',
+            `r.obj.jurisdictionId == '${adminInJurisdiction.id}'`,
+            `(${permissionActions.read}|${permissionActions.create}|${permissionActions.update}|${permissionActions.delete})`,
+          );
+        }),
+      );
```
mcgarrye, line 100: **question:** Not sure if this is necessary. @ludtkemorgan or @YazeedLoonat, can one of you check this?

api/src/services/user.service.ts
```diff
@@ -417,6 +417,7 @@ export class UserService {
     const isPartnerPortalUser =
       storedUser.userRoles?.isAdmin ||
       storedUser.userRoles?.isJurisdictionalAdmin ||
+      storedUser.userRoles?.isSupportAdmin ||
```
mcgarrye, line 420: **nitpick:** Alphabetize please

api/src/utilities/build-user-where.ts
```diff
@@ -94,8 +94,47 @@ export const buildWhereClause = (
...
         });
+        filters.push({
+          jurisdictions: {
+            some: {
+              id: {
+                in: user?.jurisdictions?.map((juris) => juris.id),
+              },
+            },
+          },
+        });
```
mcgarrye, line 137: **issue:** support admin should not have access to users. logic can be removed

mcgarrye (changes requested):
Along with the changes requested, we need to add a permission test file to the api directory under `test/integration/permission-tests` for this new role

sites/partners/src/components/users/FormUserManage.tsx
```diff
@@ -66,6 +68,12 @@ const FormUserManage = ({
     possibleUserRoles.push(RoleOption.JurisdictionalAdmin)
     possibleUserRoles.push(RoleOption.LimitedJurisdictionalAdmin)
   }
+  if (
+    !profile?.userRoles?.isPartner &&
```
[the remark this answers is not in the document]
mcgarrye: Actually, based on the checklist, support admins should not have access to create/edit users at all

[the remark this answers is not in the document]
mcgarrye: This should be adjusted so that only Admins can create support admins, currently the check only prevents Partners

api/src/permission-configs/permission_policy.csv
```diff
@@ -64,40 +74,48 @@ p, admin, listing, true, .*
...
 p, admin, mapLayers, true, .*
+p, supportAdmin, mapLayers, true, .*
 p, jurisdictionAdmin, mapLayers, true, .*
 p, limitedJurisdictionAdmin, mapLayers, true, .*
 
 p, admin, featureFlags, true, .*
 
-g, admin, jurisdictionAdmin
+g, admin, supportAdmin
+g, supportAdmin, jurisdictionAdmin
```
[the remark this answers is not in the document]
ludtkemorgan: @millord I reached out to kathy to see if it should just be the button or if we should display something else. I'll report back

mcgarrye:
Screenshot 2025-10-01 at 3 56 19 PM
Per line [404](https://github.com/bloom-housing/bloom/pull/5369/files#diff-becd400048644f760f1868e2f4332cfc98067042a50d520ffaf263e96fcb5cf7R404) in file: sites/partners/src/components/users/FormUserManage.tsx, the options should be sorting. Can we check why this isn't working for the new role?

mcgarrye:
I'm seeing a number of actions working that should not. As a support admin I was able to navigate to `/users` and `/settings` and able to do actions, I was also able to publish a listing. All of these should be disallowed actions

ludtkemorgan (approved):

sites/partners/src/components/listings/ListingFormActions.tsx
```diff
@@ -440,6 +445,9 @@ const ListingFormActions = ({
         elements.push(approveAndPublishButton)
         elements.push(requestChangesButton)
       }
+      if (profile?.userRoles.isSupportAdmin) {
```
mcgarrye, line 448: **suggestion:** should be an `else if` with the new isListingApprover logic

mcgarrye (approved):
Looks good, lgtm

Issue #5215, opened 2025-08-05
PII Removal Fields Migration


mcgarrye:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Create a migration for the Applications and Users table to add new fields and indexes. Update the DTOs accordingly.

## Technical Context
[New data schemas
](https://docs.google.com/document/d/1bCXeoQ0153h_z7hKHm3pVBAjz0xfdR3rR6dK-Zma58c/edit?tab=t.0#heading=h.1lylnmxdzhqq)
**Users table**
New columns
- Add “was_warned_of_deletion” (boolean) (do not need to expose)
New Indexes
- users.last_login_at
- users.was_warned_of_deletion

**Applications table**
New columns
- Add “is_newest” (boolean)
- Add “expire_after” (timestamp) (do not need to expose)
- Add “was_pii_cleared” (boolean) (do not need to expose)

New Indexes
- is_newest, expire_after, was_pii_cleared

Update DTOs accordingly

## Acceptance Criteria
- Migration to update the database schema
- Updated DTOs
- Updated backend-swagger

## QA Notes
- No QA needed

PR #5371, opened 2025-09-15
fix: add additional database fields for pii deletion

ludtkemorgan:
This PR addresses #5215 

## Description

This PR is the first step in removing PII data from applications/users. It adds the new fields to the `users` and `applications` tables that are required for us to accurately delete PII data.

No UI changes are done with this PR. Instead this can be tested by verifying that the new fields are correctly created in the DB. You should see the following fields

### Applications table
* `is_newest`
* `expire_after`
*  `was_pii_cleared`

### Users table
* `was_warned_of_deletion`

## Author Checklist:

## Review Process:

### DIFF
--- a/api/prisma/migrations/32_pii_deletion_fields/migration.sql
+++ b/api/prisma/migrations/32_pii_deletion_fields/migration.sql
@@ -16,15 +16,3 @@ ADD
 
 -- CreateIndex
 CREATE INDEX "applications_is_newest_idx" ON "applications"("is_newest");
-
--- CreateIndex
-CREATE INDEX "applications_was_pii_cleared_idx" ON "applications"("was_pii_cleared");
-
--- CreateIndex
-CREATE INDEX "applications_expire_after_idx" ON "applications"("expire_after");
-
--- CreateIndex
-CREATE INDEX "user_accounts_last_login_at_idx" ON "user_accounts"("last_login_at");
-
--- CreateIndex
-CREATE INDEX "user_accounts_was_warned_of_deletion_idx" ON "user_accounts"("was_warned_of_deletion");
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -236,7 +236,7 @@ model Applications {
   isNewest                     Boolean?                      @default(false) @map("is_newest")
   wasPIICleared                Boolean?                      @default(false) @map("was_pii_cleared")
   // Signifies when PII data should be cleared
-  expireAfter                  DateTime?                     @map("expire_after") @db.Timestamptz(6)
+  expireAfter                  DateTime?                     @map("expire_after") @db.Timestamp(6)
   applicantId                  String?                       @unique() @map("applicant_id") @db.Uuid
   mailingAddressId             String?                       @unique() @map("mailing_address_id") @db.Uuid
   alternateAddressId           String?                       @unique() @map("alternate_address_id") @db.Uuid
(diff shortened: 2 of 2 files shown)


api/prisma/schema.prisma
```diff
@@ -374,7 +376,7 @@ model Jurisdictions {
   publicUrl                     String                   @default("") @map("public_url")
   emailFromAddress              String?                  @map("email_from_address")
   rentalAssistanceDefault       String                   @map("rental_assistance_default")
-  whatToExpect                  String                   @default("") @map("what_to_expect")
+  whatToExpect                  String                   @default("Applicants will be contacted by the property agent in rank order until vacancies are filled. All of the information that you have provided will be verified and your eligibility confirmed. Your application will be removed from the waitlist if you have made any fraudulent statements. If we cannot verify a housing preference that you have claimed, you will not receive the preference but will not be otherwise penalized. Should your application be chosen, be prepared to fill out a more detailed application and provide required supporting documents.") @map("what_to_expect")
```
ludtkemorgan, line 379: This was missed in previous PR. It was updated in a migration but not in the schema file so whenever you generate a new migration it tried to set it back to an empty string.

api/prisma/migrations/32_pii_deletion_fields/migration.sql
```diff
@@ -0,0 +1,30 @@
...
+CREATE INDEX "applications_was_pii_cleared_idx" ON "applications"("was_pii_cleared");
+
+-- CreateIndex
+CREATE INDEX "applications_expire_after_idx" ON "applications"("expire_after");
+
+-- CreateIndex
+CREATE INDEX "user_accounts_last_login_at_idx" ON "user_accounts"("last_login_at");
+
+-- CreateIndex
+CREATE INDEX "user_accounts_was_warned_of_deletion_idx" ON "user_accounts"("was_warned_of_deletion");
```
mcgarrye, line 30: **question:** Yazeed had raised in the [design document](https://docs.google.com/document/d/1bCXeoQ0153h_z7hKHm3pVBAjz0xfdR3rR6dK-Zma58c/edit?tab=t.0#heading=h.ngdt1q2kyrfe) about whether the indices should be combination indices on each table or individual as you have them now. I didn't have time to revisit this before the work was picked up (didn't realize it was this sprint). Did you put any thought into it or just followed the ticket AC?

ludtkemorgan: You wrote up the design doc and ticket so well I didn't think about it at all.

My understanding is that we need all those indices for different reasons as in we will need to query on the table by each of those fields individually. So keeping them separate makes sense. @YazeedLoonat maybe you have more of an opinion on this.

YazeedLoonat: ah I see! If each is used to query individually I think this set up works. We may want to look at performance for a bit since I think each new index creates a cumulative performance impact?

ludtkemorgan: Yeah, since these for the most part are only going to be used on cron jobs (except the isNewest flag) I wonder if we should not do the indexes and add them if we find it is necessary after running the cron job for the first time. 

Thoughts @mcgarrye?

mcgarrye: I'm fine with that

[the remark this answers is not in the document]
ludtkemorgan: Sometime you'll need to share with me your vscode settings so that it can format on save and keep the newline

mcgarrye (approved):
LGTM

Issue #5377, opened 2025-09-16
Non-Reg - API Layer Updates for Non-Regulated Listings


nicknorena:
### Description

Update the NestJS API layer to handle non-regulated listing fields, including DTOs, validation pipes, and service layer methods.

### Acceptance Criteria

- [x]  Update listing DTOs to include new non-regulated fields
~- [ ]  Modify `ListingCreateUpdateValidationPipe` to handle different validation rules for regulated vs non-regulated listings~
- [x]  Update validation logic:
    - [x]  Make `applicationDueDate` optional when `listingType = "nonRegulated"`. This can be done by updating the listing.dto `ValidateIf` function on that field
    - [x]  Handle deposit validation based on `depositType`
      - [x] If `depositType` = `depositRange` then nothing should be set for `depositValue` and should  have a value for `depositMin` and `depositMax`
      - [x] If `depositType` = `fixedDeposit` then nothing should be set for `depositMin` and `depositMax` as well as having a value set for `depositValue`.
    - [x]  Handle rent validation based on `rentType`
- [x]  Update service layer methods to properly save/retrieve new fields
- [x]  Ensure `cocInfo` field is excluded from public API responses but included in partner portal responses
- [x]  **TO DO: CONFIRM VALIDATION RULES FOR APPLICATION TYPES FOLLOWING DESIGN GUIDANCE - https://www.figma.com/design/5WzHg5F3lLgJyNTOd37zBb/Detroit-Non-Regulated-Listings?node-id=6016-5092&t=8UQX1yOVf4ayWPKM-1**

### Technical Notes

- Validation should allow for different required fields based on listing type
- Maintain backward compatibility for existing regulated listings
- Consider jurisdiction-specific validation rules

PR #5413, opened 2025-10-02
feat: non regulated listings api layer updates

matzduniuk:
This PR addresses #5377

## Description

* Create a set of new custom decorators for rent and deposit fields validation
* Update DTOs to contain new non-regulated listings fields
* Update listings controller integration tests to test the new validation methods

## How to test?

Run the following command in the repo directory:
```
cd api/ && yarn jest --config ./test/jest-e2e.config.js -w -t 'Listing Controller Tests'
```

## Author Checklist:

## Review Process:

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,5 @@
   "lint-staged": {
     "*.{js,ts,tsx}": "eslint"
   },
-  "dependencies": {},
-  "packageManager": "yarn@1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610"
+  "dependencies": {}
 }


ludtkemorgan:
> > > > > Screenshot 2025-10-08 at 2 04 52 PM Default message isn't returning just right for the unit group decorator
> > > > 
> > > > 
> > > > I'm not sure what you used when testing. I achieved a similar result using Postman, but the backend logs themselves show the message without the escape characters, same on the FE, as proven by the integration tests, which pass for expected strings without escape characters. In terms of the nested validation prefix (ie, `unitGroups.0.<fieldName>`). I haven't found if there is a solution for this already in the repo, I found solutions on GH, but I am unsure if introducing this type of overrides is worth it
> > > 
> > > 
> > > I was most concerned with the nested prefix. Happy to pair on this if needed
> > 
> > 
> > Will happily accept the offer. From what I found, the only way to fix this is to add a custom `ValidationPipe` to the 'app' instance and override the `exceptionFactory` to enforce custom formatting. From what I tested, the problem is that if I approach this is that we would need to go through all DTOs and add the `@ValidateIf` decorator to all optional fields. Overall, it seems that this might be a bit complex to make sure that we override it without degrading the exception factory from its default behaviour.
> 
> Oof yeah definitely don't love having to change all of our DTOs for this one error message... @ludtkemorgan @YazeedLoonat have you seen this before / know a way to resolve it?

@matzduniuk I caught up with @mcgarrye about this and I think it's ok to proceed with the weird error message since the only way this can happen is if someone bypasses the frontend. But we should do a follow-up at the end of the non-regulated feature to look into this more

Issue #5217, opened 2025-08-05
Application PII Removal Functionality


mcgarrye:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Update the ApplicationService and the ListingService to support removing PII from applications based on when a listing closes.

## Technical Context
Updates to the [ApplicationService](https://docs.google.com/document/d/1bCXeoQ0153h_z7hKHm3pVBAjz0xfdR3rR6dK-Zma58c/edit?tab=t.0#heading=h.3u0ij1ir5wu1):
- Update the `create` function
- Implement `removePII` and `removePIIAfter` functions
- `removePIIAfter` should batch applications
- Add tests for all

Updates to the ListingService:
- Change update() and closeListings() functions such that when a listing closes it sets expire_after on all associated applications based on APPLICATION_DAYS_TILL_EXPIRY
- Update tests
- May need batching for high number of applications

## Acceptance Criteria
- Creating an application for a user account sets `is_newest` to true and previous application to false
- On listing close, `expire_after` is set on all associated applications
- removePIIAfter cronjob run successfully and removes the required fields

## QA Notes
Requires looking at the db after application creation, listing close and cronjob completion

Set APPLICATION_DAYS_TILL_EXPIRY to 0
Create a listing with a digital app
Apply multiple times with the same user account
Close the listing
Trigger cronjob (either by calling the endpoint or setting the cron env var within a small window)
Verify all data is as expected in the DB

ludtkemorgan:
This has been tested in core. Further testing will be done in Doorway staging once #5785 is completed

Issue #5216, opened 2025-08-05
Scripts to Set Fields for PII Removal


mcgarrye:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
New columns added to Applications and User tables need to be backfilled before the cronjobs can begin running as the queries they rely on need the fields to be set. These fields cannot be set in the migration as they are dependent on external business logic.

## Technical Context
Both should use pagination/batching
**setExpireAfterForApplications function:**

1. Pull all closed listings
2. Loop through listings: 
 - Calculate expire_after = listing.closed_at + APPLICATION_DAYS_TILL_EXPIRY
 - Call updateMany on applications table where listing.id=application.listing_id and set expire_after

**setIsNewestForApplications function:**

1. Pull all users
2. Loop through users:
 - Call ApplicationService.mostRecentlyCreated() on the user id
 - Set is_newest to TRUE for the returned application

## Acceptance Criteria
- ENV variables are set
- Scripts can be run
- Scripts backfill the data as expected

## QA Notes
Requires calling the backend API

ludtkemorgan:
QA Notes:
* I ran both script runs in core by setting the value to 10 years and the proper data is set on all applicable applications.
* This does not need any further testing and can be run in Doorway once all code is in production as well as the go-ahead has been given by the Doorway team

PR #5440, opened 2025-10-14
fix: add pii deletion population script

ludtkemorgan:
This PR addresses #5216 

## Description

In order to have the PII deletion fields be properly populated before the first cron job runs we need to have script runs to populate that data. This PR creates two script runs that do the following:
1. `setInitialExpireAfterValues` - Sets a value for expire_after on all applications tied to listings that have already been closed
2. `setIsNewestApplicationValues` - Sets the `is_newest` value to true for all applications that are the most recently submitted by an applicant

Note: We should not run this until the rest of the data population tickets are done #5217 so that there is not a gap between running these scripts and new data being updated

### Testing the expire after script
1. Reseed
2. Create several applications tied to listings
3. Close at least one of these listings 
4. Set the `APPLICATION_DAYS_TILL_EXPIRY` to a number of your choosing
5. Run the `setInitialExpireAfterValues` and verify that the applications tied to the newly closed listing have the `expire_after` field set to the correct number of days in the future

### Testing the is newest script
1. Reseed
2. login to the public site and apply to several listings
3. Run the `setIsNewestApplicationValues` script 
4. Verify that all but the newest application has the `is_newest` value set to true

## Author Checklist:

## Review Process:

### DIFF
--- a/api/test/integration/script-runner.e2e-spec.ts
+++ b/api/test/integration/script-runner.e2e-spec.ts
@@ -81,7 +81,6 @@ describe('Script Runner Controller Tests', () => {
 
     it('should set the expire_after value for applications on closed listings', async () => {
       process.env.APPLICATION_DAYS_TILL_EXPIRY = '90';
-      // Reset all closedAt values so other tests don't impact this test
       const jurisData = jurisdictionFactory();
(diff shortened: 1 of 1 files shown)


mcgarrye (approved):
lgtm

PR #5462, opened 2025-10-20
fix: add new address listings view

ludtkemorgan:
This PR addresses #5466 

## Description

For a script used by the Doorway team they need the ability to get all names and addresses of all listings in the system. This is needed because currently the only way to get this information is to call the /listings and use the base or full view. This causes major performance issues as way too much data is returned.

This PR does the following two things:
1. creates a new view supported by the get many listing and get one listing endpoints which only returns `name`, `id`, and the full building address
2. Refactors how views are constructed and used so that we can utilize `select` instead of exclusively `include`. This allows us to target the fields returned in the base `listings` table

The new view can be tested using the swagger doc
1. Reseed 
2. Startup the backend and go to the filterableList endpoint http://localhost:3100/api#/listings/filterableList
3. Do a query that has "view" of "address" . Example:
```
{
  "page": 1,
  "limit": 10,
  "filter": [
  ],
  "view": "address",
  "orderBy": [
    "mostRecentlyUpdated"
  ],
  "orderDir": [
    "desc"
  ]
}
```

Since this also touches other views we need to validate other usages of the listings endpoint to make sure the correct data is returned

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/src/types/backend-swagger.ts
+++ b/shared-helpers/src/types/backend-swagger.ts
@@ -2477,7 +2477,7 @@ export class ScriptRunnerService {
     })
   }
   /**
-   * An example of how the script runner can work
+   * A script to migrate MSQ data and options to refactored schema
    */
   migrateMultiselectDataToRefactor(options: IRequestOptions = {}): Promise<SuccessDTO> {
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


mcgarrye:
Can you run `yarn generate:client`?

Issue #5372, opened 2025-09-15
[Bug]: lottery results section is incorrect


ludtkemorgan:
### What is the issue?

In Doorway a lottery result was released that wasn't run in our system via uploading a PDF of the results. This caused a weird UI scenario on the page where the public lottery section appeared in an erroneous state with an "INVALID DATE" listed. 
<img width="376" height="441" alt="Image" src="https://github.com/user-attachments/assets/8811b1a2-344b-4a42-b3ac-38c6046ee537" />

### How can you reproduce the bug?

This listing is an example: https://housingbayarea.mtc.ca.gov/listing/abc0da18-f872-43dd-a66b-9ac4324ff156/1698_university_1698_university_avenue_berkeley_ca

* Create a listing that has applications using the common app but lottery will not be run in the system 
* Close the listing
* Upload results in partners
* Make sure the lottery to be run date is in the past 
* Go to the listing in public and you will see the above

### Is there any additional context?

The INVALID DATE portion is happening because nothing is being save to the "startDate" field, but something is being saved in the "startTime"

This is an issue in Core as well, when the new seeds designs are turned off 

### What's the correct behavior 

* When manual lottery results are posted, you should remove the section that says Public Lottery or Lottery Results and instead keep the top section with the results. 

<img width="953" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnlSQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--a1ad9d3f520dd4121175145585e60ebfebf3b27f/image.png" alt="image.png" />

### What browsers are you seeing the problem on?

Chrome

### Where are you seeing this issue?

Desktop

kathyccheng:
LGTM

PR #5492, opened 2025-10-22
fix: hide redundant lottery section

KrissDrawing:
This PR addresses #5372

## Description

It will hide lottery section for 

with public env `SHOW_NEW_SEEDS_DESIGNS=FALSE`

- Create a listing that has applications using the common app but lottery will not be run in the system
- Close the listing
- Upload results in partners
- Make sure the lottery to be run date is in the past
- Go to the listing in public and you will see the above

Make sure other flow works as before (the one when results are published from partners)

## Author Checklist:

## Review Process:


sites/public/src/components/listing/ListingView.tsx
```diff
@@ -310,7 +310,12 @@ export const ListingView = (props: ListingProps) => {
   }
 
   let lotterySection
-  if (publicLottery && (!lotteryResults || (lotteryResults && !lotteryResults.url))) {
+
+  if (
+    publicLottery &&
+    (!lotteryResults || (lotteryResults && !lotteryResults.url)) &&
+    (listing.status === ListingsStatusEnum.active || !lotteryResults)
```
ludtkemorgan: issue: I was still able to reproduce the issue by doing an abnormal workflow.
1. Edit a listing to have a public lottery start date/time in the past
2. Close that listing and upload a lottery result
3. Re-open the listing. 
4. The "Invalid date" section is still being shown.

I think your changes are good here, but I believe we should also delete the entire section starting on line 326. It's using the wrong field to show the timestamp and also the above scenario is the only time it could be shown. So I'm good with you deleting the entire second EventSection

PR #5504, opened 2025-10-28
fix: use conductor values for Common App form step checkpoints

jaredcwhite:
This PR is a bugfix follow-up to #5265. I had been using the `AppSubmissionContext` values to check the application & listing states, rather than the form conductor within the context which always had the most up-to-date values. This was causing some issues with filling out new applications after finishing previous ones, specifically when there was a community disclaimer page.

With this fix, there should be no error/redirection with filling out one application, then going to another listing and starting to fill it out, etc.


jaredcwhite:
I'm not sure why that one Netlify build keeps failing with memory issues but the others are working. Don't think it's related to this PR specifically.

Issue #3182, opened 2022-11-22
Update UX of preference selection without jurisdiction


emilyjablonski:
A the moment in the partners listing form, if you go to create a new listing but don't select a jurisdiction, you are able to get two actions deep into adding a preference before you receive a warning that you must first select a jurisdiction.

The current flow is
* Click Add Preference
* Opens Add Preferences drawer, click Select Preferences
* Opens Select preferences drawer, has this message: `You must first select a jurisdiction`

We should prevent action in terms of preference management if a jurisdiction is not selected from the root form.
 @slowbot A preference on how? I am imagining disabling the button and showing the message as helper text

emilyjablonski:
Still valid

Issue #2142, opened 2021-11-04
Update New Listing Button to Include Jurisdiction Selection


slowbot:
**What is this feature or what feature is this part of?**
A clear and concise description of what the problem is.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**What is the acceptance criteria/definition of done?**
A clear and concise description of the acceptance criteria required to close this issue.

**QA Review Instructions**
This is to be filled out by the developer who completes this issue, before passing to QA.

emilyjablonski:
Would love love love to do this

emilyjablonski:
Fixed as part of jurisdiction pre-selection

PR #5582, opened 2025-11-17
feat: preselect jurisdiction

emilyjablonski:
This PR addresses #2142 #3182

## Description

Requires pre-selecting a jurisdiction in a dialog before being able to visit the add listing page if you have access to multiple jurisdictions.

* A user with access to multiple jurisdictions should be prompted by the modal before being able to visit the add listing page
* A user with access to one jurisdiction should not be prompted by the modal
* A user should not be able to visit `listings/add` directly - it must have a jurisdiction id parameter, and should redirect to the home page

Jurisdiction selection dialog:
Screenshot 2025-11-24 at 10 44 19 AM

<hr />

A new read-only section was added at the top of the edit page that mirrors a section that already exists at the top of the view page - "Listing data". The jurisdiction does not appear at all if a user only has access to one jurisdiction.
One jurisdiction:
Screenshot 2025-11-17 at 10 58 16 PM
Multiple jurisdictions:
Screenshot 2025-11-17 at 10 58 37 PM

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/application.service.ts
+++ b/api/src/services/application.service.ts
@@ -36,6 +36,7 @@ import { MostRecentApplicationQueryParams } from '../dtos/applications/most-rece
 import { PublicAppsViewQueryParams } from '../dtos/applications/public-apps-view-params.dto';
 import { ApplicationsFilterEnum } from '../enums/applications/filter-enum';
 import { PublicAppsViewResponse } from '../dtos/applications/public-apps-view-response.dto';
+import dayjs from 'dayjs';
 
 export const view: Partial<
   Record<ApplicationViews, Prisma.ApplicationsInclude>
@@ -636,6 +637,18 @@ export class ApplicationService {
       }
     }
 
+    // If a new application comes in after close and PII needs to be deleted
+    let expireAfterDate = undefined;
+    if (
+      listing.status === ListingsStatusEnum.closed &&
+      process.env.APPLICATION_DAYS_TILL_EXPIRY &&
+      !isNaN(Number(process.env.APPLICATION_DAYS_TILL_EXPIRY))
+    ) {
+      expireAfterDate = dayjs(listing.closedAt)
+        .add(Number(process.env.APPLICATION_DAYS_TILL_EXPIRY), 'days')
+        .toDate();
+    }
+
     const transactions = [];
     // Set all previous applications for the user to not be the newest
     if (requestingUser?.id) {
@@ -775,6 +788,7 @@ export class ApplicationService {
                 },
               }
             : undefined,
+          expireAfter: expireAfterDate,
 
           // TODO: Temporary until after MSQ refactor
           applicationSelections: undefined,
--- a/api/src/services/listing.service.ts
+++ b/api/src/services/listing.service.ts
@@ -2503,38 +2503,7 @@ export class ListingService implements OnModuleInit {
       throw new HttpException('listing failed to save', 500);
     }
 
-    const listingApprovalPermissions = (
-      await this.prisma.jurisdictions.findFirst({
-        where: { id: incomingDto.jurisdictions.id },
-      })
-    )?.listingApprovalPermissions;
-
-    if (listingApprovalPermissions?.length > 0)
-      await this.listingApprovalNotify({
-        user: requestingUser,
-        listingInfo: { id: incomingDto.id, name: incomingDto.name },
-        approvingRoles: listingApprovalPermissions,
-        status: incomingDto.status,
-        previousStatus: storedListing.status,
-        jurisId: incomingDto.jurisdictions.id,
-      });
-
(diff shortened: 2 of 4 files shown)


sites/partners/src/pages/index.tsx
```diff
@@ -304,6 +350,66 @@ export default function ListingsList() {
...
+        ariaDescribedBy="listing-select-dialog-content"
+        onClose={() => setListingSelectModal(false)}
+      >
+        <Form id="listing-select-form" onSubmit={handleSubmit(onSubmit)}>
+          <Dialog.Header id="listing-select-dialog-header">
+            {"Add listing dialog header"}
+          </Dialog.Header>
+
+          <Dialog.Content id="listing-select-dialog-content">
+            {"Add listing dialog content"}
```
ludtkemorgan: suggestion: I know this is filler text until we fully implement this. So I'm putting a note here on what I think it should contain. It should explicitly reference that the jurisdiction cannot be changed after the fact.

ludtkemorgan (commented):
Love this change! And I hope that we can go forward with this as it will save us a lot of headaches. Especially around some of the feature flags we have setup for Detroit.

I do think we need to talk to the Doorway team before we could proceed with this though. Their setup is to have 1 jurisdiction per county in the system but every jurisdiction has essentially the same settings. So for example if we turn on a feature flag for one jurisdiction we will be turning it on for all of them. It is less important in that setup to pre-select the jurisdiction and I know on a couple of occasions they have switched the jurisdiction after creating it. Either because they accidentally selected the wrong county or they wanted to create it before the new county existed. That second case shouldn't happen again unless they expand to more counties which I don't anticipate them doing. We'll just want to touch base with them to see if that limitation would cause any issues.

sites/partners/src/components/listings/PaperListingDetails/sections/DetailListingData.tsx
```diff
@@ -3,24 +3,37 @@ import { t } from "@bloom-housing/ui-components"
 import { FieldValue, Grid } from "@bloom-housing/ui-seeds"
 import SectionWithGrid from "../../../shared/SectionWithGrid"
 import { ListingContext } from "../../ListingContext"
-import { getDetailFieldDate, getDetailFieldTime } from "./helpers"
+import { getDetailFieldDate, getDetailFieldString, getDetailFieldTime } from "./helpers"
 
-const DetailListingData = () => {
+type DetailsListingDataProps = {
```
mcgarrye: **nitpick:** This has an s on `Detail` when the component does not

sites/partners/src/pages/index.tsx
```diff
@@ -278,7 +309,22 @@ export default function ListingsList() {
               <div className="flex gap-2 items-center">
                 {isAdmin && (
                   <>
-                    <Button size="sm" variant="primary" href="/listings/add" id="addListingButton">
+                    <Button
+                      size="sm"
+                      variant="primary"
+                      onClick={() => {
+                        // also check if feature flags
```
ludtkemorgan: question: Is this comment applicable? It doesn't seem tied to what is below it

sites/partners/src/components/listings/PaperListingForm/sections/ListingIntro.tsx
```diff
@@ -1,36 +1,28 @@
...
+import { defaultFieldProps, fieldHasError, fieldMessage } from "../../../../lib/helpers"
 import SectionWithGrid from "../../../shared/SectionWithGrid"
-import styles from "../ListingForm.module.scss"
 
 interface ListingIntroProps {
-  jurisdictions: Jurisdiction[]
+  enableHousingDeveloperOwner: boolean
+  enableListingFileNumber: boolean
+  enableNonRegulatedListings: boolean
+  jurisdictionName: string
```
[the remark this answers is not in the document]
emilyjablonski: Updated them to be optional!

sites/partners/src/pages/index.tsx
```diff
@@ -304,6 +350,66 @@ export default function ListingsList() {
...
+
+          <Dialog.Content id="listing-select-dialog-content">
+            {t("listings.selectJurisdictionContent")}
+            <Grid>
+              <Grid.Row columns={3}>
+                <Grid.Cell className={"seeds-grid-span-2"}>
+                  <div className={`${defaultJurisdiction ? "hidden" : ""} seeds-m-bs-4`}>
+                    <Select
+                      id={"jurisdiction"}
+                      defaultValue={defaultJurisdiction}
```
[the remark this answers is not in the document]
emilyjablonski: This is me prepping for having other flags in here, where the user might only have access to one jurisdiction (like a Lakeview user where there should be no jurisdiction dropdown, but there may be a form field for non-regulated, and we want to collect both)

Issue #5478, opened 2025-10-21
Partners - New marketing flyer section


sarahlazarich:
This section should look very similar to the building selection criteria section, but be behind a feature flag, and have two upload options.

This ticket will include creating the feature flag, and adding the two new optional PDF/URL fields that mirror the existing buildingSelectionCriteria field, but for `marketingFlyer` and `accessibleMarketingFlyer`.

The new section should appear at the bottom of the `Application dates` section in partners.

<img width="515" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMlFUREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4a51b32b6f663a4751b99faad68280f9f9d88fd4/Screenshot%202025-11-10%20at%207.01.04%E2%80%AFPM.png" alt="Screenshot 2025-11-10 at 7.01.04 PM.png" />

There should be two form fields in this drawer, one to upload or link a marketing flyer, and another to upload or link an accessible marketing flyer.

<img width="628" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMllUREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b734e4e8034c5cc658651638b8330f1a6ef704fe/Screenshot%202025-11-10%20at%207.03.58%E2%80%AFPM.png" alt="Screenshot 2025-11-10 at 7.03.58 PM.png" />

AC:
- The new section shows on the partners listing edit and view pages if the feature flag is on
- Updates to this field can be saved, edited, and removed
- The field is shown in the listings export

emilyjablonski:
Okidokie recommending this be a separate section - we'll just need copy for the question itself, and then is it valid that this can be a PDF or a URL still?

emilyjablonski:
[...]

PR #5621, opened 2025-11-24
feat: partners new marketing flyer section

KrissDrawing:
This PR addresses #5478 

## Description

Adds new marketing flyer subsection in partners that is behind `enableMarketingFlyer`.

Seed data.
For angelopolis in listing `add/edit`-> `appliaction process` -> `Application dates` section. there should be new `marketing flyer` subsection.
test:
- adding same type, mixed type etc.
- editing, partial editing
- deleting

also make sure buildingSelectionCriteria is still exporting properly in csv as i changed helper a bit.

_`MarketingFlyer.tsx` is technically similar to `BuildingSelectionCriteria` but it uses form instead of state (as tbh building selection criteria doesn't populate on edit so it's no good for 2 field form). So it's a bit mix of `BuildingSelectionCriteria` and `Open house` approach. (at least it seem to work fine 😅 )_

## Author Checklist:

## Review Process:


api/test/unit/services/listing.service.spec.ts
```diff
@@ -467,6 +467,11 @@ describe('Testing listing service', () => {
       listingsLeasingAgentAddress: exampleAddress,
       listingsBuildingSelectionCriteriaFile: exampleAsset,
       listingsResult: exampleAsset,
+      marketingFlyer: 'https://example.com/marketing-flyer.pdf',
```
emilyjablonski: **nitpick:** This example of data I don't think would be possible to achieve in the UI - a URL and PDF for both of them?

emilyjablonski (changes requested):
This is looking so so so good!! Thank you for how thorough this is! I see what you mean about it having to merge multiple patterns due to having two fields in the drawer.

My main comment is coming from design after being able to test out the UX in the browser with multiple uploads - I know this wasn't an explicit ask in the ticket, but we think we should show the edit table more similar to how we do photos - because now that we have two uploads in the table, the delete button deleting both of the assets is a little awkward. Could we update the edit table view to show them as individual rows, and they can each have their own delete button as the photos table does?

Example from photos. Each uploaded asset gets its own row and delete button, and the button for `Add marketing flyer` would remain, and just change its text content to `Edit marketing flyer`.
Image

Final comment here is that I'm unable to delete an uploaded PDF from the drawer itself once I've added one (though I can delete a URL by just removing the content).
https://github.com/user-attachments/assets/82413465-a44f-4e56-aa6e-00379a68aeaf

Issue #5651, opened 2025-12-03
Pull option multi-select or exclusion out of options into main preference page 


kathyccheng:
As an admin, I should be choosing the multi-select or exclusive on the main preferences versus each option 

**Acceptance Criteria** 
* When I'm on the add preference screen I should see the question "Are these options multi-select or exclusive" with the option to choose radio buttons multi-select or exclusive underneath the Add option button 
* When I'm on the add option screen then I should no longer see the "Is this option multi-select or exclusive" 

**Notes**
* This option should now also have been moved to the top level of the preference, so should save there 

**Screenshots** 
<img width="1596" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNUpIREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6008d99d54f9fb0ff867d6f891240b2219cb79dd/image.png" alt="image.png" />

<img width="823" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVZIREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--05d07411b8d60792489ff986baf5332329abf9da/image.png" alt="image.png" />

Issue #5650, opened 2025-12-03
Preferences Field Validations MSQ


kathyccheng:
**Acceptance Criteria**
* When I enter any preference field - title, description. options, url, link - that has extra spaces at the end, then it should be trimmed before saving 
* When I enter a preference title that is over 32 characters then I should see an error 
* Verify alphanumeric special characters in preference title - causes problems for excel and CSV export - enforcing validations on UI, just a space shouldn't work, some sort of starts with an alphanumeric

sarahlazarich:
create - yes
edit - yes
retire - yes 
add to listings - yes

Issue #5597, opened 2025-11-18
Update Settings -> Preferences Table


mcgarrye:
- [ ] General
- [ ] Backend
- [ ] Public
- [x] Partners

## Background / Issue
Step 1 of updating our Preference UI. We want the table to be match the components used on the partners listings browse page and the partners users browse page

## Technical Context
**Settings -> Preferences Table**
- Convert table to UIC table used 
- Add `Status` column
- Move `Add Preference` button to the top right
- Change actions buttons to `Edit` or `View` button. When an MSQ is not in `draft` or `visible` state, the edit button is changed to a view button 
- Move copy and delete to inside the modal as `Copy` and `Delete`

**Search and Sorting are not yet implemented on the backend**, so search bar can be hidden for now

**In order to develop this and test along the way with the feature flag**, we might want to create a new settings page with this table 

## Acceptance Criteria
Table should match what is shown in the figma design; exception is column names which should be the following: `NAME`, `STATUS`, `JURISDICTION`, `LAST UPDATED`

## Figma
(V3) https://www.figma.com/design/11nqrUviyvhcXFAuA5XXZY/Preference-Management-Feature-Improvement?node-id=4880-14834&t=iSWErI3L1JYhXwtC-0

## QA Notes

** Please provide as much information as possible (screenshots, documentation, etc) **

Issue #5149, opened 2025-07-24
Update Settings -> Preferences Modals


mcgarrye:
- [ ] General
- [ ] Backend
- [ ] Public
- [x] Partners

## Background / Issue
With the new MSQ model, the UI needs to be updated to reflect the changes, i.e. the status field and status dependent requirements, moving of isExclusive and mapping correctly to new fields.

## Acceptance Criteria
* When I'm on the main preferences table, when the preference state is draft, then I'll see the action edit 
* When I'm on the main preferences table, when the preference state is active or retired, then I'll see the action view
* When I click add or edit for a preference that's in a draft, then I'll see copy and delete in the edit drawer. It's no longer in the preference table 
* When I click add or edit for a preference and the preference is in draft state, then I should see show to partners as the action next to save. Once it's been made public to partners, you'll see hide from partners. This determines whether is preference is available in listing selection. The admin can go back and forth until the preference is active. 
* When I click delete, then the preference will be deleted 
* When I click view for an active listing, then the preferences modal will show the view state below with Copy and Retire as options 
* When I click retire then the preference will change into the state retired 
* When I click view for a retired listing, then the preferences modal will show the view state below with Copy and Return to Active as options 
* When I click return to active, then the preference will change to the state active 
* When I click copy in any state, then it'll generate a copy of the preference 

## Technical Context
**Settings -> Preferences -> Create/Edit Modal & View**
- Move `Is this option multi-select or exclusive?` question to top level --> moved into new ticket 
- Sends data in new format
- If Draft status, user can make visible (Show to Partners), or delete
- If not in Draft or Visible status, user can view (non-editable UI) or retire

**Settings -> Preferences -> Delete Modal**
- True delete if in `draft`
- `toRetire` if `inUse`

## Figma
(V3) https://www.figma.com/design/11nqrUviyvhcXFAuA5XXZY/Preference-Management-Feature-Improvement?node-id=4880-14834&t=iSWErI3L1JYhXwtC-0

<img width="938" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK2hHREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0ba7459bde88db91561ac2492f69d072048a08d5/image.png" alt="image.png" />

[...]

PR #5654, opened 2025-12-03
feat: set up new Preferences page & edit/view drawers (MSQ V2)

jaredcwhite:
This PR addresses #5597, #5149, #5650, #5651

## Description

The PR adds a new, separate screen for viewing/editing Preferences (at the path `/settings/preferences-new`) which allows for managing MSQ V2 style preferences.

To seed your local DB with preferences in V2, run: `yarn setup --msqV2` in the `api` folder

Then you can log in as a Bloomington admin in Partners and access the new page via Settings. Otherwise you'll still see the old page and seed data.

It's possible to view the two seed preferences as well as add new preferences. Unfortunately there's no easy way to flip between Edit and View for the same preference. You could try manually editing the code in the `preferences-new.tsx` for the table. I'm also not sure how easy it is to test various other statuses available in the enum.

Some of the code for editing a preference is copied over and modified from the previous V1 editing, but all of the code for viewing is new, so it's possible I missed some edge cases or something in the UI could be improved. This is what I would call a "first draft".

There are also some follow-up issues which will touch on certain things in here, such as adding search, pagination, and sorting to the new preferences table.

## Author Checklist:

## Review Process:


jaredcwhite:
I don't understand why the Backend Integration test keeps failing. When I run it locally, I do get one failure…but it's a totally different failure in some other file. 😅

sites/partners/src/pages/settings/preferences-new.tsx
```diff
@@ -0,0 +1,201 @@
+import React, { useContext, useState, useMemo } from "react"
```
ludtkemorgan: nit: same as other comment about calling this `preferences-v2.tsx` to follow the naming pattern we have been using for the feature flag

ludtkemorgan (changes requested):
So far looking pretty good functionally. I'm still making my way through the changes but wanted to get my review out for what I have reviewed so far.

Can you also add new frontend tests for this new functionality? We should have a new test suite for the new edit page along with all of the drawers/modals. Let me know if you need any assistance with coming up with test cases for those

sites/partners/src/components/settings/preferences-new/PreferenceEditDrawer.tsx
```diff
@@ -0,0 +1,996 @@
...
+                    <Field
+                      id="name"
+                      name="name"
+                      label={t("t.title")}
+                      placeholder={t("t.title")}
+                      register={register}
+                      type="text"
+                      dataTestId={"preference-title"}
+                      defaultValue={questionData?.name}
+                      errorMessage={`${t("errors.requiredFieldError")}. ${t(
```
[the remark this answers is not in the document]
jaredcwhite: I was wondering about the regex I added for this…it requires an uppercase first letter but maybe I need to relax that. I think that's why you were running into an error here.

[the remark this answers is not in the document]
ludtkemorgan: I also noticed the regex doesn't allow parenthesis but when you copy a preference it defaults to adding `(copy)` to the preference name. I wonder how we should handle that.

This is the current flow:
1. click "edit" or "view" on any preference
2. Click "copy"
3. open the new preference and click "save" or "show to partners"
4. The name field shows an error

Image

api/prisma/seed-staging.ts
```diff
@@ -396,25 +400,85 @@ export const stagingSeed = async (
...
+          status: MultiselectQuestionsStatusEnum.active,
+          name: 'City Employees',
+          description: 'Employees of the local city.',
+          applicationSection:
+            MultiselectQuestionsApplicationSectionEnum.preferences,
+          options: [
+            {
+              name: 'At least one member of my household is a city employee',
+              collectAddress: false,
+              ordinal: 0,
```
ludtkemorgan: nit: the default ordinal should be 1 and not 0. Otherwise these seeded preferences show as ordinal 0 

Image

sites/partners/src/components/settings/preferences-new/PreferenceEditDrawer.tsx
```diff
@@ -0,0 +1,996 @@
...
+    ((optionData?.shouldCollectAddress && watch("shouldCollectAddress") === undefined) ||
+      watch("shouldCollectAddress") === YesNoEnum.yes) &&
+    isAdditionalDetailsEnabled
+  const isValidationRadiusVisible =
+    profile?.jurisdictions.find((juris) => juris.id === watch("jurisdictionId"))
+      ?.enableGeocodingRadiusMethod ||
+    profile?.jurisdictions.every((juris) => juris.enableGeocodingRadiusMethod)
+  const radiusExpand =
+    (optionData?.validationMethod === ValidationMethodEnum.radius &&
+      watch("validationMethod") === undefined) ||
```
ludtkemorgan: suggestion: looks like you are doing `watch("validationMethod")` 4 times here. Could it be a variable created above all of these?

jaredcwhite:
@ludtkemorgan Regarding tests, I'm thinking of adapting the unit/component-level tests to the new pages, and leaving Cypress alone for now. I feel like maintaining e2e tests for two different DB schemas would be really complicated, so maybe we can revisit if/how to do it once we sunset V1.

ludtkemorgan (approved):
After moving the translation strings to the correct file this is good!

api/src/services/multiselect-question.service.ts
```diff
@@ -295,49 +295,50 @@ export class MultiselectQuestionService {
...
-                    };
-                  }),
-                },
-              }
-            : undefined,
-        },
-        include: includeViews.fundamentals,
-      });
+    const rawMultiselectQuestion =
+      await this.prisma.multiselectQuestions.create(finalCreateData);
```
mcgarrye: **question:** Any particular reason for this change?

sites/partners/__tests__/pages/settings/preferences-v2.test.tsx
```diff
@@ -0,0 +1,295 @@
...
+import {
+  listing,
+  multiselectQuestionPreferenceV2,
+} from "@bloom-housing/shared-helpers/__tests__/testHelpers"
+import {
+  FeatureFlagEnum,
+  MultiselectQuestionsStatusEnum,
+} from "@bloom-housing/shared-helpers/src/types/backend-swagger"
+import { mockNextRouter, render } from "../../testUtils"
+import SettingsPreferences from "../../../src/pages/settings/preferences-v2"
```
mcgarrye: **nitpick:** alphabetize imports

sites/partners/src/pages/settings/preferences-v2.tsx
```diff
@@ -0,0 +1,199 @@
...
+import { Button, Tag } from "@bloom-housing/ui-seeds"
+import dayjs from "dayjs"
+import Head from "next/head"
+import Layout from "../../layouts"
+import { NavigationHeader } from "../../components/shared/NavigationHeader"
+import { useJurisdictionalMultiselectQuestionList } from "../../lib/hooks"
+import EditPreference, { DrawerType } from "../../components/settings/preferences-v2/EditPreference"
+import styles from "./preferences-v2.module.scss"
+import TabView from "../../layouts/TabView"
+import { getSettingsTabs, SettingsIndexEnum } from "../../components/settings/SettingsViewHelpers"
```
mcgarrye: **nitpick:** alphabetize imports

sites/partners/src/components/settings/SettingsViewHelpers.tsx
```diff
@@ -8,12 +8,21 @@ export enum SettingsIndexEnum {
...
+) => {
   const baseUrl = "/settings/"
+
   return (
     <Tabs
       verticalSidebar
-      onSelect={(index) => void router.push(`${baseUrl}/${SettingsIndexEnum[index]}`)}
+      onSelect={(index) => {
+        let subpath = SettingsIndexEnum[index]
+        if (v2Preferences && subpath === "preferences") subpath = `${subpath}-v2`
```
mcgarrye: **suggestion:** can we have the url be `multiselectquestions-preferences`?
```suggestion
        if (v2Preferences && subpath === "preferences") subpath = `multiselectquestions-${subpath}`
```

mcgarrye:
https://github.com/user-attachments/assets/ae2d1b1f-6bf0-4604-abe3-6b7d196a5565

I'm seeing flashes of the old table when loading the page, is there a way to avoid that? Out of my depth on this level of frontend lol

Issue #4386, opened 2024-10-09
Dependency - Upgrade ag-grid-community & ag-grid-react


emilyjablonski:
We have two dependabot security alerts on these packages: https://github.com/bloom-housing/bloom/security/dependabot?q=package%3Aag-grid-community+manifest%3Ayarn.lock+has%3Apatch

We need to upgrade a major version(s) though and it looks like the API has changed slightly.

This will first require an update to the AgTable component in UIC.

PR #5669, opened 2025-12-04
feat: data table

emilyjablonski:
This PR addresses #4386

## Description

We are 8 major versions behind on [ag-grid](https://www.ag-grid.com/), and they seem best suited to complex tables that need to be highly performant. It comes with lots of cool features but they can be harder to adjust. It is, relatively in our ecosystem, a very large package.

This PR creates a new DataTable component using [TanStack Table](https://tanstack.com/table/latest) and transitions just the listings page to use the DataTable component instead of the AgGrid component from UIC. TanStack Table is headless and very lightweight. The documentation was excellent and it was very easy to setup.

I know we chatted about the potential of moving off Next, and after this exploration I think TanStack Start is a really intriguing option. I'm also exploring TanStack Form!

You can use the [deploy preview](https://deploy-preview-5669--partners-bloom-dev.netlify.app/sign-in).
These are the main pieces of functionality in the listings table to test:
1. Pagination
2. Sorting
3. Page size
4. Search
5. Loading state

After this merges in, we should transition our other tables one by one to this new table.

## Author Checklist:

## Review Process:


mcgarrye:
Screenshot 2026-04-08 at 1 39 43 PM
Screenshot 2026-04-08 at 1 39 48 PM
Is the change from the 3 character message appearing on type to being static intentional? I personally like the appearing message but it's fine either way

mcgarrye (approved):
Looks great to me! Just left some clarifying questions

emilyjablonski:
@mcgarrye I don't think we need to get the header changes approved by clients - open to other suggestions on naming, was just trying to remove unneeded language to get back some more space. I'm not sure why we haven't been in filtering on other columns, but those would be backend changes so weren't included here. And we have sign-off to remove reordering!

sites/partners/src/pages/index.tsx
```diff
@@ -378,82 +135,298 @@ export default function ListingsList() {
...
+          plaintextName: t("listings.createdDate"),
+        },
+      }),
+      columnHelper.accessor("scheduledPublishAt", {
+        id: "scheduledPublishAt",
+        cell: (props) =>
+          props.getValue()
+            ? dayjs.utc(props.getValue() as string).format("MM/DD/YYYY")
+            : t("t.none"),
+        header: () => t("listings.scheduledListingPublishDateShort"),
```
ludtkemorgan, line 236: Suggestion: Similar to Eric's comment. I like having it shortened so that it takes up less space but Having "Publish Date" and "Published Date" is a bit confusing. Could we switch "Published Date" to "Date Published"? 

Image

sites/partners/src/components/shared/DataTable.tsx
```diff
@@ -0,0 +1,541 @@
+import React, { useEffect, useState } from "react"
```
ludtkemorgan, line 1: suggestion: Is there a way for initial load to have the "loader" not appear within the table and instead have the spinner load where the table should be. It's a big weird to see a gray square before the data loads. Especially when you can see where the footer will go

Image

sites/partners/src/components/shared/DataTable.tsx
```diff
@@ -0,0 +1,541 @@
...
+            {dataQuery.data?.totalItems === 1
+              ? t("listings.totalListing")
+              : t("listings.totalListings")}
+          </span>
+
+          <span className={styles["current-page"]}>
+            Page {dataQuery.data?.currentPage} of{" "}
+            {dataQuery.data?.totalItems &&
+              Math.ceil((dataQuery.data?.totalItems || 0) / dataQuery.data?.itemsPerPage)}
+          </span>
```
ludtkemorgan, line 391: question: It's nice to know how many pages we have but with this we are losing the ability to jump to any page. Was that intentional?

Image

vs

Image

sites/partners/src/components/shared/DataTable.tsx
```diff
@@ -0,0 +1,541 @@
...
+            <select
+              value={table.getState().pagination.pageSize}
+              onChange={(e) => {
+                setPagination((prev) => ({ ...prev, pageIndex: 0 }))
+                table.setPageSize(Number(e.target.value))
+              }}
+              className={styles["show-select"]}
+              id={"show-numbers"}
+            >
+              {/* TODO: Remove 3 before merge, just for easier pagination testing purposes */}
```
ludtkemorgan: issue: 🧹

ludtkemorgan (commented):
This is looking good. No major issues found. Just a personal preference around the naming of the headers and the loader. Let me know if that doesn't make sense

sites/partners/src/components/shared/DataTable.tsx
emilyjablonski: it was! this was many moons ago, but i did a few QA sessions with Em and given the shape of the data, the thought was that jumping to a specific page was not super useful

sites/partners/src/components/shared/DataTable.tsx
emilyjablonski: open to changes! the tricky part is if the loading state doesn't look like this in terms of having a predictable height, the whole page section can flash and jump which going between pages, or jump down on page load - but i know there are pros and cons to both

sites/partners/src/pages/index.tsx
emilyjablonski: absolutely!

mcgarrye: I think "Publish date" and "Date Published" would is what Morgan meant

ludtkemorgan (approved):

PR #5678, opened 2025-12-09
fix: filter ami charts by jurisdiction in draft listing

emilyjablonski:
This PR addresses #5679

## Description

AMI charts in the partners unit drawer are not filtering by jurisdiction on a draft listing. This means you can see every available AMI chart, save a unit with a chart that does not belong to the jurisdiction, and then when saving and re-opening the unit drawer, the chart selection is empty.

We should always filter the AMI charts by the pre-selected jurisdiction.

Create a draft listing and ensure the AMI charts in the units form are only those that belong to that jurisdiction.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/prisma/seed-staging.ts
+++ b/api/prisma/seed-staging.ts
@@ -344,19 +344,34 @@ export const stagingSeed = async (
   });
   // build ami charts
   const amiChart = await prismaClient.amiChart.create({
-    data: amiChartFactory(10, mainJurisdiction.id),
+    data: amiChartFactory(10, mainJurisdiction.id, null, mainJurisdiction.name),
   });
   await prismaClient.amiChart.create({
-    data: amiChartFactory(10, mainJurisdiction.id, 2),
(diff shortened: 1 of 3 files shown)


ludtkemorgan:
@emilyjablonski It looks like this issue is in Doorway staging as well with the latest changes. I don't think we should release without this fix. Once it is merged into main can you also create a cherry-pick PR into Doorway?

api/prisma/seed-staging.ts
```diff
@@ -355,6 +355,9 @@ export const stagingSeed = async (
   await prismaClient.amiChart.create({
     data: amiChartFactory(8, lakeviewJurisdiction.id, 2),
   });
+  await prismaClient.amiChart.create({
```
ludtkemorgan: suggestion: I wonder if we should update the seeded AMI chart names so it's easy to tell what jurisdiction they are from. Maybe all start with the jurisdiction name followed by a random word?

PR #5682, opened 2025-12-09
refactor: update generate translations

emilyjablonski:
## Description

Updates our generate machine translations script to make it easier - previously we needed to first run the missing translations script, then manually add each language one by one into a CSV and pass it into the machine translations script. Now this script does both.

In development, you can just add the English keys in the general file and then run this script.

Hoping this will help our translations dev efficiency!

There are instructions in the file, but add the google translate keys, and from within the shared-helpers script folder run `ts-node generate-machine-translations.ts > translated-strings.json`.

You'll need to manually remove some non-English keys to see the appropriate output.

I will update the existing Notion docs for these scripts post-merge.

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/src/scripts/get-machine-translations.ts
+++ b/shared-helpers/src/scripts/get-machine-translations.ts
@@ -1,9 +1,11 @@
 /* eslint-disable @typescript-eslint/no-var-requires, import/no-unresolved */
 // Finds missing translations and automatically translates them using Google Translate API
 // Prints out translations in the JSON translation file format: "key": "translated string"
-// You will need to add the environment variables for Google Translate API access from the api env file, but ensure you do not commit them!
+// You will need to add the environment variables for Google Translate API access from the api env file into this env file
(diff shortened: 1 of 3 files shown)


shared-helpers/src/scripts/get-machine-translations.ts
```diff
@@ -22,9 +16,10 @@ async function main(argv: string[]) {
     "bn" = "bn",
   }
 
-  const GOOGLE_API_EMAIL = "SECRET_VALUE"
-  const GOOGLE_API_ID = "SECRET_VALUE"
-  const GOOGLE_API_KEY = "SECRET_VALUE"
+  // DO NOT COMMIT THESE VALUES! REMOVE BEFORE COMMITTING
```
YazeedLoonat: **suggestion (non-blocking):** if we add dotenv as an import here we can make these .env variables so we never run the risk of someone committing them 

I think if someone does commit them we would have to rotate keys n stuff until we can have github delete all references to that PR/branch

PR #5690, opened 2025-12-11
fix: move non-regulated selection to jurisdiction modal

ludtkemorgan:

## Description

During the development of the non-regulated listing functionality, a number of edge cases were identified related to users switching their selection between "regulated" and "non-regulated" after partially completing the form. This is particularly problematic for unit data and could result in corrupted data, leading to listings that are a hybrid of regulated and non-regulated states.

To prevent this problem we can make the listing type selection earlier in the flow utilizing the new jurisdiction selection modal.

image

If the user only has access to one jurisdiction and that jurisdiction has `enableNonRegulatedListings` turned on they will only see the Listing Type question
image

### Multiple jurisdictions
1. Seed the staging data
2. Log into partners as an admin and click "add listing" on the listings page
3. The select jurisdiction modal should appear without the "What kind of listing is this?"  showing immediately
4. Select any jurisdiction besides "Lakeview" and the "What kind of listing is this?" question shouldn't show
5. Select "Lakeview" and the "What kind of listing is this?" question should show
6. Select "Non-regulated" and "Get started"
7. Non-regulated should be shown and you can't change it

### One jurisdiction
With the default seeded data this can be hard to test without making some changes since jurisdictional admin is disabled for Lakeview. You can do one of the following steps:
1. Enable jurisdictional admin for lakeview and create that a jurisdictional admin for that jurisdiction
2. Turn on `enableNonRegulatedListings` for a different jurisdiction (Note they will also need to have unit groups enabled)

Once the above is done you should be able to go through the "add listing" flow and just see the one question on the modal

### additional testing
The following should also be tested
* Create a "Regulated" listing
* Open already created listings and correct listing type should be selected (or not shown if enableNonRegulatedListings is turned off)

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/src/pages/index.tsx
+++ b/sites/partners/src/pages/index.tsx
@@ -431,7 +431,7 @@ export default function ListingsList() {
 
           <Dialog.Content id="listing-select-dialog-content">
             {t("listings.selectJurisdictionContent")}
-            <Grid spacing={"lg"}>
+            <Grid>
               <Grid.Row columns={3}>
                 <Grid.Cell className={"seeds-grid-span-2"}>
                   <div className={`${defaultJurisdiction ? "hidden" : ""} seeds-m-bs-4`}>
@@ -465,8 +465,8 @@ export default function ListingsList() {
                   </div>
                 </Grid.Cell>
               </Grid.Row>
-              <div aria-live="polite">
-                {isNonRegulatedEnabled && (
+              {isNonRegulatedEnabled && (
+                <div aria-live="polite">
                   <fieldset>
                     <legend className={`text__caps-spaced`}>
                       {t("listings.listingTypeTitle")}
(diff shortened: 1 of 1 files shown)


sites/partners/src/pages/index.tsx
```diff
@@ -420,6 +466,83 @@ export default function ListingsList() {
                   </div>
                 </Grid.Cell>
               </Grid.Row>
+              {isNonRegulatedEnabled && (
```
ludtkemorgan: @emilyjablonski here is the version using two "Field" and below is the commented out version with the "FieldGroup".  In that version the description is within an expandable section

emilyjablonski: Looks good! Before is `isNonRegulatedEnabled` line, could we just wrap this entire section in a div with `aria-live="polite"` so that the section appearing is announced?

emilyjablonski (commented):
Nit non-blocking (because smaller screen size on partners) we could add some padding between the options when it switches to vertical
Image


Could we have the order in which the questions appear be the same between the detail and edit pages? Imagining on the detail page we would move up the listing type question
Image
Image

emilyjablonski (approved):
This lgtm! The last piece is I think we just introduced some unnecessary padding underneath the dropdown.
PR:
Image
Prod:
Image

PR #5693, opened 2025-12-15
5577/property section backend

matzduniuk:
This PR addresses #5577 

## Description

* Updates the Prisma schema to include a new `Properties` table
* Adds a new migration file setting up the property table
* Adds Properties table related controllers, services, and dtos required to handle basic CRUD operations, including:
  * Retrieving the list of properties using a pagination query params or using a name search filter
  * Retrieving a single property by its ID
  * Creating a new Property entry 
  * Updating and deleting an existing entry by its ID
* Adds interation e2e test for the Properties controller available endpoints
* Add a new endpoint to the listings controller to retrieve listings by assigned property ID

1. Run the `yarn setup` command to apply the new migration and set up a `Properties` table
2. Run the backend application
3. Using Postman or a similar tool, perform the following operations:
   2a. Create at least two new property entries using the **POST** `/properties` endpoint.
   2b. Retrieve all created properties using the **GET** `/properties` endpoint (this step should also be repeated using different query parameters and search filters)
   2c. Using a generated ID copied from the previous query, retrieve a single property entry using the **GET** `/properties/<ID>` endpoint.
    2d. Using the same ID, perform the update operation using the **PUT** `/properties` endpoint. (The request will return an updated entry, but step `2c` can be repeated to ensure that changes have been properly applied) 
    2e. Using the same ID, perform the delete operation **DELETE** `/properties` endpoint.
> [!TIP]
> The above steps should also be performed using invalid data, like missing DTO fields, invalid URLs, invalid IDs etc. to ensure proper error and exception handling
4. Open the Prisma Studio view and assign the properties to example listings (copy the ID of one assigned property)
5. Using Postman, target the **GET** 'listings/byProperty/<ID>' endpoint to retrieve all listings that have been assigned to the given property.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/__tests__/components/listing/listing_sections/LeasingAgent.test.tsx
+++ b/sites/public/__tests__/components/listing/listing_sections/LeasingAgent.test.tsx
@@ -8,6 +8,7 @@ import {
 import { listing } from "@bloom-housing/shared-helpers/__tests__/testHelpers"
 import { AuthContext } from "@bloom-housing/shared-helpers"
 import { screen } from "../../../testUtils"
+import * as helpers from "../../../../src/lib/helpers"
 
 afterEach(cleanup)
 
@@ -27,12 +28,9 @@ describe("<LeasingAgent>", () => {
 
   it("shows all content enableLeasingAgentAltText on", () => {
     const phoneNumber = "(123) 456-7890"
+    jest.spyOn(helpers, "isFeatureFlagOn").mockReturnValue(true)
     render(
-      <AuthContext.Provider
-        value={{
-          doJurisdictionsHaveFeatureFlagOn: () => true,
-        }}
-      >
+      <AuthContext.Provider value={{}}>
         <LeasingAgent
           listing={{
             ...listing,
@@ -86,12 +84,9 @@ describe("<LeasingAgent>", () => {
   })
   it("shows all content enableLeasingAgentAltText off", () => {
     const phoneNumber = "(123) 456-7890"
+    jest.spyOn(helpers, "isFeatureFlagOn").mockReturnValue(false)
     render(
-      <AuthContext.Provider
-        value={{
-          doJurisdictionsHaveFeatureFlagOn: () => false,
-        }}
-      >
+      <AuthContext.Provider value={{}}>
         <LeasingAgent
           listing={{
             ...listing,
--- a/sites/public/src/components/listing/ListingView.tsx
+++ b/sites/public/src/components/listing/ListingView.tsx
@@ -98,9 +98,9 @@ export const ListingView = (props: ListingProps) => {
     FeatureFlagEnum.disableListingPreferences,
     listing?.jurisdictions?.id
   )
-  const enableLeasingAgentAltText = doJurisdictionsHaveFeatureFlagOn(
-    FeatureFlagEnum.enableLeasingAgentAltText,
-    listing.jurisdictions.id
+  const enableLeasingAgentAltText = isFeatureFlagOn(
+    props.jurisdiction,
+    FeatureFlagEnum.enableLeasingAgentAltText
   )
 
   const leasingAgentContactText = enableLeasingAgentAltText
--- a/sites/public/src/components/listing/ListingViewSeeds.tsx
+++ b/sites/public/src/components/listing/ListingViewSeeds.tsx
@@ -265,7 +265,7 @@ export const ListingViewSeeds = ({ listing, jurisdiction, profile, preview }: Li
       {LotteryEvent}
       {ReferralApplication}
       {WhatToExpect}
-      <LeasingAgent listing={listing} />
+      <LeasingAgent listing={listing} jurisdiction={jurisdiction} />
       {ListingUpdatedAt}
     </>
   )
--- a/sites/public/src/components/listing/listing_sections/LeasingAgent.tsx
+++ b/sites/public/src/components/listing/listing_sections/LeasingAgent.tsx
@@ -1,9 +1,14 @@
 import * as React from "react"
 import { Card, Heading, Link } from "@bloom-housing/ui-seeds"
-import { Address, AuthContext } from "@bloom-housing/shared-helpers"
+import { Address } from "@bloom-housing/shared-helpers"
 import { t } from "@bloom-housing/ui-components"
 import styles from "../ListingViewSeeds.module.scss"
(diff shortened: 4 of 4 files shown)


api/prisma/schema.prisma
```diff
@@ -736,6 +737,10 @@ model Listings {
   Listings                              Listings[]                    @relation("copy_of")
   lastUpdatedByUserId                   String?                       @map("last_updated_by_user_id") @db.Uuid
   lastUpdatedByUser                     UserAccounts?                 @relation("last_updated_by_user", fields: [lastUpdatedByUserId], references: [id], onDelete: NoAction, onUpdate: NoAction)
+  // START LA SPECIFIC
```
YazeedLoonat: **nitpick:** lets remove these references, I don't think we need to specify this is for LA

api/src/controllers/property.controller.ts
```diff
@@ -0,0 +1,139 @@
...
+@UseGuards(OptionalAuthGuard)
+@ApiExtraModels(
+  PropertyCreate,
+  PropertyUpdate,
+  PropertyQueryParams,
+  PropertyFilterParams,
+  PaginationMeta,
+  IdDTO,
+)
+@PermissionTypeDecorator('property')
```
YazeedLoonat: **question:** this would get this to leverage the old property permissions from `permission_policy.csv` is that correct or do those perms need to be updated?

api/src/dtos/properties/property-update.dto.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export class PropertyUpdate extends OmitType(Property, [
+  'createdAt',
+  'updatedAt',
+  'name',
+]) {
+  @Expose()
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @ApiPropertyOptional()
+  name?: string;
```
YazeedLoonat: **question:** since name is required in the db why is it optional here?

api/src/modules/property.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import { PropertyController } from '../controllers/property.controller';
+import { PropertyService } from '../services/property.service';
+import { PermissionService } from '../services/permission.service';
+
+@Module({
+  controllers: [PropertyController],
+  providers: [PropertyService, PermissionService],
```
YazeedLoonat: **issue (blocking):** since we have a permissionModule already you should pass that along instead of instantiating a brand new one

I think you'll also need the PrismaModule here as well

api/src/services/listing.service.ts
```diff
@@ -2744,6 +2759,26 @@ export class ListingService implements OnModuleInit {
     return mapTo(Listing, listingsRaw);
   };
 
+  findListingsWithPorperty = async (propertyId: string) => {
```
YazeedLoonat: **nitpick:** findListingsWithPorperty -> findListingsWithProperty

shared-helpers/src/types/backend-swagger.ts
```diff
@@ -406,7 +406,7 @@ export class ListingsService {
   /**
    * Get listings by multiselect question id
    */
-  retrieveListings(
+  retrieveListingsByMsq(
```
YazeedLoonat: **issue:** you're going to need to make a change in the `useListingsMultiselectQuestionList` at least on the ui side of things as this is a breaking change

api/src/services/property.service.ts
```diff
@@ -0,0 +1,258 @@
...
+      throw new NotFoundException(
+        `property with id ${propertyId} was requested but not found`,
+      );
+    }
+
+    return mapTo(Property, propertyRaw);
+  }
+
+  async create(propertyDto: PropertyCreate, requestingUser: User) {
+    const rawJurisdiction = await this.prisma.jurisdictions.findFirstOrThrow({
```
YazeedLoonat: **issue:** the errors prisma generates are very hard to parse in production environments. we should instead try a findOne and then throw our own error if we don't (there are examples of this in listings and the other services as well)

api/src/services/property.service.ts
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  async deleteOne(propertyId: string, requestingUser: User) {
+    if (!propertyId) {
+      throw new BadRequestException('a property ID must be provided');
+    }
+
+    const propertyData = await this.findOrThrow(propertyId);
+
+    const rawJurisdiction = await this.prisma.jurisdictions.findFirstOrThrow({
```
YazeedLoonat: **issue:** same as above

YazeedLoonat (changes requested):
Overall looks good, some questions and some requested changes

a must have on this would be unit tests on the service functions

a nice to have would be new permission e2e tests in the permission suites

api/src/services/property.service.ts
```diff
@@ -0,0 +1,258 @@
...
+  async list(params: PropertyQueryParams): Promise<PaginatedPropertyDto> {
+    const whereClause = this.buildWhere(params);
+
+    const count = await this.prisma.properties.count({
+      where: whereClause,
+    });
+
+    let page = params.page;
+
+    if (count && params.limit && params.limit !== 'all' && params.page > 1) {
```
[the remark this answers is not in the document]
matzduniuk: Those helpers are used later in the code, but none of those helpers verify that the given page from the request does not overshoot available pages

api/src/controllers/property.controller.ts
```diff
@@ -0,0 +1,130 @@
...
+@ApiTags('properties')
+@ApiExtraModels(
+  PropertyCreate,
+  PropertyUpdate,
+  PropertyQueryParams,
+  PaginationMeta,
+  IdDTO,
+)
+@PermissionTypeDecorator('properties')
+@UseGuards(JwtAuthGuard, PermissionGuard)
```
YazeedLoonat: **nitpick:** sorry for missing this earlier we should add the APIKeyGuard here as well

api/src/controllers/property.controller.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  @Get(':id')
+  @ApiOperation({
+    summary: 'Get a property object by ID',
+    operationId: 'getById',
+  })
+  public async getPropertyById(
+    @Param('id', new ParseUUIDPipe({ version: '4' })) propertyId: string,
+  ) {
+    return this.propertyService.findOne(propertyId);
```
YazeedLoonat: **nitpick:** missing await

Issue #5677, opened 2025-12-09
Building features - Smoking policy to radio


emilyjablonski:
LAHD would like the Smoking policy field under `Building features` to be a radio field instead of a text area. This should be behind a feature flag `enableSmokingPolicyRadio` and it should stay optional. This toggle should only be on in LA / Angelopolis.

The title of the field should stay `Smoking policy`. The three radio options should be `No smoking allowed`, `Smoking allowed`, and `Policy unknown`

We can keep the field in the backend as a text entry field, and just store the value from the radio buttons into the text field. The default should be `Policy unknown` if there is no data in the field, and this value should save as null, not as `Policy unknown`. This should automatically dynamically translate the fields on the public site, but we should ensure they transate.

AC: 
- New values reflected on partners 
- New values reflected on public 
- New values reflected on export only if all jurisdictions have the flag on

PR #5694, opened 2025-12-15
fix: add smoking policy radio

KrissDrawing:
This PR addresses #5677 

## Description

Adds `enableSmokingPolicyRadio` flag that adds radio instead of textfield for smoking policy

On Angelopolis add / edit new listing it should have `Policy unknown` as default selected. Now we should be able to switch that and see it on details page, and public page. I assumed that `Policy unknown` just won't show that section on public, as it worked before for `null` (not sure if that is correct).
For csv export it will export as `Policy unknown` just when all jurisdiction have flag on

## Author Checklist:

## Review Process:


emilyjablonski (approved):
Beautiful!! Works perfectly.

I know this wasn't in the ticket, but showed it to Em, would we be able to just swap `Services offered` with `Smoking policy` regardless of the toggle value, so that in the case the toggle is on, the radio breaks up the text areas just a wee bit less? 

Other than that looks good to merge!

Image

Issue #5676, opened 2025-12-09
New Field - Parking fee


emilyjablonski:
LA wants us to add a new parking-related field to the building features section. This field should be under a new feature toggle `enableParkingFee` and it should only be on in LA / Angelopolis. We will need to add a new field to the listing DTO.

In partners, it should be an optional currency field (like Application fee) under `Building features`.	

In public, it should live with the other building features under the Features heading at the bottom. It should appear similar to `Built`, and the value should be prepended with a $. The heading should be `Parking fee` and we will need to add manual google translations for the heading.

AC:
- Field shows on partners site
- Field shows on public site 
- Field translates on public site
- Field shows in export if every jurisdiction has the toggle enabled

PR #5709, opened 2025-12-19
feat: New Field - Parking fee

millord:
This PR addresses #5676

## Description

The purpose of this ticket was - LA wanted us to add a new parking-related field to the building features section. This field should be under a new feature toggle enableParkingFee and it should only be on in LA / Angelopolis. We will need to add a new field to the listing DTO.

Partners

1-Navigate to the partners site
2-Click on add a listing and select any jurisdiction
3-In the listing creating page look for the Building features section
4-at the bottom of that section you should see the new Parking fee field

image

Public

1-Create a listing in the partners site and add a parking fee to it (is steps above to achieve that)
2-Navigate to the public site
3-Select the listing you created before to navigate to the details page
4-Click on the features toggle (+)
5-You should see the new Parking fee value you previously added

image

## Author Checklist:

## Review Process:


api/prisma/seed-staging.ts
```diff
@@ -111,6 +112,7 @@ export const stagingSeed = async (
         FeatureFlagEnum.enableMarketingStatus,
         FeatureFlagEnum.enableNeighborhoodAmenities,
         FeatureFlagEnum.enableNonRegulatedListings,
+        FeatureFlagEnum.enableParkingFee,
```
emilyjablonski: **issue:** This flag is only supposed to be on in Angelopolis, so we can remove it from the other seeds

shared-helpers/src/locales/general.json
```diff
@@ -1132,6 +1132,7 @@
   "t.order": "Order",
   "t.orUpTo": "or up to",
   "t.other": "Other",
+  "t.parkingFee": "Parking Fee",
```
emilyjablonski: **issue:** This should be sentence case `Parking fee` - we'll need to update the translations for this as well

sites/partners/src/components/listings/PaperListingForm/sections/BuildingFeatures.tsx
```diff
@@ -36,6 +42,11 @@ const BuildingFeatures = (props: BuildingFeaturesProps) => {
     }
   }, [props.enableAccessibilityFeatures, setValue])
 
+  const enableParkingFee = doJurisdictionsHaveFeatureFlagOn(
```
emilyjablonski: **issue:** The pattern in these files is to pass feature flags in as props so that the files don't need to duplicate the contexts (similar to the existing `enableAccessibilityFeatures` prop)

Issue #5754, opened 2026-01-09
Filter by LA regions


emilyjablonski:
As listed in the [Browser / Filter PRD](https://www.notion.so/exygy/Browse-Filter-Improvements-PRD-25b3433f03d080f18886e68543dac623), we need to be able to filter by the following regions:

Antelope Valley, San Fernando Valley & Santa Clarita Valley, San Gabriel Valley, Metro Los Angeles, West Los Angeles, South Los Angeles, East Los Angeles, South Bay / Harbor

We will be able to re-use parts of our existing regions feature, which is a dropdown in partners, and we can already filter by these values!

The regions should appear in partners and have the correct values for Angelopolis / LA. If set, the value should appear on a public listing where it already does under the Neighborhood heading, as well as in the filter drawer. These values should never translate.

Currently under the existing region toggle as defined and use in Detroit, Neighborhood automatically becomes a dropdown when the region toggle is enable. However, Neighborhood should stay a text input for LA. We also want to be able to more dynamically configure regions in LA than we have done for Detroit. Detroit's regions are defined directly n core as below which we want to avoid:
```
export enum RegionEnum {
  "Greater_Downtown" = "Greater_Downtown",
  "Eastside" = "Eastside",
  "Southwest" = "Southwest",
  "Westside" = "Westside",
}
```
We will implement this under a new regions v2 flag that is separate from the existing regions toggle. We will add a text array configuration on the jurisdiction for regions. This will enable us to refactor Detroit's region and neighborhood issues separately as it might be hairy - here is that ticket! https://github.com/bloom-housing/bloom/issues/5769

emilyjablonski:
Flagging that our URL filters don't currently work with an `&` - this is not a big fix but I did not do this here, and changed the value to `San Fernando Valley and Santa Clarita Valley` - lmk if that's a concern!

Issue #4989, opened 2025-06-11
Typing is broken in test files


emilyjablonski:
Currently, all of our public/partners unit tests have typing broken in our IDEs. It was caused (weirdly) by removing an [unused import from the backend](https://github.com/bloom-housing/bloom/pull/4851/files#diff-94132bc0a9b2779baf3386a72d6a075431dae7d367f37e6bb89bd4389d764f85L11). It makes tests very difficult to write unless you manually add back in the import, ensuring you remove/add it for every commit.

PR #5772, opened 2026-01-14
feat: configurable regions

emilyjablonski:
This PR addresses #5754 #4989

## Description

Adds a new feature flag for configuration regions on a jurisdiction. It should appear on partners, on the public detail page, and in the public filter drawer.

Every time for the last few months when I need to write unit tests, I have to manually go into the script runner backend file to add back in the test import that for whatever reason breaks test typing on the frontend. I added that back in and linked the ticket, and as a result lots of typing issues in tests were more easily surfaced, so there are a number of trivial type issue fixes in various test files on partners and public, in addition to unnecessary usages of AuthContext.

Reseed, and Angelopolis will have a generic set of regions set on the jurisdiction. When creating a partner listing, these new region values should populate. We should be able to save them and see them on the detail page. Once saved the data should re-load appropriately on the edit page again. On a published listing, it should appear under the Neighborhood heading. On the public drawer, we should be able to filter by these new regions. They should never translate.

To ensure no regressions, I also tested the region feature in Lakeview, as nothing should have changed.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -102,7 +102,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^4.0.0",
     "jest": "~29.6.2",
-    "jest-mock-extended": "^3.0.7",
+    "jest-extended": "^7.0.0",
     "prettier": "^2.3.2",
     "supertest": "^6.1.3",
     "ts-jest": "~29.1.1",
@@ -127,7 +127,10 @@
       "**/*.(t|j)s"
     ],
     "coverageDirectory": "../coverage",
-    "testEnvironment": "node"
+    "testEnvironment": "node",
+    "setupFilesAfterEnv": [
+      "jest-extended/all"
+    ]
   },
   "prisma": {
     "seed": "ts-node prisma/seed.ts"
--- a/api/test/integration/property.e2e-spec.ts
+++ b/api/test/integration/property.e2e-spec.ts
@@ -119,9 +119,10 @@ describe('Properties Controller Tests', () => {
         expect.arrayContaining([
           expect.objectContaining({
             name: expect.anything(),
-            description: expect.anything(),
-            url: expect.anything(),
-            urlTitle: expect.anything(),
+            description: expect.toBeOneOf([expect.any(String), null]),
+            url: expect.toBeOneOf([expect.any(String), null]),
+            urlTitle: expect.toBeOneOf([expect.any(String), null]),
+            jurisdictions: expect.anything(),
           }),
         ]),
       );
--- a/api/test/jest-e2e.config.js
+++ b/api/test/jest-e2e.config.js
@@ -11,4 +11,5 @@ module.exports = {
       },
     ],
   },
+  setupFilesAfterEnv: ['jest-extended/all'],
 };
--- a/api/test/jest.config.js
+++ b/api/test/jest.config.js
@@ -11,4 +11,5 @@ module.exports = {
       },
     ],
   },
+  setupFilesAfterEnv: ['jest-extended/all'],
 };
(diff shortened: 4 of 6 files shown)


api/prisma/seed-helpers/listing-data/hollywood-hills-heights.ts
```diff
@@ -1,12 +1,12 @@
...
   ApplicationMethodsTypeEnum,
   ListingsStatusEnum,
+  Prisma,
   ReviewOrderTypeEnum,
 } from '@prisma/client';
-import dayjs from 'dayjs';
 import { yellowstoneAddress } from '../address-factory';
 
-export const hollywoodHillsHeights = {
+export const hollywoodHillsHeights: Prisma.ListingsCreateInput = {
```
emilyjablonski: **note:** Updated this listing to be in the Angelopolis jurisdiction so one already exists after seed like our other jurisdictions

sites/partners/src/components/listings/PaperListingForm/sections/BuildingDetails.tsx
```diff
@@ -320,41 +338,60 @@ const BuildingDetails = ({
...
-                    t("listings.yearBuilt"),
-                    requiredFields,
-                    errors,
-                    clearErrors
-                  )}
-                />
-              )
+            )}
+            {showInitialRegion && enableConfigurableRegions && (
+              <Select
```
ludtkemorgan: I'm seeing a weird screen behavior when I change regions. The page flashes and moves the content up and down. However if I have the region at a specific height within the page it doesn't do it. This video shows both cases 

https://github.com/user-attachments/assets/7263e867-2594-41c4-a6ff-b4ec79ce0668

ludtkemorgan: After a little more digging it might be because changing regions triggers a re-render and the listing images are reloaded causing a temporary smaller page as the images don't exist for a split second. Don't know how to solve that

sites/partners/__tests__/pages/listings/[id]/index.test.tsx
```diff
@@ -375,7 +375,7 @@ describe("listing data", () => {
       it("should display Building Details section - without region", () => {
         render(
           <ListingContext.Provider value={{ ...listing, region: RegionEnum.Southwest }}>
-            <DetailBuildingDetails />
+            <DetailBuildingDetails enableRegions={false} enableConfigurableRegions={false} />
```
Comment (ludtkemorgan) [inline sites/partners/__tests__/pages/listings/[id]/index.test.tsx]: nit: Both of these `<DetailBuildingDetails>` tests in this file are no longer needed with the addition of `DetailBuildingDetails.test.tsx`. 

Related but not necessary for this pr: The tests in this file in general can be split out into their own test files based on the component they are testing. we have done that with a few components but not all of them.

sites/partners/src/components/listings/PaperListingForm/sections/BuildingDetails.tsx
emilyjablonski: I can't repro this at the same screen dimensions on that seeded listing & would love to pair real q next week 🫡

sites/partners/__tests__/pages/listings/[id]/index.test.tsx
Comment (ludtkemorgan) [inline sites/partners/__tests__/pages/listings/[id]/index.test.tsx]: chore: the unused imports from this file can be cleaned up

emilyjablonski:
Screenshot 2026-01-16 at 1 21 55 PM
@ludtkemorgan Curious what you think about this - this would only be true for the core experience, any jurisdiction with the flag on would only have one, and given we are planning to merge them, I don't think there's ever going to be a jurisdiction that has both

PR #5798, opened 2026-01-27
test: typed and dynamic listing cypress test

emilyjablonski:
## Description

As a stepping stone to be able to easily test an Angelopolis listing in Cypress based on the new feature flagged fields / features, this PR types our listing data in our Cypress tests (from a JSON that was not formatted like a listing), and also changes both the data enter and verify queries to be conditional based on if the data exists in the test data, as opposed to how it works now which just expects it to look exactly like the one Bloomington test listing we have.

I also added TODOs for existing gaps.

Ensure the Cypress test still passes!

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/cypress/e2e/default/03-listing.spec.ts
+++ b/sites/partners/cypress/e2e/default/03-listing.spec.ts
@@ -2,7 +2,8 @@ import {
   ApplicationAddressTypeEnum,
   ApplicationMethodsTypeEnum,
 } from "@bloom-housing/shared-helpers/src/types/backend-swagger"
-import { bloomingtonListing, CypressListing } from "../../fixtures/bloomingtonListing"
+import { bloomingtonListing } from "../../fixtures/bloomingtonListing"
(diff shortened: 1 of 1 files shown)


sites/partners/cypress/fixtures/bloomingtonListing.ts
```diff
@@ -0,0 +1,234 @@
...
+  ListingsStatusEnum,
+  ReviewOrderTypeEnum,
+  ListingNeighborhoodAmenities,
+} from "@bloom-housing/shared-helpers/src/types/backend-swagger"
+
+export type CypressListingImage = {
+  fixtureName: string
+  altText?: string
+}
+export type CypressListing = Listing & {
```
ludtkemorgan: suggestion: I'm guessing the intention of these types is that you want to create other fixtures (e.g. one for Angelopolis) and these types will be used by those fixtures as well. Could we create a type file within the cypress directory that just has these types?

PR #5811, opened 2026-01-29
fix: allow application to load with no ami charts

emilyjablonski:
[Support issue](https://exygy.slack.com/archives/C02G1S19QA2/p1769702013187219)

## Description

If a listing has units without an AMI chart, the listing is null on the application view page, resulting in the name being null, and checks around preferences and programs that depend on the listing are sometimes incorrect depending on the data.

Understanding why some AMI charts are null I think is only related to Seed data & a recent bug. Locally, Hollywood Heights had the wrong AMI chart selected. For deployed environments, we recently had a bug where we weren't filtering AMI charts by jurisdiction on partners, which meant you could get through the form by selecting a chart that did not belong to the jurisdiction, but one would not actually appear after saving - I think this is why we are seeing this on core dev for some listings.

I don't see this issue in Detroit - it's possible it made it to some listings on Doorway while that bug was active, but I haven't seen any.

Apply while logged in to a listing with at least one unit without an AMI chart, and then view the listing on the my applications page. Ensure the listing name appears at the top.

You can either manually remove the chart, OR seed while on the main branch, have your public jurisdiction be Angelopolis, and apply to the Hollywood Angelopolis listing (we had an issue in the seed with Angelopolis AMI charts before the changes in this PR).

## Author Checklist:

## Review Process:


sites/public/src/components/applications/SubmittedApplicationView.tsx
```diff
@@ -100,15 +100,21 @@ const SubmittedApplicationView = ({
               ?.length === 0
           }
           editMode={false}
-          enableUnitGroups={doJurisdictionsHaveFeatureFlagOn(FeatureFlagEnum.enableUnitGroups)}
+          enableUnitGroups={doJurisdictionsHaveFeatureFlagOn(
```
emilyjablonski: **note:** This was pulling from all jurisdictions associated w a user's profile instead of the one associated with the listing

Issue #2760, opened 2022-05-25
Flash of Progress Nav on application start page 


slowbot:
The is a flash at application start page to load, Would prefer if this loaded at min height and title loaded progressively

https://user-images.githubusercontent.com/65367387/170302454-396c02f6-5ad3-48fd-9783-0fe25ae798f1.mov

![Screen Shot 2022-05-25 at 9 40 33 AM](https://user-images.githubusercontent.com/65367387/170302471-b48db03e-2a8d-4179-87cd-7dd834609d0a.png)
![Screen Shot 2022-05-25 at 9 40 26 AM](https://user-images.githubusercontent.com/65367387/170302473-75a8a6a8-0cf6-4566-8b05-6736b4bea9f2.png)

@dominikx96 Here is a video of the flash, it looks like that progressive navigation component while the page is still fetching the listing is rendering much smaller than it does once it has the title. Ideally it renders at its presumed height with the title to minimize any flashing. Lmk if you can't repro! I have better luck in the Netlify environments throttling the network.

dominikx96:
@slowbot a fix has been merged.

QA notes:
1. Open the public site and the first application step
2. Make sure that the card header works properly without flash

emilyjablonski:
Still valid

PR #5812, opened 2026-01-29
fix: prevent choose language flash, limit image height

emilyjablonski:
This PR addresses #2760

## Description

The choose language page in applications is flashing while the listing data loads.

This adds our loading state to the page. Additionally, adds a max height and background cover styling to the image just like we do on the directory page, so that even if the image is very tall, it always shows at the same height.

This is the after, and then the before
https://github.com/user-attachments/assets/4b2cabc9-9da9-4cc1-894c-8f182fe38b1b

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/package.json
+++ b/shared-helpers/package.json
@@ -18,7 +18,7 @@
   },
   "dependencies": {
     "@bloom-housing/ui-components": "13.0.3",
-    "@bloom-housing/ui-seeds": "3.3.0",
+    "@bloom-housing/ui-seeds": "3.3.1",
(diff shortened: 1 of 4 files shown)


ludtkemorgan (changes requested):
Love this change! The choose language page flash has bugged me for a long time. Thanks for tackling it. 

I did notice a small issue when in arabic. The loader is on the right side instead of the center of the page

https://github.com/user-attachments/assets/912a9f37-4aac-4e92-94b4-f38be110dbfe

PR #5815, opened 2026-01-30
test: adds cypress test for angelopolis features

emilyjablonski:
## Description

Introduces checking for feature flagged parts of the listing page to the Cypress listing test, and adds a test that tests an Angelopolis listing including all of the features specific to Angelopolis.

Also continues to work through some remaining TODOs of fields we weren't yet testing (mainly optional addresses).

Ensure the test suite still passes. It looks to have added 1 minute to the Cypress partners suite, imo this is worth it but very open to feedback here!

We should do this for Lakeview as well.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/cypress/e2e/default/03-listing.spec.ts
+++ b/sites/partners/cypress/e2e/default/03-listing.spec.ts
@@ -307,11 +307,11 @@ describe("Listing Management Tests", () => {
     fillIfDataExists(cy, "yearBuilt", listing.yearBuilt?.toString(), "type")
 
     if (getFlagActive(listing, FeatureFlagEnum.enableRegions)) {
-      if (getFlagActive(listing, FeatureFlagEnum.enableConfigurableRegions)) {
-        fillIfDataExists(cy, "configurableRegion", listing.configurableRegion, "select")
-      } else {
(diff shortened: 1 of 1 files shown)


sites/partners/cypress/e2e/default/03-listing.spec.ts
```diff
@@ -275,8 +306,18 @@ describe("Listing Management Tests", () => {
     )
     fillIfDataExists(cy, "yearBuilt", listing.yearBuilt?.toString(), "type")
 
+    if (getFlagActive(listing, FeatureFlagEnum.enableRegions)) {
+      if (getFlagActive(listing, FeatureFlagEnum.enableConfigurableRegions)) {
+        fillIfDataExists(cy, "configurableRegion", listing.configurableRegion, "select")
```
ludtkemorgan: question: By relying on feature flags to determine if we populate/verify the data we do run the risk of not properly setting the feature flag in the fixture and the test would still pass. For example we forget to add `enableConfigurableRegions` to the angelopolis fixture this section would be skipped and the test would still pass. Is there a reason we couldn't just fill the field in if `listing.configurableRegion` has data?

PR #5834, opened 2026-02-03
fix: validation issues properties

emilyjablonski:
PR review and debugging for [5576](https://github.com/bloom-housing/bloom/tree/5576/property-section-listing-association)


api/src/services/listing.service.ts
```diff
@@ -2678,12 +2678,18 @@ export class ListingService implements OnModuleInit {
...
             },
           },
-          property: incomingDto?.property
+          property: incomingDto.property
             ? {
                 connect: {
                   id: incomingDto.property.id,
                 },
               }
+            : storedListing.property
```
emilyjablonski: **issue:** Not introduced by your changes, but without disconnect here, we were unable to remove a property once it had been added

api/prisma/seed-helpers/listing-factory.ts
emilyjablonski: **note:** Added a few into the Seed so that the core Angelopolis listing stays valid

Issue #5605, opened 2025-11-19
fill the user account snapshot tables on edit and delete


sarahlazarich:
On edit of a user account from the partner or public view or delete of a user account we should first generate a snapshot into the user account snapshot tables

PR #5865, opened 2026-02-10
feat: user snapshots are now created

YazeedLoonat:
This PR addresses https://github.com/bloom-housing/bloom/issues/5605

## Description
User snapshots will now be created when fields that are NOT transient are altered. 

Attached to this PR are a few modifications that are needed to get the site running a windows (came out of pairing with LA)

There is not, and will not be for some time, a UI attached to this so you will have to play around with the DB to test things:

- you should be able to log in without a snapshot being created
- you should be able to create users without snapshots being created
- you should be able to favorite a listing without snapshots being created

- if a user is updated in the partner portal a snapshot should be created
- if a user updates themselves in the public site a snapshot should be created

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/user.service.ts
+++ b/api/src/services/user.service.ts
@@ -54,6 +54,7 @@ import { ModificationEnum } from '../enums/shared/modification-enum';
 import { CronJobService } from './cron-job.service';
 import { ApplicationService } from './application.service';
 import { SnapshotCreateService } from './snapshot-create.service';
+import { toAddHelper, toRemoveHelper } from '../utilities/snapshot-helpers';
 
 /*
   this is the service for users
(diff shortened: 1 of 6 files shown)


ludtkemorgan (commented):
question: I think this is expected. But when you create an account a snapshot isn't created, however when you confirm that account it creates the snapshot. 
Flow:
1. create account from the public site - no snapshot created
2. Click on the "confirm my account" link sent in the welcome email - snapshot is created.
3. Update the account on the UI - snapshot is created

Just wanted to verify that confirmation of user accounts is expected for snapshotting

api/src/services/user.service.ts
```diff
@@ -264,80 +269,153 @@ export class UserService {
...
+
+      // if listing is on dto but not in db, we need to store it
+      const toAdd = dto.listings?.reduce((acc, curr) => {
+        if (!storedUser.listings?.some((elem) => elem.id === curr.id)) {
+          acc.push(curr.id);
+        }
+        return acc;
+      }, []);
+      if (toAdd?.length) {
+        transactions.push(async (transaction: PrismaClient) => {
```
ludtkemorgan: suggestion: we could reduce the number of transactions by one for both the listing and jurisdiction connections if you combine the `connect` and `disconnect` into one

```
 if (toAdd?.length || toRemove?.length) {
        transactions.push(async (transaction: PrismaClient) => {
          return transaction.userAccounts.update({
            where: {
              id: dto.id,
            },
            data: {
              listings: {
                connect: toAdd?.map((elem) => ({
                  id: elem,
                })),
                disconnect: toRemove?.map((elem) => ({
                  id: elem,
                })),
              },
            },
          });
        });
      }
```

ludtkemorgan (commented):
Looks good. Just some suggestions and also a question around if confirmation of a user is expected in the snapshot flow

PR #5880, opened 2026-02-11
fix: display preference and programs in correct ordinal value

ludtkemorgan:
This PR addresses [support issue](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/metrotranscom/doorway/1491)

## Description

The ordinal tied to preference/programs on listings is not being used on the public and partner site to display them in the correct order. This usually doesn't cause an issue because the backend usually returns the multiselect questions in the order they were created which is the order they are supposed to be in. However, that is not deterministic and occasionally they are returned in a different order. The frontends should use the ordinal values to properly display both programs and preferences

1. Create a listing with many programs and preferences attached. Publish that listing
2. Go to the public site and verify that the programs and preferences on that listing are displaying the correct order
3. Edit the listing by changing the order of the preferences.
4. Verify that the updated order is displaying correctly on both the partner site as well as the listing details page on the public site

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/partners/__tests__/components/listings/PaperListingForm/index.test.tsx
+++ b/sites/partners/__tests__/components/listings/PaperListingForm/index.test.tsx
@@ -238,10 +238,12 @@ describe("PaperListingForm", () => {
 
       // units
       expect(
-        screen.getByRole("row", { name: "Unit # Unit type AMI Rent SQ FT ADA Actions" })
+        screen.getByRole("row", {
+          name: "Unit # Unit type AMI Rent SQ FT Accessibility priority type Actions",
+        })
       ).toBeInTheDocument()
       expect(
-        screen.getAllByRole("row", { name: "Studio 45.0 1104.0 285 Edit Delete" }).length
+        screen.getAllByRole("row", { name: "Studio 45.0 1104.0 285 n/a Edit Delete" }).length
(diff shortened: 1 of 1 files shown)


shared-helpers/src/auth/AuthContext.ts
```diff
@@ -406,8 +406,11 @@ export const AuthProvider: FunctionComponent<React.PropsWithChildren> = ({ child
       }
       // Return true only if all jurisdictions have the flag turned on
       if (onlyIfAllJurisdictionsHaveItEnabled) {
-        return jurisdictions.every(
-          (j) => j.featureFlags?.find((flag) => flag.name === featureFlag)?.active || false
+        return (
+          !!jurisdictions.length &&
```
ludtkemorgan: If for some reason the jurisdictions aren't loaded this was incorrectly returning true. This shouldn't happen in production but it does in some tests if all of the mocking isn't done correctly and the extra safeguard doesn't hurt.

jaredcwhite:
This all seems good @ludtkemorgan, but I'm wondering why we wouldn't expect that the backend API is returning the MSQs in the ordinal sort order? Does the sorting _have_ to be done on the frontend?

Issue #2241, opened 2021-11-30
Aria described by is not being associated properly on Form Groups


slowbot:
**Describe the bug**
Aria described by is not being associated properly on multiple instances of the form group component

**Expected behavior**
- Aria described by is not being  associated properly with checkboxes on preferred units
- Aria described by  not being  associated properly with radio on household changes and youth
- Aria described by not being  associated properly with radio on housing vouchers
- Aria described by not being  associated properly with radio on income
- Aria described by not being associated properly with radio on demographics
- Aria described by not being associated properly with checkbox terms
- Aria described by not being associated properly with contact type preference, address state and phone number type

**Desktop (please complete the following information):**

slowbot:
@emilyjablonski For your review

Issue #2240, opened 2021-11-30
Form labels should be associated with appropriate fields 


slowbot:
**Describe the bug**
Several form fields have missing or incorrectly associated labels

**Expected behavior**
- Short form phone number, primary applicant and household member input is not associated with its label, should have proper id and for attributes
- "No" option on ADA question is not associated with its label, should have proper id and for attributes
- Form labels on Name component of Settings has empty labels (should have label and rely on sr-only css to hide from screen reader)
- Form labels on Name component and re-enter email and password of Create Account has empty labels (should have label and rely on sr-only css to hide from screen reader)

**Desktop (please complete the following information):**
 - OS: BigSur
 - Browser Chrome
 - Version 96.0.4664.55 (Official Build) (x86_64)

slowbot:
@emilyjablonski For your review

PR #5896, opened 2026-02-16
fix: form accessibility fixes

emilyjablonski:
This PR addresses parts of related issues like #2240 #2241 

## Description

Makes some accessibility fixes in our public site forms. This was just a quick pass at the most obvious outstanding issues. Highly recommend reviewing with hiding whitespace, as the main changes were either wrapping sections in different fieldsets, or wrapping sections in ids for aria-controls, and are much more minimal than the diff looks.

Overall changes:
* Related fields should be wrapped in a fieldset (like first / middle / last name, or addresses).
* Fieldsets need to have legends (we had some missing altogether or were using labels).
* Form fields that, when selected, show additional fields (like "I have a second phone number" or indicating you work in the county) need to have aria-controls set to the ID of the section that will be toggled on, and the ID must always be in the DOM. If it's a checkbox, it should also have aria-expanded. Voiceover is not consistent about this, but JAWS relies on this more.
* We should not have multiple links on the same page with the same text (like Edit) outside of a clear item context like a list - it should be clear what each link is for to a screen reader without the visual context.
* Helper text (like error messages, or example subnotes) should be associated with the form field through aria-describedby so that it is read when the form field is read and doesn't require you to continue navigating past the field to know it's there - some of this happened here, most of this will be fixed with [this UIC PR](https://github.com/bloom-housing/ui-components/pull/195), some cannot easily be fixed until an upgrade of our form library, like the date of birth field or phone field
* Deduplicates error box code, and now includes focusing on the error banner when an error is triggered so a user can re-check the form without navigating backward (ideally we would focus on the first input with an error, but our older version of react-hook-forms doesn't support this well)
* Turns the housing members page into a list so it's clear there will be repeated sections

There are no functionality changes.

[Deploy preview](https://deploy-preview-5896--bloom-public-seeds.netlify.app/applications/start/choose-language?listingId=5d588c32-dfa7-486e-9a3c-fc2eac69b655) vs [core](https://bloom-public-seeds.netlify.app/applications/start/choose-language?listingId=5d588c32-dfa7-486e-9a3c-fc2eac69b655)
1. You can test out triggering an error in the form by not filling out a required field, hitting next, and then pressing tab to see your focus is within the alert.
2. With a screen reader on, you can focus on a field with helper text and hear it read the helper text after the input, like the example subnote on the email field.
3. With a screen reader on, you can navigate through the household members page and hear it read the members as a list.

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/package.json
+++ b/shared-helpers/package.json
@@ -17,7 +17,7 @@
     "prettier": "prettier --write \"**/*.ts\""
   },
   "dependencies": {
-    "@bloom-housing/ui-components": "13.0.4",
+    "@bloom-housing/ui-components": "13.0.6",
     "@bloom-housing/ui-seeds": "3.3.1",
     "@heroicons/react": "^2.1.1",
     "axios-cookiejar-support": "^5.0.5",
--- a/sites/partners/package.json
+++ b/sites/partners/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@bloom-housing/shared-helpers": "^7.7.1",
-    "@bloom-housing/ui-components": "13.0.4",
+    "@bloom-housing/ui-components": "13.0.6",
     "@bloom-housing/ui-seeds": "3.3.1",
     "@heroicons/react": "^2.2.0",
     "@mapbox/mapbox-sdk": "^0.13.0",
--- a/sites/public/package.json
+++ b/sites/public/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@bloom-housing/shared-helpers": "^7.7.1",
-    "@bloom-housing/ui-components": "13.0.4",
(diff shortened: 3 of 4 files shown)


sites/public/src/pages/account/edit.tsx
```diff
@@ -237,63 +237,66 @@ const Edit = () => {
...
+                    }
+                    dataTestId={"account-last-name"}
+                  />
+                  <Button
+                    type="submit"
+                    size="sm"
+                    variant="primary-outlined"
+                    loadingMessage={nameLoading ? t("t.loading") : undefined}
+                    id={"account-submit-name"}
+                    ariaLabel={`${t("account.settings.update")} ${t("application.name.yourName")}`}
```
emilyjablonski: **note:** The buttons on this page all just say "Update" - they should indicate which section they are for

PR #5948, opened 2026-02-24
feat: advocate edit account

emilyjablonski:
This PR addresses #5944

## Description

This ticket assumes an advocate is already created and approved. It covers editing the account information in the user's dashboard when logged in. The work to use to same form fields to actually create the account will be a different PR.

I separated out the sections into helper functions to reduce duplication between the public user edit account and advocate user edit account pages. We can plug any of the individual sections into account forms. We will also be able to use them for the upcoming advocate request form, and could use them on the public create form as well.

I think the per-field update button pattern we have is weird, but a redesign of this existing pattern is out of scope for these tickets.

We should test the public user edit account page and the advocate user edit account page. We should be able to update each section.

## Author Checklist:

## Review Process:


ludtkemorgan (approved):
Change looks great!!

PR #5955, opened 2026-02-25
feat: feature flag verify

emilyjablonski:
This PR addresses #5958

## Description

The verify income red light should not be on in any jurisdictions on the income step (I also think its intended behavior is buggy, as it will block you for $50/month as too low but not $50/year).

I feature flagged this off instead of fully removing it for now.

Reseed, and in Angelopolis for Hollywood Hills Heights, submit an application. Ensure you select no to vouchers and then input income as $50 / month. You should be able to proceed.

In partners, enable the new verify income feature flag in Angelopolis.

Do the same steps, ensuring you fully re-start the app from the choose language page, and the same steps should trigger the error and prevent you from continuing (below). We want this off everywhere right now.

Screenshot 2026-02-26 at 8 38 00 AM

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/__tests__/pages/applications/financial/income.test.tsx
+++ b/sites/public/__tests__/pages/applications/financial/income.test.tsx
@@ -104,7 +104,7 @@ describe("applications pages", () => {
 
       fireEvent.click(screen.getByRole("button", { name: "Next" }))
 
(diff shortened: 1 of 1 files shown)


ludtkemorgan (approved):
This looks good. I'm wondering how this works in Doorway. I don't see any code differences but I've never seen this error come up when applying to listings there.

Issue #5858, opened 2026-02-06
Make the dbinit script re-grant privileges if it has to recreate the bloom_prisma db


avrittrohwer:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue

The init SQL scripts in api/dbinit check if the bloom_prisma DB, bloom_api and bloom_readonly users exist before attempting to create them.  This allows the init script to run multiple times without failure.

The init script does not currently properly handle running again if the bloom_prisma DB has been dropped.  It properly recreates the DB, but does not re-grant permisisons to bloom_api and bloom_readonly.  We should add a check if the proper grants are in place, separate from if the users exist.

## Acceptance Criteria

1. Update rds and docker-compose init sql scripts to check if the expected permissions are granted.
2. Add a test case in .github/workflows/docker_compose_ci.yml

Issue #5796, opened 2026-01-27
Add support for setting mapbox and google translate API keys


avrittrohwer:
- [x] General
- [ ] Backend
- [ ] Public
- [ ] Partners

## Background / Issue

Bloom relies on mapbox and google translate external integrations, so the site needs API keys for each of these.

## Acceptance Criteria

The infra should set up secrets for the API keys.  Users will be responsible for putting the API keys as a secret value.

Issue #5789, opened 2026-01-23
Create public s3 bucket


YazeedLoonat:
- [x] Infra

## Background / Issue
Bloom allows for the uploading of publicly available files (think listing images, paper application pdfs, etc). We'll need an S3 bucket to act as a CDN that grants access to the objects in the bucket to anyone to read and only the API should be able to upload objects to the bucket

Issue #5788, opened 2026-01-23
Create Private S3 bucket


YazeedLoonat:
- [x] Infra

## Background / Issue
Bloom leverages a private S3 bucket to temporarily store uploads of application exports. This holds PII for all applicants to a listing so keeping it private is for the best. The way we've leverage it in the past is we would hand back a PresignedGetURL that grants temporary access to the object in the bucket and the bucket automatically removes the object after 1-2 days

anyone with the presigned get url should be able to read the object, but only the API should be able to upload an object to the bucket

PR #5969, opened 2026-02-27
feat: add infra support for API keys, add application support for s3 instead of cloudinary

avrittrohwer:
This PR addresses #5796

This PR addresses #5788

This PR addresses #5789

This PR addresses #5858

## Description

1. Adds support in infra code for creating mapbox and google translation API keys.
2. Adds support in application code for using S3 instead of cloudinary.

I've deployed it to https://core-dev.bloomhousing.dev and https://partners.core-dev.bloomhousing.dev. Verify:

1. Translations work by looking at listings pages in different languages.
2. Address validation works by adding an address to the application flow.
3. Mapbox map is correctly shown on the listing
4. Listing image upload works.
5. Listing export works.
6. Application export works. 

## Author Checklist:

## Review Process:

### DIFF
--- a/api/test/integration/lottery.e2e-spec.ts
+++ b/api/test/integration/lottery.e2e-spec.ts
@@ -497,28 +497,26 @@ describe('Lottery Controller Tests', () => {
         },
       });
 
-      const selections = await prisma.applicationSelections.createManyAndReturn(
-        {
-          data: [
-            {
-              applicationId: applicationA.id,
-              multiselectQuestionId: preferenceAId,
-            },
-            {
-              applicationId: applicationA.id,
-              multiselectQuestionId: preferenceBId,
-            },
-            {
-              applicationId: applicationB.id,
-              multiselectQuestionId: preferenceAId,
-            },
-            {
-              applicationId: applicationC.id,
-              multiselectQuestionId: preferenceBId,
-            },
-          ],
-        },
-      );
+      await prisma.applicationSelections.createManyAndReturn({
+        data: [
+          {
+            applicationId: applicationA.id,
+            multiselectQuestionId: preferenceAId,
+          },
+          {
+            applicationId: applicationA.id,
+            multiselectQuestionId: preferenceBId,
+          },
+          {
+            applicationId: applicationB.id,
+            multiselectQuestionId: preferenceAId,
+          },
+          {
+            applicationId: applicationC.id,
+            multiselectQuestionId: preferenceBId,
+          },
+        ],
+      });
 
       await request(app.getHttpServer())
         .put(`/lottery/generateLotteryResults`)
--- a/api/test/integration/permission-tests/permission-as-partner-correct-listing.e2e-spec.ts
+++ b/api/test/integration/permission-tests/permission-as-partner-correct-listing.e2e-spec.ts
@@ -86,6 +86,7 @@ describe('Testing Permissioning of endpoints as partner with correct listing', (
   let app: INestApplication;
   let prisma: PrismaService;
   let userService: UserService;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   let applicationFlaggedSetService: ApplicationFlaggedSetService;
   let cookies = '';
   let jurisdictionId = '';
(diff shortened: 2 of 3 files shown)


api/prisma/seed-helpers/asset-file-id-helper.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Returns the appropriate fileId for a Cloudinary asset based on the
+ * DBSEED_USE_FULL_CLOUDINARY_URL_AS_FILE_ID environment variable.
+ *
+ * When the env var is set, returns the full Cloudinary URL:
+ *   https://res.cloudinary.com/exygy/image/upload/w_400,c_limit,q_65/<IMAGE_ID>.jpg
+ *
+ * Otherwise, returns the image ID as-is.
+ */
+export function assetFileId(imageId: string): string {
+  if (process.env.DBSEED_USE_FULL_CLOUDINARY_URL_AS_FILE_ID) {
```
YazeedLoonat: **nitpick (non-blocking):** I feel like at some point this is going to cause us a little confusion. This would only be true when the UI has cloudinary turned off but the API is still seeding with cloudinary values (since in a production env this could never be the case)

maybe we can use the env variable `CLOUDINARY_SECRET` where if its not set then we seed with the full url and if its set then we don't have to worry about it?

api/src/modules/application-exporter.module.ts
```diff
@@ -14,7 +15,7 @@ import { PrismaModule } from './prisma.module';
     MultiselectQuestionModule,
     PermissionModule,
   ],
-  providers: [ApplicationExporterService],
+  providers: [ApplicationExporterService, S3Service],
```
YazeedLoonat: **issue (blocking):** this is a quirk of nestjs and how it does dependency injection. Whenever you attach a service like this you are instantiating a new instance of this service. I see below that there are a few other instances of this kind of line, with each line meaning that the fully running version has a few different S3 services running all at once. 

The way around this is to create an S3 module and attach the module like we do with something like the PrismaModule. Modules are all instantiated once and then handed out to other modules that need it as a singleton 

hopefully that makes sense, if not lemme know! Basically what I would like is to create an S3 Module that is then attached to all these modules that need access to the s3 service instead so that we don't have duplicate services hangin around taking up resources

api/src/modules/asset.module.ts
```diff
@@ -1,12 +1,15 @@
...
 import { AssetService } from '../services/asset.service';
+import { CloudinaryService } from '../services/cloudinary.service';
 import { PermissionModule } from './permission.module';
+import { S3Service } from '../services/s3.service';
 
 @Module({
   imports: [PermissionModule],
   controllers: [AssetController],
-  providers: [AssetService],
+  providers: [AssetService, CloudinaryService, S3Service, ConfigService],
```
YazeedLoonat: **issue (blocking):** same as above for the cloudinary service

shared-helpers/src/types/backend-swagger.ts
```diff
@@ -8913,9 +8937,6 @@ export interface PartnerUserCreate {
   /**  */
   additionalPhoneExtension?: string
 
-  /**  */
-  isAdvocate?: boolean
```
YazeedLoonat: **issue (blocking):** I think you may have resolved a merge conflict that got rid of the isAdvocate flag on users which is a breaking change to the advocate user work

you may be able to fix with just running `yarn generate:client` in the API again and that may fix up the issue

api/src/services/s3.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  async urlForPrivate(key: string): Promise<string> {
+    const command = new GetObjectCommand({
+      Bucket: this.privateBucket,
+      Key: key,
+    });
+    return getSignedUrl(this.s3Client, command, {
+      expiresIn: 60 * 5, // seconds. Expires 5 minutes after generation
```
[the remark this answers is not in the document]
ludtkemorgan: I know that this was an existing amount of time so we don't need to change it, but 5 minutes seems pretty short for this.

PR #5977, opened 2026-03-02
feat: core map

emilyjablonski:
This PR addresses #5984

## Description

Pulls over the core map functionality from Doorway, reduces some duplication, and creates some UI fixes as part of transitioning to Seeds. 

~Run `yarn setup:large` in api/. In the admin panel in partners, turn the map feature flag on. You can test in any jurisdiction.~

Run `yarn db:setup:staging-region` in the api directory. Set the public site jurisdiction to "Bay Bridge"

After merging this PR, the feature will remain behind a feature flag until the tickets below are completed. This PR should re-test the core map functionality.

Remaining work that is defined in subtasks under the [parent ticket](https://github.com/bloom-housing/bloom/issues/5983):
* Make it jurisdictionally configurable so that you could configure it to fetch from any number of jurisdictions and show a filter - this is required for Doorway's use, but not any other jurisdiction at the moment. Right now it's just taking the one public site's jurisdiction. I added some inline TODOs for this work.
* Hook up the map to pull filters from the URL params to support Doorway's homepage search - this was brought over but not tested as we don't have the homepage search
* Make the info cards under the listing list configurable - I just removed the Doorway specific ones
* Cypress test - Doorway has one, but we need to figure out how we're going to turn the feature flag on in the test

## Author Checklist:

## Review Process:


api/src/services/listing.service.ts
```diff
@@ -156,6 +156,15 @@ includeViews.base = {
...
 };
 
+includeViews.map = {
+  ...includeViews.base,
+  jurisdictions: {
+    select: {
+      featureFlags: true,
+    },
+  },
+};
```
ludtkemorgan: issue: This returns a lot of extra data. The response is fairly large since it has most data from the listing plus the feature flags for the jurisdiction on each listing (lot's of duplication).

I recommend we change this to being in the `selectViews` and explicitly select the fields we need. Something like:

```
selectViews.map = {
  ...selectViews.name,
  listingsBuildingAddress: true,
  listingImages: {
    include: {
      assets: true,
    },
  },
  reservedCommunityTypes: true,
  units: {
    select: {
      monthlyRent: true,
      unitTypes: {
        select: { numBedrooms: true, name: true },
      },
    },
  },
};
```

api/src/services/listing.service.ts
```diff
@@ -278,7 +287,10 @@ export class ListingService implements OnModuleInit {
           include: includeViews[params.view ?? 'full'],
         });
 
-    const listings = mapTo(Listing, listingsRaw);
+    const listings =
+      params.view === ListingViews.map
+        ? (listingsRaw as unknown as Listing[])
+        : mapTo(Listing, listingsRaw);
```
ludtkemorgan: question: Why is this needed? We are getting a list of listings back for all view types

api/prisma/seed-staging.ts
```diff
@@ -1495,40 +1506,66 @@ export const stagingSeed = async (
...
-          isJurisdictionalAdmin: false,
-        },
-        email: `partner-user-${savedListing.name
-          .toLowerCase()
-          .replaceAll(' ', '-')}@example.com`,
-        confirmedAt: new Date(),
-        jurisdictionIds: [savedListing.jurisdictionId],
-        acceptedTerms: true,
-        listings: [savedListing.id],
+  if (large) {
```
ludtkemorgan: suggestion: This seed file is getting pretty unwieldy with all of the data for different jurisdictions. I think we should split this file up into different files based on use case (e.g. one file for each jurisdiction). That doesn't have to happen for this PR, but I might create a refactor to do that.

ludtkemorgan (commented):
Thought I'd start reviewing this as it looks like we will be corifying Doorway and this will be an important step in that. Only got a little way into reviewing and might not get back to it for a while so leaving the comments so I don't forget. Overall, love the look and feel!

ludtkemorgan (approved):

Issue #5945, opened 2026-02-23
Waitlist Lottery Unit Table Showing Vacant Units


iph-97:
<img width="1010" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVVHRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--487a9693e5dec9ac3ce51a45b809223279d41ab9/image.png" alt="image.png" />
Our first waitlist lottery has gone live in Doorway, and Babs noticed a discrepancy on the listing page. On the unit table on the listing details page, it shows "1 vacant unit" rather than "Open waitlist" as the availability. Can we take a look and see what's going on? Medium prio :large_yellow_circle: please!

PR #6006, opened 2026-03-05
fix: rent summary table fix for waitlist lottery

ludtkemorgan:
This PR addresses #5945

## Description

One place was missed in the waitlist lottery epic. In the unit summary table on the listing detail page it still displays "X vacant units" when it should be "open waitlist" for all waitlistLottery listings

On a listing for a jurisdiction where enableWaitlistLottery is turned on do the following scenarios:

### SHOW_NEW_SEEDS_DESIGNS = true
1. Create a waitlist lottery listing with at least one unit. Verify "Availability" column has "Open Waitlist"
2. Create a waitlist lottery listing with at least two units and have different AMI chart percentages. Verify "Availability" column has "Open Waitlist" for all AMI chart percentages
3. Create a first come first serve listing and verify "Availability" column has vacant unit counts

### SHOW_NEW_SEEDS_DESIGNS = false
1. Create a waitlist lottery listing with at least one unit. Verify "Availability" column has "Open Waitlist"
2. Create a waitlist lottery listing with at least two units and have different AMI chart percentages. Verify "Availability" column has "Open Waitlist" for all AMI chart percentages
3. Create a first come first serve listing and verify "Availability" column has vacant unit counts

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/src/lib/helpers.tsx
+++ b/sites/public/src/lib/helpers.tsx
@@ -352,13 +352,19 @@ export const getListings = (listings) => {
   }
 
   const generateTableSubHeader = (listing) => {
-    if (listing.reviewOrderType !== ReviewOrderTypeEnum.waitlist) {
+    if (
+      listing.reviewOrderType !== ReviewOrderTypeEnum.waitlist &&
+      listing.reviewOrderType !== ReviewOrderTypeEnum.waitlistLottery
+    ) {
       return {
         content: t("listings.availableUnits"),
         styleType: AppearanceStyleType.success,
         isPillType: true,
       }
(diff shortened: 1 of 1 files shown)


mcgarrye (approved):
Image
The actual changes here look good, however when seeds is turned off the browse page still incorrectly shows `Available Units`. This isn't a blocker for Doorway as they don't have the same browse page but we should have a follow-up to fix it

ludtkemorgan:
@mcgarrye good catch! We don't use that page anywhere outside of HBA so it's it's not as critical. But I did make the change just to be safe

PR #6013, opened 2026-03-08
refactor: public dynamic imports

emilyjablonski:
## Description

Reduces some client page-level bundle sizes on the public site by lazy-loading two heavy dependencies - markdown and Mapbox geocoding.

Added LazyMarkdown which wraps markdown-to-jsx in next/dynamic so it is shipped in a separate chunk. We were importing markdown in the main application layout because we use it if the maintenance banner is on - but that doesn't mean we should be loading it on every single page including ones that don't actually use the package.

In ValidateAddress.tsx, we now load @mapbox/mapbox-sdk/services/geocoding via dynamic import() inside findValidatedAddress so the package is only requested when the user runs address validation vs page load.

The main shared first load is unchanged - savings are in page-specific chunks. Some examples:
`/` from 10.7kb --> 8.36kb
`/account/dashboard` from 4.62kb --> 2.48kb
`/get-assistance` from 5.24kb --> 3.08kb

These aren't super noticeable but the hope is lazy loading will allow the browser to be interactive quicker on slow connections.

To test I would submit an application. You can see markdown on the what to expect, terms, and confirmation pages. We can test the geocoding on the address verification page, and there is a new loading spinner more easily seen when throttling the network.

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/src/components/listing/ListingViewSeeds.tsx
+++ b/sites/public/src/components/listing/ListingViewSeeds.tsx
@@ -1,6 +1,5 @@
 import React, { useContext, useEffect, useState } from "react"
 import { useForm } from "react-hook-form"
-import Markdown from "markdown-to-jsx"
 import {
   EnumListingListingType,
   FeatureFlagEnum,
@@ -50,6 +49,7 @@ import styles from "./ListingViewSeeds.module.scss"
 import { ReadMore } from "../../patterns/ReadMore"
 import { OtherFeatures } from "./listing_sections/OtherFeatures"
 import { PropertyDetailsCard } from "./listing_sections/PropertyDetailsCard"
+import LazyMarkdown from "../core/LazyMarkdown"
 
 interface ListingProps {
   listing: Listing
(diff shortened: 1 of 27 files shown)


sites/public/src/components/listing/ListingViewSeeds.tsx
```diff
@@ -159,7 +159,7 @@ export const ListingViewSeeds = ({ listing, jurisdiction, profile, preview }: Li
       {listing.whatToExpect && (
         <InfoCard heading={t("whatToExpect.label")}>
           <div className={"bloom-markdown"}>
-            <Markdown>{listing.whatToExpect}</Markdown>
+            <LazyMarkdown>{listing.whatToExpect}</LazyMarkdown>
```
ludtkemorgan: I'm seeing that same hydration issue here when I add actual markdown to this section

Image

emilyjablonski: Do you mind sharing what you added? I can't repro this one 🤔

ludtkemorgan: Shoot, I've reseed several times since then. I think I added some lists and bolded something

Issue #5957, opened 2026-02-26
fill the application snapshot tables on edit and delete


YazeedLoonat:
On edit of an application from the partner view or delete of an application we should first generate a snapshot into the application snapshot tables

PR #6024, opened 2026-03-10
feat: application snapshots

YazeedLoonat:
This PR addresses https://github.com/bloom-housing/bloom/issues/5957

## Description
On application update and delete a snapshot is created

there is no UI attached to this so you are going to need to keep an eye on the db for this:
- edit an application 
- check the db to verify application snapshots were created

## Author Checklist:

## Review Process:

### DIFF
--- a/api/test/integration/listing.e2e-spec.ts
+++ b/api/test/integration/listing.e2e-spec.ts
@@ -805,7 +805,7 @@ describe('Listing Controller Tests', () => {
       });
     });
 
-    it.skip('should get listings from list endpoint when no params are sent', async () => {
+    it('should get listings from list endpoint when no params are sent', async () => {
       const res = await request(app.getHttpServer())
         .post(`/listings/list`)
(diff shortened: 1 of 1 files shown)


api/src/services/snapshot-create.service.ts
```diff
@@ -1091,4 +1091,539 @@ export class SnapshotCreateService {
...
+      userAccounts,
+
+      applicationSelections,
+      householdMember,
+      preferredUnitTypes,
+      ...rest
+    } = currData;
+
+    // create snapshot
+    await this.prisma.applicationSnapshot.create({
```
ludtkemorgan: issue: If I edit an application 2 times I run into a constraint issue
```
Unique constraint failed on the fields: (`listing_id`,`confirmation_code`)
```

This is because there is a constraint on the applicationSnapshot table `@@unique([listingId, confirmationCode])`. I think we can just remove that entirely from the snapshot table

Issue #5872, opened 2026-02-11
Configure Google translate


sarahlazarich:
We will need to add google translate on the public site in LA's required languages (list TBD)
LA's languages are:
1. Spanish
2. Korean
3. Armenian
4. Traditional Chinese
5. Filipino/Tagalog
6. Farsi (Persian)
7. Vietnamese

AC: 
- Public site can be translated in LA's required languages via google translate (List TBD) 
- Both static and dynamic content should translate

PR #6102, opened 2026-03-23
feat: default db translations

emilyjablonski:
This PR addresses #5872

## Description

Sets default translations in the null jurisdiction for all DB strings in all languages using the new db translation script.

Setup the db on main, then change to this branch and run `yarn db:migration:run`. Ensure that the translations table has default translations for the null jurisdiction, which serves as the fallback in jurisdictions if a jurisdiction row does not exist.

Strings that need customization, like to insert the jurisdiction name or email, are handled in migrations within forks.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/scripts/db-translation-input.example.json
+++ b/api/scripts/db-translation-input.example.json
@@ -1,23 +1,14 @@
 {
-  "accountRemoval": {
-    "subject": "Bloom Housing scheduled account removal due to inactivity",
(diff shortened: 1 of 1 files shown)


YazeedLoonat (commented):
Quebbin: with the way this is written now it'll update or insert null translations everywhere this migration runs. This would change the default translations pretty much everywhere, are these translations going to work everywhere? 

Like doorway, detroit, LA, and all our dev instances will all get these so wanna be sure we aren't going to be affecting the translation "stack" anywhere that already has these

Issue #6122, opened 2026-03-30
Accessible Units in Filter


sarahlazarich:
Steps to reproduce: 
- Go to https://bloom-angelopolis.netlify.app/listings
- See open the filter drawer 

Observed behavior: 
- You can not filter by mobility and/or hearing vision units?

Expected behavior: 
- You should be able to filter by mobility and/or hearing vision units, per the design 
- figma.com/design/qgL3N9ssyOMhpD5W4ULg7G/Filter-set--Accessibility?node-id=15-2793

<img width="729" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2RxRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--487ef853ffb58049f61948617a7b9f559fc15080/image.png" alt="image.png" />

sarahlazarich:
lgtm!

PR #6133, opened 2026-04-01
feat: filter on accessible units

emilyjablonski:
This PR addresses #6122

## Description

Creates a filter if a jurisdiction has accessibility unit types configured.

Listings should return if at least one unit has that accessibility priority type. Given how they are configured in LA, we don't expect users to need an `AND` filter so this is using an `OR`.

## Author Checklist:

## Review Process:


docs/feature-flags/enableReasonableAccommodations.md
```diff
@@ -0,0 +1,21 @@
...
+
+When true, the reasonable accommodations question is shown in the public and partner application flow
+
+## Additional Information
+
+## Images
+
+### Public application
+
+![image.png](./images/image%2051.png)
```
ludtkemorgan: ❤️

Issue #6067, opened 2026-03-18
Disclaimer to pets policy


sarahlazarich:
We should add a sub-heading/disclaimer on the pets policy on the public site.

<img width="1062" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOU5LRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--fe4129149c9f6a7edcfee3b7b108efd89913738c/image.png" alt="image.png" />

The copy should be: 
- “Service dogs and emotional support animals are not pets and are permitted as required by law. Pets may also be required to be allowed in certain types of housing, such as publicly financed housing developments in the City of Los Angeles.”

AC:
- Disclaimer should load on LA's public site
- Disclaimer should translate

emilyjablonski:
QA Notes
Ensure the sub-note appears next to pet policy. [Here](https://bloom-angelopolis.netlify.app/listing/2a8a6d34-b861-4c4d-85b8-003a9a6b34d4/angelopolis_lottery_listing_1000_vin_scully_ave_los_angeles_ca) is an example listing.

sarahlazarich:
lgtm

PR #6139, opened 2026-04-02
feat: t if exists

emilyjablonski:
This PR addresses #6067

## Description

The new `tIfExists` helper returns null if the key does not exist at all in the JSON. If the key exists, which we can set in the locale overrides in LA, the helper will render.

I'm including the override in core so we can see how it renders. Because all of these core environments are based on the same string files, I'm testing no key by locally removing it from the general overrides file. It works with both versions of the pet policy field, and the description comes second.

Screenshot 2026-04-02 at 3 18 17 PM
Screenshot 2026-04-02 at 3 18 56 PM

## Author Checklist:

## Review Process:


ludtkemorgan (changes requested):

ludtkemorgan:
Also as a suggestion since this PR tackles the pet policy disclaimer could we add that to the commit message when it goes in `feat: t if exists and pet policy disclaimer`

ludtkemorgan (approved):
Thanks for making those changes! Just make sure to update the message when squashing this

PR #6170, opened 2026-04-14
feat: autopopulate reserved community type description

emilyjablonski:
This PR addresses #6171

## Description

(1) Everywhere, regardless of flag values: if a description is set on a reserved community type, when the dropdown value is changed, the description should auto populate with that value.

(2) New feature flag `disableReservedCommunityTypeEdit`, that when true, will still autopopulate the description, but not allow it to be editable and just show it as a plaintext FieldValue. It will still save the value on the listing.

On a fresh reseed, set a listing's reserved community type to Seniors 62+. The description should fill with the db value. 

In Bloomington: You should be able to edit it to custom text - it will only update on dropdown change.

In Angelopolis: It will not be editable and will just appear as a FieldValue.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/enums/feature-flags/feature-flags-enum.ts
+++ b/api/src/enums/feature-flags/feature-flags-enum.ts
@@ -8,6 +8,7 @@ export enum FeatureFlagEnum {
   disableEthnicityQuestion = 'disableEthnicityQuestion',
   disableJurisdictionalAdmin = 'disableJurisdictionalAdmin',
   disableListingPreferences = 'disableListingPreferences',
+  disableReservedCommunityTypeEdit = 'disableReservedCommunityTypeEdit',
   disableWorkInRegion = 'disableWorkInRegion',
   enableAccessibilityFeatures = 'enableAccessibilityFeatures',
   enableAdaOtherOption = 'enableAdaOtherOption',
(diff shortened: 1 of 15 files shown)


api/src/enums/feature-flags/feature-flags-enum.ts
```diff
@@ -67,6 +67,7 @@ export enum FeatureFlagEnum {
   example = 'example', // sample feature flag for testing purposes
   hideCloseListingButton = 'hideCloseListingButton',
   swapCommunityTypeWithPrograms = 'swapCommunityTypeWithPrograms',
+  disableReservedCommunityTypeEdit = 'disableReservedCommunityTypeEdit',
```
ludtkemorgan: nit: we try to alphabetize these enums so they appear in that order everywhere else

sites/partners/src/components/listings/PaperListingForm/sections/CommunityType.tsx
```diff
@@ -81,6 +86,25 @@ const CommunityType = ({
...
+      const matchedType = reservedCommunityTypes.find((type) => type.id === currentCommunityType)
+      setDescription(matchedType?.description ?? "")
+    } else {
+      setDescription("")
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [currentCommunityType, loading])
+
+  useEffect(() => {
+    const isListingInitialType = currentCommunityType === listing?.reservedCommunityTypes?.id
```
ludtkemorgan: I don't know if this is intentional or not. But the description doesn't get updated in this one scenario. And I think it's because of this line.

1. Select senior 62+
2. Change the description
3. Save the listing
4. Edit the listing and change the community type to something else
5. Change the community type back to senior 62+ and no description appears

I could see this being confusing if they accidentally change the community type.

Also, there are 6 useEffects in this file and these two new ones seem very similar is there a way to combine them? It looks like different behavior should be happening when loading so maybe splitting out was trying to solve one use case.

sites/partners/src/components/listings/PaperListingForm/sections/CommunityType.tsx
```diff
@@ -116,29 +140,46 @@ const CommunityType = ({
...
+            {disableReservedCommunityTypeEdit && (
+              <FieldValue label={t("listings.reservedCommunityDescription")}>
+                {description || t("t.none")}
+              </FieldValue>
+            )}
+          </Grid.Cell>
+        </Grid.Row>
+        <Grid.Row
+          columns={3}
+          className={disableReservedCommunityTypeEdit ? styles["hidden-field"] : ""}
```
ludtkemorgan, line 160: question: Is there a reason you are hiding this field via css instead of just not rendering it? Is it because we need to save the description and this is the easiest way?

ludtkemorgan (changes requested):
Just found one bug. If that was intentional behavior I'm good with this. But will wait on approving until that is confirmed

sites/partners/src/components/listings/PaperListingForm/sections/CommunityType.tsx
emilyjablonski: Updated! The useEffects were handling different scenarios to account for on load, but I set up a ref to track initial load and merged into one useEffect

sites/partners/src/components/listings/PaperListingForm/sections/CommunityType.tsx
emilyjablonski: Yep! It all gets passed through react-hook-form this way.

ludtkemorgan (approved):
🦤

PR #6185, opened 2026-04-15
fix: swagger file update

emilyjablonski:
## Description

The recent upgrade of swagger docs required from the axios fix results in a slightly different swagger file generation.

I just ran `yarn generate:client` here!

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/src/types/backend-swagger.ts
+++ b/shared-helpers/src/types/backend-swagger.ts
@@ -5,12 +5,23 @@
 /** Generate by swagger-axios-codegen */
 /* eslint-disable */
 // @ts-nocheck
-import axiosStatic, { AxiosInstance, AxiosRequestConfig } from "axios"
+import axiosStatic from "axios"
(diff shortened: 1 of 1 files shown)


shared-helpers/src/types/backend-swagger.ts
```diff
@@ -1061,6 +1044,10 @@ export class ReservedCommunityTypesService {
 
       const configs: IRequestConfig = getConfigs("delete", "application/json", url, options)
 
+      /** 适配移动开发（iOS13 等版本），只有 POST、PUT 等请求允许带body */
+
+      console.warn("适配移动开发（iOS13 等版本），只有 POST、PUT 等请求允许带body")
```
ludtkemorgan, line 1049: note: With this change all delete endpoints now output this random string when it's called. We don't really use any of these endpoints so I'm not worried. But just calling out this weird behavior

ludtkemorgan (approved):

PR #6187, opened 2026-04-16
fix: limit partners from editing and unpublishing open listings

KrissDrawing:
This PR addresses #6148 

## Description

Will hide edit button for partners (for open listings). For attempt to access edit page from link it will redirect to listings/[id].
For backend it will give partner just read access.

Also not sure if that is something that is possible, but for seeded data partner will have access to listings, but won't have jurisdiction data for them (so will have for example access to angelopolis listings but wont have angelopolis feature flags applied). It seem like it cant be done from UI, so i just assumed it's an issue not feature.

Log in as partner with `disablePartnerOpenListingEdits` feture flag for selected jurisdiction.
for open listings we should not have edit button for listing. When listing is closed or draft it should be visible, and we should be able to edit listing. For open listing when we go to `/listings/[id]/edit` it should redirect to `/listings/[id]`.
Then check if for open listing backend will accept changes (i would just comment out:
```
 if (isEditOpenListingRestricted && listingId) {
    window.location.href = `/listings/${listingId}`
    return null
  }
```
in `ListingGuard.tsx` and then navigate via link to page `/listings/[id]/edit`.
With feature flag = `false` it should work as before, so partner should be able to edit open listings

## Author Checklist:

## Review Process:


emilyjablonski:
I only discovered this in testing, but we will need to also remove editability on closed listings, so that's what my commit is covering.

sites/partners/src/components/listings/ListingFormActions.tsx
```diff
@@ -83,11 +84,22 @@ const ListingFormActions = ({
...
-    "hideCloseListingButton",
+    FeatureFlagEnum.hideCloseListingButton,
+    listingJurisdiction?.id
+  )
+  const isPartnerPublicListingRestrictionEnabled = doJurisdictionsHaveFeatureFlagOn(
+    FeatureFlagEnum.disablePartnerPublicListingEdits,
     listingJurisdiction?.id
   )
 
+  console.log({ isPartnerPublicListingRestrictionEnabled })
```
YazeedLoonat: **nitpick:** 🧹

YazeedLoonat (approved):
LGTM tiny non blocking change

Issue #6068, opened 2026-03-18
Disclaimer to fees


sarahlazarich:
We need to add a disclaimer to the fees section in LA: This can go under "additional fees" card. We should use the new `tIfExists` function so that we don't need a feature flag. Those strings need to live only in the public overrides file. This means it should appear in core too, but we need to test locally removing the string and make sure it appears as expected too.

<img width="986" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeEJMRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--64cb1e97593f4e853325859e9acfa81aa1231a55/image.png" alt="image.png" />

The copy should read: 
- “Fees are not charged for reasonable accommodations needed due to disability, such as emotional support animals. Fees are also not charged for aids and services needed for effective communication due to a disability. With respect to credit checks, applicants receiving government rent subsidies have the option to provide “lawful, verifiable
alternative evidence” of their ability to pay rent, instead of relying on their credit history.”

AC:
- Disclaimer should load on LA's public site
- Disclaimer should translate

sarahlazarich:
hold on
- input from izzie on where this can go jurisdiction-wise
- thoughts from emily on where this should be around the fees card

KrissDrawing:
QA Notes: https://angelopolis.bloom.exygy.dev/listing/cd4d867f-e5df-47b8-bca3-7fb4dc1e2dab/angelopolitanczykowianeczka_blackburn_ave_los_angeles_ca
For example in this listing we can see this disclaimer

sarahlazarich:
lgtm

PR #6196, opened 2026-04-20
fix: add empty disclaimer to additional fees

KrissDrawing:
This PR addresses #6068

## Description

Adds empty additional fees disclaimer that will appear only after filling `"listings.additionalFeesDisclaimer"` key.
It's part one of task, those keys will be updated in LA.
Tbh not sure if thats correct approach with empty strings in overrides (but it seem right)

It should cause no changes for current implementation.
After filling`"listings.additionalFeesDisclaimer"` key for language, it should show it below additional fees.

## Author Checklist:

## Review Process:


sites/public/src/components/listing/listing_sections/AdditionalFees.tsx
```diff
@@ -40,69 +40,76 @@ export const AdditionalFees = ({
...
-              </div>
-            )}
-          </Card.Section>
-        </Card>
+            </Card.Section>
+          </Card>
+          {tIfExists("listings.additionalFeesDisclaimer") && (
+            <p className={`seeds-p-bs-4`}>{t("listings.additionalFeesDisclaimer")}</p>
+          )}
+        </>
```
KrissDrawing, line 112: basically those are the changes in this file. Rest is just due to wrapping component in fragment.

sites/public/page_content/locale_overrides/general.json
```diff
@@ -11,6 +11,7 @@
   "footer.content.projectOf": "Bloomington Housing Portal is a project of the",
   "footer.copyright": "© City & County of Bloomington",
   "footer.forGeneralQuestions": "For general program inquiries, you may call us at 000-000-0000.",
+  "listings.additionalFeesDisclaimer": "",
```
emilyjablonski: **issue:** This should have the copy in the ticket. In order to test it locally, we have to totally remove this key, but it should be merged in. This line still passes `tIfExists`

emilyjablonski (changes requested):

sites/public/page_content/locale_overrides/general.json
KrissDrawing: But shouldn't that copy be just in LA repo?
So i believe i should remove all keys here, and add them to LA `locale_overrides` 🤔
otherwise im missing the point.

emilyjablonski: When we're using the tIfExists function we want to be able to test what it will look like in core as the default, so we've been giving it generic core copy. The new disclaimer on the create account page is an example where we used Bloomington. In a fork we can then override it or remove the key entirely and nothing will show. It's sort of the opposite of how overrides usually work. I do think it's a bit awkward, so we will iterate at some point here for sure!

KrissDrawing: Oh, ok now i got it. Is there any action for me here, like i assume it will need to be removed for forks (so like doorway). Do we document such things somewhere?

emilyjablonski: No you're good for now I can do the release! I think part of the iteration should be making it so that these are opt-in instead of opt-out to your point, so that we won't need to remove it when bringing it to Doorway.

emilyjablonski (approved):

PR #6208, opened 2026-04-21
fix: reserved community type test fixes

ludtkemorgan:
This PR addresses issue with bloom-la release.

## Description

Two fixes to accommodate changes a forked jurisdiction might want and still allow tests to pass:
* reserved community types now have test id consistent rather than the custom text so testing for them in Cypress tests is consistent even if the jurisdiction has overriden the text
* When seeding data delete all community types already existing in the database. This is an issue because we create RCT with a jurisdiction attached, but on the UI we don't care what jurisdiction it is attached to. 
  * Long term solution would be to refactor how we use RCT so that either jurisdiction id does not matter or we create unique ones for every jurisdiction in the database and reference just the appropriate jurisdiction's version in the partner site

Also, there was a change recently that accidentally changed the public cypress tests in github to only run 1 test. This means that most of the tests haven't been running for the last few months and a bug slipped in. When starting an application and logging in you are redirected to the home page because the `listingId` is not getting appended to the redirect url

The tests should pass but also the one bug found should be tested for the following scenarios:
* When logging in the correct redirect url should be adhered to
  * Start the application flow by clicking "sign-in". After logging in the browser should redirect to the start of the application with the correct listing id
  * Malicious links in the redirect url should redirect to the account dashboard page

## Author Checklist:

## Review Process:


api/prisma/seed-staging.ts
```diff
@@ -1084,6 +1084,8 @@ export const stagingSeed = async (
 
   // create pre-determined values
   const unitTypes = await unitTypeFactoryAll(prismaClient);
+  // Clear all existing reserved community types and add new ones
+  await prismaClient.reservedCommunityTypes.deleteMany();
```
YazeedLoonat: **question:** I am unsure about this, in LA this is going to cause some issues (a difference between the reserved community types available with the ones created by the migrations being deleted here and the generic reserved community types created) 

I think maybe we should talk about this in a bit more depth but curious if @emilyjablonski feels similarly about this difference mattering

emilyjablonski: I could pull it over to LA to test it but my initial instinct would be that it would be ok - I think I might be missing something tho! What pieces would you expect would break?

YazeedLoonat: I don't think anything would break, but there would be differences between our local versions stood up and what LA actually has for reserved community types

emilyjablonski: Oh I see!! Yes I agree it would be ideal that LA migrations would be reflected when seeding over there - and removing the types here would cause a mismatch

ludtkemorgan: We are still going to run into mismatches in the LA repo since we don't filter by jurisdiction on the reserved community type. The way it currently works is it will display all of the LA and all of the seeded RCTs. So for example the Senior62 will show up twice. This will continue to cause headaches for our tests in a forked repo. 

I could get around the specific problem we have for LA by making the tests select community types that are not in the LA migration files. And also update which one of the seeded RCTs has a description. Would that work for everyone? We could potentially run into issues with the next jurisdiction or even LA if they change the community types they want, but it would at least save us some headaches in the near term until we can fully refactor how community types are saved and fetched.

ludtkemorgan: @YazeedLoonat @emilyjablonski Do you have any thoughts on this?

emilyjablonski: Confirming we chatted about a generic RCT for testing that doesn't exist in prod envs!

emilyjablonski: ^ Just checking in again about the generic RCT updates but no rush

ludtkemorgan: Update from after our discussion in eng workshop. I have proceeded with deleting all RCTs existing in the system before running the tests. Additionally test RCTs have been created, one for no description and one with a description. So this is ready to be reviewed

YazeedLoonat (approved):
lgtm

PR #6215, opened 2026-04-23
fix: upping 256 max units array size

YazeedLoonat:
This ups the max number of allowed units on a listing to 500

the easiest way to test this is to change the seed script to create a listing with >256 units and <500 units and verify that editing that seeded listing saves correctly


ludtkemorgan (approved):

ludtkemorgan:
@yazeed it looks like we need to up this to 600

yazeed:
@ludtkemorgan think you've got the wrong yazeed here

PR #6261, opened 2026-05-04
feat: app decline reason

emilyjablonski:
This PR addresses #6266 #6307

## Description

When the application status feature flag is on and a housing application is given the status declined, we need to show two new fields to the application edit screen, which appear conditionally.

Partners:
1. A dropdown field "Decline reason" to the right of application status, that only appears if the status is "Declined". It must be required if it is visible. The options are:

- Household income too high
- Household income too low
- Household size too large
- Household size too small
- Attempted to contact; no response
- Applicant declined unit
- Does not meet senior building requirement
- Household does not need accessible unit features
- Other

2. A free text field "Decline reason additional details" with the helper text below "For internal use only" that appears only when one of the following reasons is selected, and in those cases should be required:

- Attempted to contact; no response
- Applicant declined unit
- Other

4. The decline reason should show in the existing modal that alerts partners that saving will either send an email to an applicant or remove some information from view. If it's added, it should be included in the email update. If it's removed / the status changes away from decline, it should be clear that information will be removed for the public applicant.

5. This data should be available in the export if the flag is on.

Public:
1. The status pill should be visible on the individual application page.
2. The decline reason should be visible on the individual application page if the status is declined and the reason data exists. They should translate.

Partners:
In Angelopolis, add the decline status to an application. Ensure when you do, the reason dropdown appears, and then when a value is entered, the text area appears. The reason dropdown should be required. When saving, the reason should be included in the modal alerting partners it will be shown in the email.

Screenshot 2026-05-05 at 11 03 12 AM
Screenshot 2026-05-05 at 11 03 43 AM
Screenshot 2026-05-05 at 11 03 56 AM

<hr />

Public:
A user receiving a declined status update should see both the status and reason in the email. On the public site, the status should now show on an individual application page. If the status is declined, the reason should also show.

Screenshot 2026-05-05 at 10 56 45 AM
Screenshot 2026-05-05 at 11 02 07 AM

## Author Checklist:

## Review Process:


api/prisma/migrations/58_app_decline_reason_translations/migration.sql
```diff
@@ -0,0 +1,150 @@
+-- en
+WITH updated AS (
```
ludtkemorgan, line 2: issue: this breaks the local development flow when seeding the data.

With this change this is what is happening:
1. When you seed it first runs all of the migrations
2. This migration adds a new `en` translation entry without a jurisdiction id that only has the decline reason fields as there are no entries yet in the translation table. 
3. Then it adds all of the seeded values including running the translation-factory where it adds a new `en` entry that doesn't have a jurisdiction id. 
4. Now there are 2 english entries in that database and when it goes to fetch translations it only grabs the first one (the one with most translations missing) so emails are missing all of the fields.

My recommendation would be to do one of the following:
1. Change the way this migration is used to match how we normally just update translations in migrations rather than doing an `INSERT INTO`
2. Change the translation-factory to override existing translations

api/prisma/migrations/58_decline_reason_translations/migration.sql
```diff
@@ -0,0 +1,150 @@
+-- en
+WITH updated AS (
+  UPDATE translations
+  SET translations = jsonb_set(jsonb_set(jsonb_set(jsonb_set(jsonb_set(jsonb_set(jsonb_set(jsonb_set(COALESCE(translations, '{}'::jsonb), '{applicationUpdate,declineReasonChange}', '"Your application decline reason is %{value}"'::jsonb, true), '{applicationUpdate,declineReason,ageRestriction}', '"Age Restriction"'::jsonb, true), '{applicationUpdate,declineReason,incomeRestriction}', '"Income Restriction"'::jsonb, true), '{applicationUpdate,declineReason,unitRestriction}', '"Unit Restriction (e.g. bedroom count and/or occupancy)"'::jsonb, true), '{applicationUpdate,declineReason,programRestriction}', '"Program Restriction (e.g. CES unit, Veteran Unit, VASH unit, etc.)"'::jsonb, true), '{applicationUpdate,declineReason,attemptedToContactNoResponse}', '"Attempted to Contact; No Response"'::jsonb, true), '{applicationUpdate,declineReason,householdDoesNotNeedAccessibleUnit}', '"Household does not need accessible unit features"'::jsonb, true), '{applicationUpdate,declineReason,other}', '"Other"'::jsonb, true)
```
ludtkemorgan: issue: I think there might be an issue with the way this is getting generated. I ran this migration and this is what the email looks like 

Image

So it's adding the first line, but not any of the decline reasons.

ludtkemorgan (changes requested):
I still need to test some of the public user flow and won't get to that until tomorrow, but wanted to send my initial findings.

My recommendation would be to conform the translation migration more similar to how we update translations in other migration files

api/prisma/migrations/58_app_decline_reason_translations/migration.sql
emilyjablonski: Ope we probably need to update the whole db translations script! This was script output, I'll take a look at both

emilyjablonski: Digging in more here - can you share more about the issue you're seeing when seeding the data? All my emails are coming through with all strings / fields (an example screenshot was in the PR description) - this was all before the latest commit as well adjusting the seed

api/prisma/seed-helpers/upsert-translation.ts
```diff
@@ -0,0 +1,33 @@
+import { Prisma, PrismaClient } from '@prisma/client';
+import * as lodash from 'lodash';
+
+/**
+ * Create or merge translation JSON for one (language, jurisdictionId) pair
+ */
+export async function upsertTranslation(
```
ludtkemorgan: question: is there a reason this can't be in the `translation-factory`?

api/prisma/seed-helpers/upsert-translation.ts
```diff
@@ -0,0 +1,33 @@
...
+  const existing = await prisma.translations.findFirst({
+    where: { language, jurisdictionId },
+  });
+
+  if (!existing) {
+    await prisma.translations.create({ data });
+    return;
+  }
+
+  const merged = lodash.merge(
```
ludtkemorgan: this might cause issues in our forked repos if we are validating anything on emails because the translations could look different based on the migrations that are run. It most likely won't happen because you are choosing the seeded data as the highest priority and we should be only testing against things that are seeded.

I tested this change in the bloom-la repo and all tests passed. Just flagging it as there is a chance it could come up in the future.

api/prisma/migrations/58_app_decline_reason_translations/migration.sql
ludtkemorgan: Without your most recent commit this is what my all emails from bloom look like after seeding. 

Image

I've tested on main, without your most recent commit, and with your most recent commit. It only was happening when I was on not your most recent commit.

So good news is your most recent commit fixed it! There is a chance that the issue existed for you, but you were getting lucky in that when it fetched the english translation from the DB it was grabbing the one with all of the translations and not the one that only had the decline reason text.

api/prisma/migrations/58_decline_reason_translations/migration.sql
ludtkemorgan: For more context these are the steps that I did to test this migration without the influence of the translations added via the seeding script.

1. On the main branch run the seeding script
2. Pull down these changes and run migrations `yarn db:migration:run`
3. You will also need to run `yarn prisma generate` because of the new fields added
4. Do all of the flows to trigger this email and you can see not all of the translations were added

ludtkemorgan (commented):
I finished testing all the flows and it looks good. There is just the one issue with the translation migration file that still exists.

api/prisma/seed-helpers/upsert-translation.ts
emilyjablonski: Moved it!

ludtkemorgan (approved):
Thanks for making those changes! I'm seeing the translations now

PR #6262, opened 2026-05-05
fix: open waitlist, hide closed when no unit groups

emilyjablonski:
This PR addresses #6264

## Description

1. The open waitlist filter was not appropriately filtering on the listing status when unit groups is off
2. The closed waitlist filter should be hidden when unit groups are off

1. Ensure when unit groups are off, that the closed waitlist checkbox in the Availability section does not appear in the filter drawer
2. Ensure any listings with the gray "Open waitlist" status banner appear when filtering by open waitlist

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/listing.service.ts
+++ b/api/src/services/listing.service.ts
@@ -543,23 +543,21 @@ export class ListingService implements OnModuleInit {
                   key: ListingFilterKeys.availabilities,
                   caseSensitive: true,
                 });
-                return [
-                  {
-                    AND: builtFilter
-                      .map((filt) => ({
-                        unitGroups: {
-                          some: {
-                            [FilterAvailabilityEnum.openWaitlist]: filt,
-                          },
+                return {
+                  AND: builtFilter
+                    .map((filt) => ({
+                      unitGroups: {
+                        some: {
+                          [FilterAvailabilityEnum.openWaitlist]: filt,
                         },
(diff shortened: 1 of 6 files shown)


api/src/services/listing.service.ts
```diff
@@ -577,21 +577,39 @@ export class ListingService implements OnModuleInit {
...
+                      }))
+                      .concat(
+                        notUnderConstruction.map((filt) => ({
+                          marketingType: filt,
+                        })),
+                      ),
+                  },
+                  {
+                    AND: [
+                      { unitGroups: { none: {} } },
```
YazeedLoonat, line 598: should we be checking for non unitGroups here or would it be more accurate to check that some units exist instead? 

the reason I'm asking is if this puts us in a weird scenario where if a malformed listing with no units and no unit groups makes it in then checking for no unit groups means that it would show up when these filters are applied vs checking for units means that malformed listing would not show up since its missing both units and unit groups

hopefully that quebbin makes some sense

emilyjablonski: If there are no unit groups, then there are no unit groups with an open waitlist and it should not show, so I think this check is still ok - lmk if that doesn't resolve your question tho!

YazeedLoonat (approved):
LGTM

Issue #6246, opened 2026-04-29
Admin approval dialogs


emilyjablonski:
This should be under a new feature flag `enableAutopublish`. In the short term, we only expect to be using this in LA, which means listings approval will be on and `disablePartnerPublicListingEdits` will be enabled. We should ensure these are the configurations set when developing and testing.

[Prototype here](https://bloom-prototypes.netlify.app/scheduled-listing-publish-date.html) (still waiting for client approval)

When this new autopublish flag is on, and when a partner has submitted a listing for approval to an admin, there are a few changes to the confirmation dialog content admins should see when approving a listing.

The admin button "Approve & Publish" should just say "Approve" when this flag is on. 

There should be a new modal that appears for admins when clicking this button. Note this Approve / Approve & Publish button currently appears on BOTH the detail and edit pages. Whether or not this flag is on, we should remove it from the edit page and just show it on the detail page in all cases.

If a partner has entered a scheduled publish date, the approve modal copy should be:
`This listing is scheduled to publish automatically on X/XX/XXXX between 12:00 AM and 2:00 AM. Approving now will not publish it immediately.`

If a parter has entered a scheduled publish date, but it is now in the past, the approve modal copy should be:
`The scheduled publish date entered on the listing has passed. Approving this listing will publish it immediately. To delay publication, update the scheduled publish date before approving.`

If a partner has not entered a scheduled publish date, the approve modal copy should be:
`This listing has no scheduled publish date entered. Approving it will publish it immediately. To delay publication, set a publish date before approving.`

No other action is expected in this ticket apart from showing the right modal content when the flag is on. We can continue to just always immediately approve and publish for now and it will change to Open status.

KrissDrawing:
After QA Notes: After [this QA](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/6180) log in as an admin and check modals for approve for:
- date in the future
- date in the past
- no date at all

Also make sure approve text is updated

sarahlazarich:
looks good!

PR #6267, opened 2026-05-06
feat: add admin approve modal for enableAutopublish

KrissDrawing:
This PR addresses #6246 

## Description

Adds dialog approval modal for `enableAutopublish`.
Removes approve button from edit view.

With flag on create pending listing for review, submit it as a partner.
As admin you should have `approve` button on listing detail page (it is removed from edit page).
On click it will show different copy for different `scheduledPublishAt`.
On prisma studio you can modify it with date like `2026-06-15T00:00:00.000Z` to test:
- without this date
- date from the past (so today and before)
- date in the future (tomorrow and next)

when flag is off it should show `approve & publish` just for details page (it is removed from edit page)
check if it works as before

## Author Checklist:

## Review Process:


sites/partners/__tests__/components/listings/PaperListingForm/AdminListingApprovalDialog.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+    ).toBeInTheDocument()
+  })
+
+  it("renders future-scheduled body copy when publish window has not ended", () => {
+    render(
+      <AdminListingApprovalDialog
+        isOpen={true}
+        onClose={jest.fn()}
+        onConfirm={jest.fn()}
+        scheduledPublishAt={"2026-06-15T00:00:00.000Z"}
```
emilyjablonski, line 50: **note:** We should probably generate this to always be in the future, otherwise it might fail on June 15th - or use the system time piece below

sites/partners/page_content/locales/general.json
```diff
@@ -160,6 +160,7 @@
   "authentication.createAccount.firstName": "First name",
   "authentication.createAccount.lastName": "Last name",
   "errors.alert.emailConflict": "This email is already in use. Please contact your housing department if you're still experiencing issues",
+  "errors.alert.listingPublishError": "There are errors in this listing that must be resolved before publishing. To see the errors, please open the listing in edit view.",
```
emilyjablonski, line 178: **question:** When does this happen? How would the listing end up for an admin without having passed all the checks submitting for review by a partner?

emilyjablonski (commented):
Just some questions! Also semi-unrelated, but I'm unable to put tomorrow's date, but I should be able to:

Image

sites/partners/page_content/locales/general.json
KrissDrawing: To be honest i have not idea, i just put it here, cause it was hardcoded as a string. Also as we have no approve on edit the message was outdated, not sure if this one works, or it should be more try to save it in edit view. Or as you pointed out if we even need it at all, if so i believe it's super edge case

```
// if it is a bad request (is missing fields or incorrect data) then display an error banner
if (err.response?.status === 400) {
  setErrorAlert(
    "There are errors in this listing that must be resolved before publishing. To see the errors, please try to approve this listing from the edit view."
  )
}
```

sites/partners/__tests__/components/listings/PaperListingForm/AdminListingApprovalDialog.test.tsx
KrissDrawing: I added it in `beforeEach` above. Also double checked if it fails if i switch date to a minute before, so it should work. It might be cleaner if i add it inside test tbh

KrissDrawing:
Ok, i think it was comparing `utc`, to local time. I added time zone for backend error. I am not fully sure, but i think that frontend part should work fine, as im comparing `utc` to `utc`. But my brain really hurts from thinking about that, tho i will be definitely double checking it.
I remember checking that with some time travel chrome extension, but not sure how reliable that is 🤔

KrissDrawing:
@emilyjablonski Apparently i was wrong. But now it looks promising, frontend modal will display proper copy if user is in correct time zone (I tested it using `time travel` extension + sensors for LA) and for local timezone also with time travel extension.

emilyjablonski:
^ Just experimenting w Copilot!

sites/partners/page_content/locales/general.json
emilyjablonski: Oh interesting!! Assuming we somehow ran into this before, maybe with a data migration.

emilyjablonski:
I've pulled down the latest, and I'm still seeing that issue where I can't put in tomorrow's date. Lmk if it's not reproduceable though!

https://github.com/user-attachments/assets/5f367210-d915-48f0-9a67-1905838fb72c

KrissDrawing:
I added backend parsing for selected timezone for scheduledPublishedAt so it will check frontend dialog variant base on backend timezone time.

emilyjablonski (approved):
Tysm for those changes!

PR #6273, opened 2026-05-06
feat: prevent public site from rendering draft listings

YazeedLoonat:

## Description
Draft listings were able to be rendered on the public site as if they were a fully published listing

this pr will make it so that on the public site when you try to get to a draft listing the site doesn't render the page and instead scoots you back to the home back with an alert letting you know that listing was not available

on the public site do the follow:
- navigate to an open listing, it should render properly
- navigate to a closed listing, it should render properly
- in the url try and navigate to a draft listing, you should get scooted back to the home page with an alert

^ try that last step in a few different languages to make sure that the translations all work

## Author Checklist:

## Review Process:


shared-helpers/src/locales/general.json
```diff
@@ -792,6 +792,7 @@
   "housingCounselors.languageServices": "Language services: ",
   "housingCounselors.subtitle": "Talk with a local housing counselor specific to your needs.",
   "housingCounselors.visitWebsite": "Visit %{name}",
+  "invalidListing.redirect": "This listing is unavailable",
```
emilyjablonski: **note:** Let's put this in the section with `errors.`

sites/public/src/pages/listing/[id]/[slug].tsx
```diff
@@ -16,6 +17,7 @@ import dayjs from "dayjs"
 import { fetchJurisdictionByName } from "../../../lib/hooks"
 import { Jurisdiction, Listing } from "@bloom-housing/shared-helpers/src/types/backend-swagger"
 import { ListingViewSeeds } from "../../../components/listing/ListingViewSeeds"
+import { useRouter } from "next/router"
```
Comment (emilyjablonski) [inline sites/public/src/pages/listing/[id]/[slug].tsx]: **note:** Nit: This import should be moved up

YazeedLoonat:
Hey @emilyjablonski updated per your comments, lemme know what you think

emilyjablonski (approved):
LGTM but curious for other feedback about if this is extensive enough, vs not even letting the URL be hit at all!

sites/public/src/pages/listing/[id]/[slug].tsx
```diff
@@ -52,11 +56,22 @@ export default function ListingPage(props: ListingProps) {
     listing?.status,
     profile,
   ])
+  useEffect(() => {
+    if (!(listing?.status === "active" || listing?.status === "closed")) {
```
Comment (ludtkemorgan) [inline sites/public/src/pages/listing/[id]/[slug].tsx]: suggestion: since this is used twice and there is a chance we might want different statuses in the future to be accessible this could be changed to have the following:

Above the component have `const VALID_STATUSES = ["active", "closed"]`

and then in both of these places you could do
`!VALID_STATUSES.includes(listing?.status)`

ludtkemorgan (approved):

ludtkemorgan:
@emilyjablonski Do you have a suggestion on how we could implement not letting the URL be hit at all?

emilyjablonski:
@ludtkemorgan I don't personally, I thought I had interpreted some of your messages that that was desired so wanted to check in!

PR #6277, opened 2026-05-07
feat: agency mgmt

emilyjablonski:
This PR addresses #6288

## Description

We need to add an agency management feature in partners. It will function similar to how property management does. It should be enabled in settings in partners as another tab if the housing advocate feature is on. 

Users should be able to see a similar table with a filter input and an “Add agency” button. The only inputs in the “Add agency” drawer should be name which is required, and jurisdiction if the user has access to multipe jurisdictions which would also be required. Otherwise the jurisdiction field is hidden and automatically set to be the user’s only jurisdiction. 

We need similar success and failure toasts. 

There should only be edit and delete actions. The delete modal needs similar logic as well, where you are not able to delete an agency at all if it is associated with any advocate users. In that case, the heading should say “Agency in use” and the content “This agency is currently associated with user(s) and is unable to be deleted.” and the table can have just First/Last name and email address of the users. 

As an admin user, open the settings page and ensure you see the agencies tab. You should be able to see existing agencies, add new ones, edit agencies, and delete agencies without any associated users. You should not be able to delete agencies with associated users.

## Author Checklist:

## Review Process:


sites/partners/src/components/settings/PropertyDrawer.tsx
```diff
@@ -158,27 +160,29 @@ export const PropertyDrawer = ({
...
-                      validation={{ required: true }}
-                      inputProps={{
-                        onChange: () => {
-                          clearErrors("jurisdictions.id")
-                        },
-                        "aria-required": true,
-                        "aria-hidden": !!defaultJurisdiction,
-                      }}
-                    />
+                    {editedProperty ? (
```
emilyjablonski, line 163: **note:** This shifts from using a disabled select to a field value

sites/partners/src/components/settings/AgencyDrawer.tsx
```diff
@@ -0,0 +1,143 @@
...
+}: AgencyDrawerProps) => {
+  const { profile } = useContext(AuthContext)
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const { register, errors, clearErrors, trigger, getValues } = useForm<AgencyFormTypes>()
+
+  const jurisdictionOptions: SelectOption[] =
+    profile.jurisdictions.length !== 0
+      ? [
+          { label: "", value: "" },
+          ...profile.jurisdictions.map((jurisdiction) => ({
```
ludtkemorgan, line 40: issue: this is displaying all jurisdictions even if they don't have the feature flag enabled

sites/partners/src/components/settings/SettingsViewHelpers.tsx
```diff
@@ -5,35 +5,70 @@ import { Tabs } from "@bloom-housing/ui-seeds"
...
 }
 
-export const getSettingsTabs = (selectedIndex: SettingsIndexEnum, enableV2MSQ: boolean) => {
+type SettingsTabsFeatureFlags = {
+  enablePreferences: boolean
+  enableProperties: boolean
+  enableAgencies?: boolean
+}
+
+export const getEnabledSettingsTabCount = ({
```
ludtkemorgan, line 17: suggestion: this function is only ever being used to know if we should hide the tabs. Could we re-work this so that it returns true if at least one is enabled and that way you don't need to do `<= 1` every time it's called. I'd recommend to change the name of this function as well if we go that route

ludtkemorgan (changes requested):
Love it! Just found one issue

sites/partners/src/components/settings/AgencyDrawer.tsx
emilyjablonski: Oh we have the same problem w properties!! Will fix in both places

ludtkemorgan (approved):

Issue #6241, opened 2026-04-29
Doorway feature: filter by number of bathrooms


ludtkemorgan:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
Doorway allows for filtering results by number of bathrooms. This should be added to the search form in core. The backend logic already exists so this is just a frontend change.

Open questions:
* what numbers should we display as options?

## Technical Context
* Backend already exists for filter by bathrooms 
* The logic only does exact match whereas Doorway searches for that number + following numbers. 
* logic should be behind a new feature flag

## Acceptance Criteria
* When feature flag is enabled a filter by number of bathrooms appears in the search form

ludtkemorgan:
@hannan-abdi The initial round of this has been merged and you can find it in the core Bloomington environment https://bloom.exygy.dev/listings. It has the functionality, but it was implemented in the most basic way. You have to select each bathroom type you want to include in your search

We would like your input on the best way to implement this feature from a UI perspective. Specifically how could we have a design that you just need to select one option and all larger bathroom counts is included

This implementation:
<img width="372" height="167" alt="Image" src="https://github.com/user-attachments/assets/1811314b-7a41-44ed-bef7-8d27b009dbcb" />

Doorway's implementation:
<img width="487" height="130" alt="Image" src="https://github.com/user-attachments/assets/903e027e-5597-4656-8aa6-c4257e16a654" />

ludtkemorgan:
QA Notes:
* The initial round only did checkboxes as shown above. An additional round will be done to convert to radio buttons
* This has been enabled on [bridge bay environment](https://bridgebay.bloom.exygy.dev/listings)
* The "Bathroom" filter should appear on the filter drawer. When selected the appropriate bathroom selections should be filtered on

sarahlazarich:
lgtm

PR #6278, opened 2026-05-08
fix: add filter by number of bathrooms

ludtkemorgan:
This PR addresses #6241

## Description

Adds the "Bathrooms" filter to the public search form

image

I chose to have it be checkboxes rather than Doorway's radio button because that fits better with the patterns we have in core. Also, it is behind a feature flag

Open question:
* Is up to 4 bedrooms enough for options? If so, should the 4 bedroom option be changed to 4+ and have it be anything greater than or equal to 4?

After seeding the staging data on the public site using Bloomington you should see the "Bathrooms" filter and it should properly filter the listings when selections are chosen.

The filter should not appear for any other jurisdictions

## Author Checklist:

## Review Process:

### DIFF
--- a/sites/public/src/components/browse/FilterDrawerHelpers.tsx
+++ b/sites/public/src/components/browse/FilterDrawerHelpers.tsx
@@ -38,6 +38,7 @@ export interface FilterData {
   monthlyRent?: { [K in "maxRent" | "minRent"]?: string }
   regions?: { [K in RegionEnum]: BooleanOrBooleanString }
   configurableRegions?: string
+  bathrooms?: { [K in "1" | "2" | "3" | "4"]?: BooleanOrBooleanString }
   section8Acceptance?: BooleanOrBooleanString
   reservedCommunityTypes?: { [K in ReservedCommunityTypes]?: BooleanOrBooleanString }
   multiselectQuestions?: Record<string, BooleanOrBooleanString>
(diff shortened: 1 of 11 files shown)


sites/public/src/components/browse/FilterDrawerHelpers.tsx
```diff
@@ -444,6 +448,12 @@ export const encodeFilterDataToBackendFilters = (data: FilterData): ListingFilte
         filter[ListingFilterKeys.monthlyRent] = userSelections["maxRent"]?.replace(",", "")
         filters.push(filter)
       }
+      // } else if (filterType === ListingFilterKeys.bathrooms && userSelections) {
```
emilyjablonski: **note:** 🧹

emilyjablonski:
Screenshot 2026-05-11 at 4 11 09 PM
Do we need shared here, or no?

ludtkemorgan:
@emilyjablonski I have added Shared and 5 to match the partner site

image

emilyjablonski (commented):
As a user, I would personally expect if I am searching for 1 bathroom that listings with more than that would come back, but that's not the case here. Any thoughts on your expectations? As one example, Zillow resolves this with a +:

Screenshot 2026-05-15 at 12 42 13 PM

emilyjablonski (approved):

ludtkemorgan:
Merging this as is and will follow-up with design to come up with long term solution

PR #6292, opened 2026-05-12
feat: adds winston logger to the API to help with AWS logs

YazeedLoonat:
## Description
This adds the option of turning on the winston logger for the API so that the logs in AWS a better off.
The winston logger is turned on in production via the new env variable `USE_WINSTON`

In order to test this you'll need to change 2 env variables:
- Locally set `NODE_ENV` => `production`
- Locally set `USE_WINSTON` => `TRUE`

This will turn the winston logger on and then you can play around throwing errors and moving through the regular logs which should now show up as winston logs

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -82,7 +82,9 @@ async function bootstrap() {
     }
     cb(null, options);
   });
-  app.use(logger);
+  if (process.env.USE_WINSTON === 'TRUE') {
+    app.use(logger);
+  }
   app.use(
     cookieParser(),
     compression({
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -82,9 +82,6 @@ async function bootstrap() {
     }
     cb(null, options);
   });
-  if (process.env.USE_WINSTON === 'TRUE') {
-    app.use(logger);
-  }
   app.use(
     cookieParser(),
     compression({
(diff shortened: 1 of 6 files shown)


api/src/main.ts
```diff
@@ -9,12 +9,18 @@ import { AppModule } from './modules/app.module';
...
 import * as opentelemetry from '@opentelemetry/api';
+import { WinstonModule } from 'nest-winston';
+import { winstonLogger } from './logger/winston.logger';
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule, {
     logger:
       process.env.NODE_ENV === 'development'
         ? ['error', 'warn', 'log', 'debug']
+        : process.env.USE_WINSTON === 'TRUE'
```
ludtkemorgan, line 20: question: is there a reason we couldn't use winston for all production environments? Is it because winston doesn't work well with the Heroku setup?

YazeedLoonat: yeah winston logs I think don't play nice with papertrail so wanted to keep them separate, but also its possible that someone may not want to have winston in a prod env for whatever reason so figured we'd make it a controllable var

api/yarn.lock
```diff
@@ -1296,13 +1296,27 @@
   resolved "https://registry.yarnpkg.com/@colors/colors/-/colors-1.5.0.tgz#bb504579c1cae923e6576a4f5da43d25f97bdbd9"
   integrity sha512-ooWCrlZP11i8GImSjTHYHLkvFDP48nS4+204nGb1RiX/WXYHmJA2III9/e2DWVabCESdW7hBAEzHRqUn9OUVvQ==
 
+"@colors/colors@1.6.0", "@colors/colors@^1.6.0":
```
ludtkemorgan, line 1299: We are now importing all colors?! I hope we also have numbers and letters as well.

ludtkemorgan (approved):

PR #6293, opened 2026-05-12
feat: no demographics fixes

YazeedLoonat:
## Description
Currently if an application export is requested but one or more applications does not have demographics data the export will error

This change means that the export will generate and treat those missing demographics fields as null

the easiest way to test this is to delete demographics data tied to applications on a listing then request an applications export 

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/utilities/application-export-helpers.ts
+++ b/api/src/utilities/application-export-helpers.ts
@@ -409,6 +409,9 @@ export const getExportHeaders = (
         path: 'preferredUnitTypes',
         label: 'Requested Unit Types',
         format: (val: UnitType[]): string => {
+          if (!val?.map) {
+            return '';
+          }
           return val.map((unit) => unitTypeToReadable(unit.name)).join(',');
         },
       },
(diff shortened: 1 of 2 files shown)


api/src/utilities/application-export-helpers.ts
```diff
@@ -486,8 +489,14 @@ export const getExportHeaders = (
       {
         path: 'demographics.race',
         label: 'Race',
-        format: (val: string[]): string =>
-          val.map((race) => convertDemographicRaceToReadable(race)).join(','),
+        format: (val: string[]): string => {
+          if (!val?.map) {
```
ludtkemorgan: question: why is the .map needed? is it to validate that it is an array? I don't think any besides an array or null could exist in the database so you could just do `!val`

ludtkemorgan (approved):

api/src/utilities/application-export-helpers.ts
YazeedLoonat: good call out here, updated!

PR #6317, opened 2026-05-18
fix: make demographics race "Decline to respond" option exclusive

### LINKED ISSUE #2884: Demographics race checkbox set decline should be exclusive
If a user selects some checkboxes and then selects "decline to respond" the other checkboxes should deselect. If a user selects "decline to respond" and then another checkbox the decline should uncheck. We have this exclusive box behavior in preferences in the application.

youhaowei:
## Issue

Closes #2884

On the application **demographics** step, the race question allowed "Decline to respond" to be selected at the same time as other race options. It should be mutually exclusive — the same exclusive-checkbox behavior already used in the application preferences/programs step.

## Change

`sites/public/src/pages/applications/review/demographics.tsx`:

- Checking **"Decline to respond"** clears every other race option.
- Checking any other race option clears **"Decline to respond"**.
- Reuses the shared `setExclusive` helper (`shared-helpers/.../multiselectQuestions.tsx`) that powers the same behavior in `ApplicationMultiselectQuestionStep`. The per-field `onChange` is wired through `FieldGroup`'s existing `inputProps` — `FieldGroup` itself is unchanged.

## Testing

- New unit tests in `demographics.test.tsx` cover both directions of exclusivity; full demographics suite is 14/14 passing.
- Manually verified in a browser on the running demographics step — checking "Decline to respond" clears selected race options, and checking another option clears "Decline to respond".
- `tsc`, ESLint, and Prettier are clean.

## Known issue — not fixed here, needs discussion

`FieldGroup` (in `@bloom-housing/ui-components`) decides sub-option visibility from an internal **click-tracked** array (`checkedInputs`) rather than from form state. Because this fix unchecks options programmatically via `setValue` (which fires no click event), a parent option's sub-options stay **visually expanded** after the parent is cleared by the exclusive logic — e.g. expand "Asian", then check "Decline to respond": "Asian" and its sub-options uncheck correctly, but the sub-option rows remain on screen until the next interaction.

- Checked state and submitted data are **correct** — this is purely visual.
- It only occurs on the programmatic uncheck path; a normal user click still collapses sub-options correctly.
- Root cause is in `ui-components` `FieldGroup` and affects every consumer. The proper fix is to derive sub-option visibility from watched form state instead of the click-tracked mirror. That is a cross-repo change and is left for team discussion on how best to handle.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sites/public/__tests__/pages/applications/review/demographics.test.tsx
+++ b/sites/public/__tests__/pages/applications/review/demographics.test.tsx
@@ -233,6 +233,64 @@ describe("Demographics", () => {
       fireEvent.click(screen.getByRole("checkbox", { name: "Other / Multiracial" }))
       expect(await screen.findAllByTestId("otherMultiracial")).toHaveLength(2)
     })
+
+    describe("'Decline to respond' race exclusivity", () => {
+      const renderDemographics = () => {
+        const conductor = new ApplicationConductor({}, {})
+        conductor.config.featureFlags = [
+          { name: FeatureFlagEnum.enableLimitedHowDidYouHear, active: false } as FeatureFlag,
+        ]
+        conductor.config.raceEthnicityConfiguration = defaultRaceEthnicityConfiguration
+
+        render(
+          <AppSubmissionContext.Provider
+            value={{
+              conductor,
+              application: JSON.parse(JSON.stringify(blankApplication)),
+              listing: {} as unknown as Listing,
+              syncApplication: () => {
+                return
+              },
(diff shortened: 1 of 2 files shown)


youhaowei:
Re-triggering CI — the failures are in submit-application.spec.ts (unhandled auth-client error at the address step, before demographics) and map-address-popup (map view flake). Both are unrelated to this PR's changes. Re-running.

Issue #5985, opened 2026-03-03
Make map jurisdictions configurable


emilyjablonski:

ludtkemorgan:
QA Notes:
* This can be tested in the new Bloom-doorway site https://bloom-doorway.bloom.exygy.dev/ and https://partners-bloom-doorway.netlify.app/. 
* All jurisdiction's listings should appear on the public site

Note: The jurisdiction names are tied to the made up names in core and won't match Doorway's jurisdictions

sarahlazarich:
wow!!

PR #6323, opened 2026-05-18
fix: add sub jurisdictions to allow multiple jurisdictions in search

ludtkemorgan:
This PR addresses #5985 

## Description

With a region based search we need to be able fetch listings from multiple jurisdictions tied to the higher level jurisdiction.

This PR adds a new relationship to the jurisdiction table for "sub jurisdictions". This allows for a jurisdiction to have subjurisdictions attached to it that all appear in the search. For example in Doorway the "Bay Area" jurisdiction would have sub jurisdictions of all of the individual counties in the bay area. 

Additionally, I noticed we were both returning more data than needed in the get jurisdictions endpoint as well as calling this endpoint in more places than we should have. This PR does the following:
* switches the call to use a `select` of just the fields we need rather than an `include`
* All of the calls within the partner site are now getting the jurisdiction data from the user's profile instead of making an additional call to the backend

This can be tested with the new "Bridge Bay" and sub jurisdictions
1. in the api directory run `yarn setup:region`
2. Set the jurisdiction to Bay Bridge in the public .env file and start up the site
3. On the public site you should see listing from all jurisdictions 
4. turn off the `enableListingMap` feature flag and restart the public site. In the list view you should also see listings from all jurisdictions

## Author Checklist:

## Review Process:


api/src/dtos/jurisdictions/jurisdiction.dto.ts
```diff
@@ -254,4 +255,10 @@ export class Jurisdiction extends AbstractDTO {
   @Type(() => RaceEthnicityConfiguration)
   @ApiPropertyOptional({ type: RaceEthnicityConfiguration })
   raceEthnicityConfiguration?: RaceEthnicityConfiguration;
+
+  @Expose()
+  @Type(() => RaceEthnicityConfiguration)
```
mcgarrye: **issue:** Incorrect type

api/src/services/jurisdiction.service.ts
```diff
@@ -62,16 +90,18 @@ export class JurisdictionService {
...
       throw new BadRequestException(
         'a jurisdiction id or jurisdiction name must be provided',
       );
     }
+    const viewToUse = condition.view || JurisdictionViews.public;
 
     const rawJurisdiction = await this.prisma.jurisdictions.findFirst({
       where: this.buildWhere(condition),
-      include: view,
+      select: selectViews[viewToUse],
```
mcgarrye, line 130: **question:** Passing `full` returns an empty object?

mcgarrye:
The seed is for `Bridge Bay` not Bay Bridge, had me thinking I couldn't read based on your testing instructions lol

mcgarrye (approved):
Looks good, no major issues. Few callouts
1. Home page of public site does not display the correct jurisdiction name
2. If the sub jurisdictions have the enableListingMap set to true, but the top level jurisdiction does not, we follow the top level. This makes sense but we should document
3. We don't need to account for non-seeds going forward, correct?

api/src/services/jurisdiction.service.ts
ludtkemorgan: I didn't find any use case where we would need the full response. But I should probably add the correct selects so we have it for the future. I'll do that now!

ludtkemorgan: I added a `full` view and made it the default for the list endpoint. As far as I can tell it's only used on the partners /admin page

mcgarrye (approved):
Approved again!

Issue #5837, opened 2026-02-03
LA Listing Notifications - Email copy updates


emilyjablonski:

KrissDrawing:
@emilyjablonski  For doorway, below listing name, we have:
```text
THIS INFORMATION MAY CHANGE
Please view listing for the most updated information
```
should that stay or should i remove it?
Also this email will be just in english? With button also just in english?
As with all of them it looks like that :D
<img width="269" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL3FyRFE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--736be76bdf33ecc565b6a587fa3deca2869cf591/image.png" alt="image.png" />

emilyjablonski:
@sarahlazarich I might direct the above questions to you!

PR #6340, opened 2026-05-25
feat: email unsubscribe and settings link

matzduniuk:
This PR addresses #5954 

## Description

* Updates the rental opportunity email translation strings
* Updates the rental opportunity email template
* Adds the sign-in with  email settings redirect link to the rental opportunity email template

1. Run the API `setup` script to apply new migrations
2. Make sure your test user account (with the preferred email has at least one of the `UserNotificationPreferences` set
3. Go to the partner's site and publish/save a published listing with a configuration matching the preferences set in step 2 to trigger the email send.
4. On the received email, scroll down below the buttons section and click the `Unsubscribe and manage email setting`
5. The user should be redirected to the jurisdiction public sign-in page with the `redirectUrl=/account/notifications` URL param attached
6. Sign in as a valid user
7. The user should be automatically rediercted to the account email notification preferences settings page

## Author Checklist:

## Review Process:

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


emilyjablonski (changes requested):
Some things I'm noticing, not all of which are caused by these changes, so happy to split these up if we need to! Specific to this PR, it's only the below, otherwise the redirect looks great:
* The string is `Unsubscribe and manage email setting` but it should be plural `settings`

The below are unrelated to this change, but matter for the notifications feature as a whole. I'm 100% okay with making these into a separate ticket if you would prefer.
* Locally if I create a new account and update notification settings, I'm unable to get any emails to send unless I remove `userPreferences: { sendEmailNotifications: true }` from the emailUsers query. It doesn't look like that is being set anywhere - not sure if it can be removed or if it's having some side effect I'm not aware of?
* I'm not able to get notifications for waitlist emails. We have a check for lottery, we also need one for waitlist.
* I'm not able to get notifications for regions in Angelopolis. I think this is because we're checking the `region` field, but LA is using `configurableRegion` (confusing I know). We should be fine to check both like `const listingRegion = listing?.configurableRegion || listing?.region;`
* The listing template needs to show region (configurableRegion) if the value exists which was part of [this ticket](https://github.com/bloom-housing/bloom/issues/5837), but it's not displaying.
* The listing template is always showing accessible marketing flyer even if one does not exist on the listing.
Image

* We're not using consistent naming for the accessibility types. The email should match what is in the form, which is to use `Hearing/Vision`. The email should not say `Hearing and Vision`.
Image

Image

api/src/services/listing.service.ts
```diff
@@ -2958,18 +2958,29 @@ export class ListingService implements OnModuleInit {
...
+    let preferenceFilters = [];
+
+    preferenceFilters = priorityTypes.map((priorityType) => {
       return {
         [priorityType]: true,
-      } as Prisma.UserNotificationPreferencesWhereInput;
+      };
     });
 
     if (listing?.region) {
```
emilyjablonski: **issue:** This still has the configurable region issue. An LA listing will have a null `listing.region` but may have a value in `listing.configurableRegion`.

api/src/services/listing.service.ts
```diff
@@ -2988,11 +3005,11 @@ export class ListingService implements OnModuleInit {
           email: '',
         },
         AND: [
-          {
-            userPreferences: {
-              sendEmailNotifications: true,
-            },
-          },
+          // { TODO: consider enabling this when this flag will be available to update from user account page
```
emilyjablonski, line 3003: **question:** Is there a future state where we expect a new flag to be available on accounts?

emilyjablonski (commented):
Tysm for all the updates!! It's just the remaining configurable region filtering piece.

There is also an unrelated issue I made a new ticket for, where on an open listing, if you just make a change and save it, it sends out the notification email again as if a new listing has been published.

matzduniuk:
> Tysm for all the updates!! It's just the remaining configurable region filtering piece.
> 
> There is also an unrelated issue I made a new ticket for, where on an open listing, if you just make a change and save it, it sends out the notification email again as if a new listing has been published.

This is something we talked about when this ticket originally went for a PR (raised by @YazeedLoonat), but we decided to leave it as be till we do performance testing. This mainly depends on what updates we expect to be applied to published listings. For example, if a partner adds new units to an existing listing, such that they provide new accessibility opportunities, such an update would not send a notification to users who selected such a preference.
The most balanced solution would be a set of checks like:
1. If the status changes to active (i.e., first publish or republish), we always send an email
2. If the status is already active (i.e., edit to publish listing), perform a check against the currently stored listing to verify that the notification preference targeted fields have changes. If not, don't send the email.

Tho this solution would require additional computation to validate if an updated listing is valid for notification + another source of truth we would need to keep updated (other than the email formatting helper), where we would need to keep track of in case of any changes to make sure the service works as expected (integration tests should handle this).

api/src/services/listing.service.ts
matzduniuk: I assumed that this flag does not exist for no purpose. 
Depending on the results of the service performance tests, we could think about including this (either as a toggle in the APP or automatic update based on whether any notification preferences have been selected). 
If we come to a conclusion that the user query is too time-consuming, we could consider sacrificing memory and create a partial index on the UserAccounts table for only those that have sendEmailNotifications set as true.

emilyjablonski (approved):
So excited!!!

Issue #5219, opened 2025-08-05
User Deletion Functionality


mcgarrye:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Public users who have been inactive for 3 years need to have their account deleted and the PII removed from any applications.

## Technical Context

Add `deleteUserAfterInactivity` function to the [UserService](https://docs.google.com/document/d/1bCXeoQ0153h_z7hKHm3pVBAjz0xfdR3rR6dK-Zma58c/edit?tab=t.0#heading=h.nqeit7fswuef)
Add `deleteAfterInactivity` function/cronjob to the [UserService](https://docs.google.com/document/d/1bCXeoQ0153h_z7hKHm3pVBAjz0xfdR3rR6dK-Zma58c/edit?tab=t.0#heading=h.xzquylmehypi)

## Acceptance Criteria
If a public user has been inactive for `USERS_DAYS_TILL_EXPIRY` (3 years) and has receive the warning email and did not sign in, their account should be deleted and application pii should be removed.

## QA Notes
Requires looking at the db after application creation, listing close and cronjob completion

Set USERS_DAYS_TILL_EXPIRY to 0
Create a public user account with your email
Apply to a listing
Sign out
Trigger email cronjob (either by calling the endpoint or setting the cron env var within a small window)
Should receive warning email
Trigger deletion cronjob (either by calling the endpoint or setting the cron env var within a small window)
Verify expected data is deleted from the DB
Attempt to sign in, should fail

kathyccheng:
Just public users, thanks!

PR #6354, opened 2026-05-27
fix: add last login at to update password flow

ludtkemorgan:
This PR addresses https://github.com/bloom-housing/bloom/issues/5218 and https://github.com/bloom-housing/bloom/issues/5219

## Description
NOTE: this has already been merged into Doorway

During last minute testing it was identified that there are users in the system that the last_login_at field is not up to date. This happens when the last time a user logged in was via the forgot password flow. In that flow we were not updating the last_login_at field.

This causes issues because that is the field we are using to determine if a user should be both deleted as well as warned of deletion. The fix is to both update that flow to start setting that field as well as updating both of those cron jobs to look at the password_updated_at field

Also, after doing checks on how many people will get deleted in the first run I realized it would be close to 13K users. So I added a LIMIT to the number. The first time it is run in production I will run it 13 times to get everyone

Go through the whole warn user and delete user flows:

1. Create several public users
2. Change their last_login_at to be well in the past
3. Change their password_updated_at to be well in the past
4. Run both cron jobs and they should successfully warn and delete those users

Go through the forgot password flow and make sure the last_login_at field is updated

## Author Checklist:

## Review Process:

### DIFF
--- a/api/prisma/seed-helpers/user-factory.ts
+++ b/api/prisma/seed-helpers/user-factory.ts
@@ -17,6 +17,7 @@ export const userFactory = async (optionalParams?: {
   password?: string;
   phoneNumber?: string;
   phoneNumberVerified?: boolean;
+  passwordUpdatedAt?: Date;
   roles?: Prisma.UserRolesUncheckedCreateWithoutUserAccountsInput;
   singleUseCode?: string;
   lastLoginAt?: Date;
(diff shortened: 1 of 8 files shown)


mcgarrye (approved):
Looks good, solves issue found.

I did realize when testing that if someone signs in after the email and then hypothetically time passes until they are up for deletion again, we never set `wasWarnedOfDeletion` back to false so they will just be deleted on the second go instead of warned again. This can be fixed in a later PR or added in here

PR #6358, opened 2026-05-28
fix: allow setting seed jurisdiction name override, default seed

emilyjablonski:
## Description

Allows for Angelopolis seed name to be overridden, such that LA can set the new environment variables when testing locally and can just run the staging seed and have one jurisdiction with a custom name be added.

Adds two new backend env vars - `DBSEED_JURISDICTION` and `DBSEED_JURISDICTION_NAME`. When the jurisdiction is set, only that jurisdiction is seeded. When the name is set, that seeded jurisdiction uses that as the name (moving underscores to spaces so it is script friendly).

LA would set:
DBSEED_JURISDICTION=Angelopolis 
DBSEED_JURISDICTION_NAME=Los_Angeles

Without these, ensure all 4 jurisdictions are seeded.
With these two values, ensure only the Angelopolis-like jurisdiction is seeded and named `Los Angeles`.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/prisma/seed-staging.ts
+++ b/api/prisma/seed-staging.ts
@@ -19,15 +19,16 @@ export const stagingSeed = async (
     publicSiteBaseURL,
     msqV2,
     asRegion = false,
-    angelopolisJurisdictionName,
+    jurisdiction,
   }: {
     jurisdictionName: string;
     publicSiteBaseURL: string;
     msqV2: boolean;
     asRegion: boolean;
-    angelopolisJurisdictionName?: string;
+    jurisdiction?: string;
   },
 ) => {
+  const resolvedJurisdictionName = jurisdictionName?.replace(/_/g, ' ');
   // Seed feature flags
   await createAllFeatureFlags(prismaClient);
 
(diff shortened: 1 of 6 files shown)


mcgarrye (approved):
lgtm

ludtkemorgan (commented):
I know this is already approved and I'm ok merging this as is so this could be a follow-up. Rather than making this be a jurisdiction specific override we could make it a bit more generic. I haven't fleshed it out fully but I was thinking something like this:

* Add a new flag to the seed script `jurisdiction` (e.g --jurisdiction Angelopolis)
* In seed-staging take that new flag in and if it exists only seed that specific jurisdiction.
* Use the jurisdictionName field as an override for that passed in jurisdiction. 

So the flow would be like this:
1. call this command: `yarn db:setup:staging --jurisdiction Angelopolis --jurisdictionName Los_Angeles`
2. The seed script only populates Angelopolis with name Los Angeles (note we will have to remove the underscore, the terminal command doesn't like spaces)

emilyjablonski:
@ludtkemorgan This didn't have a big rush so I made some changes here to make it more configurable! Lemme know what you think

api/prisma/seed-staging.ts
```diff
@@ -19,13 +19,16 @@ export const stagingSeed = async (
...
+    jurisdiction,
   }: {
     jurisdictionName: string;
     publicSiteBaseURL: string;
     msqV2: boolean;
     asRegion: boolean;
+    jurisdiction?: string;
   },
 ) => {
+  const resolvedJurisdictionName = jurisdictionName?.replace(/_/g, ' ');
```
ludtkemorgan: since this is implemented via an environment variable we don't have the issue with spaces. It's only an issue when adding a parameter from the terminal command

api/prisma/seed.ts
```diff
@@ -20,6 +20,8 @@ async function main() {
     values: { environment, jurisdictionName, msqV2, asRegion },
   } = parseArgs({ options });
   const publicSiteBaseURL = env.DBSEED_PUBLIC_SITE_BASE_URL;
+  const seedJurisdiction = env.DBSEED_JURISDICTION;
+  const seedJurisdictionName = env.DBSEED_JURISDICTION_NAME ?? jurisdictionName;
```
ludtkemorgan, line 24: This works and I'm cool with it. My original idea would be just call the seeding script with the desired parameters. So the changes where would be to add the new parameter in the options above and pass those directly to the stagingSeed below.

```
yarn db:setup:staging --jurisdiction Angelopolis --jurisdictionName Los_Angeles
```

The nice thing about the way you did it is it's more discoverable.

ludtkemorgan (approved):

PR #6343, opened 2026-05-26
fix: add scheduled action buttons with dialogs

KrissDrawing:
This PR addresses #6181 

- [x] Addresses the issue in full
- [ ] Addresses only certain aspects of the issue

## Description

Adds actions and dialogs for scheduled state (Make sure that [#6247](https://github.com/bloom-housing/bloom/pull/6295) is reviewed as it base on that)

## How Can This Be Tested/Reviewed?

Set listing status to `scheduled`, then on partners check listing actions for details and edit.
Also test save, unapprove and publish dialogs and actions they are triggering

## Author Checklist:

- [ ] Added QA notes to the issue with applicable URLs
- [x] Reviewed in a desktop view
- [x] Reviewed in a mobile view
- [ ] Reviewed considering accessibility
- [x] Added tests covering the changes
- [ ] Made corresponding changes to the documentation
- [ ] Ran `yarn generate:client` and/or created a migration when required

## Review Process:

- Read and understand the issue
- Ensure the author has added QA notes
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Either (1) explicitly ask a clarifying question, (2) request changes, or (3) approve the PR, even if there are very small remaining changes, if you don't need to re-review after the updates

Issue #6213, opened 2026-04-22
Autopublish cron job


emilyjablonski:
For listings in a scheduled status, a new cron job should run at midnight and automatically publish listings that are scheduled for that date. For now, we can publish them all at midnight, but there are some unresolved questions detailed below.

Question: Should we send out an email notification to admins that a listing was automatically published?
Question: The PRD indicated we need to either retry and/or alert if there was a failure. How should we plan to do that?
Question: Pending Notifications load testing results, we may need to stagger when we send out the notifications for performance, i.e. if we have 10 listings going live at midnight and need to send many thousands of emails. Our current window is 12:00 AM - 2:00 AM

sarahlazarich:
Sep ticket to outline what emails send when, we do want an email to go to adminds at published 
Sep ticket to adjust approval emails to say "will be scheduled to approve at x "

sarahlazarich:
dont want to retry - but could run chron job twice to try to pick up everything/anything that was missed

sarahlazarich:
ticket just to make chron job run as 12:01 and 1 am, make sure the logs are good

KrissDrawing:
Not sure where to test that Cron Job

Issue #6181, opened 2026-04-15
Admin updates when in scheduled status


sarahlazarich:
This should be under a new feature flag `enableAutopublish`. In the short term, we only expect to be using this in LA, which means listings approval will be on and `disablePartnerPublicListingEdits` will be enabled. We should ensure these are the configurations set when developing and testing.

[Prototype here](https://bloom-prototypes.netlify.app/scheduled-listing-publish-date.html) (still waiting for client approval)

We need to ensure when a listing is in the scheduled state, these are the listing actions available. There is no partner/edit set because we are assuming `disablePartnerPublicListingEdits` will be enabled.

In a scheduled state, listing detail page partner actions should be "Copy" and "Preview"
In a scheduled state, listing detail page admin actions should be "Publish" "Edit" "Copy" "Preview"
In a scheduled state, listing edit page admin actions should be "Save" "Unapprove" "Exit"

Unapprove is new - this should move the status from `scheduled` to `pendingReview`. This should show a modal with this copy:
`This listing will be returned to under review status and will no longer be scheduled to automatically publish. It will need to be re-approved before it can be published.`

Publish in this context is new - this will override the scheduled date and publish it. This should show a modal with this copy:
`This listing will go live immediately, overriding the scheduled publish date.`

Save in this context should show this copy in a dialog:
`This listing is approved and scheduled to publish on X/XX/XXXX between 12:00 AM and 2:00 AM. Saving will not affect its scheduled publication on this date.`

In a scheduled state, if an admin removes the scheduled publish date, and then Saves, the dialog should have this copy:
`This listing no longer has a scheduled publish date entered. As it is already approved, saving will publish it immediately. To delay publication, set a publish date before saving, or Unapprove the listing.`

sarahlazarich:
May be edge cases that need to be explored 
We will need copy for warnings

PR #6359, opened 2026-05-29
fix: add cron job for autopublish

KrissDrawing:
This PR addresses #6213 

## Description

Review after [#6181 ](https://github.com/bloom-housing/bloom/pull/6343)
Adds publish listing cron job that will publish listing when in scheduled state and scheduledPublishedAt date is in the past.

I added some quicker strings sth like:
```
LISTING_SCHEDULED_PUBLISH_CRON_STRING=* * * * *
LISTING_SCHEDULED_PUBLISH_RETRY_CRON_STRING=* * * * *
```
So there we need to test that it wont publish listings with scheduledPublishedAt in the future.
and that it publish it when its in the past.
It requires `scheduled` state.

## Author Checklist:

## Review Process:


api/.env.template
```diff
@@ -38,6 +38,10 @@ PARTNERS_PORTAL_URL=http://localhost:3001
 EMAIL_API_KEY=SG.ExampleApiKey
 # controls the repetition of the listing cron job
 LISTING_PROCESSING_CRON_STRING=0 * * * *
+# controls the primary scheduled listing auto-publish cron job
+LISTING_SCHEDULED_PUBLISH_CRON_STRING=1 0 * * *
+# controls the retry scheduled listing auto-publish cron job
+LISTING_SCHEDULED_PUBLISH_RETRY_CRON_STRING=0 1 * * *
```
YazeedLoonat: **suggestion (non-blocking):** my recommendation is to make the retry have a longer start up time, lets bump this to start at 2 instead of 1 just so we can make sure the initial run has completed

YazeedLoonat (approved):
LGTM with a small suggestion

api/src/services/listing.service.ts
```diff
@@ -3335,6 +3349,91 @@ export class ListingService implements OnModuleInit {
...
+    return this.publishScheduledListings('publishScheduledListingsCron');
+  }
+
+  async publishScheduledListingsRetryCronJob(): Promise<SuccessDTO> {
+    this.logger.warn('publishScheduledListingsRetryCron job running');
+    await this.cronJobService.markCronJobAsStarted(
+      LISTING_SCHEDULED_PUBLISH_RETRY_CRON_JOB_NAME,
+    );
+
+    return this.publishScheduledListings('publishScheduledListingsRetryCron');
```
ludtkemorgan: @YazeedLoonat @KrissDrawing I know this is already approved and I'm not opposed to this solution. But my original thought was we could just use the same job and have the cron string have multiple time. For example if we want to run at midnight and 1 am it would look like this `0 0,1 * * *`

KrissDrawing: I went for this solution for slightly more informative logs, but not sure if they are needed (as we can probably tell by the time of log what job it was). And i can see that even if it could be useful to have that separation, in our case it probably wont be used. This said i can change it to one function if there are no objections.

ludtkemorgan: Thanks. My preference would be just the one cron job for code and env variable simplicity. If we need to add more logging I recommend we add more error logging for failures

api/src/services/listing.service.ts
```diff
@@ -3347,6 +3354,77 @@ export class ListingService implements OnModuleInit {
...
+        ],
+      },
+    });
+
+    const listingIds = listings.map((listing) => listing.id);
+    this.logger.warn(
+      `${logName} found ${listingIds.length} listing(s) ready to publish`,
+    );
+
+    if (listingIds.length > 0) {
```
ludtkemorgan: nit: we could wrap everything below this in this if statement since we don't need to do anything if there are no listings meeting the requirement. Or even better yet have a quick return here if listingIds is empty
```suggestion
if (!listingIds.length) {
  return { success: true }
}
```

ludtkemorgan (approved):

PR #6391, opened 2026-06-04
fix: set default form submit as post

emilyjablonski:
Some form submissions if done before hydration are being submitted as `GET` instead of `POST` because we are not setting post on our internal Form component.

## Description

The `form` element has no method set and defaults to `GET` which results in all form values being added as URL parameters. `react-hook-forms` will call preventDefault before we run our custom submit handlers which use `POST`. With Next.js, the server renders the HTML first and then the JS bundles asynchronously load, and if you submit before that finishes, the browser handles it natively and would `GET` in this scenario.

On a very slow connection and submitting very quickly is when submitting before hydration is most likely.

We were using a mix of a shared-helpers Form and one from UIC. They both had the issue, but I migrated us to the shared-helpers one so that we could make the change in core.

On main to see this happen, you can bypass the form itself and see the behavior, you can run `document.getElementById('sign-in').submit()` directly in the browser console, to mimic bypassing React's event handling, to see the parameters added into the URL. That no longer happens on this branch.

For us, only filter/search experiences should be `GET`, so those are set explicitly and we otherwise default to `POST`.

Most of our forms should also suppress submit on Enter as they are longer and more involved. Only our account access flows like signing in or the forgot password flow are passing that flag, allowing the form to submit on Enter.

## Author Checklist:

## Review Process:

### DIFF
--- a/shared-helpers/__tests__/Form.test.tsx
+++ b/shared-helpers/__tests__/Form.test.tsx
@@ -1,0 +1,12 @@
+import React from "react"
+import { render, cleanup } from "@testing-library/react"
+import { Form } from "../src/views/forms/Form"
+
+afterEach(cleanup)
+
+describe("<Form>", () => {
+  it("renders with method post", () => {
+    const { container } = render(<Form>content</Form>)
+    expect(container.querySelector("form")).toHaveAttribute("method", "post")
+  })
+})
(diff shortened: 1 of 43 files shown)


YazeedLoonat (approved):
LGTM

Issue #6225, opened 2026-04-27
"This is a testing site" Banner for dev public sites


YazeedLoonat:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
For both Doorway and LA we have run into an issue where real applicants have found the dev public sites and submitted applications. 

Doorway has gone with a "Test Listing" image but a banner flagged on with an env variable for the dev sites would hopefully dissuade that without having to require all images being the same

PR #6399, opened 2026-06-10
feat: creating a dev site banner

YazeedLoonat:
This PR addresses https://github.com/bloom-housing/bloom/issues/6225

## Description
This allows through an environment variable the ability to to add a banner on the public site indicating it is not a production environment and applications will not be considered.

This also pulls in missing behavior from the deprecated site header that will also allow for a maintenance banner.

Set `NON_PROD_BANNER` to `TRUE` to see the new banner on the public site.

You can also set the maintenance environment variable and can see both banners are able to be applied at the same time if needed (`MAINTENANCE_WINDOW=2024-02-05 18:00 -08:00,2027-02-05 18:55 -08:00`).

These should still be after the skip to main content link.

The string was also added in the overrides file as the expectation is a jurisdiction will need to add their production URL.

Screenshot 2026-06-23 at 3 42 36 PM

## Author Checklist:

## Review Process:


sites/public/src/patterns/SiteHeader.tsx
```diff
@@ -436,7 +436,14 @@ export const SiteHeader = (props: SiteHeaderProps) => {
       )}
       {props.showMessageBar && (
```
emilyjablonski, line 465: **issue:** We had chatted in slack about re-using the existing banner - let's just pass in the message from the <SiteHeader> usage itself, as opposed to duplicating the banner - is there different guidance from design?

shared-helpers/src/locales/general.json
```diff
@@ -1325,6 +1325,7 @@
   "t.delete": "Delete",
   "t.deposit": "Deposit",
   "t.description": "Enter description",
+  "t.devSiteBanner": "This site is for testing only. All data on this site is for testing only. Please do not apply",
```
emilyjablonski: **issue:** Is this copy approved by product / design? I would suggest removing the middle sentence, but at the least being consistent about formatting and including a period for the last one

sites/public/.env.template
```diff
@@ -47,3 +47,6 @@ CACHE_REVALIDATE=30
 # for listings map
 GOOGLE_MAPS_API_KEY=
 GOOGLE_MAPS_MAP_ID=
+
+# When TRUE
```
emilyjablonski: **issue:** Let's complete this comment, and make the default value false

emilyjablonski (changes requested):
The spacing around this message is awkward - left a comment below but would prefer to use the existing banner - let me know what you think! Would also love to see a test
Image
Image

ludtkemorgan (approved):
Looks great!

emilyjablonski (approved):

PR #6400, opened 2026-06-10
fix: dependabot multer

emilyjablonski:
## Description

Upgrading this transitive dependency will allow [these 4 dependabot vulnerabilities](https://github.com/bloom-housing/bloom/security/dependabot?q=package%3Amulter+manifest%3Aapi%2Fyarn.lock+is%3Aopen) to be resolved.

It's only a patch upgrade so risk is low. Testing simple requests/responses work like our Cypress tests do will cover the main functionality of this package.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -51,7 +51,7 @@
     "@nestjs/core": "^10.4.4",
     "@nestjs/jwt": "^10.2.0",
     "@nestjs/passport": "^10.0.3",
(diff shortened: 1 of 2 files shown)


ludtkemorgan (approved):
Looks like this version of multer is already a year old but it doesn't get updated very often so that's probably expected

PR #6408, opened 2026-06-11
fix: clear temp files cron job crash

emilyjablonski:
[Code scanning alert](https://github.com/bloom-housing/bloom/security/code-scanning/37)

## Description

This PR fixes an issue with the temp file cron job.

The previous implementation used fs.readdir and fs.rm with throw inside the callbacks. Because the await on fs.readdir resolves immediately (before the callback fires), any thrown exception escapes the async context as an unhandled rejection, which crashes the Node process. This affected both the scheduled cron job and the admin HTTP endpoint.

This rewrites the method using fs.promises.readdir and fs.promises.rm so all operations are properly awaited, with a single try/catch that surfaces errors as InternalServerErrorException without crashing the server.

Happy path: hit PUT /clearTempFiles as an admin while files exist in api/src/temp/. Verify the files are gone and the response is { success: true }.

Error path: the old code would crash the server - verify the new code doesn't. You can simulate a read error by temporarily renaming src/temp/ to something else before hitting the endpoint, then confirm the API returns a 500 and the server keeps running and handles subsequent requests normally.

Cron path: set TEMP_FILE_CLEAR_CRON_STRING to run every minute in a local env, let it fire, and confirm it completes without crashing.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/app.service.ts
+++ b/api/src/services/app.service.ts
@@ -48,7 +48,7 @@ export class AppService implements OnModuleInit {
     );
     try {
       const files = await fs.promises.readdir(join(process.cwd(), 'src/temp/'));
-      const filesToDelete = files.filter((f) => !f.includes('.git'));
+      const filesToDelete = files.filter((f) => f !== '.gitignore');
       await Promise.all(
         filesToDelete.map((f) =>
           fs.promises.rm(join(process.cwd(), 'src/temp/', f), {
@@ -60,6 +60,7 @@ export class AppService implements OnModuleInit {
         `listing csv clear job completed: ${filesToDelete.length} files were deleted`,
       );
     } catch (e) {
+      this.logger.error('listing csv clear job failed', e);
       throw new InternalServerErrorException(e);
     }
     return {
--- a/api/src/services/app.service.ts
+++ b/api/src/services/app.service.ts
@@ -46,35 +46,22 @@ export class AppService implements OnModuleInit {
     await this.cronJobService.markCronJobAsStarted(
       TEMP_FILE_CLEAR_CRON_JOB_NAME,
     );
-    let filesDeletedCount = 0;
-    await fs.readdir(join(process.cwd(), 'src/temp/'), (err, files) => {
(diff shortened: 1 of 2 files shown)


api/src/services/app.service.ts
```diff
@@ -46,35 +46,22 @@ export class AppService implements OnModuleInit {
...
-                },
-              );
-            } catch (e) {
-              throw new InternalServerErrorException(e);
-            }
-          }
-        }),
+    try {
+      const files = await fs.promises.readdir(join(process.cwd(), 'src/temp/'));
+      const filesToDelete = files.filter((f) => !f.includes('.git'));
```
ludtkemorgan: question: I know this was existing behavior, but do you know why we ignore .git files? I don't believe any git files should be in the temp directory.

api/src/services/app.service.ts
```diff
@@ -46,35 +46,22 @@ export class AppService implements OnModuleInit {
...
+          }),
+        ),
       );
       this.logger.warn(
-        `listing csv clear job completed: ${filesDeletedCount} files were deleted`,
+        `listing csv clear job completed: ${filesToDelete.length} files were deleted`,
       );
-    });
+    } catch (e) {
+      throw new InternalServerErrorException(e);
```
ludtkemorgan, line 64: suggestion: It could be useful to have an additional log here that the job failed before throwing the exception.

ludtkemorgan (approved):
Just one nit, otherwise looks good!

api/src/services/app.service.ts
emilyjablonski: Looks like we have a .gitignore file in there just to keep the folder tracked by git, we could probably filter explicitly on gitignore

emilyjablonski: Updated!

api/src/services/app.service.ts
emilyjablonski: Updated!

PR #6410, opened 2026-06-11
fix: create notification preferences if they do not exist

emilyjablonski:
## Description

Ensure we create notification preferences if they don't yet exist, and ensure the string is translated.

In deployed environments, with only existing users, you were able to add notification preferences because the default preferences did not yet exist and we were not creating them.

Additionally, the error toast needs to be wrapped in the translation string.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/user.service.ts
+++ b/api/src/services/user.service.ts
@@ -924,9 +924,6 @@ export class UserService {
         middleName: dto.middleName,
         lastName: dto.lastName,
         dob: dto.dob,
-        notificationPreferences: {
-          create: {},
-        },
         jurisdictions: jurisdictionsToConnect
           ? {
               connect: jurisdictionsToConnect.map((juris) => ({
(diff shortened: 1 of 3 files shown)


ludtkemorgan (approved):
This change is good and I'm ok merging as is. However, I'd argue we go even one step further and don't create a user's preference until it is explicitly added by the user. That way we have the following advantages:
* Users that don't utilize this feature don't create extra data in the database for a bunch of fields with a value of `false`
* Jurisdictions that don't have this feature on don't have unnecessary data in the database

The way this would look is by making these two changes:
* updating `user.service.ts` to not set `notificationPreferences` on all three user create functions (createPublicUser, createPartnerUser, and createAdvocateUser)`
* remove the NotFoundException in `retrievePreferences` endpoint within the user service.

I'm ok with us doing that as a separate PR if you'd prefer so approving this

emilyjablonski:
@ludtkemorgan Easy change, implemented that!

PR #6412, opened 2026-06-12
fix: notifications translations

emilyjablonski:
## Description

Screenshot 2026-06-11 at 5 47 28 PM
Screenshot 2026-06-11 at 5 47 24 PM

Some notifications db translations are not appearing despite being added in previous migrations.

This PR restores translation keys dropped by migration 59's shallow JSONB merge.

Migration 59 (59_update_rental_opportunity_email_translations) updated rentalOpportunity.accessibilityType.hearingAndVision and introduced rentalOpportunity.footer.unsubscribeAndEmailSettings. It did this with a shallow || merge at the {rentalOpportunity} level, passing accessibilityType and footer as plain JSON objects. Because || is a shallow merge, it replaced those two nested objects entirely rather than patching individual keys, silently dropping:

* rentalOpportunity.footer.accessibleMarketingFlyer
* rentalOpportunity.accessibilityType.hearing, mobility, vision, mobilityAndHearing, mobilityAndVision, mobilityHearingAndVision

This migration restores both dropped key groups for all languages.

It also fixed the same latent bug in the script scripts/generate-db-translation-sql.ts. Note that Migration 59 was not generated by the script, so no previously-generated migrations are affected, but the bug seemed present there as well.

This is more difficult to testing locally because the translation factory after a seed will apply the values.

On main: reseed and verify the issue exist: these should be null:

SELECT language,
  translations->'rentalOpportunity'->'footer'->'accessibleMarketingFlyer' AS flyer,
  translations->'rentalOpportunity'->'accessibilityType'->'mobility' AS mobility
FROM translations
WHERE language NOT IN ('en', 'es') AND jurisdiction_id IS NULL
ORDER BY language;

Then switch to this branch and run migration 61
Re-run the same query - all rows should now be populated

## Author Checklist:

## Review Process:

### DIFF
--- a/api/prisma/migrations/61_restore_accessible_marketing_flyer_translation/migration.sql
+++ b/api/prisma/migrations/61_restore_accessible_marketing_flyer_translation/migration.sql
@@ -3,6 +3,7 @@
 -- "footer" and "accessibilityType", replacing those nested objects entirely
 -- rather than merging into them. This dropped accessibleMarketingFlyer from
 -- footer and all accessibilityType subtypes except hearingAndVision.
+-- Only targets global translations (jurisdiction_id IS NULL).
 
 -- Restore footer.accessibleMarketingFlyer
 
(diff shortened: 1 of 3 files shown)


ludtkemorgan (approved):
👍 

I sure am looking forward to when we don't need to manage these via migrations!

PR #6425, opened 2026-06-16
fix: call full view to get email from in email service

ludtkemorgan:
This PR addresses [bug created with this PR](https://github.com/bloom-housing/bloom/pull/6323)

## Description

With the refactor of jurisdiction views one place was missed and emails are no longer able to be sent. This is because the emailFromAddress is not returned in the default view. The PR changes the view to be the "full" view. 

A better long term solution would be to only select the specific fields that are needed. This can be done by either creating a new view specific to emails or calling the database directly from the email service.

Go through any flow that would trigger an email and it should be sent

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/email.service.ts
+++ b/api/src/services/email.service.ts
@@ -25,6 +25,7 @@ import { Listing } from '../dtos/listings/listing.dto';
 import { IdDTO } from '../dtos/shared/id.dto';
 import { User } from '../dtos/users/user.dto';
 import { FeatureFlagEnum } from '../enums/feature-flags/feature-flags-enum';
+import { JurisdictionViews } from '../enums/jurisdictions/view-enum';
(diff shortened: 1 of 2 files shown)


api/src/services/email.service.ts
```diff
@@ -35,6 +35,7 @@ import {
 } from '../utilities/listing-data-formatters';
 import { unitTypeMapping } from '../../prisma/seed-helpers/unit-type-factory';
 import { UnitAccessibilityPriorityTypeEnum } from '../enums/units/accessibility-priority-type-enum';
+import { JurisdictionViews } from 'src/enums/jurisdictions/view-enum';
```
emilyjablonski: **note:** I think this needs to be relative

mcgarrye (approved):
email received

Issue #6228, opened 2026-04-28
API integration with SMS Service


YazeedLoonat:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Coming from: https://docs.google.com/document/d/1W6Ts3mwiSkb-DBPKJ0dnpf_SPbuq69hMbHUe1VDAHBY/edit?tab=t.0#heading=h.8g27may77w31

we are going to want to tie our AWS SMS Service to our API so our API can send messages through AWS via our API

PR #6427, opened 2026-06-16
feat: add aws sms integration

chadbrokaw:
This PR addresses #6228 

## Description

 - Adds AWS End User Messaging (Pinpoint SMS Voice V2) as an alternative SMS provider alongside the existing Twilio integration
 - Provider is selected at runtime via a `SMS_PROVIDER` env var (`twilio` or `aws`); defaults to `twilio` so existing behavior is unchanged

This is part 1 of the ticket and will need part 2 in order to be fully operable. For now, please inspect the code and ensure it looks good.

## Author Checklist:

## Review Process:

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -38,6 +38,7 @@
     "translations:sql": "ts-node scripts/generate-db-translation-sql.ts"
   },
   "dependencies": {
+    "@aws-sdk/client-pinpoint-sms-voice-v2": "^3.1003.0",
     "@aws-sdk/client-s3": "^3.1003.0",
     "@aws-sdk/client-sesv2": "^3.1003.0",
     "@aws-sdk/lib-storage": "^3.1003.0",
--- a/api/src/services/sms.service.ts
+++ b/api/src/services/sms.service.ts
@@ -1,12 +1,35 @@
 import { Injectable } from '@nestjs/common';
+import { AwsCredentialIdentity } from '@aws-sdk/types';
+import {
+  PinpointSMSVoiceV2Client,
+  SendTextMessageCommand,
+} from '@aws-sdk/client-pinpoint-sms-voice-v2';
 import twilio from 'twilio';
 import TwilioClient from 'twilio/lib/rest/Twilio';
 
 @Injectable()
 export class SmsService {
   client: TwilioClient;
+  awsClient: PinpointSMSVoiceV2Client;
   public constructor() {
-    if (process.env.TWILIO_ACCOUNT_SID && process.env.TWILIO_AUTH_TOKEN) {
+    if (process.env.SMS_PROVIDER === 'aws') {
+      let credentials: undefined | AwsCredentialIdentity;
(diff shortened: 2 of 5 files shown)


api/package.json
```diff
@@ -38,6 +38,7 @@
     "translations:sql": "ts-node scripts/generate-db-translation-sql.ts"
   },
   "dependencies": {
+    "@aws-sdk/client-pinpoint-sms-voice-v2": "^3.1003.0",
```
ludtkemorgan, line 41: This is really confusing. According to [this document](https://docs.aws.amazon.com/pinpoint/latest/userguide/migrate.html) pinpoint will no longer be supported. But it appears like the SDK is still the pinpoint one. So that isn't confusing at all

api/src/services/sms.service.ts
```diff
@@ -1,12 +1,37 @@
...
 import TwilioClient from 'twilio/lib/rest/Twilio';
 
 @Injectable()
 export class SmsService {
   client: TwilioClient;
+  awsClient: PinpointSMSVoiceV2Client;
   public constructor() {
-    if (process.env.TWILIO_ACCOUNT_SID && process.env.TWILIO_AUTH_TOKEN) {
+    if (process.env.SMS_PROVIDER === 'aws') {
+      if (process.env.AWS_SMS_REGION) {
```
ludtkemorgan: suggestion: It would be useful to add a WARN log here if the region is not defined but SMS provider is AWS.

api/src/services/sms.service.ts
```diff
@@ -17,13 +42,27 @@ export class SmsService {
     phoneNumber: string,
     singleUseCode: string,
   ): Promise<void> {
-    if (!this.client) {
-      return;
+    if (process.env.SMS_PROVIDER === 'aws') {
```
ludtkemorgan, line 51: suggestion: I'm cool with the way this is setup for this PR since we only have one use case for sending SMS messages. But when we get even one more use case it would be good to have just one function to call that sends the SMS no matter what SMS provider we have. A good example is how we are doing emails with the email-provider service.

But that can hold off to future PRs if you want to isolate that change

api/src/services/sms.service.ts
```diff
@@ -17,13 +42,27 @@ export class SmsService {
...
+      if (!this.awsClient) {
```
ludtkemorgan, line 52: Same here about logging. We should have a log that warns an SMS message was attempted but no client exists

api/src/services/sms.service.ts
chadbrokaw: Agreed! I think my vote is to hold off on that until a future PR.

api/package.json
chadbrokaw: Agreed... I was confused by this but based off of everything I read it still seems like the correct path. We probably want to keep an eye on this just in case they end up deprecating this package as well.

ludtkemorgan (approved):

PR #6429, opened 2026-06-17
fix: notifications bash issues

emilyjablonski:
## Description

- ✨ This is the only blocking issue, and is just UX feedback ✨ Renames the "Unit type" label in rental opportunity notification emails from "Unit type" to "Available accessible units"
Screenshot 2026-06-16 at 9 08 47 PM

- When enableCustomListingNotifications is on, the account settings dashboard card now shows "Account settings, email, password and notifications" instead of the generic subtitle
Screenshot 2026-06-16 at 9 09 14 PM

- Sign-in auto-redirect for authenticated users for the "unsubscribe and manage email settings" link in notification emails — logged-in users visiting /sign-in?redirectUrl=... are now redirected immediately to the target URL instead of always being shown the login form

- Fixes missing partner admin and public notification emails when a listing is published directly from draft (bypassing the approval workflow) - unlikely to ever happen in practice

-  Receive a rental opportunity notification email with any accessible unit and confirm the label reads "Available accessible units" - publishing directly from draft
-  On the account dashboard, with enableCustomListingNotifications on confirm subtitle changes
-  While logged in, click the "unsubscribe and manage email settings" link from a notification email - confirm redirect to /account/notifications without hitting login screen

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/listing.service.ts
+++ b/api/src/services/listing.service.ts
@@ -1468,6 +1468,7 @@ export class ListingService implements OnModuleInit {
         id: true,
         featureFlags: true,
         listingApprovalPermissions: true,
+        publicUrl: true,
       },
       where: {
         id: dto.jurisdictions.id,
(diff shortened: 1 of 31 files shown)


shared-helpers/src/locales/general.json
```diff
@@ -1,6 +1,7 @@
 {
   "account.accountSettings": "Account settings",
   "account.accountSettingsSubtitle": "Account settings, email and password",
+  "account.accountSettingsSubtitleWithNotifications": "Account settings, email, password and notifications",
```
ludtkemorgan: nit: I'm a fan of the oxford comma (so a comma between "password" and "and"). But I realize that wasn't the standard here already so not a big deal.

api/src/services/listing.service.ts
```diff
@@ -2964,6 +2982,32 @@ export class ListingService implements OnModuleInit {
...
+        select: { publicUrl: true },
+        where: { id: rawJurisdiction.id },
+      });
+      await this.emailService.listingPublished(
+        { id: rawJurisdiction.id },
+        { id: mappedListing.id, name: mappedListing.name },
+        userInfo.emails,
+        jurisdiction.publicUrl || '',
+      );
+    }
```
ludtkemorgan, line 3010: question: I'm guessing the answer is yes, but are we ok with this email now getting triggered for all jurisdictions in the scenario a listing goes from draft to active?

ludtkemorgan (approved):
Change looks good. But we'll need to make sure this is ok for all other jurisdictions. At the moment that would just be doorway and they won't get this change until Fall. They currently get emails for when it is approved, but I don't believe they would get one if an admin publishes a "draft" listing

api/src/services/listing.service.ts
emilyjablonski: Confirming I will chat w Izzie!

shared-helpers/src/locales/general.json
emilyjablonski: Agreed, updated!

PR #6403, opened 2026-06-11
feat: add scheduledApplicationOpenAt for public

KrissDrawing:
This PR addresses #6335 

- [x] Addresses the issue in full
- [ ] Addresses only certain aspects of the issue

## Description

Public site of #6334 (make sure that one is already reviewed).
Shows coming soon section when scheduledApplicationOpenAt date is in future

## How Can This Be Tested/Reviewed?

Check listing with `enableAutopublish` Flag. With `scheduledApplicationOpenAt`:
- missing (We should be able to apply)
- in the past (We should be able to apply, as above)
- in the future (We should see coming soon banner with correct date)

## Author Checklist:

- [ ] Added QA notes to the issue with applicable URLs
- [ ] Reviewed in a desktop view
- [ ] Reviewed in a mobile view
- [ ] Reviewed considering accessibility
- [x] Added tests covering the changes
- [ ] Made corresponding changes to the documentation
- [ ] Ran `yarn generate:client` and/or created a migration when required

## Review Process:

- Read and understand the issue
- Ensure the author has added QA notes
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Either (1) explicitly ask a clarifying question, (2) request changes, or (3) approve the PR, even if there are very small remaining changes, if you don't need to re-review after the updates

Issue #6336, opened 2026-05-20
Scheduled open date - Notification & Cron job


emilyjablonski:
(1) When a listing with a scheduled application open date is published, a new kind of email called "pre-marketing" is sent to folks opted in to the appropriate listing notifications for that listing. It informs potential applicants that a new listing is coming soon, and communicates the date applications will open. If no scheduled open date is set, when a listing is published then the standard new listing should just be sent.

Email content:

Subject: `Coming soon - [Listing name]`
Header: `Coming soon`
Header 2: `[Listing name]`
Content: The same table as the standard notification email, but the table displays `Applications open` instead of `Applications due` with the appropriate open date.

(2) When the scheduled application open date & time is reached, a new cron job must trigger a notification to folks who have opted into the appropriate listing notifications for that listing. This is the standard notification email. We should not make any changes to the listing itself. It should run between 9-11am. We can start just at 9am, and if we run into performance concerns can add staggering later.

KrissDrawing:
Not sure where to test that Cron Job

PR #6430, opened 2026-06-17
feat: add notifications for scheduled application open date

KrissDrawing:
This PR addresses #6336 

## Description

Make sure that #6403 and #6397 are reviewed before (as any changes there might change approach here)
it adds cron job for scheduled application open date (which will send an email, same as when publishing without scheduled application open date).
Also adds another variant of email (coming soon) that will be send when scheduledApplicationDate is in future and we publishing listing (so either when we publish automatically, or scheduledPublishAt cron job)

Sign for notifications with your email.
have scheduled listing with `scheduledApplicationOpenAt` in the future.
1. publish it manually
2. Use cron job for publish
3. use cron job for scheduledApplicationOpenAt (LISTING_OPEN_DATE_NOTIFICATION_CRON_STRING this one)

## Author Checklist:

## Review Process:


api/src/services/listing.service.ts
```diff
@@ -243,6 +243,8 @@ includeViews.full = {
 const LISTING_CRON_JOB_NAME = 'LISTING_CRON_JOB';
 const LISTING_SCHEDULED_PUBLISH_CRON_JOB_NAME =
   'LISTING_SCHEDULED_PUBLISH_CRON_STRING';
+const LISTING_OPEN_DATE_NOTIFICATION_CRON_JOB_NAME =
+  'LISTING_OPEN_DATE_NOTIFICATION_CRON_STRING';
```
ludtkemorgan: nit: This is the "name" of the cron job. It both shows up in the logs as well as in the database. I recommend it is named `LISTING_OPEN_DATE_NOTIFICATION_CRON_JOB` instead of `STRING`. The one above could also change to that format

api/src/services/listing.service.ts
```diff
@@ -3619,6 +3637,72 @@ export class ListingService implements OnModuleInit {
...
+      },
+    });
+
+    // listings published after their open date already received the standard
+    // email from the publish path
+    const toNotify = listings.filter(
+      (listing) =>
+        !listing.publishedAt ||
+        listing.publishedAt < listing.scheduledApplicationOpenAt,
+    );
```
ludtkemorgan: suggestion: could we make this part of the query instead of doing it post the database fetch? Something like this within the `AND`?

```
{
            OR: [
              {
                publishedAt: {
                  is: null,
                },
              },
              {
                publishedAt: {
                  lt: this.prisma.listings.fields.scheduledApplicationOpenAt,
                },
              },
            ],
          },
```

Note: I didn't test the above so it might not be possible

api/src/services/listing.service.ts
```diff
@@ -3619,6 +3637,72 @@ export class ListingService implements OnModuleInit {
...
+    });
+    await this.cronJobService.markCronJobAsStarted(
+      LISTING_OPEN_DATE_NOTIFICATION_CRON_JOB_NAME,
+    );
+
+    const now = new Date();
+    const windowStart =
+      previousRun?.lastRunDate ?? dayjs(now).subtract(1, 'day').toDate();
+
+    const listings = await this.prisma.listings.findMany({
```
ludtkemorgan, line 3709: question: This might be an @emilyjablonski question. There is an edge case that won't happen in any of our current production site. I also don't think it will happen in a future jurisdiction but we might want to protect against it. If we do protect against it it should be done in the logic that actually sends the notifications to users rather than here

A user could create a public account in jurisdiction X and select their preferences. Then a listing in jurisdiction Y goes live and triggers the notification. That listing met the user's preference selections so they receive the notification.

Right now that wouldn't happen and I don't foresee one of our setups to have one  backend with multiple public sites all having different feature flags turned on.

ludtkemorgan (commented):
I'm struggling to get the steps required to trigger this email. Can you lay out the steps you do to get the cron job to trigger it? Such as what settings on the listing, etc

api/src/services/listing.service.ts
emilyjablonski: Hm good question! I think there might actually be farther-reaching consequences to jurisdictions with differing feature flags on one backend serving multiple public sites beyond notifications. Is the thinking to filter by a public user's jurisdiction matching that of the listing?

ludtkemorgan: The recommendation would be to check if the feature flag is on here before sending the notifications out. It would also give us the option to turn the feature off at any point if needed. But I'm also ok with that being a future change because I don't anticipate it happening. At least not anytime soon.

KrissDrawing: Ok, i added a check for `enableCustomListingNotifications` inside `sendListingPublishNotification` function. If it's off it should not send those notifications

KrissDrawing:
@ludtkemorgan So first thing i would do is to seed my email for some users (i will just modify seed-staging).
Then i will go to `http://localhost:3000/account/notifications` and select all.
Then i will set scheduledApplicationOpenAt on partners (for example for Hollywood Hills Heights), and update it for past date in prisma studio.
Then i will close prisma studio (as i feel that it mess somehow with cron jobs, or im just crazy) set cron env for something short and restart server. For some reason it won't run instantly (even though cron string is set for it) it will need like minute or two.

ludtkemorgan:
@KrissDrawing the reason it needs a minute to run is that all cron jobs add a random number of seconds to end of the cron string so that in the deployed environment it doesn't get triggered multiple times (one for each server instance) within the same second. We then check if one has started in the last 5 minutes to decide if it should run. It also has the downside of you can trigger it twice within a 5 minute period.

api/src/services/listing.service.ts
```diff
@@ -3619,6 +3651,74 @@ export class ListingService implements OnModuleInit {
...
+    const previousRun = await this.prisma.cronJob.findFirst({
+      where: { name: LISTING_OPEN_DATE_NOTIFICATION_CRON_JOB_NAME },
+    });
+    await this.cronJobService.markCronJobAsStarted(
+      LISTING_OPEN_DATE_NOTIFICATION_CRON_JOB_NAME,
+    );
+
+    const now = new Date();
+    const windowStart =
+      previousRun?.lastRunDate ?? dayjs(now).subtract(1, 'day').toDate();
```
ludtkemorgan, line 3707: question: Is there a reason we only want to go back 1 day as the default?

api/test/unit/services/listing.service.spec.ts
```diff
@@ -7596,6 +7605,44 @@ describe('Testing listing service', () => {
       process.env.PROXY_URL = undefined;
     });
 
+    it('should send comingSoon subscriber notification when scheduledApplicationOpenAt is in the future', async () => {
+      const futureOpenAt = new Date(Date.now() + 7 * 24 * 60 * 60 * 1000);
```
ludtkemorgan: suggestion: rather than adding a number of milliseconds here you could use dayJS to add hours or days

ludtkemorgan (approved):
Just one question

api/src/services/listing.service.ts
KrissDrawing: The idea is that if for some reason we have dates that are in the past, after migration or sth, and we run it first time then it should just send those emails for today. So technically we don't want to send overdue notifications i think.

emilyjablonski:
@copilot resolve the merge conflicts in this pull request

emilyjablonski:
^ Yikes that did not work at all 😭

Issue #5987, opened 2026-03-03
Dynamic listing cards underneath map


emilyjablonski:

ludtkemorgan:
QA notes:
* This can be tested in [Bridge Bay environment](https://bridgebay.bloom.exygy.dev/listings)
* On the listing map view you should see 3 cards:
  * Email sign up link card
  * View resources card
  * "Looking for housing elsewhere" card (equivalent to the "Looking for housing in San Francisco" card in Doorway)

PR #6431, opened 2026-06-17
fix: map view dynamic cards

ludtkemorgan:
This PR addresses #5987 

## Description
On the listing map search below listing results a set of content cards can now be displayed
image

This is similar to what Doorway currently has 
image

The three resource cards from the home page will appear here if they are turned on for the jurisdiction. Additionally a new custom content card can be added. This card(s) will display if the text string exist in the general overrides file. 

This is only visible in the map view so the easiest way to do that is to use the Bridge Bay jurisdiction:
1. Seed using the "region" configuration: `yarn db:setup:staging-region`
2. Set the jurisdiction env variable for the public site to be "Bridge Bay".
3. Go to the listings page and see the cards at the bottom of the right rail.
4. To test all of the cards you will want to also turn on `enableAdditionalResources` and `enableResources`

## Author Checklist:

## Review Process:

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


api/prisma/seed-staging/seed-bridge-bay.ts
```diff
@@ -1505,12 +1505,12 @@ export const realisticAddressesForActive = [
...
   {
-    street: '1120A Madera Avenue, Menlo Park, CA 94025',
+    street: '1120A Madera Avenue',
     city: 'Menlo Park',
     state: 'CA',
     zipCode: '94025',
-    latitude: 36.961738,
-    longitude: -120.064487,
+    latitude: 37.47197516388515,
+    longitude: -122.15722059781827,
```
ludtkemorgan, line 1514: This was adding a pin outside of the bay area region making the map slightly more zoomed out.

sites/public/src/components/browse/map/MapListingCard.module.scss
```diff
@@ -30,14 +30,17 @@
...
         cursor: pointer;
         @media (--md-down) {
           width: 100%;
+          padding: 0;
         }
       }
 
       .image-background {
         height: 100%;
+        max-height: var(--seeds-s48);
```
ludtkemorgan, line 49: Uniform image heights for each card. Cards with more content will now have the image inset

sites/public/page_content/locale_overrides/general.json
```diff
@@ -14,6 +14,10 @@
   "footer.forGeneralQuestions": "For general program inquiries, you may call us at 000-000-0000.",
   "listings.additionalFeesDisclaimer": "Fees are not charged for reasonable accommodations needed due to disability, such as emotional support animals. Fees are also not charged for aids and services needed for effective communication due to a disability. With respect to credit checks, applicants receiving government rent subsidies have the option to provide \"lawful, verifiable alternative evidence\" of their ability to pay rent, instead of relying on their credit history.",
   "listings.petPolicyDescription": "Service dogs and emotional support animals are not pets and are permitted as required by law. Pets may also be required to be allowed in certain types of housing, such as publicly financed housing developments.",
+  "listingResource.additionalCard1.title": "Looking for housing elsewhere?",
+  "listingResource.additionalCard1.link": "https://www.exygy.com",
+  "listingResource.additionalCard1.linkLabel": "See Listings",
```
emilyjablonski: **nitpick:** Sentence case `See listings`

sites/public/page_content/locale_overrides/general.json
emilyjablonski, line 1: **praise:** ✨

sites/public/src/components/browse/map/ListingsCombined.module.scss
```diff
@@ -115,6 +115,12 @@
       width: 55%;
     }
   }
+
+  h3 {
```
emilyjablonski: **note:** The cards take `headingClass` which would prevent us from having to change this later if we adjust the heading levels structure

sites/public/src/components/browse/map/ListingsCombined.module.scss
```diff
@@ -167,3 +173,13 @@
 .info-window {
   max-width: var(--seeds-s64);
 }
+
+.listings-list-info-card {
+  padding: 0 var(--seeds-s4) var(--seeds-s4) var(--seeds-s4);
```
emilyjablonski: **nitpick (non-blocking):** This is the same as `padding: 0 var(--seeds-s4) var(--seeds-s4);`

sites/public/src/components/browse/map/ListingsCombined.module.scss
```diff
@@ -167,3 +173,13 @@
...
+}
+
+.listings-list-info-cards {
+  background-color: var(--seeds-bg-color-muted);
+  padding-top: var(--seeds-s4);
```
emilyjablonski: **note:** It's only mattered for us so far for left/right in RTL languages, but we're moving toward logical properties, so padding-block-start and border-block-start instead of top/bottom/left/right etc

sites/public/src/components/browse/map/MapListingCard.module.scss
emilyjablonski: Screenshot 2026-06-30 at 8 40 30 PM

Could be misunderstanding, but I don't think this shows a uniform image height

emilyjablonski (commented):
(1) Looks like there's a visual regression on the pin popup
Main:
Screenshot 2026-06-30 at 8 45 13 PM

Branch:
Image

(2) Testing this in the browser, I really prefer the behavior on main - can you remind me what is driving this change?

Main:
Screenshot 2026-06-30 at 8 45 06 PM

Branch:
Screenshot 2026-06-30 at 8 47 06 PM

ludtkemorgan:
@emilyjablonski The change is a recommendation from Hannan. The furthest right version is what was decided. 
listing card layout in map view

It looks like I need to fine tune the results because there are a couple of cases the images are shorter than the max height when addresses are short and there are no tags

sites/public/src/components/browse/map/ListingsCombined.module.scss
ludtkemorgan: Can you explain what your recommendation here would be? Is it that I create a new class name and put it here while also passing that class into the BloomCard in reference?

ludtkemorgan:
@emilyjablonski this is ready to be re-reviewed. Let me know if you have any questions

sites/public/src/components/browse/map/ListingsCombined.module.scss
ludtkemorgan: I did what I think you mean, but let me know if it's incorrect

emilyjablonski (changes requested):
Image

Those updates look good! I'm seeing a regression on the padding on the mobile card though

emilyjablonski:
@ludtkemorgan It looks like it's still the case that not all images in the list are the same height, not sure if that's an issue or acceptable given the updates to design, but this otherwise lgtm! Lmk if you wanna move forward, it's only like a 10% difference in height
Screenshot 2026-07-14 at 9 56 14 AM

ludtkemorgan:
@emilyjablonski I should have updated the PR description. Having every image be the exact same height ran into a few weird cases depending on the content. So instead it was decided that there is a little bit of wiggle room. The images can now be between 176px and 192px. That way it's more likely to align with the bottom of the content but the small content cards and large content cards don't have it aligned

emilyjablonski (approved):
Sounds great!

PR #6433, opened 2026-06-19
feat: include external listings filter

mcgarrye:
This PR addresses [#(6363)](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/6363)

## Description

Adds new listing filter to include external listings when calling the list endpoint.

Create a new listing in the partners portal
Add a `externalListingId` to the listing through the DB
Listing should no longer appear in partners or on public
Export listings and confirm listing does not appear
Using the API, call POST `listings/list` with `includeExternal`: true in the filters
Returned list should contain new listing with externalListingId set

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/listing.service.ts
+++ b/api/src/services/listing.service.ts
@@ -543,6 +543,24 @@ export class ListingService implements OnModuleInit {
       caseSensitive: true,
     });
 
+    const includeExternal = params?.find(
+      (f) => f[ListingFilterKeys.includeExternal] !== undefined,
+    );
+    if (
+      includeExternal === undefined ||
+      includeExternal[ListingFilterKeys.includeExternal] === false
+    ) {
+      filters.push({
+        OR: [
+          {
+            externalListingId: {
+              equals: null,
+            },
+          },
+        ],
+      });
+    }
+
     // detect combined >=/<= monthlyRent filters and add one range filter
     const rentParams =
       params?.filter((f) => f[ListingFilterKeys.monthlyRent] !== undefined) ||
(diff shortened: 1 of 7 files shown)


api/src/services/listing.service.ts
```diff
@@ -543,6 +543,24 @@ export class ListingService implements OnModuleInit {
...
+      filters.push({
+        OR: [
+          {
+            externalListingId: {
+              equals: null,
+            },
+          },
+        ],
+      });
+    }
```
ludtkemorgan, line 558: question: Is there a reason this filter can't be in the params.forEach section below?

api/test/unit/services/listing.service.spec.ts
```diff
@@ -632,7 +632,17 @@ describe('Testing listing service', () => {
...
         where: {
-          AND: [],
+          AND: [
+            {
+              OR: [
+                {
+                  externalListingId: {
+                    equals: null,
+                  },
+                },
```
ludtkemorgan: Now that I see this laid out with the full query. Is the "OR" necessary?

ludtkemorgan (approved):
Works great! Just two questions

api/src/services/listing.service.ts
mcgarrye: Unfortunately it needs to live outside of that section because we need to look for when the param is not passed as well as when it is false

api/test/unit/services/listing.service.spec.ts
mcgarrye: Ya know I was just following the existing pattern but the "OR" is not necessary

Issue #6384, opened 2026-06-02
Bulk CSV Upload - Background jobs


emilyjablonski:
This ticket will introduce `background_jobs`. We need to create a job after the S3 upload, return `jobId` immediately, poll for status, enforce limits on concurrency, and kick off the processing work. The actual processing is in a later ticket. This ticket is for the new table and APIs.

**DB updates**
enum BackgroundJobStatusEnum {
  pending
  validating
  processing
  sendingEmails
  completed
  failed
}

model BackgroundJob {
  id                String @id @default(uuid()) @db.Uuid
  listingId         String @db.Uuid
  requestedByUserId String @db.Uuid
  status            BackgroundJobStatusEnum
  inputS3Key        String
  errorMessage      String?
  errorRow          Int?
  createdAt         DateTime @default(now())
  updatedAt         DateTime @updatedAt
  completedAt       DateTime?
}

**API updates**

BackgroundJobService:
- create({ listingId, inputS3Key, requestedByUserId }) → row status: pending
- getById(jobId) → DTO; 403 if job’s listingId not accessible to caller
- findActiveForListing(listingId) → latest job in pending | validating | processing | sendingEmails, else null
- findActiveJob() -> returns true if there is any job running (i.e. in status other than completed or failed)

AC
- New background_jobs db updates
- POST …/jobs — authorized { listingId, s3Key } → { jobId, status: "pending" } immediately (does not wait for CSV work); kicks off runner in background - does not create if there is a current job for that listing in any status other than completed or failed
- GET …/jobs/:jobId — returns poll DTO (status, errors, counts, initiator); 403 without listing access; 404 if missing
- GET …/jobs?listingId= — returns current in-flight job for listing (with initiator name), or empty when none; ignores completed/failed

emilyjablonski:
Notes from moving forward with one-by-one approach:
* We need less statuses. After more load testing we think validation actually will be quick enough to do synchronously which means we actually don't need `pending` or `validating`. We no longer need longer need the `sendingEmails` step either as it will be happening inline.
* We'll be able to do some better job tracking with this method, so we can add: totalRecords (Int?) and processedCount (Int default of 0)

PR #6448, opened 2026-06-24
feat: bulk CSV background jobs setup

matzduniuk:
This PR addresses #6384 

## Description

* Adds new BackgroundJobs controller, service, and module files.
* Adds new database table for BackgroundJobs records and enums
* Adds new prisma migration script
* Adds integration and e2e tests for the new changes

1. Re-set up the backend application to apply new migrations.
2. Using Postman, perform the following operations:
   2a. Create a new background job using the POST /jobs endpoint, providing a valid listingId (copy one from Prisma Studio → listing table) and any string value for inputS3Key. The response should return a BackgroundJob object with status: "processing".

   2b. Using the same listingId, attempt to create a second job using the POST /jobs endpoint again — this should return a 409 Conflict error, confirming that duplicate active jobs for the same listing are blocked.

   2c. Using the listingId from step 3a, retrieve the currently active job using the GET /jobs?listingId=<ID> endpoint. The response should return the same job object created in 3a.

   2d. Copy the id field from the response in 3a and retrieve the job directly using the GET /jobs/<ID> endpoint. Confirm the returned object matches.

> [!TIP]
> The above steps should also be performed using invalid data — missing or malformed body fields on POST, a non-> UUID listingId, an unrecognised job ID on the GET by ID endpoint, and requests with the passkey or Authorization > > header omitted — to verify proper error and validation handling.

4. Open Prisma Studio, navigate to the background_job table, and manually update the status of the job created in step 3a from processing to completed
5. Using Postman, repeat the GET /jobs?listingId=<ID> request from step 3c — the response should now return null, confirming that completed jobs are not surfaced as active
6. Using Postman, send the POST /jobs request again with the same listingId — this should succeed and return a new 7. BackgroundJob with status: "processing", confirming that a new job can be created once the previous one is no longer active

## Author Checklist:

## Review Process:


api/prisma/schema.prisma
```diff
@@ -2619,3 +2636,9 @@ enum YesNoEnum {
 
   @@map("yes_no_enum")
 }
+
+enum BackgroundJobStatusEnum {
+  processing
+  completed
+  failed
+}
```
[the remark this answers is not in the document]
emilyjablonski: You can ignore this one, it didn't take into account the comment on the issue

api/src/services/background-jobs.service.ts
```diff
@@ -0,0 +1,123 @@
...
+    const hasActiveListing =
+      !!(await this.prismaService.backgroundJob.findFirst({
+        select: {
+          id: true,
+        },
+        where: {
+          listingId: listingId,
+          status: BackgroundJobStatusEnum.processing,
+        },
+      }));
```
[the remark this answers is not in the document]
emilyjablonski: You can ignore this one, it didn't take into account the comment on the issue

api/src/services/background-jobs.service.ts
```diff
@@ -0,0 +1,123 @@
...
+      const jobData = await this.prismaService.backgroundJob.findFirstOrThrow({
+        where: {
+          id: jobId,
+        },
+      });
+
+      return mapTo(BackgroundJob, jobData);
+    } catch {
+      throw new NotFoundException(`Job with id: ${jobId} was not found`);
+    }
```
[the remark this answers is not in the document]
emilyjablonski: Not actually huge prio but we can update anyway

api/src/services/background-jobs.service.ts
```diff
@@ -0,0 +1,123 @@
...
+   * @param listingId - Id of the listing for which the job should be retrieved
+   * @returns Details on the currently processed job for the desired listing
+   */
+  async findActiveForListing(listingId: string): Promise<BackgroundJob | null> {
+    const activeJob = await this.prismaService.backgroundJob.findFirst({
+      where: {
+        listingId: listingId,
+        status: BackgroundJobStatusEnum.processing,
+      },
+    });
```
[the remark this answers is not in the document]
emilyjablonski: We can ignore this one for now!

api/test/unit/services/background-jobs.service.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+      expect(result.listingId).toBe(listingId);
+      expect(result.status).toBe(BackgroundJobStatusEnum.processing);
+      expect(prisma.backgroundJob.create).toHaveBeenCalledWith({
+        data: {
+          listingId,
+          requestedByUserId: userId,
+          inputS3Key: createDto.inputS3Key,
+          status: BackgroundJobStatusEnum.processing,
+        },
+      });
```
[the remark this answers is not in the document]
emilyjablonski: Same can ignore

shared-helpers/src/types/backend-swagger.ts
```diff
@@ -10854,6 +10962,12 @@ export enum MfaType {
   "sms" = "sms",
   "email" = "email",
 }
+
+export enum BackgroundJobStatusEnum {
+  "processing" = "processing",
+  "completed" = "completed",
+  "failed" = "failed",
+}
```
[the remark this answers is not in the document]
emilyjablonski: Same can ignore

shared-helpers/src/types/backend-swagger.ts
```diff
@@ -2537,6 +2537,69 @@ export class AuthService {
...
+  ): Promise<BackgroundJob> {
+    return new Promise((resolve, reject) => {
+      let url = basePath + "/jobs"
+
+      const configs: IRequestConfig = getConfigs("get", "application/json", url, options)
+      configs.params = { listingId: params["listingId"] }
+
+      axios(configs, resolve, reject)
+    })
+  }
```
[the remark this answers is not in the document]
emilyjablonski: Autogenerated, so can ignore, can just make sure latest has been generated

api/src/services/background-jobs.service.ts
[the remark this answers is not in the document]
matzduniuk: I added a permission check for overall 'read' permissions for jobs. Let me know if we also want to limit this by read permission on the listing to which the job is tied to.

api/src/services/background-jobs.service.ts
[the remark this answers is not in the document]
matzduniuk: I think that as we ensure that during creation, only one active job can exist at a time for a listing, this issue is technically negligible, as no more than one item should ever be found either way.

api/prisma/schema.prisma
```diff
@@ -2240,6 +2240,23 @@ model UserRoleSnapshot {
   @@map("user_role_snapshot")
 }
 
+model BackgroundJob {
+  id                String                  @id @default(dbgenerated("uuid_generate_v4()")) @db.Uuid
+  listingId         String                  @map("listing_id") @db.Uuid
+  requestedByUserId String                  @map("requested_by_user_id") @db.Uuid
```
YazeedLoonat, line 2247: should this be foreign keyed to the user account table?

api/src/services/background-jobs.service.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    return activeJob ? mapTo(BackgroundJob, activeJob) : null;
+  }
+
+  /**
+   * Returns true if there is any job running (i.e. in status other than completed or failed)
+   * @returns True if any active job exists (false otherwise)
+   */
+  async findActiveJob(requestingUser: User): Promise<boolean> {
```
ludtkemorgan: question: I know this is laid out in the ticket as a required function in this service. But since there is no usage of it from either a controller or other service what is it's purpose?

api/src/services/background-jobs.service.ts
```diff
@@ -0,0 +1,144 @@
...
+  Injectable,
+  NotFoundException,
+} from '@nestjs/common';
+import { PrismaService } from './prisma.service';
+import { BackgroundJobStatusEnum } from '@prisma/client';
+import { mapTo } from '../utilities/mapTo';
+import { BackgroundJob } from '../dtos/background-jobs/background-job.dto';
+import { S3Service } from './s3.service';
+import { User } from 'src/dtos/users/user.dto';
+import { BackgroundJobCreate } from 'src/dtos/background-jobs/background-job-create.dto';
```
YazeedLoonat: need relative paths here

api/src/controllers/background-jobs.controller.ts
```diff
@@ -0,0 +1,88 @@
...
+  @ApiOperation({
+    summary: 'Get an active background job for a listing',
+    operationId: 'findActiveJobForListing',
+  })
+  @ApiOkResponse({ type: BackgroundJob })
+  public async getListingActiveJob(
+    @Request() req: ExpressRequest,
+    @Query('listingId', new ParseUUIDPipe({ version: '4' })) listingId: string,
+  ): Promise<BackgroundJob | null> {
+    return await this.backgroundJobsService.findActiveForListing(
```
ludtkemorgan: idea: This is built exactly how it is defined in the ticket but to me this job feels like it should return an array of active jobs for that listing. The endpoint below this is where we are looking for a single job and this one is the one where we are looking for any active job connected to the listing. I know in the create endpoint we prevent multiple jobs getting triggered for the same listing, however I feel like this background job service could be useful for other use cases and we might down the road have a scenario where multiple jobs could get triggered.

My recommendation would be to have this endpoint return an array of BackgroundJob. And if none exist it would be an empty array.

api/src/services/background-jobs.service.ts
```diff
@@ -0,0 +1,144 @@
...
+   */
+  async getById(jobId: string, requestingUser: User): Promise<BackgroundJob> {
+    await this.permissionService.canOrThrow(
+      requestingUser,
+      'jobs',
+      permissionActions.read,
+    );
+
+    try {
+      const jobData = await this.prismaService.backgroundJob.findFirstOrThrow({
```
YazeedLoonat: we prefer not to use the prisma OrThrow functions because the errors that they throw are pretty tough to parse

you should switch this to remove the try catch and instead do a findFirst and if no value is found then throw the not found error

YazeedLoonat (dismissed):
looks good but a few tweaks

api/test/integration/background-jobs.e2e-spec.ts
```diff
@@ -0,0 +1,281 @@
...
+    });
+
+    it('should return 401 when passkey header is incorrect', async () => {
+      await request(app.getHttpServer())
+        .post('/jobs')
+        .set({ passkey: 'wrong-key' })
+        .set('Cookie', cookies)
+        .send({ listingId, inputS3Key: 'test-s3-key' })
+        .expect(401);
+    });
```
ludtkemorgan, line 122: suggestion: lets delete these two tests. The passkey is not unique to this controller so no need to test it here. They also fail for me locally because I don't have the `API_PASS_KEY` env variable set. Meaning this test is reliant on having a specific configuration

ludtkemorgan: Same with the other passkey tests below

api/src/services/background-jobs.service.ts
emilyjablonski: This will be used on the frontend to know if there's a job active for another listing

api/src/controllers/background-jobs.controller.ts
emilyjablonski: There is only one allowed to run at a time - if we want to prep for the potential future of having more jobs though could add that functionality in here.

api/src/services/background-jobs.service.ts
ludtkemorgan: Ok, in that case there will need to be a corresponding controller endpoint so that the frontend can call it.

matzduniuk: Added a new endpoint and e2e tests

ludtkemorgan (approved):
Thanks for making those updates!

Issue #5986, opened 2026-03-03
Pull filters in from URL to map search


emilyjablonski:

ludtkemorgan:
QA Notes:
* This can be tested in the Bridge Bay environment https://bridgebay.bloom.exygy.dev/
* On the [listing map view](https://bridgebay.bloom.exygy.dev/listings) select different filters. 
  * Check the URL and make sure each filter has been added after the `?`
  * Verify that filtering is still working as expected

sarahlazarich:
lgtm

PR #6452, opened 2026-06-26
fix: update query when filtering in map view

ludtkemorgan:
This PR addresses #5986 

## Description

When filtering on the browse listing map page all filtered parameters are added to the page URL query params. These are also read on page load to determine the preselected filters.

This gives the following benefits:
* Links to the map search page (such as the home page) can have a filter pre-applied
* Reloading the page doesn't lose your applied filters

I recommend seeding the bridge bay jurisdiction `yarn db:setup:staging-region`

The full map browse experience should be tested. But specifically these flows should be considered
* Opening the filter drawer and applying filters
* Clearing filters
* Change a filter on the URL and make sure it is accurately applied
* Switching to mobile view
* Filtering on mobile view
* Moving the map around

## Author Checklist:

## Review Process:


mcgarrye (approved):
lgtm

Issue #6235, opened 2026-04-28
Doorway feature: data removal disclaimer on edit account page


ludtkemorgan:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
When editing a public account on Doorway there is a disclaimer "You can contact BAHFA to have your account information and associated data deleted at any time. To do so, please reach out to doorway@bayareametro.gov.". This needs to be ported over to core.

<img width="644" height="255" alt="Image" src="https://github.com/user-attachments/assets/64fdcee7-eb45-4f0f-8974-d3bc6fa7ff0c" />

## Technical Context
We can do this functionality without adding a feature flag. Instead we can use the tIfExists functionality to display that section if the string exists

ludtkemorgan:
@ludtkemorgan add generic content

ludtkemorgan:
QA Notes:
* This can be tested in the bloom-doorway environment https://bloom-doorway.bloom.exygy.dev/.
* Create a public account and login
* Go to account settings https://bloom-doorway.bloom.exygy.dev/account/edit and verify that the message appears at the bottom of the form

PR #6455, opened 2026-06-30
fix: add optional disclaimer to account edit pages

ludtkemorgan:
This PR addresses #6235 

## Description

Doorway has a disclaimer at the bottom of the public edit account page. This PR adds the ability to have a disclaimer show on the page if the translation string exists.

image

The following scenarios should be tested:
### No disclaimer:
1. Login as a public user. Go to edit your account http://localhost:3000/account/edit and verify that there is no disclaimer
2. Login as an advocate user. Go to edit your account http://localhost:3000/account/edit and verify that there is no disclaimer
### Disclaimer on 
Delete the `"account.settings.disclaimer": "",` line from the public override file
4. Login as a public user. Go to edit your account http://localhost:3000/account/edit and verify that the disclaimer is at the bottom of the page
5. Login as an advocate user. Go to edit your account http://localhost:3000/account/edit and verify that the disclaimer is at the bottom of the page

## Author Checklist:

## Review Process:


shared-helpers/src/locales/general.json
emilyjablonski, line 1: **praise:** In order for us to be able to remove these in a fork and not mess with alignment in the core translations files, these strings have to live only in the public overrides files

sites/public/src/components/account/EditPublicAccount.tsx
```diff
@@ -122,6 +129,8 @@ export const EditPublicAccount = () => {
     user
   )
 
+  console.log("tIfExists", tIfExists("account.settings.disclaimer"))
```
emilyjablonski: **issue:** 🧹

sites/public/src/components/account/EditAccountView.module.scss
```diff
@@ -16,3 +16,19 @@
...
+    --card-content-padding-inline: var(--seeds-s4);
+    --card-content-padding-block: var(--seeds-s6);
+  }
+  a:not(:global(.seeds-button)) {
+    text-decoration: underline;
+  }
+}
+
+.account-settings-disclaimer {
+  font-size: var(--seeds-font-size-xs);
```
emilyjablonski: **note:** Can this be at least one size bigger since it's presented as body text and not a sub-label to another element? Our xs pushes the boundaries of a11y limits and is usually only used for really intentional hierarchy

emilyjablonski (dismissed):
Left a few comments!

sites/public/src/components/account/EditAccountView.module.scss
ludtkemorgan: For sure! This was ported over directly from Doorway. Do you have documentation on when to use each font-size?

shared-helpers/src/locales/general.json
ludtkemorgan: Oops, it looks like my last commit didn't have the files in the override folder despite me updating that in the PR description 🤦 . I put empty strings in all of the languages and tested it out. It works as expected. The `tIfExists` returns null as expected so nothing gets rendered (That's why I accidentally pushed up that console log to test it) and if removed it renders the default string.

My preference would be to have default text in the main translation files and then strings that we anticipate to have turned off by default to override to empty in the override folder. That way we know that it exists with an example of the text, but it won't impact any other jurisdiction. I know you mentioned earlier there is an issue with this route so let me know if I'm missing something

ludtkemorgan:
@emilyjablonski this is ready for re-review

mcgarrye (approved):
lgtm

Issue #6444, opened 2026-06-24
Remove published at when returning to draft


emilyjablonski:
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBK1JORGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0a10bfb9d7584f99aa552fbcbace947c2d9d2dab/Screenshot%202026-06-22%20at%2011.26.22%E2%80%AFAM.png" alt="Screenshot 2026-06-22 at 11.26.22 AM.png" />

When a listing goes from any status back to draft, we should remove the published at date if it exists. It's particularly awkward now that we the new scheduled open date field, as a listing can show both a scheduled publish date in the future, and then a published at date in the past.

sarahlazarich:
If we remove this, can we still get a notice that it was published & unpublished?
Check the activity log before starting - point assuming it's in there 
EDIT: It's there!

KrissDrawing:
QA Notes: On https://partners.bloom.exygy.dev/ Unpublish published listing, on table now `published date` should be `None`

PR #6457, opened 2026-07-01
fix: clear publishedAt while switching to draft listing

KrissDrawing:
This PR addresses #6444 

## Description

Clears `publishedAt` while listing status changes to `pending`

on partners unpublish listing. It should clear publishedAt.
It should add it on publish (and not modify it while editing published listing) <- so it should work as before

## Author Checklist:

## Review Process:

### DIFF
--- a/api/test/unit/services/listing.service.spec.ts
+++ b/api/test/unit/services/listing.service.spec.ts
@@ -6178,6 +6178,7 @@ describe('Testing listing service', () => {
           section8Acceptance: false,
           unitsAvailable: 0,
           isVerified: false,
+          publishedAt: null,
         },
         where: {
           id: expect.anything(),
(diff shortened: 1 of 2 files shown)


api/test/unit/services/listing.service.spec.ts
```diff
@@ -8016,4 +8019,80 @@ describe('Testing listing service', () => {
...
+    it('active → closed: returns stored value', () => {
+      const result = ListingService['resolvePublishedAt'](
+        ListingsStatusEnum.closed,
+        ListingsStatusEnum.active,
+        storedDate,
+      );
+      expect(result).toBe(storedDate);
+    });
+
+    it('active → changesRequested: returns stored value', () => {
```
emilyjablonski: **issue:** We don't have a pathway in the application for this status change or the following case - but if we did, it should be the case that the date would get removed. To go from active to pendingReview or changesRequested would mean it was unpublished. It's not possible for this to happen though nor should we have a path in the future, so we can safely remove/update these test cases

emilyjablonski (approved):
One small cleanup item, otherwise lgtm!

PR #6462, opened 2026-07-01
feat: implement external listing service

mcgarrye:
This PR addresses [#(6364)](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/6364)

## Description

Implements the externalize and ingest endpoints for the external listing service.

Through the API, call GET `/externalListings`
Should return an object that includes ids and names of jurisdictions, reservedCommunityTypes, unitRentTypes, unitTypes and ids, contentUpdatedAt and jurisdictionIds for all active listings.
Change data in the db or through partners portals and confirm changes are reflected in the data

Using the data returned in the above GET response, call PUT `/externalListings/ingest` with the following payload
`{
    externalURL: 'http://localhost:3100',
    jurisdictionId: '{jurisdictionId}',
    targetName: '{jurisdictionName}',
}`
Should duplicate all active listings in the target jurisdiction with external fields filled out. Verify in the database.
Call again, nothing should happen
For listings on the target jurisdiction, 
- close one active listing, 
- create a new listing, 
- edit an existing listing 
- make sure one listing has not been changed

Run ingest again and validate the following respectively:
- duplicate is deleted
- new duplicate is created
- duplicate is deleted and re-created
- no change
To test negative flows, use junk data in payload

## Author Checklist:

## Review Process:


api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
     private logger = new Logger(ExternalListingService.name),
   ) {}
+
+  async externalize(): Promise<ExternalizedDetails> {
+    const jurisdictions = await this.prisma.jurisdictions.findMany({
+      select: { id: true, name: true },
+    });
+    const listings = await this.prisma.listings.findMany({
+      select: { id: true, contentUpdatedAt: true, jurisdictionId: true },
+      where: { externalListingId: null, status: ListingsStatusEnum.active },
```
ludtkemorgan, line 35: question: I know we don't have the use case now, but do you think we should make it possible to get non-active listings as a passed in parameter?

api/src/dtos/external-listings/ingest-params.dto.ts
```diff
@@ -0,0 +1,24 @@
+import { Expose } from 'class-transformer';
+import { IsString, IsDefined } from 'class-validator';
+import { ApiProperty } from '@nestjs/swagger';
+import { ValidationsGroupsEnum } from '../../enums/shared/validation-groups-enum';
+
+export class IngestParams {
+  @Expose()
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
```
ludtkemorgan, line 10: suggestion: There is a class-validator called `IsUrl` which would help validate that the parameter is valid

api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
+    const {
+      jurisdictions,
+      listings,
+      reservedCommunityTypes,
+      unitRentTypes,
+      unitTypes,
+    } = data;
+    const externalJurisdiction = jurisdictions.find(
+      (juris) => juris.name === targetName,
+    );
```
ludtkemorgan, line 86: suggestion: there is a chance that the "data" response doesn't have 1 or all of those values. My recommendation would be to add null checks to all of these. ex: `jurisdictions?.find(...` and `reservedCommunityTypes?.forEach`

api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
+          (internalRCT) => internalRCT.name === rct.name,
+        );
+        if (matchingRCT) {
+          combinedRCTs.push({ internalId: matchingRCT.id, externalId: rct.id });
+        } else {
+          this.logger.error(
+            `External reserved community type ${rct.name} could not be matched`,
+          );
+        }
+      });
```
ludtkemorgan: nit: this is a good use case for a reduce function
```suggestion
     const combinedRCTs = reservedCommunityTypes.reduce((acc, rct) => {
        const matchingRCT = internalReservedCommunityTypes.find(
          (internalRCT) => internalRCT.name === rct.name,
        );
        if (matchingRCT) {
          acc.push({ internalId: matchingRCT.id, externalId: rct.id });
        } else {
          this.logger.error(
            `External reserved community type ${rct.name} could not be matched`,
          );
        }
        return acc;
      }, []);
```

api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
+
+      // query and compare unit types
+      const internalUnitTypes = await this.prisma.unitTypes.findMany({
+        select: { id: true, name: true },
+      });
+
+      const combinedUTs = [];
+
+      unitTypes.forEach((unitType) => {
```
ludtkemorgan: same here for the reduce function

api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
+    }
+    if (reservedCommunityTypeId === undefined) {
+      this.logger.error(
+        `Listing with external id ${externalListing.id} could not be created. Reserved community type ${data.reservedCommunityTypes.name} does not exist.`,
+      );
+      return;
+    }
+
+    await this.prisma.listings.create({
+      data: {
```
ludtkemorgan, line 288: question: is there a reason you are explicitly doing every field instead of `...data` for all of these 1 to 1 fields? I'm worried that if we add a new listing we forget to add it here

api/src/main.ts
```diff
@@ -112,7 +112,12 @@ async function bootstrap() {
       ];
     });
   });
-  SwaggerModule.setup('api', app, document);
+  SwaggerModule.setup('api', app, document, {
+    swaggerOptions: {
+      tagsSorter: 'alpha',
```
ludtkemorgan, line 117: 🅰️ 🥳

api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
 export class ExternalListingService {
   constructor(
+    private httpService: HttpService,
     private prisma: PrismaService,
-    private permissionService: PermissionService,
     @Inject(Logger)
     private logger = new Logger(ExternalListingService.name),
   ) {}
+
+  async externalize(): Promise<ExternalizedDetails> {
```
ludtkemorgan, line 29: question: I could be mis-remembering this but I thought for the externalize endpoint we were only going to request a specific jurisdiction (by name). It would still need to fetch the jurisdiction, but that way we only need to return the listings that match the jurisdiction we are looking for.

api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
+          listingIdsToDelete.push(existingListing.id);
+        }
+      }
+
+      for (const newExternalListing of listings) {
+        if (
+          newExternalListing &&
+          newExternalListing.jurisdictionId === externalJurisdiction.id
+        ) {
+          this.logger.error('This should be impossible');
```
ludtkemorgan: question: Should this actually be impossible? When I ran it this error was outputted 3 times

api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
+    );
+
+    let reservedCommunityTypeId: string;
+    if (data.reservedCommunityTypes) {
+      reservedCommunityTypeId = combinedRCTs.find(
+        (rct) => rct.externalId === data.reservedCommunityTypes.id,
+      )?.internalId;
+    }
+    if (reservedCommunityTypeId === undefined) {
+      this.logger.error(
```
ludtkemorgan: issue: Every time I run it I run into an error here that crashes the backend. data.reservedCommunityTypes doesn't exist so when it tries to grab `name` in this error message it hits a null pointer.

I'm not sure how that is the case though

ludtkemorgan: Oh! I figured it out. This if statement should be within the previous if statement. It's null because that listing doesn't have a reserved community type

api/src/services/external-listing.service.ts
```diff
@@ -9,9 +20,478 @@ import { PermissionService } from './permission.service';
...
+                floor: unit.floor,
+                maxOccupancy: unit.maxOccupancy,
+                minOccupancy: unit.minOccupancy,
+                monthlyIncomeMin: unit.monthlyIncomeMin,
+                monthlyRent: unit.monthlyRent,
+                monthlyRentAsPercentOfIncome: unit.monthlyRentAsPercentOfIncome,
+                numBedrooms: unit.numBedrooms,
+                numBathrooms: unit.numBathrooms,
+                number: unit.number,
+                sqFeet: unit.sqFeet,
```
ludtkemorgan, line 463: same here about explicitly setting each one for the unit and unitGroups. can you do `...unit`?

api/test/integration/permission-tests/permission-as-no-user.e2e-spec.ts
```diff
@@ -64,24 +67,52 @@ import {
...
   listingPublished: jest.fn(),
 };
 
-const testCronJobService = {
-  startCronJob: jest.fn().mockResolvedValue(undefined),
-  markCronJobAsStarted: jest.fn().mockResolvedValue(undefined),
+const testHttpService = {
+  pipe: jest.fn(),
+  get: jest.fn(() => {
+    const response = {
```
ludtkemorgan, line 90: suggestion(non-blocking): since this is mocked in several files. Could we add it to a helper file. Similar to how we have a mock data for frontend tests in testHelpers.ts. 

No worries if you don't want to tackle that for this PR though

api/test/integration/external-listing.e2e-spec.ts
```diff
@@ -0,0 +1,443 @@
...
+    it('should do nothing if external listing has not been updated', async () => {
+      const jurisdiction = await prisma.jurisdictions.create({
+        data: jurisdictionFactory('doNothingExternalListing'),
+      });
+      await reservedCommunityTypeFactoryAll(jurisdiction.id, prisma);
+      const listingData = await listingFactory(jurisdiction.id, prisma);
+      const listingCreated = await prisma.listings.create({
+        data: {
+          ...listingData,
+          contentUpdatedAt: dayjs(new Date()).add(1, 'days').toDate(),
```
ludtkemorgan: nit(very nit): Since a contentUpdateAt can never be in the future could this value be update to be now and the mocked http response be a date in the past?

ludtkemorgan (changes requested):
Mostly minor, but one issue found which should be an easy fix

api/test/integration/permission-tests/permission-as-no-user.e2e-spec.ts
mcgarrye: I was just following the pattern we already had in the permission tests. I would rather move them all out in a different PR

api/src/services/external-listing.service.ts
mcgarrye: Totally possible, was left in when debugging something

api/src/services/external-listing.service.ts
mcgarrye: This is because we don't want to accept all fields. Foreign ids and fields that require us to create data cannot just be passed in directly, so it is safer for this to be explicit

api/src/services/external-listing.service.ts
mcgarrye: I found a real nifty solution

api/src/services/external-listing.service.ts
mcgarrye: I think it would be possible but I hesitate to take the time to implement it and rework all of the code/tests to do so at this time. I think it could be added to the backlog as a nice to have. But I will leave the final decision to you fearless leader

api/src/services/external-listing.service.ts
mcgarrye: What was implemented follows the implementation laid out in the TDD

ludtkemorgan (approved):
Thanks for the update!

Issue #6240, opened 2026-04-29
Doorway feature: filter by county


ludtkemorgan:
- [x] General
- [ ] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Doorway has the ability to search by "county". This functionality needs to be added to core. The backend API already supports searching by county so this will just be a frontend change to the listing search form, however some backend changes might be required to make counties configurable.

<img width="642" height="249" alt="Image" src="https://github.com/user-attachments/assets/9e264dbe-bb2b-4c55-8761-aeee41700939" />

## Technical Context
* "counties" is a query parameter in the listing list endpoint. This queries by the building address "county" field
  * This is not a very optimized query. It would be better to have these done by jurisdiction since Doorway has each county tied to a jurisdiction. To do that we should add a `jurisdictions` query param.
* The list of counties will need to be configurable
  * One option is to have the counties be listed on the jurisdiction (for Doorway's case it would be the "bay area" jurisdiction)
  * Another option is to have it be an env variable on the public site
* The link to Dahlia should be an additional line of text that is conditionally rendered if the string exists
* The ability to search by county should be a feature flag

## Acceptance Criteria
* Ability to search by county is added to the search form on the public site. There is one checkbox per configured county
* A conditionally rendered string e.g `(For San Francisco listings, go to [DAHLIA](https://housing.sfgov.org/))` can be displayed below the county checkboxes

ludtkemorgan:
@hannan-abdi for this functionality we want to remove the disabled "San Francisco" checkbox and instead just rely on the text below the checkboxes to signify that San Francisco listings are on DAHLIA

ludtkemorgan:
QA Notes:
* This can only be tested in the [new doorway environment](https://bloom-doorway.bloom.exygy.dev/listings)
* The "County" filter should appear and properly filter the listings by the correct county
* The sub text with a link to DAHLIA should appear

sarahlazarich:
yay!

PR #6463, opened 2026-07-02
fix: add county filter on listing search based on subjurisdictions

ludtkemorgan:
This PR addresses #6240 

## Description

Adds the "County" filter to the filter drawer on the public site. This is dynamically built based on the "sub jurisdictions" of the site's jurisdiction

There is also an optional sub-note that can be displayed. In order to view it you need to add a string to the translation file
```
 "listingFilters.countyFilterNote": "(For San Francisco listings, go to <a className='lined' href='https://housing.sfgov.org/' target='_blank'>DAHLIA</a>)",
```

Open questions:
1. ~Should the filter be labeled "County" or "Counties"?~ - A: confirmed "County" is the correct label
2. ~What should be the flow if they unselect all counties? Right now it filters by all counties and doesn't give any indication that is happening~ - A: switch to having none selected by default to follow pattern of other multi-select filters

This only works with the multi-jurisdiction setup. So it is recommended you seed with the "region" command `yarn db:setup:staging-region` and set the site's jurisdiction to "Bridge Bay".

The following flows should work:
* Opening the filter drawer pre-selects all counties
* Changing the selected counties appropriately filters the listings
* Adding the "listingFilters.countyFilterNote" shows the sub note

## Author Checklist:

## Review Process:


api/src/dtos/listings/listings-filter-params.dto.ts
```diff
@@ -86,14 +86,6 @@ export class ListingFilterParams extends BaseFilter {
   @IsString({ groups: [ValidationsGroupsEnum.default] })
   [ListingFilterKeys.city]?: string;
 
-  @Expose()
```
ludtkemorgan, line 89: Removing the ability to filter by "county" on the backend as we will be filtering by "jurisdictions". This functionality can be added back if we ever get a different setup where each county isn't it's own "jurisdiction"

mcgarrye:
To respond to the two open questions, I think we should have none of the counties selected on filter open and for the label to be `County`. This follows the same pattern we have for the other filters which are not selected initially (availability, bedroom size and bathrooms) [Though I do think it should be `Availability type`, `Number of bedrooms`, `Number of bathrooms`]. This works because we are already sending all subjurisdictions back with the `jurisdiction` filter and when we filter on the subjurisdictions we send them back with the `jurisdiction**s**` filter and get the desired result

mcgarrye (approved):
LGTM!

Issue #6458, opened 2026-07-01
Add 6 and 7+ bedroom types


emilyjablonski:
To support requirements for LA's land-use listings, we need to add two new bedroom sizes: 6 and 7+

We'll just add 6 and 7 into the enum and have the string display 7+

We need to ensure they appear where expected across both sites, like the public filter and rent tables, and partners dropdown

sarahlazarich:
Cool in core to add 6 and 7+
Doorway will inherit this as a part of corification

ludtkemorgan:
@KrissDrawing I checked the database of core and there isn't the 6 and 7 bedroom types in the unit_type table. Relooking at your PR it looks like your migration added the enum but didn't add the unit_type entry.

I realized that the new unit types aren't originally added as a migration so we have a couple of options:
1. We could manually add these new ones to core and write a migration for just LA
2. Add a migration to core that will be picked up everywhere. But this means any new jurisdiction spun up will get 6 and 7 by default and not any of the smaller bedroom types.

Unfortunately Sarah is out this week so we might have to make a decision ourselves. My recommendation is we go with option 1. If that sounds good I can add them in core if you want to create an LA specific migration in the bloom-la repo?

KrissDrawing:
@ludtkemorgan Oh thanks! I added those to seeded data and then forgot about database.
But isn't that Sarah comment above align more with 2nd option? As technically on public we will have anyway 6,7 bedrooms for example in filters.
Aside of that not super sure what do you mean by this: `But this means any new jurisdiction spun up will get 6 and 7 by default and not any of the smaller bedroom types.`. Are those added manually? Can we maybe use some `where` condition to handle that, for example add those only if `fiveBdrm` exist?

ludtkemorgan:
The unit types have been added manually in the past, usually by seeding the database.

Ok, new plan. How about we create a migration in core that adds all of the unit types but put where clauses on all of them to make sure we aren't inserting values that already exist. That way we solve both problems. It will add the 2 values everywhere and also handle new jurisdictions

example: 
```
INSERT INTO unit_types (num_bedrooms, name)
SELECT 2, 'twoBdrm'
WHERE NOT EXISTS (
    SELECT 1 FROM unit_types WHERE name = 'twoBdrm'
);

INSERT INTO unit_types (num_bedrooms, name)
SELECT 6, 'sixBdrm'
WHERE NOT EXISTS (
[...]

PR #6493, opened 2026-07-09
fix: add six and seven bedrooms

KrissDrawing:
This PR addresses #6458 

## Description

Adds six and seven + bedrooms options.
Using google translate for this seemed wrong (as it returned inconsistent translations), so i used common sense and some AI so i cant guarantee if they are correct.

run migration.
Units should have now 2 new options.
they will be visible on public listing site.

## Author Checklist:

## Review Process:


api/src/utilities/application-export-helpers.ts
```diff
@@ -1091,6 +1091,8 @@ export const typeMap = {
   threeBdrm: 'Three Bedroom',
   fourBdrm: 'Four+ Bedroom',
```
ludtkemorgan: question: I know this is already existing, but shouldn't this be changed to `Four Bedroom`?

ludtkemorgan (approved):

api/prisma/migrations/66_add_six_and_seven_bedroom_unit_types/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterEnum
```
ludtkemorgan, line 1: This file name needs to be updated to the sequential number system we are using before merging

Issue #6377, opened 2026-06-01
Bulk CSV Upload - Private bucket presigned upload, upload API


emilyjablonski:
We need to update our s3 service to allow partner users to directly upload CSVs to the private bucket, and for the API to stream that data. 

We can use the convention `bulk-application-updates-{listingId}-{userId}-{time-generated}.csv` for filepaths

API
* We will need a new POST endpoint for uploading a URL at `/applications/bulk-update/upload-url` with body `{ uploadUrl: string, s3Key: string }`
* `s3Key` matches `bulk-application-updates-{listingId}-{userId}-{time-generated}.csv` and is the key used in the presigned URL
* Missing/invalid `listingId` → 400
* Unknown listing / no listing access → 403 or 404 per existing permission patterns
* Listing jurisdiction must have `enableApplicationStatus` set to true

S3 Presign
- Presigned URL targets private bucket (not public assets bucket)
- PUT with header `Content-Type: text/csv` and a small CSV body returns 200 from S3
- For testing, check for file's existence in private bucket - will require AWS access so may need to be completed by or paired w an Exygy dev

PR #6494, opened 2026-07-09
Feat: S3 private bucket pre-signed URL generation integration

matzduniuk:
This PR addresses #6377

## Description

* Sets up a new DTO object for the presigned URL generation endpoint
* Sets up the new application controller `'bulk-update/upload-url'` endpoint 
* Sets up the base service method for validating the new DTO data and creating a template s3Key

## Author Checklist:

## Review Process:


api/src/controllers/application.controller.ts
```diff
@@ -327,6 +331,18 @@ export class ApplicationController {
     );
   }
 
+  @Post('bulk-update/upload-url')
+  @ApiOperation({
+    summary: 'Generates an presigned URL link to a private S3 bucket',
+    operationId: 'uploadBulkUpdate',
+  })
+  @ApiOkResponse({ type: Boolean })
```
YazeedLoonat: nit: this should be the same type as the promised response

api/src/services/application-bulk-upload.service.ts
```diff
@@ -295,6 +304,70 @@ export class ApplicationBulkUploadService {
     return stringData;
   }
 
+  // TODO Yazeed - Integrate the uploadUrl method with S3 presigned URL generation logic
+  async uploadUrl(
+    dto: ApplicationBulkUrl,
+  ): Promise<ApplicationBulkPresignedUrl> {
```
YazeedLoonat, line 308: this should be a DTO returned

api/src/services/application-bulk-upload.service.ts
```diff
@@ -295,6 +304,70 @@ export class ApplicationBulkUploadService {
...
+    const { userId, listingId, uploadUrl } = dto;
+
+    const listingData = await this.prisma.listings.findUnique({
+      select: {
+        jurisdictionId: true,
```
YazeedLoonat, line 314: we should specify the listing id is responded here so we aren't returning the full listing object

api/src/dtos/applications/application-bulk-url.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+  @IsDefined({ groups: [ValidationsGroupsEnum.applicants] })
+  @ApiProperty()
+  userId: string;
+
+  @Expose()
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @MinLength(1, { groups: [ValidationsGroupsEnum.default] })
+  @IsDefined({ groups: [ValidationsGroupsEnum.applicants] })
+  @ApiProperty()
+  uploadUrl: string;
```
YazeedLoonat: I dont think we need to include the uploadUrl on the incoming DTO

api/src/services/application-bulk-upload.service.ts
```diff
@@ -295,6 +304,70 @@ export class ApplicationBulkUploadService {
...
+        mapTo(Jurisdiction, listingData.jurisdictions),
+        FeatureFlagEnum.enableApplicationStatus,
+      )
+    ) {
+      throw new ForbiddenException(
+        `Jurisdiction with id: ${listingData.jurisdictionId} does not have the enableApplicationStatus flag enabled`,
+      );
+    }
+
+    const s3KeyTemplate = `bulk-application-updates-${listingId}-${userId}-${formatLocalDate(
```
YazeedLoonat: I dont think we should use the local date here, we should use the server time here instead so just new Date formatted

api/src/services/application-bulk-upload.service.ts
```diff
@@ -295,6 +304,70 @@ export class ApplicationBulkUploadService {
...
+      new Date(),
+      'YYYY-MM-DD-hh:mm:ss',
+    )}.csv`;
+
+    return {
+      key: s3KeyTemplate,
+      presignedUrl: '', // TODO: Yazeed - Add S3 generated presigned URL
```
YazeedLoonat: you should be able to use the s3.service.ts `uploadURLForPublic()` as a template to create a new function that aims at the private bucket instead of the public one and then use that here to get the upload url 

you should be able to test this too as long as you have the private s3 bucket creds that we use in our hosted core env locally if you dont you can reach out to me and I can get those to ya

YazeedLoonat (commented):
looks good overall left you some comments and the next steps for ya
lemme know if you have any questions or ran into anything

api/src/services/application-bulk-upload.service.ts
matzduniuk: We are already limiting the listing response object to only have the `jurisdictionId` and linked jurisdiction feature flags array

api/src/controllers/application.controller.ts
```diff
@@ -327,6 +329,18 @@ export class ApplicationController {
...
+  @Post('bulk-update/upload-url')
+  @ApiOperation({
+    summary: 'Generates an presigned URL link to a private S3 bucket',
+    operationId: 'uploadBulkUpdate',
+  })
+  @ApiOkResponse({ type: ApplicationBulkPresignedUrl })
+  async uploadBulkUpdate(
+    @Body() dto: ApplicationBulkUrl,
+  ): Promise<ApplicationBulkPresignedUrl> {
+    return this.applicationBulkUploadService.uploadUrl(dto);
```
YazeedLoonat: by standard we await these responses

api/src/services/application-bulk-upload.service.ts
YazeedLoonat: sorry to specify here, this is getting the full listing object out of the db which is then held in memory until it goes through the mapping to get the DTO out 

we can save mem but only specifying the fields we actualyl need on this select

YazeedLoonat (commented):
looks good, just a few tiny nitpicks

YazeedLoonat (approved):
this is still a WIP PR but looks good to me

Issue #6378, opened 2026-06-01
Bulk CSV Upload - Validate CSV


emilyjablonski:
After a partner uploads a CSV for bulk application editing, when the job is kicked off, we need to validate the CSV synchronously and return an error if there are any issues. This ticket is just for the backend.

For now, updatable fields are only:
- status
- applicationDeclineReason
- applicationDeclineReasonAdditionalDetails
- accessibleUnitWaitlistNumber
- conventionalUnitWaitlistNumber
- manualLotteryPositionNumber

This ticket is to implement validation for a new `POST /applications/bulk-update` endpoint. Validation runs inside the request before any job is created. If validation fails, return a 400 with the error message and do not create a job. If validation passes, the job is created in `processing` and async processing would then begin (handled in a separate ticket).

The endpoint should take in `{ s3Key: string }`

Request flow

1. Check if an active job exists globally → 409 if so (before any S3 work or validating)
2. Stream CSV from S3
3. Run all validation checks below
4. On any failure → 400 with error message, no job created
5. On success → create job in processing with listingId extracted from the s3Key, requestedByUserId, inputS3Key, totalRecords, and then return { jobId }

Validation checks are all or nothing, meaning the first failing check should short circuit the flow. Row numbers in error messages should be 1-indexed to account for the header row (i.e. first data row is row 2).

| Failure condition | HTTP response |
|---|---|
| File is not a CSV | 400 "Upload Failed: file must be a CSV format" |
| Headers don't exactly match expected columns | 400 "Upload Failed: CSV has additional or missing columns" |
| File has headers but zero data rows | 400 "Upload Failed: CSV contains no application records" |
| Duplicate application IDs in CSV | 400 "Upload Failed: One or more rows beginning on row X contain duplicate application IDs" |
| Application ID not found or belongs to a different listing | 400 "Upload Failed: One or more rows beginning on row X have incorrect application identification numbers" |
| Context fields (First Name, Last Name, Submission Date) don't match DB | 400 Same as above (hard failure — bad ID match signals a stale export) |
| Status string not a valid option (Human readable strings) | 400 "Upload Failed: Could not match one or more application status inputs beginning on row X with accepted system options" |
| Decline reason is not a valid option (Human readable strings) | 400 "Upload Failed: Could not match one or more application decline reason inputs beginning on row X with accepted system options" |
| Status is "Declined" but no decline reason provided | 400 "Upload Failed: One or more rows beginning on row X have a declined status without a decline reason" |
[...]

PR #6524, opened 2026-07-16
6378/validate csv

matzduniuk:
This PR addresses #6378 

## Description

* Adds a new endpoint to the application controller.
* Adds a comprehensive validation method for the incoming bulk application CSV file to the bulk upload service.
* Add integration tests for CSV validation methods
* Adds a new S3 service method for retrieving a CSV file from a private S3 bucket

> [!NOTE]
> At the moment, testing of this branch is reduced to verifying the coverage of the included integration tests or by preparing a mock CSV file directly in the S3 bucket

## Author Checklist:

## Review Process:


api/src/services/application-bulk-upload.service.ts
```diff
@@ -30,42 +91,40 @@ export class ApplicationBulkUploadService {
...
-      case ApplicationStatusEnum.submitted:
-        return 'Submitted';
-      case ApplicationStatusEnum.waitlist:
-        return 'Wait list';
-      case ApplicationStatusEnum.waitlistDeclined:
-        return 'Wait list - Declined';
-      default:
-        return statusEnum;
-    }
+  private convertApplicationStatusToReadable(
```
YazeedLoonat: suggestion: make these public functions, they are much easier to unit test that way

api/src/services/application-bulk-upload.service.ts
```diff
@@ -30,42 +91,40 @@ export class ApplicationBulkUploadService {
...
+    statusEnum: ApplicationStatusEnum,
+  ): string {
+    return APPLICATION_STATUS_MAP[statusEnum] ?? statusEnum;
   }
 
+  private convertReadableToApplicationStatus = (
```
YazeedLoonat: suggestion: same as above

api/src/services/application-bulk-upload.service.ts
```diff
@@ -317,4 +376,296 @@ export class ApplicationBulkUploadService {
       },
     );
   }
+
+  private validateFileFormat(s3Key: string): void {
```
YazeedLoonat: suggestion: same as above

api/src/services/application-bulk-upload.service.ts
```diff
@@ -317,4 +376,296 @@ export class ApplicationBulkUploadService {
...
+    if (!s3Key.toLowerCase().endsWith('.csv')) {
+      throw new BadRequestException('Upload Failed: file must be a CSV format');
+    }
+  }
+
+  private validateHeaders(actualHeaders: string[]): void {
```
YazeedLoonat: same as above

api/src/services/application-bulk-upload.service.ts
```diff
@@ -317,4 +376,296 @@ export class ApplicationBulkUploadService {
...
+   * Kept separate from the per-row checks so the DB is only hit once,
+   * regardless of the number of rows.
+   */
+  private async fetchDbApplications(
+    rows: CsvRow[],
+    listingId: string,
+  ): Promise<ApplicationContextFields[]> {
+    const ids = rows.map((r) => r[bulkUploadHeaderNames.applicationId]);
+
+    return this.prisma.applications.findMany({
```
YazeedLoonat, line 487: issue: as the number of ids climbs this is going to break down, I think you're going to need to adjust this so that instead of trying to grab every application all at once you validate chunks of the set at a time

I think you can mirror what we do in the application export builder in terms of how we chunk it out

api/src/services/application-bulk-upload.service.ts
```diff
@@ -317,4 +376,296 @@ export class ApplicationBulkUploadService {
...
+    this.validateHeaders(headers);
+    this.validateHasDataRows(rows);
+
+    const dbApps = await this.fetchDbApplications(rows, dto.listingId);
+    const foundIds = new Set(dbApps.map((a) => a.id));
+    const dbMap = new Map(dbApps.map((a) => [a.id, a]));
+
+    this.validateNoDuplicateId(rows);
+
+    for (let i = 0; i < rows.length; i++) {
```
YazeedLoonat: suggestion: if I understand correctly these are all static analysis of the incoming CSV (like it doesn't hit our DB at all)

if thats true lets do this analysis first. Its much less time intensive and allows us to kick requests out faster

YazeedLoonat (changes requested):
looks good, some requested changes and suggestions

api/src/services/application-bulk-upload.service.ts
matzduniuk: Not exactly. Some tests are pure semantics, like proper strings for decline reasons, valid integer cell values, etc but we also do comparative validation against existing records like:
1. Verifying if the application with the given ID exists in the records 
2. If step one is true, we compare if the applicant's data (First and Last Name) as well as the submission date have not changed (if any of those differ, the row is invalid).

YazeedLoonat (approved):
`This is still a WIP so not sure if it'll fully approve but this looks good to me

very minor nitpick but by convention we don't really use private functions in the service files to make unit testing a bit easier please make the privated functions in `/application-bulk-upload.service.ts` public and then I think this is good to go

emilyjablonski:
@matzduniuk Is this one ready to move forward?

ludtkemorgan:
@matzduniuk as discussed in standup, feel free to merge this PR

PR #6530, opened 2026-07-20
fix: msq visible to draft check

mcgarrye:

## Description

A bug was found in Doorway where one could assign a visible MSQ to a draft listing and then change the status of the MSQ back to `draft` causing problems for the listing. A check has been added to prevent returning a `visible` MSQ to `draft` once it is associated with a listing.

Open question: do we want a better error message for this than the generic one on the FE?

Run `yarn setup --msqV2`
Login to the partners site
Navigate to the Settings page
Create a Preference and `Show to partners`
Choose or create a draft listing and attach the visible MSQ
Return the the Settings page and attempt to return the MSQ to draft
Verify error occurs

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/services/multiselect-question.service.ts
+++ b/api/src/services/multiselect-question.service.ts
@@ -208,12 +208,32 @@ export class MultiselectQuestionService {
   }
 
   /*
-    this will return 1 multiselect question or error
+    this will return 1 mapped multiselect question or error
   */
   async findOne(
     multiselectQuestionId: string,
     view: MultiselectQuestionViews = MultiselectQuestionViews.fundamentals,
   ): Promise<MultiselectQuestion> {
+    const rawMultiselectQuestion = await this.findOrThrow(
+      multiselectQuestionId,
+      view,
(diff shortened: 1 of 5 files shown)


api/src/services/multiselect-question.service.ts
```diff
@@ -576,6 +595,13 @@ export class MultiselectQuestionService {
           throw new BadRequestException(
             "status 'visible' can only change to 'draft' or 'active'",
           );
+        } else if (
+          nextState === MultiselectQuestionsStatusEnum.draft &&
+          listings.length > 0
+        ) {
+          throw new BadRequestException(
```
ludtkemorgan, line 602: suggestion: I know this is an open question in your PR. I think we should display this error message to the user if possible so we don't get paged when they try to do this

ludtkemorgan (approved):
Works as expected. I just recommend we surface the error to the user

ludtkemorgan (approved):
👍 👍

Issue #6519, opened 2026-07-15
Seed / data migration command


seanmalbert:
## #6519 — Seed / data migration command

- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Existing static translation and content files (and any existing `translations` blob rows) must be
moved into the new key-level tables once per environment, idempotently, including converting
markdown bodies to HTML.

## Technical Context
- TDD: §7.1.
- A `script-runner` command that: explodes existing `translations` blobs into per-key rows; loads
  base and per-site overrides into key rows (Partners overrides as global `null`-jurisdiction rows)
  with `sourceHash` baselines; extracts structured content into `JurisdictionContent`, converting
  markdown bodies to sanitized HTML. Idempotent, with a dry-run flag.
- Write `updated_at` at millisecond precision (use `prisma.create`/`createMany`, not raw SQL with
  `now()`/`CURRENT_TIMESTAMP`). Prisma reads `Timestamp(6)` back at millisecond precision, so a row
  stored with sub-millisecond microseconds can never match the value an editor client echoes back,
  and #6510's per-key optimistic lock would report a permanent false conflict on the first edit of
  that key.
- **Seed one fork-specific key.** The staging seed should create a single override row whose key has
  no value in the base locale files, so the "this will hide a section" path in the editor (#6516)
  can be exercised without hand-writing a row. Reviewing #6516, the only way to reach that dialog
  was to add a key through the API. `listings.petPolicyDescription` is the kind of key this covers:
  present in some forks, absent from the base.
- Depends on: #6508 (data model); #6513 (to sanitize the converted HTML).

## Acceptance Criteria
- Re-running upserts rather than duplicates; a dry-run reports changes without writing.
- Base and public/Partners overrides land in the correct key rows; non-English rows carry a
  `sourceHash` of their English source.
- A seeded jurisdiction has at least one override whose key has no base value, and reverting or
  emptying it in the editor produces the hide-section confirmation.
- Markdown content bodies are converted to sanitized HTML and open cleanly in the editor; the
  dry-run report lists each converted body.

## QA Notes
Backend. Run the dry-run and confirm the reported changes; run for real and confirm idempotency on a
second run. Spot-check converted markdown bodies render correctly and open in Tiptap.

Issue #6514, opened 2026-07-15
Public site hydration and ISR


seanmalbert:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
The public site must render each jurisdiction's DB-driven translations and structured content, keep
its static/SEO serving profile, and reflect admin edits promptly. Overridable pages move to ISR
rather than SSR.

## Technical Context
- TDD: §5.2, §5.3, §5.6, §8.
- Fetch per-site overrides into the polyglot at generation; keep overridable pages on
  `getStaticProps` with a short `revalidate` (ISR). Add an on-demand revalidation API route
  (`res.revalidate`) triggered on admin save, paired with a CDN path purge (or short `s-maxage`).
- Wire structured-content rendering (replace the null-returning `jurisdiction_*.tsx` stubs).
- Confirm `tIfExists` compatibility: it resolves against the same polyglot, and the three-state
  semantics hold (no override / value / empty-to-hide).
- Depends on: #6509 (read/merge service); #6511 (content service).

## Acceptance Criteria
- Overridable public pages render as ISR with content read at generation; SSR HTML carries
  translated strings with no flash of raw keys.
- An admin save triggers `res.revalidate()` for the affected routes plus a CDN purge (or short
  `s-maxage`), and the change appears within seconds.
- Footer/FAQ/resources render from DB content matching the current static output (parity).
- `tIfExists` hides a section for an absent key or an empty override value; a base section can be
  hidden with an empty override and shown by reverting.

## QA Notes
Public site. Verify pages are statically served (ISR), an admin content edit appears within seconds
after revalidation + CDN purge, and a `tIfExists`-driven section shows/hides correctly for
value/empty/absent. Include a rendering-parity check against current static content.

Issue #6508, opened 2026-07-15
Key-level translation data model and migrations


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Phase 1 moves per-jurisdiction translations out of build-time files and into the database at
key-level granularity, so each string can carry the metadata later work needs (a staleness source
hash, machine/human provenance) and support per-key editing, locking, and history. This ticket lays
the foundation: the `TranslationStrings` table, its enums, and the `JurisdictionContent` table, plus
migrations.

## Technical Context
- TDD: §3.1, §3.2, §3.3.
- `TranslationStrings`: one row per `(jurisdictionId, language, site, key)` with `value`,
  `sourceHash`, `origin`. Add `SiteEnum` (`public`, `partners`) and `TranslationOrigin`
  (`machine`, `human`).
- The unique index must be `NULLS NOT DISTINCT` (Postgres 15+; Bloom runs 18) so base rows (`NULL`
  jurisdiction and `site`) and global Partners rows (`NULL` jurisdiction) cannot duplicate. Prisma
  `@@unique` documents the tuple; the `NULLS NOT DISTINCT` clause goes in the migration SQL. Add a
  lookup index on `(jurisdiction_id, language, site)`.
- `JurisdictionContent`: typed JSON columns, unique on `(jurisdiction_id, language)`, FK to
  `jurisdictions` with `ON DELETE CASCADE`.

## Acceptance Criteria
- `TranslationStrings` exists with the `NULLS NOT DISTINCT` unique index and the lookup index.
- Inserting a duplicate `(NULL, en, NULL, key)` or `(NULL, en, partners, key)` row is rejected.
- `JurisdictionContent` exists per §3.2.
- Migrations apply cleanly on a copy of production and are additive (nothing reads the new tables yet).

## QA Notes
Backend-only. Verify by running the migrations against a production copy and via DB-level tests:
confirm the duplicate-NULL insert is rejected (the `NULLS NOT DISTINCT` behavior) and that app
behavior is unchanged since nothing reads the new tables yet.

PR #6536, opened 2026-07-21
fix: translation read/merge service and public read endpoints

### LINKED ISSUE #6509: Translation read/merge service and public read endpoints
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
The public and Partners sites need to read a jurisdiction's per-site translation overrides from the
API, and the email service must keep working unchanged. This ticket reimplements the merge over the
new key rows and adds the public read endpoints.

## Technical Context
- TDD: §4.1, D2.
- Reimplement `getMergedTranslations` to aggregate key rows in the existing four-level precedence
  (generic-default, generic-language, jurisdictional-default, jurisdictional-language). The generic
  (null-jurisdiction) lookups filter `site = null` so the global Partners rows do not leak into the
  base/email merge.
- Add `getJurisdictionOverrides(jurisdictionId, language, site)` and the endpoints:
  `GET /jurisdictions/:id/translations?site=&language=` and the non-jurisdiction-scoped
  `GET /translations?site=partners&language=`.
- Depends on: #6508 

## Acceptance Criteria
- `getMergedTranslations(jurisdictionId, language)` returns the same merged output as today, now
  assembled from key rows; equivalent to the current output on a seeded dataset.
- The per-site read returns only that scope's keys, aggregated to a flat object; no partner-only
  keys appear in a public response, and vice versa.
- `GET /translations?site=partners&language=` returns the global Partners keys.
- Cache-Control headers per §4.1.

## QA Notes
Backend-only. Cover with unit/integration tests: the four-level fallback matrix, per-site scope
isolation (public response has no partner-only keys), and the email path producing the same result
as before on seeded data.

seanmalbert:
This PR addresses #6509

## Description

Reimplements the translation merge over the key-level `translation_strings` table (added in #6508) and adds the public read endpoints the public and Partners sites will use for per-site overrides. Backend only.

- **`getMergedTranslations`** keeps its signature and nested output, now assembled from the key rows instead of the whole-blob `translations` rows. It reads the four scopes (generic-default, generic-language, jurisdictional-default, jurisdictional-language), all constrained to `site IS NULL` so public and Partners overrides (including the global-Partners `null` jurisdiction rows) do not enter the base/email merge, then reconstructs the nested object from the flat dot-path keys and merges in the existing four-level precedence.
- **Legacy fallback.** When a scope has no key rows, it falls back to the legacy `translations` blob for that scope. The email path is not feature-flag gated, so this keeps it working on any environment before the data migration (#6519) backfills the key rows. The fallback and the legacy read are removed when the `translations` table is retired.
- **`getJurisdictionOverrides(jurisdictionId, language, site)`** returns a flat key/value object of one site's jurisdictional overrides, two-level (the requested language over the jurisdiction's English default). Callers layer it over their statically-bundled base. A `byName` convenience resolves the jurisdiction id first.
- **Endpoints** (`TranslationController`, `TranslationModule`):
  - `GET /jurisdictions/:jurisdictionId/translations?site=&language=`
  - `GET /jurisdictions/byName/:jurisdictionName/translations?site=&language=`
  - `GET /translations?language=` for the global Partners layer (jurisdiction-independent).
  - Reads allow anonymous access (`ApiKeyGuard` + `OptionalAuthGuard`) and set `Cache-Control: public, s-maxage=300, stale-while-revalidate=600`.

The change is additive: the email caller's output is unchanged, and the legacy `translations` table is untouched.

The frontend hydration (#6514/#6515) and the client regeneration are out of scope for this PR, so the generated client types are not updated here.

1. Unit tests:
   ```
   cd api && yarn test
   ```
   `translation.service.spec.ts` covers the four-level fallback matrix, nested reconstruction from dot-path keys, the `site IS NULL` filter, the legacy-blob fallback, and the override methods (two-level, flat, per-site isolation, byName not-found).
2. Integration tests:
   ```
   yarn test:backend:new:e2e
   ```
   `translation.e2e-spec.ts` covers the three endpoints, the `Cache-Control` header, per-site scope isolation (a public response contains no partner-only key), and the global Partners read.
3. Manual, with the migration applied and some `translation_strings` rows present:
   - `GET /jurisdictions/:id/translations?site=public&language=es` returns a flat overrides object with a `Cache-Control` header, and no partner-only keys.
   - `GET /translations?language=es` returns the global Partners overrides.
   - Confirm the email path is unchanged: `getMergedTranslations(jurisdictionId, language)` returns the same merged result as before on a seeded dataset (and on an environment with no key rows yet, via the legacy fallback).

## Author Checklist:

- [ ] Reviewed in a desktop view (N/A: backend service and endpoints only)
- [ ] Reviewed in a mobile view (N/A: backend service and endpoints only)
- [x] Added tests covering the changes (unit tests for the service; integration tests for the endpoints)
- [x] Ran `yarn generate:client` and/or created a migration when required (N/A: no schema change; client regeneration is deferred to the frontend hydration tickets)

## Review Process:


api/src/controllers/translation.controller.ts
```diff
@@ -0,0 +1,92 @@
...
+import { OptionalAuthGuard } from '../guards/optional.guard';
+
+const OVERRIDES_OK_RESPONSE = {
+  schema: {
+    type: 'object',
+    additionalProperties: { type: 'string' },
+  },
+} as const;
+
+const CACHE_CONTROL = 'public, s-maxage=300, stale-while-revalidate=600';
```
ludtkemorgan: thought: I could see us needing to fine tune these numbers. since we aren't changing the translations very often we could probably do an even higher number here. But since we don't know those numbers these are fine for now.

thought number 2: There are other endpoints that could use more of a cache control (I'm thinking jurisdiction, AMI charts, etc). Would there be a good place to put this configuration somewhere where more controllers could utilize it?

api/src/controllers/translation.controller.ts
```diff
@@ -0,0 +1,92 @@
...
+
+@Controller()
+@ApiTags('translations')
+@UsePipes(new ValidationPipe(defaultValidationPipeOptions))
+@UseGuards(ApiKeyGuard, OptionalAuthGuard)
+export class TranslationController {
+  constructor(private readonly translationService: TranslationService) {}
+
+  @Get('jurisdictions/:jurisdictionId/translations')
```
ludtkemorgan: issue: Our normal pattern is to have each controller determine the first string of the path. This is done within the `@Controller()` on each controller file. We already have a controller with the prefix "jurisdictions" (jurisdiction.controller.ts) so this could cause some confusion when looking for these endpoints.

My suggestion would be to have this file use `@Contoller("translations")`. then change both of the public endpoints to remove the trailing "translations". This would make the full path to be `translations/jurisdictions/:jurisdictionId` and `translations/byName/:jurisdictionName`

api/src/controllers/translation.controller.ts
```diff
@@ -0,0 +1,92 @@
...
+const OVERRIDES_OK_RESPONSE = {
+  schema: {
+    type: 'object',
+    additionalProperties: { type: 'string' },
+  },
+} as const;
+
+const CACHE_CONTROL = 'public, s-maxage=300, stale-while-revalidate=600';
+
+@Controller()
```
ludtkemorgan: issue: in order for the frontend to consume this it needs to be added to the backend-swagger. This can be done with the the api command `yarn generate:client` while the API is running.

However, as of some point last week the script is failing when trying to read the file and we are still digging into it. Let us know if you happen to solve the problem. But if we aren't able to solve it let's make sure that the backend-swagger step is added for the frontend tickets

ludtkemorgan (changes requested):
Functionally looks good. Just some thoughts on the controller

api/src/controllers/translation.controller.ts
ludtkemorgan: The issue has been resolved. https://github.com/bloom-housing/bloom/pull/6539. After rebasing with main you should be able to run the yarn command above

api/src/controllers/translation.controller.ts
seanmalbert: Good idea. I moved this into a cache-control util for reuse (api/src/utilities/cache-control.ts), and added `PUBLIC_CACHE_S_MAXAGE` and `PUBLIC_CACHE_STALE_WHILE_REVALIDATE` env vars so they're not hardcoded.

api/src/controllers/translation.controller.ts
seanmalbert: Thanks for flagging this. Updated per your suggestion.

api/src/controllers/translation.controller.ts
seanmalbert: All the things I've forgotten! I fixed the issue in api/src/controllers/background-jobs.controller.ts, but looks like you got to it first on main, and ran `yarn generate:client` to update swagger.

ludtkemorgan (approved):

Issue #6485, opened 2026-07-08
Configure optional fields for listing type


emilyjablonski:
For land-use listings, the following fields become optional instead of required: housing developer (developer), leasing agent address (listingsLeasingAgentAddress), community type (reservedCommunityTypes/reservedCommunityDescription), and per-unit minimum income (maxIncomeHouseholdSize{1-8} on the AMI chart). The Preferences section should be hidden inline entirely (same pattern as the existing disableListingPreferences prop on ProgramsAndPreferences).

On the backend, branch on listingType === landUse inside ListingCreateUpdateValidationPipe.transform() and strip those fields from the resolved requiredFields array before validation runs. Don't touch jurisdiction.requiredListingFields itself. We could consider a per-listing type fields configuration at a later time if we run into something similar.

Partners listing form UI: mirror the same exclusion wherever requiredFields gets passed into PaperListingForm sections so asterisks match what the server actually enforces.

PR #6538, opened 2026-07-22
fix: configure optional fields for land use listing type

KrissDrawing:
This PR addresses #6485 

## Description

Disables required fields for land use listings

On partners listing form
```
'developer',
'listingsLeasingAgentAddress',
'reservedCommunityTypes',
'reservedCommunityDescription',
```
should not be required on listing validation for land use listing.
Additionally `Preferences` section should be hidden.
`maxIncomeHouseholdSize` works kind of weird as if i click edit and change a value then save and then again edit then it will restore initial value, which is a bit counterintuitive. I can technically overwrite it by clicking save on save listing, but then it will go back to AMI chart value if i click edit on unit and save without changing anything 🤔 
But aside of that it should not be required 😄 

## Author Checklist:

## Review Process:


mcgarrye:
I'm a bit confused by the ticket. 'listingsLeasingAgentAddress', 'reservedCommunityTypes', 'reservedCommunityDescription' are never required to begin with. We really only need to remove the requirement for `developer` but even that field isn't required on LA

mcgarrye (changes requested):
The AMI chart change is not working correctly. If you clear a `Maximum annual income` it sends an empty string to the backend which is not saved in the db so the actual AMI values are still presented on the public site. How is this supposed to appear for public users? This is against how AMI is currently setup so the backend will not understand empty maximum fields

KrissDrawing:
@emilyjablonski Do you have any answers for above questions. So that will be:
- consideration about reason of removing those required fields (is that like preemptive work for sth?)
- What should we do with removing AMI value, should we make override with income 0 ?

Then this comment from PR:
`maxIncomeHouseholdSize works kind of weird as if i click edit and change a value then save and then again edit then it will restore initial value, which is a bit counterintuitive. I can technically overwrite it by clicking save on save listing, but then it will go back to AMI chart value if i click edit on unit and save without changing anything 🤔
`
That seem like a bug, should i fix that here?

emilyjablonski:
@mcgarrye That's definitely a true picture of what the required fields look like right now, but they have been unstable in LA - for example the other non-land-use team plans to use preferences soon, and have discussed using the reserved community field to indicate a listings accessibility status and it would in that case become required. I personally think it's valuable to define these as it seemed important for the team - but am definitely still open to feedback if it doesn't feel appropriate.

emilyjablonski:
@KrissDrawing That's definitely a bug! I will make a separate ticket for that, good find. https://github.com/bloom-housing/bloom/issues/6587

ludtkemorgan:
@mcgarrye has the above addressed all of your concerns?

KrissDrawing:
Should work now fine with minimum income beeing optional. And if no minimal income for any unit it should hide the column

mcgarrye:
Can you update the PR description please!

mcgarrye (approved):
Looks good, just update the PR description/testing instructions before merging

Issue #6518, opened 2026-07-15
Staleness check in the missing-translations job


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Today the missing-translations job only catches missing keys, so a changed English string with
un-updated translations passes silently. The job should also flag stale translations.

## Technical Context
- TDD: §6 (staleness), §9.
- Extend the job to flag keys whose stored `sourceHash` no longer matches the current English
  source, alongside missing keys.
- Depends on: #6508; #6510; #6511

## Acceptance Criteria
- The job reports stale keys in addition to missing ones and fails CI when stale keys exist.
- Editing an English key marks its non-English keys stale; re-saving a non-English value clears it.

## QA Notes
Backend/CI. Seed a changed English value with un-updated translations and confirm the job fails on
stale; re-save the translation and confirm it passes.

Issue #6510, opened 2026-07-15
Translation admin CRUD (per-key)


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
The Partners translation editor needs endpoints to read and write a jurisdiction's overrides per key,
with safe concurrent editing and the staleness/provenance metadata maintained on save.

## Technical Context
- TDD: §4.2.
- Raw list/get endpoints, a batch per-key PUT, and a per-key DELETE (revert to base). Each edit
  carries its own `lastUpdatedAt` for per-key optimistic locking.
- On a non-English save, record the current English key's `sourceHash` and set `origin = human`.
- Depends on: #6508

## Acceptance Criteria
- A batch save upserts each edited key; a stale per-key `lastUpdatedAt` returns `409` naming only
  that key, while non-conflicting keys in the batch are written.
- A non-English save records the current English key's `sourceHash` and sets `origin = human`.
- DELETE removes a key's override row (revert to base).
- Writes are scoped to the admin's jurisdiction; the global Partners scope is `superAdmin`-only.

## QA Notes
Backend-only. Test the per-key `409` (only conflicting keys reported, others written), revert via
DELETE, and that a non-English save updates `sourceHash`/`origin`. Confirm cross-jurisdiction writes
are rejected.

PR #6542, opened 2026-07-23
feature: per-key translation admin CRUD

seanmalbert:
This PR addresses #6510

## Description

Adds the write endpoints the Partners translation editor needs: read a jurisdiction's per-key overrides, save a batch of edits, and delete a key. Builds on the #6508 data model and #6509 read services. Backend only.

- **Endpoints** (`TranslationController`, `translations` prefix):
  - `GET /translations/jurisdictions/:jurisdictionId/raw` — list the jurisdiction's overrides across all sites and languages.
  - `GET /translations/jurisdictions/:jurisdictionId/raw/:site/:language` — the editable keys for that scope, each with `value`, `updatedAt`, `origin`, and a `stale` flag.
  - `PUT /translations/jurisdictions/:jurisdictionId/raw/:site/:language` — save a batch of edits.
  - `DELETE /translations/jurisdictions/:jurisdictionId/raw/:site/:language/:key` — delete one key.
- **Per-key locking.** Each edit includes the `updatedAt` the client last read. A key saves only if no one else has changed it since then. Any key that was changed first is reported as a conflict and skipped; the rest still save. The response is a `409` listing only the conflicting keys, so edits to different keys (or an English and a Spanish edit) don't block each other.
- **Staleness.** Saving a non-English value stores a hash of the current English value (`sourceHash`) and sets `origin = human`. A non-English key reads as `stale` when its stored hash no longer matches the current English hash, meaning the English text changed after it was translated. English keys are never stale. The English value used is the first that exists: `(jurisdictionId, en, site)`, then `(null, en, site)`, then `(null, en, null)`.
- **Permissions.** The service checks `canOrThrow(user, 'translation', action, { jurisdictionId })`. A new `'translation'` rule in `permission.service.ts` (like the `'listing'` one) lets a jurisdictional admin edit only their own jurisdiction. Only a top admin can edit the global (null-jurisdiction) scope.
- **New pieces.** A `sourceHash` helper (reused later by the seed #6519 and the staleness job #6518), and the request/response DTOs (`TranslationKeyEdit`, `TranslationUpdate`, `TranslationRawKey`, `TranslationOverrideRow`).

This is additive over #6509; the legacy `translations` table is untouched.

1. Unit tests (`cd api && yarn test`): `translation.service.spec.ts` covers the per-key 409 (only the conflicting key reported, the rest saved), new-key create, delete, a non-English save setting `origin = human` and the English `sourceHash`, staleness from a hash mismatch, and the jurisdiction permission check. `translation-source-hash.spec.ts` covers the hash.
2. Integration tests (`yarn test:backend:new:e2e`): `translation.e2e-spec.ts` covers the per-key 409 with partial writes, delete, the staleness lifecycle (editing the English key makes a Spanish key stale; re-saving it clears the flag), and a `403` on a cross-jurisdiction write.
3. Manual, as an admin (passkey header + logged-in admin cookie), against a local API on `http://localhost:3100`:
   - `PUT http://localhost:3100/translations/jurisdictions/<jurisdictionId>/raw/public/en` a batch with one stale `updatedAt` and one new key: expect a `409` naming only the stale key, with the new key saved.
   - `GET http://localhost:3100/translations/jurisdictions/<jurisdictionId>/raw/public/es` to see each key's `value`, `updatedAt`, `origin`, and `stale`.
   - Edit the English key (`PUT http://localhost:3100/translations/jurisdictions/<jurisdictionId>/raw/public/en`), then `GET http://localhost:3100/translations/jurisdictions/<jurisdictionId>/raw/public/es` to see `stale` turn true, then re-save the Spanish value to clear it.
   - `DELETE http://localhost:3100/translations/jurisdictions/<jurisdictionId>/raw/public/en/<key>` and confirm it's gone.

## Author Checklist:

- [ ] Reviewed in a desktop view (N/A: backend endpoints only)
- [ ] Reviewed in a mobile view (N/A: backend endpoints only)
- [x] Added tests covering the changes (unit tests for the service and the hash util; integration tests for the endpoints)
- [x] Ran `yarn generate:client` and/or created a migration when required (regenerated backend-swagger for the new admin endpoints; no schema change)

## Review Process:


api/src/services/translation.service.ts
```diff
@@ -160,6 +181,270 @@ export class TranslationService {
...
+    const results = await Promise.all(
+      dto.edits.map((edit) =>
+        this.applyEdit(
+          { jurisdictionId, language, site },
+          edit,
+          englishHashes,
+          isEnglish,
+        ),
+      ),
+    );
```
seanmalbert, line 295: question (non-blocking): this can handle up to 1000 edits, but a large "save all" would briefly occupy the whole pool and slow other requests during that window. Typical editor saves should only be a few keys, so likely not a real issue. 

Do you think this is a concern we should handle now? If so, I can update it to apply edits in batches (e.g. 25 at a time).

api/src/dtos/translations/translation-update.dto.ts
```diff
@@ -0,0 +1,21 @@
...
+  ValidateNested,
+} from 'class-validator';
+import { ValidationsGroupsEnum } from '../../enums/shared/validation-groups-enum';
+import { TranslationKeyEdit } from './translation-key-edit.dto';
+
+export class TranslationUpdate {
+  @Expose()
+  @IsDefined({ groups: [ValidationsGroupsEnum.default] })
+  @IsArray({ groups: [ValidationsGroupsEnum.default] })
+  @ArrayMaxSize(1000, { groups: [ValidationsGroupsEnum.default] })
```
ludtkemorgan, line 16: question: I don't anticipate updating more than 1000 entries at once, but I'm curious why this max size was chosen?

api/test/integration/translation.e2e-spec.ts
```diff
@@ -134,4 +175,183 @@ describe('Translation Controller Tests', () => {
       .set(passkey)
       .expect(404);
   });
+
+  describe('admin raw CRUD', () => {
```
ludtkemorgan: nit: sorry I didn't catch this on your last PR. Can you update this file to be broken down by the controller endpoints? It's easier to read through if there is a describe block for each endpoint.

api/src/services/translation.service.ts
```diff
@@ -160,6 +181,270 @@ export class TranslationService {
...
+  ): Promise<string | null> {
+    const where = { ...scope, key: edit.key };
+    const data = {
+      value: edit.value,
+      origin: TranslationOrigin.human,
+      sourceHash: isEnglish ? null : englishHashes.get(edit.key) ?? null,
+    };
+
+    if (edit.lastUpdatedAt) {
+      const result = await this.prisma.translationStrings.updateMany({
```
ludtkemorgan, line 327: question: why is this an update many? I could be misunderstanding this, but wouldn't there only ever be 1 value with that where clause?

api/src/services/translation.service.ts
```diff
@@ -160,6 +181,270 @@ export class TranslationService {
...
+    language: LanguagesEnum,
+    key: string,
+    user: User,
+  ): Promise<SuccessDTO> {
+    await this.authorizeJurisdiction(
+      user,
+      jurisdictionId,
+      permissionActions.delete,
+    );
+    await this.prisma.translationStrings.deleteMany({
```
ludtkemorgan, line 360: Same here. Aren't these 4 values in the where clause the unique key? So could there ever be more than 1 that it is deleting?

api/test/integration/translation.e2e-spec.ts
```diff
@@ -134,4 +175,183 @@ describe('Translation Controller Tests', () => {
...
+      const res = await request(app.getHttpServer())
+        .get(enScope())
+        .set('Cookie', adminCookies)
+        .set(passkey)
+        .expect(200);
+
+      const row = res.body.find((r) => r.key === 'footer.title');
+      expect(row.value).toEqual('Footer Title');
+      expect(row.origin).toEqual('human');
+      expect(row.stale).toBe(false);
```
ludtkemorgan: nit: These 4 lines can be changed to be one expect statement
```
      expect(res.body).toContainEqual(
        expect.objectContaining({
          key: 'footer.title',
          value: 'Footer Title',
          origin: 'human',
          stale: false,
        }),
      );
```

api/src/services/translation.service.ts
```diff
@@ -160,6 +181,270 @@ export class TranslationService {
...
+    isEnglish: boolean,
+  ): Promise<string | null> {
+    const where = { ...scope, key: edit.key };
+    const data = {
+      value: edit.value,
+      origin: TranslationOrigin.human,
+      sourceHash: isEnglish ? null : englishHashes.get(edit.key) ?? null,
+    };
+
+    if (edit.lastUpdatedAt) {
```
ludtkemorgan, line 326: question: Am I reading this right that if you are inserting a new key using the `put` endpoint it is recommended to not send a lastUpdatedAt?

api/src/services/translation.service.ts
ludtkemorgan: hmm, good question. I personally don't think this would be an issue. As a reference point LA overrides about 175 values on the public site. We would only be doing that large chunk before the site first goes live which means it wouldn't be getting much other traffic. Let's keep it as is for now and can do some testing in a deployed environment.

ludtkemorgan (commented):
Still doing some testing but so far haven't found any issues and will continue testing later today. Just some nit comments

api/src/services/permission.service.ts
```diff
@@ -100,6 +100,12 @@ export class PermissionService {
             `r.obj.jurisdictionId == '${adminInJurisdiction.id}'`,
             `(${permissionActions.read}|${permissionActions.create}|${permissionActions.update}|${permissionActions.delete})`,
           );
+          await enforcer.addPermissionForUser(
+            user.id,
+            'translation',
+            `r.obj.jurisdictionId == '${adminInJurisdiction.id}'`,
+            `(${permissionActions.read}|${permissionActions.create}|${permissionActions.update}|${permissionActions.delete})`,
+          );
```
ludtkemorgan: issue: This allows jurisdictional admins to create, update, and delete translations. We should limit these actions to only admins

api/src/dtos/translations/translation-update.dto.ts
seanmalbert: I figured since there were almost 1500 keys, and a client probably wouldn't override all of those, that 1000 was a safe limit. We could up it to 2000 to be safe, or cap it at something more reasonable/realistic that's closer to the 175 that LA overrides.

api/test/integration/translation.e2e-spec.ts
seanmalbert: Done. thanks for flagging

api/src/services/translation.service.ts
seanmalbert: Good question. `updateMany` returns a count, so if `count === 0`, we know that the record was changed or removed. This could be an `update` with a try/catch to handle the conflict instead.

api/src/services/translation.service.ts
seanmalbert: Same reason applies here. It's a matter of `deleteMany` returning a count of 0 vs a singular delete and handling a record not found error.

api/src/services/translation.service.ts
seanmalbert: Correct, since it's a new key there's no prior version to lock to and check against.

api/src/services/permission.service.ts
seanmalbert: Done. The jurisdictional admin perm is now removed, so only admins can CRUD.

api/src/dtos/translations/translation-update.dto.ts
ludtkemorgan: Makes sense. Lets keep with this for now

api/src/services/translation.service.ts
ludtkemorgan: Thanks for the clarity. After some google, it sounds like the best practice is what you did because the try/catch route is normally if the catch scenario is abnormal. Which in our case is not.

api/test/integration/translation.e2e-spec.ts
```diff
@@ -67,71 +84,295 @@ describe('Translation Controller Tests', () => {
...
+          .set('Cookie', adminCookies)
+          .set(passkey)
+          .expect(200)
+      ).body;
+      expect(es.find((r) => r.key === 'greeting').stale).toBe(false);
+    });
+
+    it('forbids a jurisdictional admin from writing translations', async () => {
+      // Writing is admin-only for now; jurisdictional-admin access is added later, gated by the
+      // enableDbDrivenContent flag. Both its own jurisdiction and another are denied.
```
ludtkemorgan: question: is this statement accurate? I don't think we should be giving jurisdictional admin's access long term unless we start deploying all jurisdictions in one database.

ludtkemorgan (approved):
Thanks for the changes. This looks good and I'm going to approve it, but lets hold off on merging it until after the next LA release goes out.

api/test/integration/translation.e2e-spec.ts
seanmalbert: I thought the long-term plan was to allow jurisdictional admins to handle translations.  If not, that removes one less issue we need to cover. Aren't admins now only the Bloom team, and you want folks at the jurisdiction level to be able to do this work?

ludtkemorgan: Oh I see where the confusion came from. Here are the roles broken down:
* `Partner` - Partner portal user who has access to individual listings
* `Jurisdictional Admin` - Partner portal user who has admin like access for only 1 jurisdiction in the system
* `Limited Jurisdictional Admin` - Jurisdictional admin that doesn't have access to PII data
* `Support Admin` - Currently only used by Doorway. Admin users with some permissions taken away. Used by jurisdiction employees that handle user support issues
* `Admin` - Partner portal user who has access to all jurisdictions in the system. For most implementations this includes a small number employees of the jurisdiction
* `Super Admin` - Admin users that have additional technical abilities. This is only used by engineers that need access to changing things (e.g turning on feature flags).

So for this we still only want "Admins" and "Super Admins" (all super admins are also Admins). That could change if we decide to host all jurisdictions on one partner portal.

seanmalbert: Thank you for the breakdown and clarification!

PR #6540, opened 2026-07-23
fix: show contact through leasing agent section

KrissDrawing:
This PR addresses #6486 

- [x] Addresses the issue in full
- [ ] Addresses only certain aspects of the issue

## Description

Adds `Is the application through leasing agent contact only?` radio for application section for land use lisitngs.

## How Can This Be Tested/Reviewed?

Go to partners listing with land use listing type (you can change it in prisma studio) application tab.
Application types will have new radio, make sure it also clears fields when switching to `yes`.
Also there might be reasonable to test this after [this one](https://github.com/bloom-housing/bloom/pull/6537) to test for new listings.

## Author Checklist:

- [ ] Added QA notes to the issue with applicable URLs
- [ ] Reviewed in a desktop view
- [ ] Reviewed in a mobile view
- [ ] Reviewed considering accessibility
- [x] Added tests covering the changes
- [ ] Made corresponding changes to the documentation
- [ ] Ran `yarn generate:client` and/or created a migration when required

## Review Process:

- Read and understand the issue
- Ensure the author has added QA notes
- Review the code itself from a style point of view
- Pull the changes down locally and test that the acceptance criteria is met
- Either (1) explicitly ask a clarifying question, (2) request changes, or (3) approve the PR, even if there are very small remaining changes, if you don't need to re-review after the updates

Issue #6487, opened 2026-07-08
Public land use application section


emilyjablonski:
When a listing's applicationMethods includes LeasingAgent, the public listing detail page should show contact info in the Apply section instead of the normal apply-online/download-application buttons. Update getHasNonReferralMethods() in ListingViewSeedsHelpers.tsx so a leasing-agent-only listing isn't treated as having no applicable methods, and add the new render branch in listing_sections/Apply.tsx.

If the leasing agent method exists, don't show any other content in there, and the section should look like:

How to apply
Heading: Contact the leasing agent or property manager
Content: Refer to the leasing agent or property manager contact section below to apply.

<img width="300" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNldJRGc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--035f606e3d37af8c8b5dc6a2ffe0b618754bb309/Screenshot%202026-07-16%20at%208.17.02%E2%80%AFAM.png" alt="Screenshot 2026-07-16 at 8.17.02 AM.png" />

PR #6544, opened 2026-07-27
fix: update land use public aplication section

KrissDrawing:
This PR addresses #6487

## Description

Displays different application section for land use listing.
Just not fully sure if `If the leasing agent method exists, don't show any other content in there, and the section should look like:` from ticket refers to that apply card, or all those between `Apply` and `LeasingAgent` (i left them for now, as it seem it refers to data in apply section like address etc.)
image

Set application method for listing to the one with type: `leasingAgent` (in prisma studio for example or like [here](https://github.com/bloom-housing/bloom/pull/6540) if merged). Then visit listing on public site, it's apply section should be adjusted.

## Author Checklist:

## Review Process:


ludtkemorgan (approved):
This looks good and I believe you understood it correctly, this is to only update that one block not all of the other ones.

Issue #6511, opened 2026-07-15
JurisdictionContent service and admin CRUD


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
Structured content (footer, FAQ, resources, contact, disclaimers) moves to the `JurisdictionContent`
model with English-row fallback and per-item merge, read by the public site and edited in Partners.

## Technical Context
- TDD: §4.3, §4.4, OQ5.
- `getMergedContent`: two-level merge (language over English) with `mergeListById` and tombstones
  for list items. Nested content DTOs; a read-time shape guard that logs (does not throw) on
  mismatch.
- Public read `GET /jurisdictions/:id/content?language=`; per-language admin CRUD with row-level
  optimistic locking (the content row's fields live together, so it locks per row).
- Depends on: #6508; #6513

## Acceptance Criteria
- Language row merges over English field by field; list items merge by stable id; a tombstoned item
  is skipped (reads as "not present"); items added only in a non-English row are appended.
- Malformed content is rejected by the nested-DTO `ValidationPipe`; a read-time shape mismatch logs
  a warning without throwing.
- `GET /jurisdictions/:id/content?language=` returns the merged content; `204` when no row exists.
- Admin PUT upserts the `(jurisdiction, language)` row with an optimistic-lock check.

## QA Notes
Backend-only. Test the merge cases (field fallback, list merge-by-id, tombstone skip, non-English
additions), the 204-on-empty read, and the row-level `409`. Malformed HTML/shape payloads should be
rejected on write.

Issue #5734, opened 2026-01-05
Prisma upgrade introduced new prisma config file


YazeedLoonat:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
With a recent upgrade of prisma a new prisma config file can be leverage to help us configure prisma.
https://www.prisma.io/docs/orm/reference/prisma-config-reference#engine

we should look through that doc and switch our config to the new config file (along with creating the new config file) instead of keeping our config in the schema file

## Acceptance Criteria
- Create a prisma config file that is configured to work for us
- Ensure that the api is still able to connect to the db across our different envs

## QA Notes
We should still be able to connect to the database but locally and in production envs

YazeedLoonat:
We should also be able to tackle: https://github.com/bloom-housing/bloom/issues/5451 as part of this change

PR #6547, opened 2026-07-28
feature: jurisdiction content crud

seanmalbert:
This PR addresses #6511

## Description

Adds the backend for jurisdiction structured content (footer, FAQ, resources, contact, disclaimers): a public read that merges a language over the English default, and per-language admin CRUD. Backend only. No schema change.

> Stacked on #6510 (translation admin CRUD); this branch merges that base in.

- **Endpoints** (`JurisdictionContentController`, `jurisdictionContent` prefix):
  - `GET /jurisdictionContent/jurisdictions/:jurisdictionId?language=` — the merged content for that language; `204` when the jurisdiction has no content row at all; `404` for an unknown jurisdiction.
  - `GET /jurisdictionContent/byName/:jurisdictionName?language=` — the same read resolved by jurisdiction name.
  - `GET /jurisdictionContent/jurisdictions/:jurisdictionId/admin` — list the rows across languages.
  - `GET /jurisdictionContent/jurisdictions/:jurisdictionId/admin/:language` — one language's row, with `updatedAt` for the lock check.
  - `PUT /jurisdictionContent/jurisdictions/:jurisdictionId/admin/:language` — upsert one language's row.
- **Two-level merge.** A language row folds over English field by field; unset fields fall back to English. Objects recurse. Object lists with a stable `id` merge by id (`mergeListById`) — partial translation and reorders work, English order is kept, and a `{ id, _deleted: true }` tombstone drops an item. Primitive lists (e.g. `footer.textSectionsHtml`) are fully replaced.
- **Row-level locking.** The five content fields share one `(jurisdiction, language)` row, so it locks as a unit: a save sends the `lastUpdatedAt` it read and writes only if the row hasn't moved, else `409` (also `409` if the row is deleted between the write and the read-back). English and Spanish are separate rows and don't contend.
- **Validation and sanitization.** Nested DTOs enforce the shape via the existing `ValidationPipe` (`400` on malformed input); list fields are capped (`@ArrayMaxSize(256)`) and URL fields (link `href`, logo `src`/`url`) are scheme-checked (`@IsSafeUrl` — rejects `javascript:`/`data:` and other non-http(s) schemes, including whitespace-obfuscated ones). HTML leaves run through the existing `@SanitizeHtml()` on write. A read-time guard validates the merged document and logs rather than throws, so a hand-edited bad row warns (and the merge preserves its items rather than dropping them) instead of breaking the page.
- **Permissions.** The service checks `canOrThrow(user, 'jurisdictionContent', action, { jurisdictionId })`. Content writes are limited to the admin role, which has access to every jurisdiction in the system. A jurisdictional admin is denied its own jurisdiction and any other. Public reads are anonymous.

Out of scope by design: the content source hash for staleness stays with #6518.

1. Unit (`cd api && yarn test`): `content-merge.spec.ts` (merge matrix — fallback, merge-by-id incl. nested resources, tombstone, append, primitive replace, id-less/duplicate handling, prototype guard), `jurisdiction-content.service.spec.ts` (`204`/`404`, `byName`, shape-guard warning, permission check, create / update / stale-`409` / concurrent-`409` / delete-race-`409`), and `is-safe-url.spec.ts` (URL scheme validation).
2. Integration (`yarn test:backend:new:e2e`): `jurisdiction-content.e2e-spec.ts` — merged read + cache header, language fallback, `byName`, `204`, `404`, admin create/read-back, list, optimistic-lock `409`/`200`, `400` on malformed input (including an unsafe URL scheme and an over-cap list), HTML sanitization on write, partial-PUT field clearing, and `403` for anonymous, a non-admin user, and a jurisdictional admin (writes are admin-only).
3. Manual, against a local API on `http://localhost:3100` (passkey header; an admin cookie for `/admin` routes). Base path: `jurisdictionContent/jurisdictions/<jurisdictionId>`.
   - `GET .../?language=es` — Spanish merged over English (translated fields win, the rest fall back); `GET` a jurisdiction with no row — `204`.
   - `PUT .../admin/es` a body, then `GET .../admin/es` to read it back; re-`PUT` with a stale `lastUpdatedAt` — `409`, with the current one — `200`.
   - As a jurisdictional admin, `PUT` any jurisdiction — `403` (content writes are admin-only).

Note: the reused `@SanitizeHtml()` normalizes an absent optional HTML field to `null` on read (e.g. `contact.addressHtml: null`).

## Author Checklist:

- [ ] Reviewed in a desktop view (N/A: backend endpoints only)
- [ ] Reviewed in a mobile view (N/A: backend endpoints only)
- [x] Added tests covering the changes (unit tests for the merge utility and the service; integration tests for the endpoints)
- [x] Ran `yarn generate:client` and/or created a migration when required (regenerated backend-swagger for the new endpoints; no schema change, the table is from #6508)

## Review Process:


api/src/services/jurisdiction-content.service.ts
```diff
@@ -0,0 +1,267 @@
...
+import { permissionActions } from '../enums/permissions/permission-actions-enum';
+import { ValidationsGroupsEnum } from '../enums/shared/validation-groups-enum';
+
+const CONTENT_SELECT = {
+  footer: true,
+  faq: true,
+  resources: true,
+  disclaimers: true,
+  contact: true,
+} as const;
```
ludtkemorgan: Oh interesting, I didn't know about the `as const` functionality in Typescript. That could be useful in other places.

question: would we want to type this as `Prisma.JurisdictionContentSelect` to verify that only valid selects are here? That would mean the removal of the `as const`

api/src/services/jurisdiction-content.service.ts
```diff
@@ -0,0 +1,267 @@
...
+    const contentFor = (lang: LanguagesEnum): MergeableContent => {
+      const match = rows.find((row) => row.language === lang);
+      return match
+        ? ({
+            footer: match.footer,
+            faq: match.faq,
+            resources: match.resources,
+            disclaimers: match.disclaimers,
+            contact: match.contact,
+          } as MergeableContent)
```
ludtkemorgan: nit: I don't think the `as MergeableContent` is required here

api/test/integration/jurisdiction-content.e2e-spec.ts
```diff
@@ -0,0 +1,465 @@
...
+        .expect(201)
+    ).headers['set-cookie'];
+  });
+
+  afterAll(async () => {
+    await prisma.$disconnect();
+    await app.close();
+  });
+
+  it('returns the English content and caches the response', async () => {
```
ludtkemorgan: nit: These tests look great, could you split them up by controller endpoint by wrapping them in `describe` blocks? That helps readability on knowing which endpoints are being tested

api/src/dtos/jurisdiction-content/footer-content.dto.ts
```diff
@@ -0,0 +1,99 @@
...
+  id: string;
+
+  @Expose()
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @ApiProperty()
+  text: string;
+
+  @Expose()
+  @IsString({ groups: [ValidationsGroupsEnum.default] })
+  @IsSafeUrl({ groups: [ValidationsGroupsEnum.default] })
```
ludtkemorgan, line 41: thought: I like this new IsSafeUrl decorator. We typically use the `@IsUrl` decorator for inputted urls. The nice thing about that one is you can limit it to specific protocols. Would you want to use it here as well?

ludtkemorgan: Oh actually now that I think about it, the footer can link to pages within the site so it won't have a protocol.

api/src/dtos/jurisdiction-content/faq-content.dto.ts
```diff
@@ -0,0 +1,79 @@
...
+  @SanitizeHtml()
+  @ApiProperty()
+  answerHtml: string;
+
+  // Tombstone: a non-English row sets this to drop the English item from the merge.
+  @Expose()
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @IsBoolean({ groups: [ValidationsGroupsEnum.default] })
+  @ApiPropertyOptional()
+  _deleted?: boolean;
```
ludtkemorgan, line 38: question: Can you explain the flow of how the tombstoned entries would happen? I see that you can pass this in when doing a put but I don't follow the intended path. Specifically can you explain when this scenario would happen vs just fully deleting the entry?

ludtkemorgan (commented):
This is looking great. I'll need to take another pass at it before I approve it.

Question: would it be possible to add some jurisdictionContent to one of the seeded jurisdictions? That would help with testing the endpoints. I think it would be fine to do that since nothing is using these new endpoints. However, I understand if you don't want to until you are working on the front-end tickets.

api/src/dtos/jurisdiction-content/faq-content.dto.ts
seanmalbert: `_deleted` removes the item from the languages output, so that it doesn't use the English fallback, where as fully deleting the translated entry allows for the English fallback.

This allows for per-language divergence. So let's say there's a list in English and Spanish, and for some reason, maybe linguistically or culturally, an item in the list doesn't make sense to have in Spanish (and the jurisdiction doesn't want the fallback), then setting `_deleted: true` allows a jurisdiction to intentionally hide the item.

If that's not something that will ever be used, then we can remove that field on all the DTOs that have it, and always fallback to showing the English if the language item doesn't exist.

ludtkemorgan: I don't think we have a current use case, but that is not to say that it would never happen. I'm ok keeping this, we'll just need to make sure the UI is very clear on what that action does to not cause any confusion

api/src/services/jurisdiction-content.service.ts
seanmalbert: Good idea, updated to use `Prisma.validator<Prisma.JurisdictionContentSelect>()({ ... })`, which gives us the benefit of what the `as const` provided and `Prisma.JurisdictionContentSelect`.

api/src/services/jurisdiction-content.service.ts
seanmalbert: You're right, removed.

api/test/integration/jurisdiction-content.e2e-spec.ts
seanmalbert: Done, and is now split out into five blocks, one per endpoint.

seanmalbert:
@ludtkemorgan , this is ready for you re-review. I added jurisdiction content to the seeds here too.

api/src/services/jurisdiction-content.service.ts
```diff
@@ -0,0 +1,269 @@
...
+import { JurisdictionContentFields } from '../dtos/jurisdiction-content/jurisdiction-content-fields.dto';
+import { JurisdictionContentUpdate } from '../dtos/jurisdiction-content/jurisdiction-content-update.dto';
+import { mergeContent, MergeableContent } from '../utilities/content-merge';
+import { mapTo } from '../utilities/mapTo';
+import { permissionActions } from '../enums/permissions/permission-actions-enum';
+import { ValidationsGroupsEnum } from '../enums/shared/validation-groups-enum';
+
+// Validated against the model so an invalid field is a compile error, while keeping the exact
+// selected-field type on the rows the read path works with.
+const CONTENT_SELECT = Prisma.validator<Prisma.JurisdictionContentSelect>()({
```
ludtkemorgan: issue: I really like using the validator here. I just read the docs on it and unfortunately it looks like it is [deprecated](https://www.prisma.io/docs/orm/v6/prisma-client/type-safety/prisma-validator). If we move off of prism-client-js it won't work. We want to do that at [some point](https://github.com/bloom-housing/bloom/issues/5734) to get more configuration options. Looks like Prisma has a [recommendation](https://www.prisma.io/blog/satisfies-operator-ur8ys8ccq7zb) to use Typescript's "satisfies". I've never used it but it looks pretty nifty and I think it satisfies (pun intended) our usecase.

```
const CONTENT_SELECT = {
  footer: true,
  faq: true,
  resources: true,
  disclaimers: true,
  contact: true,
} satisfies Prisma.JurisdictionContentSelect;
```

ludtkemorgan (approved):
Thanks for adding the seeded data. It helped me for testing. We'll want to update it once we start using this on the front-end, but for now it works.

I'm going to approve this, just one comment that should be addressed before merge

api/src/services/jurisdiction-content.service.ts
seanmalbert: Thanks for looking into the docs for it. Updated per your suggestion. I hope you're "satisfied" 🤣

Issue #6459, opened 2026-07-01
Autofill read-only minimum occupancy


emilyjablonski:
For LA's land-use listings, they want the minimum occupancy fields on units automatically populated and read-only for compliance.

SRO 1
Studio  1
1-Bedroom 1
2-Bedroom 2
3-Bedroom 4
4-Bedroom 6
5-Bedroom 8
6-bedroom 10

For the 7+ bedrooms, because it's an unusual situation, they don't want that pre-filled. It will need to remain editable.

For now, let's just add exactly these numbers in core and make it not configurable. If the type is land use, then apply this behavior, otherwise don't.

If the type does not exist, we can put it behind a main land use feature flag, and later ensure it's only applied for the new land-use listings. We can keep the maximum occupancy editable as-is.

PR #6552, opened 2026-07-30
fix: set min occupancy for type for land use

KrissDrawing:
This PR addresses #6459 

## Description

Locks min occupancy for land use units.
There are two things: 
1. Do we want to lock min occupancy when no unit type selected? (it does not change how it works, as unit type is required so we won't be able to pick wrong value, its just ux thing) For now it wont block it as it was simplest solution.
2. I don't understand this one from task: "If the type does not exist, we can put it behind a main land use feature flag, and later ensure it's only applied for the new land-use listings. We can keep the maximum occupancy editable as-is."

Create new land use listing (Angelopolis). Add unit, it should block min occupancy field with pre-filled value, and allow to edit it for 7+ bedrooms.
It should to work as before for every other (non land-use) listing

## Author Checklist:

## Review Process:


ludtkemorgan (approved):
This looks good. 

@emilyjablonski I am also unsure what the one point means 
```
If the type does not exist, we can put it behind a main land use feature flag, and later ensure it's only applied for the new land-use listings. We can keep the maximum occupancy editable as-is.
```
Can you give clarity on what that means?

Issue #6253, opened 2026-04-30
Doorway feature: Search form on the public home page


ludtkemorgan:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
Doorway's home page utilizes a search form instead of the standard text Hero. We need to add this as an option for the home page in the core repo.

<img width="1174" height="916" alt="Image" src="https://github.com/user-attachments/assets/8fd2ac99-fa11-4537-9103-ab7c206365ab" />

Open questions:
* Design help is needed to know what it should look like in core
* How can we handle configurations?
* How do we handle additional custom content for jurisdictions (e.g the county map)

hannan-abdi:
This feature needs to be completely redesigned to meet accessibility needs.

sarahlazarich:
Can we go ahead and pull in this design?

hannan-abdi:
@ludtkemorgan @mcgarrye 
Please find the [ready-for-dev designs](https://www.figma.com/design/7txDwTSN3ITUWI14598NyE/Homepage-search-form?node-id=4787-16197) for the new homepage search hero, complete with development annotations for accessibility and interaction.

Note, this new iteration (without an image) has been added to the Core Design System [here](https://www.figma.com/design/xqyamlgdsASkLTYlcifBwq/core-patterns?node-id=4692-3459&t=SruNVhEBHA23Gw5W-1).

ludtkemorgan:
QA Note:
* The Home page should have the new search form in bloom doorway: https://bloom-doorway.bloom.exygy.dev/

ludtkemorgan:
issue:
* Image does not appear in the Netlify environment. Will test in AWS environment so keeping this in the QA column

PR #6553, opened 2026-07-30
feat: add home search form in hero

ludtkemorgan:
This PR addresses #6253 

## Description

Adds the home page search filter behind a feature flag. This includes all of the possible variations that can be done by a jurisdiction:
* Optional image - by default an image next to the search form does not exist. But can be added if image exists in the proper directory (Note: the image only exists in large desktop and mobile view)
* Filter by county - The "County" field only exists if the `enableByCounty` feature flag is turned on
* Sub note: a sub note can appear under the search form if the string exists

Also fixed the following bugs:
* Filtering by rent was not accurate because it is doing string comparison as the monthlyRent is saved as TEXT instead of a number
  * This changes a database field meaning there could be other downstream affects that need to be tested
* When units are of type "percentage of income" the map card was showing "n/a". This is because that field was missing from the query to the database

The following scenarios should be tested:
1. When `enableHomePageSearchForm` feature flag turned off the normal home page hero should exist
2. When `enableFilterByCounty` feature flag is turned on the "County" filter should appear in the "Search by Filter" tab.
  a. When nothing is entered in this field no filtering by counties should happen when clicking "View listings" is clicked
  b. When a county is selected only listings from that county should appear after clicking "View listings"
3. Image appropriately appears when an image file called `hero-image.jpg` is added to the `sites/public/public/images` directory. I recommend taking the image from the current Doorway home page. Note: you may need to restart the app when adding/removing the image
4. sub note appears when translation string "welcome.searchSubNote" is not empty. This can be tested by removing it from `sites/public/page_content/locale_overrides/general.json`
5. All filters work on the "Search by filter" tab
  a. County
  b. Bedrooms
  c. Max rent
  d. "Availability" - Clicking "Show available units only (hide waitlist)" should add the availability filter of "unitsAvailable"
6. Property search name filter should work when switching to the "Search by Property" tab

## Author Checklist:

## Review Process:


shared-helpers/src/locales/general.json
```diff
@@ -1475,8 +1477,13 @@
   "welcome.allApplicationClosed": "All applications are currently closed, but you can view closed listings.",
   "welcome.cityRegions": "City regions",
   "welcome.findRentalsForYou": "Find rentals for you",
+  "welcome.heroAltText": "Hero",
```
ludtkemorgan, line 1480: All jurisdictions that require an image should also override this text to describe the image

mcgarrye: **question:** How can we document this somewhere so we don't lose this?

ludtkemorgan: I can add it to the additional info in the feature flag file

sites/public/src/components/home/Home.tsx
```diff
@@ -38,6 +40,11 @@ export const Home = (props: HomeProps) => {
 
   const metaDescription = t("pageDescription.welcome", { regionName: t("region.name") })
 
+  const enableHomePageSearchHero = isFeatureFlagOn(
```
mcgarrye, line 43: **question:** doJurisdictionsHaveFeatureFlag vs isFeatureFlagOn?

mcgarrye (changes requested):
Image
Image
Missing translations for # of bedrooms, link to external site does not seem accessible, `All Counties` is cut off.

sites/public/src/components/home/Home.tsx
ludtkemorgan: `doJurisdictionsHaveFeatureFlag` is used for checking if the logged in users to see if jurisdictions they have access to have the feature flag turned on. It really only needs to be used by the partner site. I see it's used in two places on the public site. Those should probably be switched to `isFeatureFlagOn` which is a public site only utility function

ludtkemorgan:
@mcgarrye I've addressed your comments so it's ready for re-review. One call out is that the "Any county" text can still get cut off in certain languages and screen sizes. But we can't make the field any bigger without messing with the button

mcgarrye (approved):
lgtm

emilyjablonski:
@ludtkemorgan Working through testing, and enabling the county feature flag is not turning on the county filter for me? Not sure if there are additional instructions

emilyjablonski (changes requested):
I'd like to see 2 and 4 fixed before merge, but am open to feedback about the others. Here is a PR with some of these suggestions! https://github.com/bloom-housing/bloom/pull/6588

1. In my opinion, there is too much empty white space here, and the distance between the fields and the button is not ideal. Afaik this is going to be the much more common state compared to the county version if we turn this on in future jurisdictions with the map.Image

2. We are not using placeholder text anymore, so these Anys should be removed
Image
Image

3. With the county filter on so we have less wiggle room and in other languages, we have this wrapping going on:
Image

4. We have inconsistent spacing between the fields when the county filter is on on mobile
Image


5. The top left radius is incorrect on mobile
Image

ludtkemorgan:
Thanks @emilyjablonski for the thorough review. I pulled in your commit and I like all of the changes. The only thing that now is a little weird is on the search by listing view the image is further away from the search box 
image

emilyjablonski:
@ludtkemorgan Oh boy I didn't test the image state at all I missed that so sorry! I will peep Monday

emilyjablonski:
Screenshot 2026-08-19 at 4 46 46 PM
Super open to design suggestions, but I actually am not bothered by the screenshot - esp seeing it in Filipino as the text will expand - if that feels blocking to you I might reach out to Hannan and get some more ideas!

ludtkemorgan:
@emilyjablonski good point. I'm good with this and can get a second opinion from Hannan after it is in a deployed environment

Issue #6520, opened 2026-07-15
Rollout gating and staging-to-production promotion


seanmalbert:
## #6520 — Rollout gating and staging-to-production promotion

- [ ] General
- [x] Backend
- [x] Public
- [x] Partners

## Background / Issue
Rollout is staged behind the per-jurisdiction `enableDbDrivenContent` flag, and admin-authored
changes are promoted between environments through the editor rather than direct DB sync.

## Technical Context
- TDD: §7.2, §7.3.
- Switch the public render path on `enableDbDrivenContent`: with it off, the site keeps using the
  existing static-import path. The flag gates rendering and editor visibility only; it is not
  checked in the permission layer, since editing is limited to `admin` regardless (§4.5).
- Each editor gates its own visibility on the flag as it is built (#6516, #6517). This issue
  confirms the flag behaves consistently across both sites rather than implementing that gating.
- Add Export/Import actions to the editors (download the current screen's JSON; import PUTs it,
  reusing the editor's conflict resolution).
- Depends on: #6510 (translation CRUD); #6511 (content service); #6516 (translation editor);
  #6517 (structured-content editor).

## Acceptance Criteria
- With the flag off, the static-import render path remains active and both editors are hidden;
  flipping it on switches the jurisdiction to DB-driven content and reveals the editors to `admin`.
- Export downloads the current screen's JSON; Import PUTs it in the target environment and surfaces
  the same conflict resolution as the editor on a since-changed target.
- Rollback (flag off) restores static rendering.

## QA Notes
Public + Partners. Toggle the flag and confirm the public site and editor visibility switch
accordingly; export from staging and import to a target, including the conflict path when the target
changed after export. Confirm rollback restores static rendering.

Issue #6516, opened 2026-07-15
Partners translation editor


seanmalbert:
- [ ] General
- [ ] Backend
- [ ] Public
- [x] Partners

## Background / Issue
Admins (not engineers) need to edit translation overrides safely, over a ~1,477-key base, with
awareness of fallback, staleness, and section visibility.

## Technical Context
- TDD: §6.1, §5.6, §6 (staleness).
- Per-key editor: search matching key paths and values, site labels derived from existing overrides,
  fallback awareness and revert, per-key optimistic-lock conflict UI, interpolation-token-preserving
  validation, the staleness flag, and section-visibility awareness for `tIfExists` keys (hide via
  empty vs revert via delete).
- Visibility keys are derived, not listed. `tIfExists` is used only for keys that may not exist in
  every fork, and every current call site refers to a key with no value in the base locale files. A
  key with no base value is therefore one whose presence controls whether a section renders, and
  the editor already holds the base value for the fallback indicator. Warn on saving an empty value
  or reverting such a key. This also covers template keys such as
  `listingResource.additionalCard${index}.title`, which no static list could enumerate.
- Values are plain strings; 9 of about 1,492 contain inline HTML (`<a>`, `<strong>`, `<br>`). No
  rich-text editor is needed here. Tiptap belongs to the structured-content editor (#6517, §6.2).
- Scoped to the `public` site only, for the `admin` role (§4.5): the selected jurisdiction's
  `(jurisdiction, language, public)` rows. The `partners` scope is split out, because the global
  `(null, language, partners)` rows it would edit have no admin endpoint. Every admin route on the
  translation controller is jurisdiction-scoped, and the only route reading the global layer is the
  public `GET /translations`. Editing it needs new endpoints, and nothing renders those rows until
  #6515 anyway.
- Editor visibility is gated on the per-jurisdiction `enableDbDrivenContent` flag, which controls
  whether a jurisdiction is DB-driven. It does not gate who may edit.
- Depends on: #6510 (translation admin CRUD); #6512 (permissions).

## Acceptance Criteria
- An admin can find a key by searching its value or path, edit it, and save; results show key path,
  value, and site label.
- A stale non-English key shows an "English changed since translated" flag; re-saving clears it.
- A save that drops a `%{token}` or `smart_count` structure is blocked with the offending key
  flagged.
- A per-key conflict presents keep-mine / take-theirs; non-conflicting edits are preserved.
[...]

Issue #6512, opened 2026-07-15
Permission model and flag gating


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
The new translation and content endpoints need a settled permission model, and the staged rollout
needs a per-jurisdiction flag so a jurisdiction can be verified before its content goes live.

## Technical Context
- TDD: §4.5.
- Editing `translation` and `jurisdictionContent` is limited to the `admin` role, which has access
  to every jurisdiction in the system and is normally held by the jurisdiction's own staff.
  `jurisdictionAdmin` only distinguishes anything when several jurisdictions share one deployment,
  and each jurisdiction runs its own fork and database today. Decided in the #6542 review.
- No new policy lines are needed. `p, admin, translation, true, .*` and
  `p, admin, jurisdictionContent, true, .*` already cover both resources, including the global
  (`null` jurisdiction) Partners overrides, since the admin row carries no jurisdiction constraint.
- Two policy lines are deliberately not added, both recorded in TDD §4.5:
  `p, anonymous, translation, true, read` would grant read on the type to every signed-in user via
  `g, user, anonymous`, and `p, jurisdictionAdmin, translation, true, (read|create|update)` would
  allow any jurisdictional admin against any jurisdiction, because a `true` sub-rule cannot be
  narrowed by a per-user object rule.
- Add the per-jurisdiction `enableDbDrivenContent` flag to `FeatureFlagEnum` and `featureFlagMap`.
  It gates whether a jurisdiction reads its content from the database and whether the editors are
  shown for it. It does not gate permissions, so no flag check belongs in `permission.service.ts`.
  The first consumer is #6516, which hides the translation editor behind it so a jurisdiction does
  not see the editor before its content is ready. #6520 covers the public render path and the
  staging-to-production promotion.
- Depends on: #6508 (data model).

## Acceptance Criteria
- Anonymous can read the public translation and content endpoints.
- Editing both resources is limited to `admin`. `jurisdictionAdmin`, `supportAdmin`,
  `limitedJurisdictionAdmin`, a signed-in non-admin, and an anonymous request are all denied, on
  their own jurisdiction and on any other.
- The global Partners overrides are editable only by `admin`.
- `enableDbDrivenContent` exists as a per-jurisdiction flag, defaulting off, ready for #6516 to
  gate the editor on.

## QA Notes
Backend-only. Test each role against each resource. The flag has no observable effect until #6516
[...]

PR #6557, opened 2026-07-31
feat: adds enableDbDrivenContent feature flag

seanmalbert:
This PR addresses #6512

## Description

Adds the `enableDbDrivenContent` feature flag and pins translation and structured-content editing to the `admin` role with a test. Backend only. No schema change and no behavior change: the flag has no consumer yet.

> Stacked on #6511 (jurisdiction content CRUD); this branch merges that base in.

- **New flag.** `enableDbDrivenContent` is added to `FeatureFlagEnum` and `featureFlagMap`, with the generated docs and client. It is the per-jurisdiction rollout switch: when on, a jurisdiction reads its translations and structured content from the database instead of the bundled override files. The first consumer is #6516, which hides the translation editor behind it so a jurisdiction does not see the editor before its content is ready. The public render path and the promotion workflow are #6520.
- **Permissions unchanged.** `p, admin, translation, true, .*` and `p, admin, jurisdictionContent, true, .*` already limit editing to `admin`. A new unit test pins that down: a jurisdictional admin is denied on its own jurisdiction and on the global (null-jurisdiction) Partners scope, and an admin is allowed on the global scope.

The issue originally asked for jurisdictional-admin edit access gated on the flag. Editing stays with `admin` instead.

1. Unit (`cd api && yarn test`): `permission.service.spec.ts` gains a test that only an admin can write translations.
2. Integration (`yarn test:backend:new:e2e`): no new tests. The existing `403` cases in `translation.e2e-spec.ts` and `jurisdiction-content.e2e-spec.ts` already cover a jurisdictional admin, a signed-in non-admin, and an anonymous request. Those specs plus the ten `permission-tests` specs were run against a fresh database to confirm the flag changes nothing: 951 passing.
3. Manual: on the hidden `/admin` page in Partners, confirm `enableDbDrivenContent` appears in the flag list for a jurisdiction and can be toggled. Toggling it has no effect yet, which is expected until #6516 gates the editor on it.

Note: `yarn generate:client` also picks up 40 lines for the `applications/bulk-update/upload-url` endpoint. That drift is pre-existing; regenerating on the base branch with this branch's changes stashed produces the same lines.

## Author Checklist:

- [x] Added tests covering the changes (a unit test pinning translation writes to the admin role)
- [x] Ran `yarn generate:client` and/or created a migration when required (regenerated `backend-swagger.ts` and the feature-flag docs; no schema change, and flags are not auto-added to the database)

## Review Process:


shared-helpers/src/types/backend-swagger.ts
```diff
@@ -9475,6 +9497,24 @@ export interface ApplicationUpdateEmail {
...
+
+/** ApplicationBulkPresignedUrl */
+export interface ApplicationBulkPresignedUrl {
+  /**  */
+  key: string
+
+  /**  */
+  presignedUrl: string
+}
+
```
seanmalbert, line 9520: these are not from my changes

ludtkemorgan: There was a commit recently merged to main that accidentally didn't run the generate client command. Other open PRs also include this change so there is a chance that this will be resolved before you merge

ludtkemorgan (approved):

PR #6568, opened 2026-08-06
feature: partners translation editor

seanmalbert:
This PR addresses #6516

## Description

Adds the Partners translation editor at `/settings/translations`: a per-key editor over the ~1,492-key base merged with a jurisdiction's stored overrides, with fallback and staleness indicators, per-key conflict resolution, token validation, and a warning before a change removes a section. Limited to the `admin` role and hidden behind `enableDbDrivenContent`.

> Stacked on #6512 (permissions and flag gating); this branch merges that base in.

- **The editor.** One row per key, sorted, from the union of the bundled English base, the selected language's base, and the jurisdiction's override rows. Columns are the key path, the base value, the current value (edited inline), a status, and a revert action. Jurisdiction and language selectors scope the view; the jurisdiction selector is always shown, disabled when the admin has only one, since the jurisdiction being edited cannot be inferred otherwise. Search and paging are client-side, because every row is already in the browser.
- **Fallback and staleness.** Each row shows what the site renders today: the override when one exists, otherwise the language base falling back to English. Status is one of "Using base", "Overridden", "English changed", or "Edited". Reverting deletes the override so the key falls back again.
- **Conflicts.** Each edit captures the version its key held when first edited, and sends it as the per-key optimistic lock. A partial save returns a 409 naming only the keys someone else changed first; everything else in the batch is written. The dialog offers keep-mine or take-theirs per key, and a kept edit is re-locked against the version on the row now so the retry does not conflict again.
- **Validation.** An entered value is compared against its English source for the structure polyglot needs: a dropped `%{token}` or a dropped `||||` blocks the save with the offending keys named. The number of plural forms is not compared, since plural rules vary by language.
- **Section visibility.** A key with no base value renders only when an override supplies it, so emptying or reverting one takes its section off the site. Both paths ask for confirmation first. Emptying (hide) stays distinct from reverting (delete and fall back). Visibility keys are derived from the absence of a base value rather than listed, so template keys are covered.
- **Unsaved changes.** Edits live in component state until saved, so an in-app route change or a browser reload warns first.
- **Write-side validation (api).** `TranslationKeyEdit.value` gains a length cap and a validator refusing tags that execute or load by existing (`script`, `style`, `iframe`, `object`, `embed`, `link`, `meta`, `base`, `form`). Translation values reach the public site through a markdown renderer that parses raw HTML, so a stored `<script>` would run on every page rendering that key. The rule names refused tags rather than allowed ones because base values legitimately include `<a>`, `<strong>`, and `<span class>`, and several include a `<REGION>` placeholder an allowlist would reject. Checked against all 21 locale files: nothing currently stored is refused, and nothing exceeds the cap.

Two changes fall outside the issue:

- **Settings tab gating.** Every settings tab was shown based on its feature flag alone, while each page applied its own role redirect, so three tabs disagreed with the page they linked to. Translations requires `admin`, and Properties and Agencies both reject a limited jurisdictional admin, which the layout does show Settings to. `getVisibleSettingsTabs` now narrows each tab by role to match its page, and both the tab list and the tab count read from it. The Properties and Agencies mismatches are pre-existing; both are fixed here because they share one gate.
- **`seed-dev`.** Adds the unit rent type lookup rows, without which `unitFactoryMany` throws a `TypeError`. Its own commit (`22d764f`).

Scope deliberately left out: this edits the `public` site scope only. The `partners` scope is split out, since the global `(null, language, partners)` rows have no admin endpoint. Nothing on the public site reads these rows yet; that is #6520.

There is an open comment thread about moving this from the settings tab to the hidden `/admin` path so it reaches super admins only. That is unresolved and would be a page move plus removing the settings tab.

1. **Partners unit** (`yarn test:app:partners:unit`): 84 suites, 803 passing. New or extended here: `translationEditor.test.ts` (80), `translations.test.tsx` (18), `SettingsViewHelpers.test.ts` (9), `useUnsavedChangesWarning.test.tsx` (7), `TranslationConflictDialog.test.tsx` (7), `useRawTranslations.test.tsx` (6), `TranslationHideSectionDialog.test.tsx` (5).
2. **API unit** (`cd api && yarn test`): 63 suites, 1,208 passing, including 7 new tests for the markup validator.
3. **shared-helpers** (`yarn test:shared-helpers`): 22 suites, 191 passing, including the flattening helper.
4. **Manual**, as an admin with `enableDbDrivenContent` on for the jurisdiction:
   - Filter by a key path or a value, edit a cell, press Enter, and save. Confirm the status moves to "Edited" then "Overridden", and that the filter can stay applied throughout.
   - Empty a value for a key with no base and confirm the warning, then confirm reverting that same key warns separately.
   - Drop a `%{token}` from a value and confirm the save is blocked with the key named.
   - In two browser tabs, edit the same key in both and save the second. Confirm the conflict dialog names only that key, that other edits in the batch were written, and that keep-mine then saves successfully.
   - Switch jurisdiction and language and confirm the grid refetches. Switching is disabled while there are unsaved edits.
   - Navigate away with unsaved edits and confirm the warning.
   - Confirm a jurisdictional admin sees no Translations tab and is redirected if they visit the URL.

## Author Checklist:

- [x] Reviewed considering accessibility (the settings tabs render as navigation links; the filter, selectors, and dialogs are labelled, and the confirmation dialogs name the affected keys)
- [x] Added tests covering the changes (132 tests across seven partners suites, plus the API validator and the flattening helper)
- [x] Ran `yarn generate:client` and/or created a migration when required (no schema change; the added `@MaxLength` and custom validator do not appear in the generated client, matching the five DTOs that already use `@MaxLength`)

## Review Process:


ludtkemorgan (commented):
I haven't looked at all of the code yet, but wanted to give some initial UI feedback.
1. I think long term having it in the "settings" tab makes sense. But I'm worried giving all "admin" users this access right away could cause some issues. My recommendation would be to put it in the hidden "/admin" path. That is only accessible to super admins and it also is where we are putting things that change the actual site such as the feature flag management. That way we can continue to test this with a few of our jurisdictions before we hand it over to all admins. The most likely scenario is that jurisdictions won't be changing their translations much if at all post the initial launch so it doesn't need to be a super accessible page.
2. I like the table design. Just as a heads up we are planning to slowly migrate off of AgTable to a better table in the nearish future called [Tanstack table](https://tanstack.com/table/latest). Since we don't have that in place yet it's ok to use AgTable
3. If you filter to find a translation and then edit it the save button is disabled. This means you have to remove your filter before you can save which is not intuitive.

seanmalbert:
Thanks for the feedback you already provided, @ludtkemorgan . Item 3 is already addressed. Regarding item 1, I can move the page under /admin, or we can lock it down so only a super-admin can access it. The rest should be good to review now.

ludtkemorgan:
This is looking awesome. I'm super excited for it. I have asked copilot to do an initial review. All of its comments are valid, but none of them seem that important to me so I'll let @seanmalbert decide if they should be addressed or not.

Here are some issues I have found:
1. After you have entered into at least one field to edit the "actions" column gets misaligned. I don't know if that is an AgGrid issue or something we can control. Screenshot for reference.
image

2. The filtering functionality fails if you haven't left saved one of your changes. Video reference here for two different scenarios that cause the issue
https://github.com/user-attachments/assets/fd7f2df8-f32d-415b-a36a-2b04b2add218

3. All jurisdictions are appearing in the dropdown even if they don't have the feature flag turned on. I don't think we will run into a scenario where we have a partner site and all jurisdictions don't utilize the functionality. But just in case we should get ahead of it and limit the jurisdiction dropdown to just show the jurisdictions that have it enabled.

Here are some Ideas. None of which are important enough to hold up this pr or the feature for that matter:
1. I don't know if this would cause more harm than what it is trying to solve. Since the "revert" button immediately deletes the override it is relatively easy to accidentally click it and not realize. Would we want a confirmation before it goes through?
2. This would make the table probably too big so probably not the best idea. It would be nice to be able to see the english translation when you are on a different language. That way you can get more context of what the override is for. Now that I wrote that out I realized that we would most likely be referencing the key instead of the english translation, so it's probably not necessary.

My next step will be going over all of the code so expect another round of review either today or tomorrow

ludtkemorgan (commented):
One more bug. The next button doesn't seem to be working. After the new data loads it jumps back to the first page

https://github.com/user-attachments/assets/f565b5d5-5671-420c-8fb7-24ecf00bf3b3

It also happens if you use the "jump to" functionality

sites/partners/src/components/settings/TranslationHideSectionDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+  onClose: () => void
+  onConfirm: () => void
+}
+
+/**
+ * Confirms an edit that removes a section rather than changing its wording.
+ *
+ * These keys have no base value, so nothing falls back once the override is emptied or reverted.
+ */
+export const TranslationHideSectionDialog = ({
```
ludtkemorgan: Question: the only way I was able to test this was to add a new key via the API. Is there a plan to add a "add row" functionality to the UI?

sites/partners/src/lib/translationEditor.ts
```diff
@@ -0,0 +1,271 @@
...
+ */
+export const validateValue = (
+  row: TranslationEditorRow,
+  value: string
+): TranslationIssue | null => {
+  if (row.englishValue === null) return null
+
+  const englishTokens = tokensIn(row.englishValue)
+  const valueTokens = tokensIn(value)
+  const missingTokens = [...englishTokens].filter((token) => !valueTokens.has(token))
```
ludtkemorgan, line 233: issue: I think there might be use cases where a jurisdiction wants to not include a token in the overriden value.

For example there is a theoretical base string of `You have applied to %{applications} applications`. A jurisdiction doesn't want that for some reason and instead they want it to say `You have applied to more than 1 application`. With this validation in place it would allow it.

Another example: Base string = `For more information call %{phoneNumber} or email us at %{emailAddress}`. The jurisdiction doesn't want people calling via a phone number so they want it to be changed to `For more information email us at %{emailAddress}`

Another example: A jurisdiction doesn't want to display a string on the site at all but it contains a dynamic value. The current UI prevents that from saving if you delete the string

ludtkemorgan (changes requested):
I did not find any issues directly with the code, just found one more additional bug (it's working as intended, but I think we might need to not have that functionality). 

I'm gonna ask the team about if we should keep this in the settings tab or move it to the admin panel. I'll report back once I get some more feed back.

Lastly, I think some more frontend integration tests for this functionality would be useful. I'm ok if you want to hold off on those until you get an answer on if this functionality should live in the settings or admin panels.

sites/partners/src/pages/settings/translations.tsx
```diff
@@ -0,0 +1,530 @@
...
+  const settingsTabsFeatureFlags = {
+    enablePreferences: atLeastOneJurisdictionEnablesPreferences,
+    enableProperties,
+    enableAgencies,
+    enableTranslations,
+  }
+
+  if (!enableTranslations || !profile?.userRoles?.isAdmin) {
+    void router.push("/unauthorized")
+  }
```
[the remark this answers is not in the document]
seanmalbert: The redirect can't fire before the profile loads: RequireLogin returns null until it's present, and this route isn't in skipLoginRoutes.

Added the early return. It now skips the fetch and returns before rendering.

sites/partners/src/pages/settings/translations.tsx
```diff
@@ -0,0 +1,530 @@
...
+          data.editedValue !== null
+            ? `${styles["editable-cell"]} ${styles["edited-cell"]}`
+            : styles["editable-cell"],
+        cellEditorSelector: ({ value }: { value: string }) => {
+          const text = value ?? ""
+
+          return text.length > INLINE_EDITOR_MAX_CHARACTERS || text.includes("\n")
+            ? { component: "agLargeTextCellEditor", params: { maxLength: 5000 } }
+            : { component: "agTextCellEditor" }
+        },
```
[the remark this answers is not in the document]
seanmalbert: updated to enforce the same character limit

api/src/decorators/no-executable-markup.decorator.ts
```diff
@@ -0,0 +1,63 @@
...
+  'script',
+  'style',
+  'iframe',
+  'object',
+  'embed',
+  'link',
+  'meta',
+  'base',
+  'form',
+];
```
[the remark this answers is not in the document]
seanmalbert: added img and svg to the list

seanmalbert:
> This is looking awesome. I'm super excited for it. I have asked copilot to do an initial review. All of its comments are valid, but none of them seem that important to me so I'll let @seanmalbert decide if they should be addressed or not.
> 
> Here are some issues I have found:
> 
> 1. After you have entered into at least one field to edit the "actions" column gets misaligned. I don't know if that is an AgGrid issue or something we can control. Screenshot for reference.
> 
> image
> 2. The filtering functionality fails if you haven't left saved one of your changes. Video reference here for two different scenarios that cause the issue
>    https://github.com/user-attachments/assets/fd7f2df8-f32d-415b-a36a-2b04b2add218
> 3. All jurisdictions are appearing in the dropdown even if they don't have the feature flag turned on. I don't think we will run into a scenario where we have a partner site and all jurisdictions don't utilize the functionality. But just in case we should get ahead of it and limit the jurisdiction dropdown to just show the jurisdictions that have it enabled.
> 
> Here are some Ideas. None of which are important enough to hold up this pr or the feature for that matter:
> 
> 1. I don't know if this would cause more harm than what it is trying to solve. Since the "revert" button immediately deletes the override it is relatively easy to accidentally click it and not realize. Would we want a confirmation before it goes through?
> 2. This would make the table probably too big so probably not the best idea. It would be nice to be able to see the english translation when you are on a different language. That way you can get more context of what the override is for. Now that I wrote that out I realized that we would most likely be referencing the key instead of the english translation, so it's probably not necessary.
> 
> My next step will be going over all of the code so expect another round of review either today or tomorrow

On the issues:
1. I have not been able to reproduce this one. I tested in chrome, firefox and safari. Are you still seeing this?
2. updated so that, if a user leaves a cell open and tries to filter or paginate, it commits as edited
3. updated so that only the jurisdictions that have the flag enabled show

On your ideas:
I think they're both good ideas, and I would like to do as follow-ups (if you still want to do them) since this is a pretty big PR already.

seanmalbert:
> One more bug. The next button doesn't seem to be working. After the new data loads it jumps back to the first page
> 
> It also happens if you use the "jump to" functionality

This is now fixed.

sites/partners/src/lib/translationEditor.ts
seanmalbert: Thanks for the good examples. I updated, so that users will get a warning if they remove a token or plural form from the cell. They can cancel or confirm the change.

sites/partners/src/components/settings/TranslationHideSectionDialog.tsx
seanmalbert: ~~This is one I missed an issue for. I'll create it, and likely do that one next.~~
I'm not sure this is needed. If users can add a key with no base value, there is no way for that to render, unless the frontend code already has it. This only becomes useful if custom pages can be created that would use them.

ludtkemorgan: Oh fair point. Yeah, you are right. We should still keep the dialog in case we run into that case in the future though

package.json
```diff
@@ -96,5 +96,6 @@
     "react": "19.2.3",
     "react-dom": "19.2.3",
     "@fortawesome/react-fontawesome": "3.1.0"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
ludtkemorgan: question: What is this needed for?

ludtkemorgan:
The misalignment on the "actions" column is now only a few pixels so I don't think we need to worry about it. 
image

ludtkemorgan (approved):
Thanks for making those updates. I agree that if we want to make those suggested enhancements they can be in follow-up PRs.

I'm still leaning towards putting this functionality in the /admin page as we don't know if we will be giving this to all admins. However, since this is behind a feature flag I'm ok with us merging this and moving it before we turn it on in a production environment. 

We can also poll the team on Thursday during your demo

package.json
seanmalbert: this was added every time I ran `yarn dev` by [Corepack](https://yarnpkg.com/corepack). I added `COREPACK_ENABLE_AUTO_PIN=0` to my ~/.zshenv, so it doesn't get added again and removed.

Issue #6550, opened 2026-07-29
Red/Yellow - Setup backend infrastructure


ludtkemorgan:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
https://docs.google.com/document/d/1vGctGuebk5LK6otAhhFBY6BobX22ppb2uBKF2BH4dtk/edit?tab=t.0#heading=h.ma9413cpihrs - tickets 0,1,and 2

## Technical Context
This ticket encompasses the backend work required to support the feature: establishing the feature toggle, creating the core type/registry structure, and implementing the persistence strategy for enabling rules on a per-jurisdiction basis. Please refer to the Technical Design Document for more information on the overall technical architecture.

## Acceptance Criteria
- [ ] Create a feature flag for the work, ensuring it is disabled in production.
- [ ] Create stopLights/stopLightRules.ts containing the StopLightColor type, StopLightRule interface, and an initialized empty stopLightRules array (check TDD for exact structure).
- [ ] Implement an example red and yellow rule to verify the registry structure.
- [ ] Add an enabledStopLightRuleKeys column (String[]) to the Jurisdictions table, including the necessary database migration and DTO update.
- [ ] Thread the enabledStopLightRuleKeys configuration onto conductor.config, following the existing pattern used for raceEthnicityConfiguration and visibleSpokenLanguages.
- [ ] Add testing to ensure that enabled rules are correctly threaded through to the conductor, and disabled rules are not threaded through

ludtkemorgan:
@chadbrokaw to fill in the details

PR #6575, opened 2026-08-12
Feat/red yellow setup backend infrastructure

KrissDrawing:
This PR addresses #6550

## Description

Adds Red and yellow lights backend work.
There is this task in task AC `Implement an example red and yellow rule to verify the registry structure.` but im not sure if im supposed to add that to `stopLightRules.ts` 
```
// Example
const seniorBuildingAgeExample: StopLightRule = {
  key: "seniorBuildingMinimumAge",
  step: "primaryApplicantName",
  light: "red",
  evaluate: (application, listing) => {
    const minimumAge = getSeniorBuildingMinimumAge(listing) // helper
    if (!minimumAge) return false
    return getApplicantAge(application) < minimumAge
  },
  heading: "stopLights.seniorBuildingMinimumAge.heading",
  body: "stopLights.seniorBuildingMinimumAge.body",
  editFieldAnchor: "applicant.dateOfBirth",
}
```
I then would not know what to really do with that at this moment, so for now i ignored it.

Run migration. 
You can add `console.log("conductor.config", conductor.config)` to `chose-language.tsx` and test if it will show `enabledStopLightRuleKeys` with flag on and with data there, in other cases it should return empty array

## Author Checklist:

## Review Process:


sites/public/src/lib/applications/stopLights/stopLightRules.ts
```diff
@@ -0,0 +1,15 @@
+import { Application, Listing } from "@bloom-housing/shared-helpers/src/types/backend-swagger"
+
+export type StopLightColor = "red" | "yellow"
+
+export interface StopLightRule {
```
ludtkemorgan, line 6: suggestion: you called this out in the PR description. I don't think it's a bad idea to have an "example" yellow or red light in this file. At minimum it could help with testing. And we could always remove it when we add actual rules

ludtkemorgan (approved):
Looks good!

Issue #6515, opened 2026-07-15
Partners site hydration


seanmalbert:
- [ ] General
- [ ] Backend
- [ ] Public
- [x] Partners

## Background / Issue
The Partners portal renders the same global overrides for all users regardless of jurisdiction
(Partners is jurisdiction-independent), sourced from the DB rather than the bundled override file.

## Technical Context
- TDD: §5.5.
- Load the global `(null, language, partners)` overrides into the Partners polyglot via the
  jurisdiction-independent read (`GET /translations?site=partners&language=`), layered over the
  bundled base. No per-jurisdiction resolution on the Partners side.
- Depends on: #6509

## Acceptance Criteria
- The Partners portal applies the global Partners overrides for all users, layered over the bundled
  base.
- No per-jurisdiction resolution is attempted on the Partners side.

## QA Notes
Partners portal. Confirm the portal reflects the global Partners overrides and that users in
different jurisdictions see the same overrides.

PR #6583, opened 2026-08-12
feat: Partners site hydration

seanmalbert:
This PR addresses #6515

## Description

The Partners portal now reads the global `(jurisdictionId = null, language, partners)` translation overrides from the API and layers them over its bundled base.

> Stacked on #6590 (global Partners translation scope), which is itself stacked on #6568 (partners translation editor).

- **Layering.** `applyTranslations(locale, storedOverrides)` in `sites/partners/src/lib/translations.ts` loads the polyglot layers with every English layer below every layer for the selected locale, so a string overridden in English does not replace a translation that already exists for that locale.
- **Fetching.** `usePartnersOverrides(locale)` in `sites/partners/src/lib/partnersOverrides.ts` requests `/api/adapter/translations?language=<locale>` on mount and whenever the locale changes, aborting a request in flight when it does. It goes through the existing adapter route, which adds the `passkey` header server-side, the same way every other Partners request satisfies `ApiKeyGuard`. It uses `fetch` rather than SWR because `_app` calls it from its own body, which sits outside the `SWRConfig` that `_app` renders.
- **First paint waits.** The portal already renders nothing until mount (`hasMounted && <Component/>`); that gate becomes `hasMounted && settled &&`. The overrides include `nav.siteTitlePartners`, which appears on every page, so applying them after paint would change the header in view. A failed or timed-out request settles too, so the portal comes up on its bundled values.
- **No feature flag.** `enableDbDrivenContent` is per jurisdiction and Partners resolves no jurisdiction, so there is nothing to read it from. Before any rows exist the endpoint returns `{}`, which is already a no-op.
- **The bundled override file stays.** `sites/partners/page_content/overrides/general.json` (17 keys) sits below the stored overrides rather than being removed. #6519 migrates its contents into the database; until that has run in an environment, this file is what keeps those strings correct. Removing it is a cleanup for after #6519. `overrideTranslations.en` also combines `page_content/locales/general.json`, which is the Partners base additions rather than overrides, so it stays regardless.

### Also in this PR

Fixes from review:

- `GET /translations` returns one map per language rather than merging English underneath, so an English override no longer replaces a translation the locale already has. The editor derives its stale flag for a key overridden in English but not in the language being edited.
- The adapter returns 502 when the backend does not answer. It previously returned without sending a response, so the request hung until the socket timed out and the portal stayed blank.
- The overrides request times out after 5 seconds, rather than holding the first paint for as long as the API takes to answer.
- Switching language no longer hides the page while the new overrides load. `settled` was reset to false on every request, so the page unmounted and came back, discarding whatever the admin had entered.
- `usePartnersOverrides` moved to its own module, so `_app` no longer pulls swr, qs, and dayjs into the chunk every page loads.
- The translation value validator rejects `style` attributes, inline event handlers, `javascript:` urls, and markdown images.

1. **Partners unit** (`yarn test:app:partners:unit`)
2. **API unit** (`cd api && yarn test`) and **API e2e** (`yarn test:backend:new:e2e`)
3. **Manual.** #6590 adds the editor scope that writes these rows.

As an admin, open `/settings/translations`, switch the Site selector to Partners, and edit `nav.siteTitlePartners`. Then: reload the portal and confirm the header title changes; confirm it is correct on first paint rather than switching a moment later; revert the key and confirm the bundled value returns. Sign in as users in different jurisdictions and confirm they see the same override. With the API stopped, confirm the portal still loads on bundled values.

## Author Checklist:

- [x] Reviewed considering accessibility (no new UI; the change is which strings are loaded and when the first paint happens)
- [x] Added tests covering the changes (15 tests across two new suites)
- [x] Ran `yarn generate:client` and/or created a migration when required (N/A: frontend only, no schema or API change)

## Review Process:


sites/partners/src/pages/api/adapter/[...backendUrl].ts
```diff
@@ -5,12 +5,6 @@ import { wrapper } from "axios-cookiejar-support"
 import { getConfigs } from "@bloom-housing/shared-helpers/src/types/backend-swagger"
 import { maskAxiosResponse } from "@bloom-housing/shared-helpers"
 
-/*
```
Comment (ludtkemorgan) [inline sites/partners/src/pages/api/adapter/[...backendUrl].ts:8]: question: is there a reason you removed this comment. I think it's useful information

sites/partners/src/lib/partnersOverrides.ts
```diff
@@ -0,0 +1,41 @@
...
+      }
+      setSettled(true)
+    }
+
+    void load()
+    return () => {
+      superseded = true
+      clearTimeout(timeout)
+      controller.abort()
+    }
```
ludtkemorgan, line 37: question: could you explain what this superseded functionality is for? Since this partnersOverride is called at the top level _app.tsx I never saw the cleanup function get called

ludtkemorgan (approved):
This is awesome. Can you tell which values I changed for testing?

Image

sites/partners/src/pages/api/adapter/[...backendUrl].ts
Comment (seanmalbert) [reply sites/partners/src/pages/api/adapter/[...backendUrl].ts:8]: This was a mistake. I was cleaning up comments in other files, and I got a little greedy. I'll restore this and the other comments here before merging.

sites/partners/src/lib/partnersOverrides.ts
seanmalbert: This doesn't come into play in the partners portal, because there's only one locale, so we could take this out. But the cleanup runs on every locale change, not only on unmount. If a locale/language could be changed twice in quick succession that would leave two requests in flight, and whichever responds last would win. `superseded` and the `AbortController` ensure that a stale response is dropped rather than applied over a newer one.

To be thorough:
- `controller.abort()` stops the request that is no longer wanted.
- `superseded = true` covers the window where the response has already arrived and is being parsed
- `clearTimeout(timeout)` drops the five-second abort timer

It doesn't hurt to leave it in, but let me know if you want it pulled out until Partners supports language changing.

ludtkemorgan: Oh gotcha! That makes sense. Thanks for the explanation

sites/partners/src/pages/api/adapter/[...backendUrl].ts
Comment (seanmalbert) [reply sites/partners/src/pages/api/adapter/[...backendUrl].ts:8]: restored

Issue #6584, opened 2026-08-12
Global Partners translation scope (endpoints and editor)


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [x] Partners

## Background / Issue
The translation editor covers the `public` scope only. The `partners` scope, which is the global
`(jurisdictionId = null, language, partners)` rows the Partners portal renders, has no way to be
edited: every admin route on the translation controller is jurisdiction-scoped, and the editor
hardcodes `SiteEnum.public`. Those rows are the database replacement for
`sites/partners/page_content/overrides/general.json`, so until this lands, changing a Partners
string still means an engineer editing a file, which is what Phase 1 set out to remove.

#6516 split this scope out for two reasons: no endpoint could write the rows, and nothing rendered
them. #6515 removed the second reason.

## Technical Context
- TDD: §6.1 (editing target and site scope), §5.5.
- **Endpoints.** Add admin routes for the global scope alongside the jurisdiction-scoped ones added
  in #6510, reusing the per-key optimistic locking, conflict reporting, and value validation rather
  than adding a second write path. The site is fixed to `partners` rather than taken from the path,
  so the global scope cannot write `public` rows or the `(null, null)` base rows, which nothing
  needs:
  - `GET /translations/partners/raw/:language`
  - `PUT /translations/partners/raw/:language`
  - `DELETE /translations/partners/raw/:language/:key`

  `authorizeJurisdiction` (`api/src/services/translation.service.ts`) called `resolveJurisdictionId`
  after the permission check, and the global scope has no jurisdiction to resolve. It is split so
  the permission check always runs and the existence check runs only when there is an id, which lets
  the three admin service methods take `jurisdictionId: string | null` and serve both scopes.
- **Permissions.** No policy change is needed. `p, admin, translation, true, .*` has the literal
  `true` as its rule, so the matcher never reads the request object and `admin` is permitted with or
  without a jurisdiction. No role below `admin` has a `translation` policy row, and
  `addUserPermissions` grants none. What was missing was the assertion:
  `permission.service.spec.ts` passed `jurisdictionId: 'juris id'` in every content case and
  asserted nothing about the null case. It now runs its role list against the global scope as well.
  (The #6512 PR description claims this was already covered; it was not.)
- **Editor.** Add a site scope selector to the translations settings page. Per the TDD the
  `partners` scope is labelled as applying to every jurisdiction. The jurisdiction selector is
[...]

PR #6590, opened 2026-08-17
feat: global partners translation scope

seanmalbert:
This PR addresses #6584

## Description

The global `(jurisdictionId = null, language, partners)` translation rows drive the Partners portal's strings for all jurisdictions.

Stacked on #6568 (partners translation editor); this branch continues that chain.

**Endpoints.** Three routes with the site fixed to `partners`.

| Route | operationId |
| --- | --- |
| `GET /translations/partners/raw/:language` | `getRawPartnersTranslations` |
| `PUT /translations/partners/raw/:language` | `updateRawPartnersTranslations` |
| `DELETE /translations/partners/raw/:language/:key` | `deleteRawPartnersTranslation` |

They call the same `getRawOverrides`, `updateOverrides`, and `deleteOverride` the jurisdiction routes use, with a null jurisdiction.

**Permissions.** No policy change was needed. `p, admin, translation, true, .*` has `true` as its rule. What was missing was the assertion, so the permission spec now runs its existing role list against the global scope as well.

**Editor.** A site scope selector, with the Partners option labelled as applying to all jurisdictions.

### Also in this PR

**The public scope was comparing against the wrong base.** The public site layers `sites/public/page_content/locale_overrides` over the shared locale file, the same way Partners layers its own.

1. **API unit** (`cd api && yarn test`)
2. **API e2e** (`yarn test:backend:new:e2e`)
3. **Partners unit** (`yarn test:app:partners:unit`)
4. **Manual, as an admin, on a jurisdiction with `enableDbDrivenContent` on.**
   1. Open `/settings/translations` and switch the scope to Partners. The jurisdiction selector goes away and the row list changes to the Partners base.
   2. Edit `nav.siteTitlePartners`, save, and confirm the row reports as overridden.
   3. `GET /translations?language=en` returns the new value.
   4. Sign in as users in different jurisdictions and confirm they see the same override.
   5. Revert the key and confirm the row falls back to the bundled value.
   6. Sign in as a jurisdictional admin and confirm the API refuses a global write, not only the interface.
   7. In the `public` scope, filter for `pageDescription.faq` and confirm the base value reads "…affordable housing in Bloomington." rather than the shared wording.

## Author Checklist:

- [x] Reviewed considering accessibility (the new control is a labelled `Select` alongside the two existing ones)
- [x] Added tests covering the changes (53 across the API and Partners suites)
- [x] Ran `yarn generate:client` (three new methods on `TranslationsService`; no schema change, so no migration)

## Review Process:


ludtkemorgan (commented):
I couldn't find any issues with this. I was able to reproduce that row alignment issues. It only happens if you enter into the bottom row. 
Image
However, I think this is an AgGrid problem and since we are moving away from that we don't need to fix it

sites/partners/src/lib/publicTranslations.ts
```diff
@@ -0,0 +1,33 @@
+import generalOverrides from "../../../public/page_content/locale_overrides/general.json"
```
ludtkemorgan, line 1: question: This means the partner site always has to stay in the same repo as the public site without some changes on where these files live. We don't have a plan to split them apart like we did at one point so I don't think this will ever be an issue. However, my question is will this have any more consideration when building the docker images?

Dockerfile.sites.partners
```diff
@@ -21,6 +21,9 @@ RUN yarn install --frozen-lockfile
 COPY --chown=2002:2002 tsconfig.json /app/tsconfig.json
 COPY --chown=2002:2002 sites/partners .
 COPY --chown=2002:2002 sites/middleware.ts /app/sites/partners/src/middleware.ts
+# The translation editor compares its public scope against what the public site renders, so it
+# reads that site's locale overrides. src/lib/publicTranslations.ts is the only thing that needs them.
+COPY --chown=2002:2002 sites/public/page_content/locale_overrides /app/sites/public/page_content/locale_overrides
```
ludtkemorgan, line 26: Ope, this answers my question

ludtkemorgan (approved):

Issue #6594, opened 2026-08-18
Overrides on public pages that have no data path


seanmalbert:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
#6514 sends stored overrides to the public site as page props. Each page reads them in
`getStaticProps` and `_app` applies them. This only works for pages that have a `getStaticProps` or
`getServerSideProps` function.

40 pages under `sites/public/src/pages` have neither, so they get no props and keep rendering the
strings bundled at build time. Those 40 pages use 285 different translation keys.

An admin can edit any of those 285 keys and save, but the pages will keep showing the old one.

Of the 40 pages, 29 are the application flow under `applications/**` and 2 are under `account/**`.
The other 9 are `disclaimer.tsx`, `privacy.tsx`, `404.tsx`, `create-account.tsx`,
`create-advocate-account-confirmation.tsx`, `forgot-password.tsx`, `reset-password.tsx`,
`verify.tsx`, and `redirect.tsx`.

## Technical Context
There are two ways to fix this. Either can be used on its own, or one can be used for some pages and the other for the rest.
  1. **Add `getStaticProps` to each page**, returning the overrides the same way #6514 does for the
     pages that already have it. The override is then in the HTML the server sends, so it is correct
     on the first render and search engines see it. The cost is 40 files, and the application flow
     pages become statically generated when they are currently rendered in the browser.
  2. **Fetch the overrides in the browser from `_app`** when a page has no props, the same way the
     Partners portal does in `usePartnersOverrides`. This is one change rather than 40. The page
     renders its bundled string once and then replaces it, and search engines see the bundled
     string.

A reasonable split is option 1 for the 9 standalone pages, which can be linked to directly, and
option 2 for the application and account flows, which a user only reaches after signing in or
starting an application. The editor could mark a key that renders only on one of these pages.

Depends on: #6514 (adds the props path and the layering this extends).

## Acceptance Criteria
- A stored override changes what renders on every public page that uses the key, or the pages where
  it does not are listed and the reason is recorded.
- A page still renders its bundled strings when the override fetch fails.
- Pages that search engines index have the override in the HTML the server sends, rather than
  applying it after the page loads.

## QA Notes
Public site. Edit a key that appears in the application flow and a key that appears on
`disclaimer.tsx`. Open both pages and confirm each shows the new value. Stop the API and confirm
both pages still show their bundled strings.

PR #6609, opened 2026-08-18
feat: public site hydration

seanmalbert:
This PR addresses #6514

## Description

This updates the public site to fetch the overrides from the translation endpoint, and layers them over the bundled base.

Stacked on #6583 (Partners site hydration), which introduced the layered response shape this uses.

**Layering.** `applyTranslations(locale, publicOverrides)` in `sites/public/src/lib/translations.ts` puts every English layer below every layer for the selected locale, so a string overridden in English does not replace a translation that already exists for that locale. This moves one existing key, `advocateAccount.resendLinkSuccess`, which currently shows English in all nine non-English locales; its bundled English override is identical to the shared base.

**Fetching.** `fetchPublicOverrides(language)` in `sites/public/src/lib/hooks.ts` reads the jurisdiction's overrides by name.

**Pages.** All 20 pages with a data function pass the overrides through. `listing/[id].tsx` is the exception: it renders only a `meta` refresh tag. The 13 `getStaticProps` pages also set `revalidate`, where 3 did before.

**Build request volume.** A build fetches the overrides once per page per locale. That came to 120 requests, 12 generated pages across 10 locales, and the API allows 100 an hour. A build would therefore stop receiving overrides partway through and finish the remaining pages on bundled strings. The helper now fetches once per language while `NEXT_PHASE` is set to the build phase, which brought this build down to 20 requests. Regeneration does not use that cache, so a page rebuilt later still picks up any edit made since the build. The request count is the number of build workers times the number of locales, so a machine with more cores could still pass 100. Raising the limit for this endpoint, or exempting it from the throttle, would prevent that.

### Not included

- The `jurisdiction_*_content.tsx` stubs.
- The `res.revalidate()` route, its trigger from the Partners save, and the CDN purge.
- Overrides on the 40 pages with no data function, tracked in #6594.
- Removing the bundled `locale_overrides/*.json`, handled in #6519.

1. **Public unit** (`yarn test:app:public:unit`)
2. **Build** (`cd sites/public && yarn build`), then confirm `revalidate` is set on the prerendered routes in `.next/prerender-manifest.json`.
3. **Manual, with the API running.**
   1. In the Partners editor, public scope, edit a key that appears on `/faq`.
   2. Wait for `revalidate`, and confirm the page shows the new value.
   3. View source and confirm the value is in the HTML the server sends, not applied after the page loads.
   4. Switch to Spanish and confirm the page keeps its Spanish translation, since only the English value was edited.
   5. Stop the API, rebuild, and confirm the site still renders its bundled strings.

## Author Checklist:

- [x] Reviewed considering accessibility (no UI change; the change is which strings load and when)
- [x] Ran `yarn generate:client` and/or created a migration when required (N/A: no API change)

## Review Process:


sites/public/__tests__/lib/hooks/fetchPublicOverrides.test.ts
ludtkemorgan, line 1: suggestion: since this function is in hooks.ts rather than a separate typescript file would it make sense to put this in a subfolder called "hooks" `sites/public/__tests__/lib/hooks/fetchPublicOverrides.test.ts`

sites/public/src/lib/hooks.ts
```diff
@@ -343,6 +343,45 @@ export async function fetchJurisdictionByName(req?: any) {
...
 }
 
+const publicOverridesByLanguage = new Map<string, Record<string, Record<string, string>>>()
+
+export const OVERRIDES_TIMEOUT_MS = 5000
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function fetchPublicOverrides(language?: string, req?: any) {
+  const key = language ?? "en"
+  const duringBuild = process.env.NEXT_PHASE === "phase-production-build"
```
ludtkemorgan: suggestion: is it possible to use [next/constants](https://nextjs.org/docs/pages/api-reference/config/next-config-js#phase) here so that the enum is a type rather than a string?

sites/public/src/pages/account/applications/open/index.tsx
```diff
@@ -23,10 +23,14 @@ const OpenApplications = ({ jurisdiction }: { jurisdiction: Jurisdiction }) => {
...
+export async function getStaticProps({ locale }: { locale?: string }) {
+  const [jurisdiction, publicOverrides] = await Promise.all([
+    fetchJurisdictionByName(),
+    fetchPublicOverrides(locale),
+  ])
 
   return {
-    props: { jurisdiction },
+    props: { jurisdiction, publicOverrides },
+    revalidate: Number(process.env.cacheRevalidate),
```
ludtkemorgan, line 34: thought: I think the cacheRevalidate value we have set on production sites is fairly aggressive. Something like 30 seconds. For most of these pages we could have it be much longer. Especially since it is ok if a content update change takes a few minutes to appear. That way we could avoid unnecessary backend calls when the data is rarely updating. One idea is to have two different cache revalidate values. One for dynamic content pages like the listings page and one for more static pages like this one.

I do remember you mentioning you wanting to looking into ISR. If that is the case then you can ignore my comment.

sites/public/src/pages/index.tsx
```diff
@@ -25,14 +29,18 @@ export default function HomePage(props: HomePageProps) {
   )
 }
 
-export const getStaticProps: GetStaticProps = async () => {
-  const underConstructionListings = await fetchLimitedUnderConstructionListings(undefined, 3)
-  const jurisdiction = await fetchJurisdictionByName()
+export const getStaticProps: GetStaticProps = async ({ locale }) => {
+  const [underConstructionListings, jurisdiction, publicOverrides] = await Promise.all([
+    fetchLimitedUnderConstructionListings(undefined, 3),
```
ludtkemorgan, line 38: nit: this is existing so if you'd prefer not to change it that's fine. I didn't realize that we are always fetching under construction listings. We only need this data if the `enableUnderConstructionHome` feature flag is turned on. So we could split this out so that we fetch the jurisdiction and public overrides. And then using the jurisdiction data we could see if the `enableUnderConstructionHome` feature flag is turned on before making the call to fetch the under construction listings.

for some more context this functionality is currently only need/used by Detroit (and that site is no more) so it's an unnecessary backend call for a vast majority of the time.

ludtkemorgan (approved):
This looks great. None of my comments are blocking so I'm approving it.

ludtkemorgan:
Oh also, I don't remember why the throttler default value is set for only 100 calls in an hour. That was set years ago. It seems like we should allow more than that and I'm surprised we haven't run into more issues around that. Maybe @YazeedLoonat remembers how we chose that for the .env template. Locally I have it set to 100 calls in a 3 minute period.

The throttler has a lot more configuration than what we are currently doing. We could have multiple rules as well as skipping it for certain endpoint https://docs.nestjs.com/security/rate-limiting#customization. So we could skip it for the translation override fetching endpoint if we are concerned about builds

sites/public/__tests__/lib/hooks/fetchPublicOverrides.test.ts
seanmalbert: Agreed, moved to `sites/public/__tests__/lib/hooks/fetchPublicOverrides.test.ts`

sites/public/src/lib/hooks.ts
seanmalbert: nice call. Now using `PHASE_PRODUCTION_BUILD` from next/constants

sites/public/src/pages/account/applications/open/index.tsx
seanmalbert: I agree that 30 seconds is really short. Here's [an issue](https://github.com/bloom-housing/bloom/issues/6634) to introduce on-demand revalidation. Once that's in I think we can have a much longer revalidation window everywhere, instead of tuning different paths.

sites/public/src/pages/index.tsx
seanmalbert: I think this should be done as a separate issue. You want me to file it?

seanmalbert: Filed at https://github.com/bloom-housing/bloom/issues/6640

Issue #6428, opened 2026-06-16
Bulk CSV Upload - Polling and dialog states


emilyjablonski:
This ticket will wire up job status polling and implement the remaining dialog states for the bulk upload modal. We can assume the upload and validation tickets are complete.

Polling
- On successful upload when a jobId is returned, we can begin polling `GET /background-jobs?listingId=` every 30 seconds
- We also want to call this on page load and when the dialog opens
- When the job reaches `completed` or `failed` then stop polling
- The response to this will tell us if the active job is for this listing or not

<hr />

Dialog states

### **In-progress (there's an active job and it is for this listing)**
Bulk update in progress

Your file is being processed. This may take up to an hour for many applications.

[progress indicator: X of Y records processed with a spinner]

You'll receive an email when the job completes. You can close this dialog and processing will continue in the background.

### **Completed**
Bulk update complete

Successfully updated [N] application records.

### **Failed**
Bulk update failed

[errorMessage from job]

If you are unable to correct the issue, contact the support team.

### **Blocked (only if globallyLocked: true and it's not for this listing)**

Bulk update unavailable

A bulk update is currently in progress for another listing. Only one bulk update can run at a time across all listings. A bulk update may take up to an hour for many applications.

Please try again later. This page will update automatically when the current job completes and this feature becomes available.

<hr />

=When we know that the listing has an active job, add a spinner to the end of the bulk update button.

Notes
- We can start with a simple useEffect and setInterval
- Consider NestJs [server sent events](https://docs.nestjs.com/techniques/server-sent-events)

PR #6615, opened 2026-08-19
Feat: Background job processing notifiction service

matzduniuk:
This PR addresses #6428 

## Description

* This tickets adds a new endpoint at `/applications/bulk-update/notifications` integrating an Server-Side-Event (SSE) based notification system for the client. 

> [!WARNING]
> As the processing job is not yet implemented this featured can be tested locally by adding the following job mocking function in the `application-bulk-upload.service.ts` service file:
>```
> private mockProcessUploadJob(jobId: string): void {
>    const MOCK_DURATION_IN_MS = 10_000;
>    const MOCK_TOTAL_RECORDS = 42;
>    const MOCK_ERROR_ROW = 7;
>
>    const timer = setTimeout(() => {
>      void (async () => {
>        const hasFailed = Math.random() < 0.5;
>        const completedAt = new Date();
>
>        const result = {
>          status: hasFailed
>            ? BackgroundJobStatusEnum.failed
>            : BackgroundJobStatusEnum.completed,
>          totalRecords: hasFailed ? null : MOCK_TOTAL_RECORDS,
>          errorMessage: hasFailed
>            ? 'Mocked failure while processing the uploaded file'
>            : null,
>          errorRow: hasFailed ? MOCK_ERROR_ROW : null,
>          completedAt,
>        };
>
>        try {
>          await this.prisma.backgroundJob.update({
>            where: { id: jobId },
>            data: result,
>          });
>        } catch (error) {
>          this.notifications$.next({
>            jobId,
>            status: BackgroundJobStatusEnum.failed,
>            errorMessage: `Mock processor could not update job with id: ${jobId}`,
>            completedAt: completedAt.toISOString(),
>          });
>          console.error('mockProcessUploadJob update failed', error);
>          return;
>        }
>
>        this.notifications$.next({
>          ...result,
>          jobId,
>          completedAt: completedAt.toISOString(),
>        });
>      })();
>    }, MOCK_DURATION_IN_MS);
>
>    timer.unref();
>  }
>```
> This function should be executed directly in the `getUploadJobNotification` service methods us such:
> ```
> getUploadJobNotification(): Observable<BulkUploadJobNotification> {
>    this.mockProcessUploadJob(jobId);
>    [...]
> }
> ```

1. Run the `yarn setup` command in the `/api` directory to apply all new changes.
2. Using tools (like Postman) create a job for a selected listing:
```
Endpoint: POST /jobs
Body: 
{
    "listingId": "<valid_listing_id>",
    "inputS3Key": "test-s3-key"
}
```
3. Using prisma studio validate the job exists and has the `processing` status.
4. Connect to the SSE notification event source using the `applications/bulk-update/notifications` endpoints as such:
```
Endpoint: GET applications/bulk-update/notifications?jobId=<job_id>
Body: ---None---
```
5. The connection should be created and after 10seconds and status update should be provided and connection 

> [!NOTE]
> If you want to re-run the tests above go to step 3 and using prisma studio reset the job status to "processing"

## Author Checklist:

## Review Process:


sites/partners/src/lib/hooks.ts
```diff
@@ -867,69 +885,97 @@ export function usePropertiesList({ page, limit, search, jurisdictions }: UsePro
...
+    }
 
-  return {
-    cacheKey,
-    data,
-    loading: !!cacheKey && !error && !data,
-    error,
-  }
-}
+    const eventSource = new EventSource(url, { withCredentials })
```
YazeedLoonat, line 1078: **issue (non-blocking):** This is going to be tough to test until we go to deploy it, but I think we may run into some issues with how traffic is allowed into/out of the VPC 

Currently all traffic is set up to go from the UI -> the next js server -> the api -> the next js server -> the ui 

I think this breaks this by trying to connect up directly with the API. This is just a flag that this is a possible point of failure when we go to deploy

api/src/services/application-bulk-upload.service.ts
```diff
@@ -447,302 +419,76 @@ export class ApplicationBulkUploadService {
...
-    const dateMatch =
-      row[bulkUploadHeaderNames.applicationSubmissionDate] === expectedDate;
-
-    if (!firstNameMatch || !lastNameMatch || !dateMatch) {
-      throw new BadRequestException(
-        `Upload Failed: One or more rows beginning on row ${
-          index + 2
-        } have incorrect application details (Applicant first name, last name or submission date)`,
-      );
+  private mapJobToNotification(
```
YazeedLoonat: **nitpick:** drop the private here, we don't really have a reason to control instance functions and it is not a pattern we maintain in the API

api/src/services/application-bulk-upload.service.ts
```diff
@@ -447,302 +419,76 @@ export class ApplicationBulkUploadService {
...
+    // listing/update scoped to the job's listingId) once the authentication
+    // story for server side events is settled - EventSource cannot send the
+    // passkey header the rest of this controller relies on.
+    jobId: string,
+  ): Observable<BulkUploadJobNotification> {
+    const currentState$ = defer(() =>
+      from(this.backgroundJobService.findById(jobId)),
+    ).pipe(map((job) => this.mapJobToNotification(jobId, job)));
+
+    const updates$ = this.notifications$.pipe(
```
YazeedLoonat, line 842: **question:** why the $ in the var name?

YazeedLoonat (commented):
looks good a few minor nitpicks and something for us to resolve once we get to the testing in a deployed env

api/src/services/application-bulk-upload.service.ts
matzduniuk: It's just a common pattern in the RxJS (coming from Angular where it was very widely used) to indicate the the variable is a Observable or other stream based source of data.

sites/partners/src/lib/hooks.ts
matzduniuk: Would need to make further research but I doubt it would be that problematic and maybe just require additional configuration as SSE are commonly in use for quite some time.

YazeedLoonat (approved):
LGTM

Issue #6513, opened 2026-07-15
HTML sanitization for rich-text content


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
`JurisdictionContent` rich-text bodies are admin-authored HTML, so stored HTML must be sanitized to
prevent stored XSS. This is a precondition for enabling the structured-content editor for any
jurisdiction.

## Technical Context
- TDD: §3.2, Open Question 3.
- Resolve OQ3 (sanitizer library and tag/attribute allowlist). Apply sanitization on write and as a
  defense-in-depth pass on render.
- Depends on: none. Prerequisite for the structured-content editor (#6517) go-live; also used by the
  JurisdictionContent write path (#6511) and the seed migration's markdown conversion (#6519).

## Acceptance Criteria
- A documented tag/attribute allowlist is enforced on write; a payload with disallowed tags/attrs
  (including `javascript:` link targets) is stripped or rejected.
- Render passes stored HTML through the sanitizer as a defense-in-depth step.
- The sanitizer is in place before `enableDbDrivenContent` is flipped on for any jurisdiction.

## QA Notes
Backend/shared. Test known XSS vectors (script tags, `on*` handlers, `javascript:` hrefs, disallowed
tags) are removed on write and render; confirm allowed formatting survives.

seanmalbert:
Closing. The allowlist is enforced on write by @SanitizeHtml() on the seven rich-text DTO fields, plus sanitizeArray for footer.textSectionsHtml, and javascript: targets are stripped by allowedSchemes. #6623 added the scheme allowlist and its tests. The allowlist itself is ALLOWED_TAGS, ALLOWED_ATTRIBUTES and ALLOWED_SCHEMES in api/src/decorators/sanitize-html.decorator.ts.

The defence-in-depth pass ended up in the API rather than at render: getMergedContent maps the merged document through JurisdictionContentFields, which re-runs the transform, so every public read is sanitised a second time. #6629 relies on that rather than adding a sanitiser to the public bundle.

PR #6623, opened 2026-08-19
feat(api): define and test the rich-text sanitizer allowlist

seanmalbert:
This PR addresses #6513

## Description

Sets the URL schemes the rich-text sanitizer accepts, and covers the allowlist with tests.

The sanitizer itself was already in place. What was missing was an explicit scheme list, tests for the vectors in the issue, and a test on the write and read passes.

- **Schemes are set rather than inherited.** A link in rich text may use `http`, `https`, `mailto`, or `tel`.
- **Tags and attributes are unchanged**, and now sit in named constants alongside the schemes.
- **Sanitization runs in the API on both write and read.** `@SanitizeHtml()` is a class-transformer `@Transform`, and the content DTOs map both request bodies and responses. On write the `ValidationPipe` runs it before the service sees the payload, so what reaches the database is clean. On read `mapTo` runs it again, which covers a row the API did not write: a seed, the #6519 data migration, or direct SQL.
- **The write test now reads the row.** It previously wrote through the endpoint and read back through the endpoint, which the read pass would clean regardless. It would have passed even if nothing sanitized the value before it was stored. It now asserts against the stored row with prisma.
- **Text direction is not stored.** `dir` stays out of the allowlist. The editor and the public page set direction on their container from the active locale, so a translated body cannot disagree with the language of the row it sits in.

1. **Unit** (`cd api && yarn test`). `sanitize-html.spec.ts` gains six cases: event handler attributes, `javascript:` targets in four evasion forms (upper case, an embedded tab, an HTML entity, and `data:`), script and style contents, `iframe` and `form`, presentation attributes, and stored `dir`. It also pins the three schemes the editor offers. `jurisdiction-content.service.spec.ts` gains two cases proving `getMergedContent` and `getContent` clean a poisoned stored row. All 64 suites pass, 1287 tests.
2. **Integration** (`yarn test:backend:new:e2e`). `jurisdiction-content.e2e-spec.ts` gains a case that writes a row straight to the database with prisma and reads it back through the public endpoint clean. The existing write case now covers an event handler and a `javascript:` href, and asserts against the stored row. 20 passing against a fresh database.
3. **Manual.** Not required. There is no UI and no endpoint change.

## Author Checklist:

- [x] Ran `yarn generate:client` and/or created a migration when required (N/A: no schema change and no DTO shape change)

## Review Process:

### DIFF
--- a/api/src/decorators/sanitize-html.decorator.ts
+++ b/api/src/decorators/sanitize-html.decorator.ts
@@ -1,12 +1,15 @@
 import { Transform, TransformFnParams } from 'class-transformer';
 import sanitizeHtml from 'sanitize-html';
 
+const ALLOWED_TAGS = ['b', 'strong', 'a', 'hr', 'p', 'ol', 'ul', 'li', 'br'];
+const ALLOWED_ATTRIBUTES = { a: ['href'] };
+const ALLOWED_SCHEMES = ['http', 'https', 'mailto', 'tel'];
+
 export const sanitize = (content: string) => {
   return sanitizeHtml(content, {
-    allowedTags: ['b', 'strong', 'a', 'hr', 'p', 'ol', 'ul', 'li', 'br'],
-    allowedAttributes: {
-      a: ['href'],
-    },
+    allowedTags: ALLOWED_TAGS,
+    allowedAttributes: ALLOWED_ATTRIBUTES,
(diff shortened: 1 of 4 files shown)


ludtkemorgan (approved):
🍇

PR #6625, opened 2026-08-19
feature: translation data migration

seanmalbert:
This PR addresses #6519

## Description

Adds `yarn translations:migrate`, which moves an environment's existing translation data into `translation_strings`, and seeds a few override rows for dev and staging. Backend only. No schema change and no API change.

- **Two sources.** The legacy `translations` blob becomes one row per key at the same jurisdiction and language with no site, which is the scope the email path reads. The bundled `sites/public/page_content/locale_overrides/*.json` become rows for the jurisdiction named by `--jurisdiction`, and `sites/partners/page_content/overrides/general.json` becomes rows shared by every jurisdiction.
- **Dry run by default.** `--commit` writes. Both print the same report of what is created, updated, and unchanged.
- **The blob write is one transaction**, with an explicit timeout. `getMergedTranslations` switches from the blob to rows as soon as the first base row exists, so a partial write would leave emails resolving against an incomplete set.
- **`--skip-existing`** writes only rows that do not exist yet, for re-running against an environment whose overrides an admin has edited.
- **Source hashes** are computed across all three scopes at once, resolving an English source the way `englishSourceHashes` does at edit time. A row whose English source is not in the database keeps a null hash, since the bundled locale files are not something the API can read.
- **Seeds.** `listingFilters.countyFilterNote` is supplied by no locale file, so it is the key that exercises the editor's hide-section confirmation.

### Not included

- Markdown conversion. Core has no markdown body to convert: `jurisdiction-content-factory.ts` already writes HTML and the generic content is JSX.
- Removing the bundled override files. The sites keep rendering them when the API is unreachable.
- A delete pass. A key removed from a bundled override file keeps its row, and its value keeps rendering.
- The validation the API applies to the same column. `NoExecutableMarkup` and `MaxLength` guard the editor's write path but not this one.

1. **Unit** (`cd api && yarn test`)
2. **Local end to end**: `yarn db:setup`, then `yarn translations:migrate --jurisdiction "Bloomington"`. The report should show 1078 blob rows, 17 Partners rows, and 245 public rows to create, with nothing to update. Run again with `--commit`, then a third time to confirm everything reports unchanged.
3. **Emails still resolve**: after the commit run, confirm `getMergedTranslations` returns rows rather than the blob and that the keys and values match what the blob held.
4. **Editor**: `GET /translations/jurisdictions/:id/raw/public/en` returns the seeded rows, including the fork-only key.
5. **Arguments**: `--skip-exsiting` and `--languages` with no value both stop the run rather than changing what it writes.

Note for whoever runs this in an environment: migrations that patch the `translations` blob stop reaching email once this has run. The README documents re-running as the interim step, and the longer fix has its own issue.

## Author Checklist:

- [x] Made corresponding changes to the documentation (API README)
- [x] Ran `yarn generate:client` and/or created a migration when required (N/A: no schema or DTO change)

## Review Process:


ludtkemorgan:
@seanmalbert Is there a reason you chose to write this as a script rather than a new script runner endpoint?

seanmalbert:
> @seanmalbert Is there a reason you chose to write this as a script rather than a new script runner endpoint?

@ludtkemorgan , yes, the main reason is because the API doesn't have access to the JSON files in sites/public and sites/partners. This means that we'd either have to assemble the JSON to pass in the body or fetching the files from GitHub, so I figured this was little more straightforward. Happy to move it to use the script runner though.

api/prisma/seed-helpers/translation-strings-factory.ts
```diff
@@ -0,0 +1,67 @@
...
+  PrismaClient,
+  SiteEnum,
+  TranslationOrigin,
+} from '@prisma/client';
+import { sourceHash } from '../../src/utilities/translation-source-hash';
+
+const FORK_ONLY_KEY = 'listingFilters.countyFilterNote';
+const FORK_ONLY_ENGLISH =
+  'Select a county to see the listings available there.';
+const OVERRIDDEN_KEY = 't.getDirections';
```
ludtkemorgan, line 13: suggestion: One string on the site that would be great to be dynamic when we are seeing is the welcome string. More specifically the "region.name". Could we use that one as our OVERRIDDEN_KEY and go even one step further where we can pass in the jurisdiction name into the factory to dynamically populate it?

Image

api/README.md
```diff
@@ -80,6 +80,30 @@ Examples:
 
 By default the script generates for `en, es, tl, vi, zh, ar, bn, ko, hy, fa` and machine-translates missing non-English values using the same Google env vars used by other translation scripts.
 
+## Moving translations into key rows
+
+`yarn translations:migrate` moves an environment's existing translation data into the `translation_strings` table, which the editor and the sites read. Run it once per environment, from this repository, with `DATABASE_URL` pointing at the target. It reads the bundled override files, which a deployed API does not have access to, so it cannot run as a script-runner command.
```
ludtkemorgan, line 85: suggestion: Ideally we will only need to run this script for our existing few sites and all others will go directly to using the database for their management. Maybe add a disclaimer at the top where we explain the purpose of this?

ludtkemorgan (approved):
This worked great! I'm not exactly sure how we are supposed to run this in the production site being that this is a yarn command. Do you have steps how to do that? Specifically how this can be run in an AWS environment since the AWS container has no knowledge of the public and partner code

seanmalbert:
closing, because #6655 moves it to use the script runner, and also addresses the feedback on this PR.

Issue #6577, opened 2026-08-12
Toggle off release to partners lottery step


emilyjablonski:
AC
- With the `enableNonAdminLotteries` flag off (every jurisdiction's current default), lottery behavior is identical to today — "Release lottery" still required, and running is still Admin-only.
- With the flag on, "Release lottery" is gone from the UI and a freshly-run lottery by an admin reaches a publishable, exportable state for partners with no release Admin action in between.
- Ensure history panel — including "retracted" state — is still accurate regardless of whether or not the feature flag is on (activity log expects `ran` after `releaseToPartners` or `publishedToPublic` - need to ensure it displays correctly still under this new flag)

PR #6633, opened 2026-08-25
Feat: toggle off release to partners lottery step

matzduniuk:
This PR addresses #6577 

## Description

* Adds the new `enableNonAdminLotteries` feature flag.
* Update the partners listing lottery page to hide the "Release lottery" button and add the publish action under the "Run Lottery" button

1. Run the `yarn setup` command in the `api/` directory to apply all new changes
2. Using Prisma Studio set the `enableNonAdminLotteries` to true on at least one Jurisdiction
3. Go to partners site and login as a partner user for the jurisdiction configured in step 2.
4. Select a listing for the jurisdiction configured in step 2 that uses lottery review order. 
> [!NOTE]
> If the selected listing is not closed, click the `Edit` button and close the listing
5.  Go to the `Lottery` tab in the listing view and click the `Run Lottery` button
6. Analyse the "History" component to ensure the lottery has been both `ran` and `published`

> [!IMPORTANT]
> During the review process the correct order of logged history events should be validated as per the ticker description:
>
> "_Ensure history panel — including "retracted" state — is still accurate regardless of whether or not the feature flag is on (activity log expects ran after releaseToPartners or publishedToPublic - need to ensure it displays correctly still under this new flag_)"

## Author Checklist:

## Review Process:


api/src/enums/feature-flags/feature-flags-enum.ts
```diff
@@ -53,6 +52,7 @@ export enum FeatureFlagEnum {
   enableMultiselectVoucherQuestion = 'enableMultiselectVoucherQuestion',
   enableNeighborhoodAmenities = 'enableNeighborhoodAmenities',
   enableNeighborhoodAmenitiesDropdown = 'enableNeighborhoodAmenitiesDropdown',
+  enableNonAdminLotteries = 'enableNonAdminLotteries',
```
ludtkemorgan, line 57: suggestion: lets have angelopolis have this feature flag turned on by default. So add it to seed-angelopolis.ts

ludtkemorgan (changes requested):
issue: When clicking the "publish lottery" button as a partner I get a 403 because the backend prevents any non admin from updating the lottery status.

I think we need to update lottery.service roughly line 524 to check the new feature flag

ludtkemorgan:
discussed via slack, but commenting here for my own reference. The permissioning issue is only happening when `disablePartnerPublicListingEdits` feature flag is turned on and logic within `permissionService.canOrThrow`.

api/src/services/lottery.service.ts
```diff
@@ -499,16 +499,23 @@ export class LotteryService {
...
 
-    await this.permissionService.canOrThrow(
-      requestingUser,
-      'listing',
-      permissionActions.update,
-      {
-        id: storedListing.id,
-        jurisdictionId: storedListing.jurisdictionId,
-      },
+    const enableNonAdminLotteries = doJurisdictionHaveFeatureFlagSet(
```
ludtkemorgan: issue: this is always returning undefined because the featureFlags are not part of the "storedListing". Also we still want to verify that the user has the ability to edit the specific listing so I think we can't just fully skip the canOrThrow function.

My recommendation is to update the permission service to handle this case. One idea is to pass in an additional context to the canOrThrow where we know we are only updating the lottery status. And then in the "addUserPermissions" function to not add the blocked listing list if it's a lottery status change and this new feature flag is enabled.

ludtkemorgan (changes requested):

ludtkemorgan (approved):
Thanks for making the change. The partners can now successfully release the lottery.

I did find an existing issue though where partners can't export the applications or lottery results if `disablePartnerPublicListingEdits` is enabled because in the `application-exporter.service` we check the authorize as if it's a listing update. @YazeedLoonat Is that supposed to be the case. If so we will need to follow the same pattern that was added here.

```
 await this.permissionService.canOrThrow(
      user,
      'listing',
      permissionActions.update,
      {
        id: listingId,
        jurisdictionId,
      },
    );
``` 

I'm ok merging this PR as is and we can create a separate ticket to address the above issue

ludtkemorgan:
@YazeedLoonat the above issue has been created as a ticket for this epic https://github.com/bloom-housing/bloom/issues/6675

Issue #6630, opened 2026-08-24
Render the disclaimers and contact documents on the public site


seanmalbert:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
`JurisdictionContent` holds five documents: footer, FAQ, resources, disclaimers, and contact. #6517
lets an admin edit all five. #6514 renders three of them, because it replaces the
`jurisdiction_*_content.tsx` stubs and stubs exist only for footer, FAQ, and resources.

Nothing on the public site reads `disclaimers` or `contact`. #6629 filters
those two out of the fetch.

## Technical Context
- TDD: §5.3, which enumerates the three existing stubs rather than the five documents.
- `disclaimers.privacyHtml` renders on `privacy.tsx` and `disclaimers.disclaimerHtml` on
  `disclaimer.tsx`. Both pages render from translation keys today and neither has a data function,
  so they need #6594 or their own.
- The `contact` document (`phone`, `email`, `addressHtml`, `hours`) renders on `get-assistance.tsx`,
  which already has a data function, so it does not wait on #6594. It is a separate document from
  `resources.contactCard`, which #6629 already renders.
- Follow the pattern #6629 established: a mapper in `sites/public/src/static_content/stored_content.tsx`
  that sets a key only for a field the stored document sets, merged over the bundled content field by
  field, with rich text set as HTML rather than parsed.
- Remove `disclaimers` and `contact` from `RENDERED_DOCUMENTS` in `sites/public/src/lib/hooks.ts` so
  they reach the pages.
- Depends on: #6511 (content service); #6594 for the two pages with no data function.

## Acceptance Criteria
- An admin edits the privacy policy in Partners and `/privacy` shows it, in English and in a
  non-English locale.
- The same for the disclaimer on `/disclaimer` and the contact details on `/get-assistance`.
- A jurisdiction with no stored value for a field keeps the bundled text for that field.
- Emptying a field hides it rather than rendering an empty block, matching the other three documents.
- No page ships a content document it does not render.

## QA Notes
Public site. Edit each of the three fields in the Partners content editor and confirm the matching
page reflects it. Empty one and confirm the section disappears. Clear the stored document and confirm
the page returns to its bundled text. Repeat one case in Spanish.

PR #6635, opened 2026-08-25
fix: overrides on remaining pages for database content

seanmalbert:
This PR addresses #6594

## Description

The 39 public pages with no data function now pass the jurisdiction's translation overrides and stored content down as props, so their strings and their footer come from the jurisdiction rather than from the values bundled at build time. #6629 wired the 20 pages that already had a data function; these are the rest.

Stacked on #6629, which introduced `fetchSharedPageProps`.

- `src/lib/sharedPageProps.ts` holds the two data functions, and each page re-exports one of them. The block is otherwise identical in every page, and this keeps the props shape and the revalidate policy in one place. The module sits outside `hooks.ts` so the existing spy on `fetchSharedPageProps` still intercepts the call.
- 8 standalone pages and the 29 application-flow pages use `getStaticProps`.
- The two `account/application/[id]` pages use `getServerSideProps`, matching the other account pages. Static generation would need `getStaticPaths` with a blocking fallback, which caches one shell per application id for pages that are authenticated and fetch client-side.
- `redirect.tsx` and `listing/[id].tsx` are left alone: both render only a meta refresh, so they have no strings to override.

This also closes the footer gap #6629 listed as known: the footer no longer changes as a visitor moves between pages of the same site.

**`500.tsx` is new.** `_error.tsx` renders outside the routing table, so a data function on it is accepted and then ignored: it produces no prerender manifest entry and no page data. That matters because the error copy interpolates `resources.contactEmail`, whose bundled value is the placeholder `email@email.com`, so a server error told the visitor to write to a made-up address on every jurisdiction. A real `pages/500.tsx` takes `getStaticProps` and is prerendered per locale, which fixes the server-rendered case.

It renders inside `Layout`, as `404.tsx` does, so it has the header, the footer and a page title.

Uncaught client-side render errors still fall through to `_error.tsx`, which cannot take these props, so they still show the bundled strings. `_error` is also rendered directly as a listing-not-found state by `ListingViewSeeds` and `listing/[id]/[slug]`, which is why its copy is left alone here. Both are worth a follow-up issue rather than a change in this PR.

**Application steps now prefetch the next step.** Adding `getStaticProps` marks these pages `__N_SSG`, so the client router blocks each transition on the step's page data. The flow navigates with `Router.push` rather than `Link`, so nothing prefetched them and every Next press waited on a round trip. `ApplicationConductor.prefetchNextUrl` warms the next url on mount.

Two consequences worth knowing when reviewing:

- Working out the next url asks each step whether it would be skipped, which now happens on mount as well as on navigation. That is safe because deciding to skip does not write to the application, with one exception that reads the jurisdiction's config rather than the form. `ApplicationConductor.test.ts` asserts both, across every step definition in the routes table.
- The prefetch swallows its own failure. It runs the same code the Next button does, and on mount a throw would take the step down rather than only the navigation.

**Payload.** A page's data is `jurisdiction`, `publicOverrides` and `jurisdictionContent`. Measured against a seeded jurisdiction, an application step's page data is 3,697 bytes, or 1,669 gzipped. The application pages move from automatically statically optimized to statically generated with props, roughly 290 more prerendered pages across 10 locales on the revalidate timer. Backend load does not scale with that, since `fetchSharedPageProps` caches per language for the length of the window, so what grows is regeneration work rather than API calls. Worth watching build time.

**Jest.** Four content-page components import `static_content/generic_content.md`, which `next.config.js` handles with an `asset/source` webpack rule. Jest had no equivalent, so any test requiring those pages failed on the markdown. `.jest/markdown-transform.js` is the Jest counterpart.

1. **Public unit** (`yarn test:app:public:unit`). 109 suites, 991 tests.
   - `pageOverrideProps.test.ts` is 122 cases, up from 41. It enumerates every page with a data function and asserts it returns the jurisdiction, the overrides and the stored content, asks for the rendered locale, and sets `revalidate` where it should. It also compares the enumeration against the pages on disk, across all four extensions Next treats as pages, so a new page that misses its data function fails rather than passing silently.
   - `ApplicationConductor.test.ts` covers the prefetch target, the skip-decision rule described above, and that a prefetch failure does not propagate.
   - `500.test.tsx` covers the new page rendering inside the site layout.
   - `DisclaimerSeeds.test.tsx` renders the markdown so the transform is checked for the text it produces, not only for loading without throwing.
2. **Manual.** With a jurisdiction that has stored content, confirm `/privacy`, `/disclaimer`, `/404` and an application step render the jurisdiction's footer rather than the bundled one, and that a translation override edited in Partners appears on those pages.
3. **Build** (`cd sites/public && yarn build`). Confirm the application routes appear in `.next/prerender-manifest.json` with a revalidate value, which is what proves Next detects the re-exported data function. Check the prerender count and build time against the base branch.

## Author Checklist:

- [x] Reviewed considering accessibility (no UI change; the change is which strings load and when)
- [x] Ran `yarn generate:client` and/or created a migration when required (N/A: no API change)

## Review Process:


sites/public/src/pages/disclaimer.tsx
ludtkemorgan, line 1: question: Sorry if I missed this on the other PR, doesn't this page need to also display the database version of the disclaimer content?

sites/public/src/pages/privacy.tsx
ludtkemorgan, line 1: Same here about the privacy policy

ludtkemorgan (approved):
Just the one question about if this PR is supposed to cover disclaimer and privacy content from the db

sites/public/src/pages/disclaimer.tsx
seanmalbert: This is tracked as a separate issue #6630

sites/public/src/pages/privacy.tsx
seanmalbert: same as above, tracked in #6630

PR #6649, opened 2026-08-31
fix: add optional income (PoC)

KrissDrawing:
This PR addresses #6485 

## Description

It's PoC for making income optional. Just to test (last commit).
This one will make AMI chart income optional for all listings overrides (from backend perspective), on frontend it has additional protection.
There are 2 main questions:
1. should we allow overrides to have optional income on backend?
2. Should optional income be also present for AmiChartItem (not override). Not sure how that should work i never uploaded AMI chart, but thats for jurisdiction if im not mistaken, so there is no way to decide if thats for land use listing (so i guess answer is no 🤔 ).

Go to partners listing Unit AMI chart land use listing it should allow to fully remove ami chart income value.
For non land use it should still be required (on frontend).

## Author Checklist:

## Review Process:

### DIFF
--- a/api/src/dtos/units/ami-chart-override-item.dto.ts
+++ b/api/src/dtos/units/ami-chart-override-item.dto.ts
@@ -1,0 +1,15 @@
+import { ApiPropertyOptional, OmitType } from '@nestjs/swagger';
+import { Expose } from 'class-transformer';
+import { IsNumber, IsOptional } from 'class-validator';
+import { ValidationsGroupsEnum } from '../../enums/shared/validation-groups-enum';
+import { AmiChartItem } from './ami-chart-item.dto';
+
+export class AmiChartOverrideItem extends OmitType(AmiChartItem, [
+  'income',
+] as const) {
+  @Expose()
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @IsNumber({}, { groups: [ValidationsGroupsEnum.default] })
+  @ApiPropertyOptional()
+  income?: number;
+}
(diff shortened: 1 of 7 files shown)


mcgarrye:
For open question #2, it should not be optional. There must be a base value

mcgarrye:
Screenshot 2026-08-31 at 4 44 22 PM

Partner side detail view. If string is not numeric, we should remove the `$`

mcgarrye:
Screenshot 2026-08-31 at 4 44 48 PM

Can we get confirmation from Lily that this is the expected output on the public side? This makes sense to me but we need sign off. For context, I set the override for household size of 1 to null as above

ludtkemorgan:
@lillianvail question above ^

lillianvail:
@ludtkemorgan @mcgarrye 
So I understand, this will remove the specific unit size (& correlating columns) of the land-use listings? 

How will this interact with having income ranges?

KrissDrawing:
@lillianvail 
image
For now it displays it as `$0 - $...` (which seem wrong), but i think it's not a problem to omit it fully, and just display other present values, and treat that `null` as it is non existent.
But i don't know what is expected solution here

lillianvail:
I brought this up to LA in today's synch and they pointed out 1) they were under the impression that it was a different table and 2) it was removing a column, not a row. 

This makes much more sense to me, but let me know if you have any questions @ludtkemorgan @mcgarrye @KrissDrawing 

Screenshot 2026-09-09 at 17 47 01

mcgarrye:
> I brought this up to LA in today's synch and they pointed out 1) they were under the impression that it was a different table and 2) it was removing a column, not a row.
> 
> This makes much more sense to me, but let me know if you have any questions @ludtkemorgan @mcgarrye @KrissDrawing
> 
> Screenshot 2026-09-09 at 17 47 01

Ahhhhhh this makes much more sense. Great news cause now we won't need to mess with AMI charts at all. This should hopefully be much easier for Kriss to handle when he returns next week

Issue #6650, opened 2026-08-31
Drop the legacy translations table


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue

Nothing reads the `translations` table. #6632 made `api/src/locales/email-translations.ts` the base layer for email strings and moved overrides onto `translation_strings` rows with a `site` of `email`. Its migration 71 moved each jurisdiction's rows onto that scope, and its seed change writes new jurisdictions there. The script runner commands and the `translations:sql` generator that wrote the table were removed in the same PR.

What remains is the table, the `Translations` Prisma model, the `translations` relation on `Jurisdictions`, and two seed helpers kept only so existing specs keep compiling.

## Technical Context

- `prisma/seed-helpers/translation-factory.ts` still exports `upsertTranslation` and `translationFactory`. Neither seed calls them. Thirteen spec files use them.
- Twelve integration specs call `prisma.translations.create({ data: translationFactory() })` as fixture setup: `application.e2e-spec.ts`, `lottery.e2e-spec.ts`, and the ten files under `test/integration/permission-tests/`. Check whether any assertion depends on the row before removing the call, since they appear to create it as setup only.
- `test/unit/services/email.service.spec.ts` uses `translationFactory().translations` to mock `getMergedTranslations`. That mock returns a nested object, while the service returns a flat map of dotted keys. Polyglot accepts both shapes, so the tests pass either way, but the mock no longer matches what it stands in for. Replace it with `translations(undefined, LanguagesEnum.en)` from the module.
- The generic rows left in the table are covered by the module. #6632 verified that every key the migrations produce is present in the module with the same value, across all ten languages.
- The migrations that wrote the table stay as they are. They are history and still run against a fresh database.
- Dropping the table needs a migration, removal of the `Translations` model, and removal of the `translations` field on `Jurisdictions` at `prisma/schema.prisma:849`.

## Acceptance Criteria

- The `translations` table, the `Translations` model, and the `Jurisdictions.translations` relation are removed.
- Nothing outside `prisma/migrations` references the table.
- The unit and integration suites pass.
- A fresh setup produces a working environment, and a jurisdiction's email footer still shows that jurisdiction's name.

## QA Notes

Backend only. Run a fresh setup, then send an email for a jurisdiction that has an override and confirm both the override text and the footer. Confirm that editing an email string in Partners still takes effect.

## Dependencies

[...]

Issue #6632, opened 2026-08-24
Make email translations editable in Partners


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [x] Partners

## Background / Issue
`getMergedTranslations` reads email strings from `translation_strings` where `site` is null, in four
layers: generic English, generic language, jurisdiction English, jurisdiction language. #6519's
migration puts them there, so the read path already works and a jurisdiction's own strings already
act as overrides.

No one can edit them. `SiteEnum` is `public` and `partners` only, and every write endpoint on the
translation controller requires one, so a row with a null site cannot be addressed. Email copy is
still changed by an engineer writing a migration.

The generic values also sit apart from every other scope. Public and Partners have a bundled base in
the locale files with database overrides on top; email's base is a hardcoded object in
`api/prisma/seed-helpers/translation-factory.ts`, written to the legacy blob and exploded into
generic rows by the migration.

## Technical Context
- The email templates under `api/src/views` use 105 distinct keys, and `email.service.ts` resolves
  55 directly. 103 of the 105 do not exist in `shared-helpers/src/locales/general.json`, so they are
  email-only despite some sitting in the generic `t.` namespace. The two that are shared with the
  sites are `leasingAgent.officeHours` and `t.seeListing`.
- Add `email` to `SiteEnum` and move the 103 email-only keys onto it. Decide the two shared keys
  either way: leave them unscoped and accept that editing them changes the sites too, or give email
  its own copy under a new name.
- `getMergedTranslations` filters `site: null` and needs to read the email scope instead, keeping the
  unscoped rows as the layer beneath it so the base/partners merge is unchanged.
- The editor gains a third scope alongside public and partners. #6584 already generalized it beyond a
  single jurisdiction scope.
- Move the generic values out of `translation-factory.ts` so the seed sources them from a locale
  file, giving base strings one home. The API does not read `shared-helpers/src/locales/general.json`
  today, so this is a change to the seed rather than to the read path.
- Depends on: #6519 (its migration is what puts email strings in the new model); #6631 (email
  translation tooling still writes the legacy blob), which has to land first or its writes will keep
  bypassing whatever this adds.

## Acceptance Criteria
- An admin edits an email string for their jurisdiction in Partners and a sent email reflects it,
  in English and in a non-English language.
- A jurisdiction with no override for a key still receives the generic value.
- The public and Partners translation scopes are unaffected by the new scope.
- Base email values live in a locale file rather than in the seed helper.

[...]

PR #6593, opened 2026-08-18
Refactor: translation scope hook

seanmalbert:
Refactor only. No behaviour change and no issue behind it.

- [x] Addresses the issue in full
- [ ] Addresses only certain aspects of the issue

## Description

`settings/translations.tsx` computed its own scope selection: three pieces of state, the derived jurisdiction and language, the option lists, the service calls for each scope, and both reads. That left the page mixing scope selection with rendering the editor, and none of it was reachable from a test without rendering the whole page.

`useTranslationScope` now owns that. The page destructures what it needs and renders.

What moved, unchanged:

- The `jurisdictionId`, `language` and `site` state.
- `isGlobal`, `selectedJurisdiction`, `activeJurisdictionId`.
- `partnersLanguages`, `languageOptions`, `activeLanguage`, including the fallback when a jurisdiction does not offer the selected language.
- The `scope` object: which rows to read, which base overrides to compare against, and the save and revert calls for the jurisdiction and global endpoints.
- Both `useRawTranslations` reads, including the second English read that collects `englishOverrideKeys`.

Two renames came with the move: `scopeReady` is now `ready`, and the page's `authorized` flag is passed in as `enabled`.

The page keeps everything else. It still owns the tabs, the table, the unsaved changes warning, the conflict handling and the save and revert mutations.

## How Can This Be Tested/Reviewed?

1. **Partners unit** (`yarn test:app:partners:unit`). `useTranslationScope.test.tsx` is new and covers 17 cases: the initial scope, switching jurisdiction, the global scope naming no jurisdiction, holding the read until a jurisdiction exists, holding every read while the page is not authorized, the language options for each scope, the English fallback when a site lists no usable locales, keeping a selected language when the new scope still offers it, both write paths, and the English override key collection.
2. **Manual.** In Partners, open Settings then Translations. Switch jurisdiction, switch between the public and Partners scopes, and switch language. Confirm the rows reload for the right scope, that the Partners scope offers the site's own locales while a jurisdiction scope offers only that jurisdiction's, and that selecting a language a new jurisdiction does not offer falls back rather than leaving the editor on an unavailable language. Save an override and revert one in each scope.

## Author Checklist:

- [ ] Added QA notes to the issue with applicable URLs
- [ ] Reviewed in a desktop view
- [ ] Reviewed in a mobile view
- [x] Reviewed considering accessibility (no UI change; the page renders the same markup)
- [x] Added tests covering the changes
[...]

PR #6651, opened 2026-08-31
feature: email translation editing

seanmalbert:
This PR addresses #6632

## Description

Email strings were stored as one JSON blob per language in the `translations` table. The seed wrote them, and changing email copy meant an engineer writing a migration.

They now follow the arrangement the public and Partners scopes already use: a base that ships with the api, plus overrides in the database that an admin edits in Partners. The base is `api/src/locales/email-translations.ts`. Overrides are `translation_strings` rows on a new `email` value of `site_enum`.

`getMergedTranslations` layers them from lowest to highest precedence: the English base, generic English overrides, then the recipient's language base and its overrides, then the jurisdiction's English and language rows. A language sits above the English it was translated from, so editing an English string does not replace its translations. A key a language does not translate still falls through to English.

### Migrations

`70_add_email_translation_scope` adds `email` to `site_enum`. Postgres cannot use a new enum value in the transaction that adds it, so everything that writes the value lives in the next migration.

`71_move_jurisdiction_email_translations` moves a jurisdiction's email overrides onto the email scope. It moves only the keys whose value differs from the generic row.

A jurisdiction's row in the legacy table contains its own overrides plus a copy of the base. Of the 141 `UPDATE translations` statements in earlier migrations, 108 have no `jurisdiction_id IS NULL` clause and use `jsonb_set`, which creates the key when it is absent, so every jurisdiction row accumulated the base as well. Moving all of it would have written around 117 rows per jurisdiction per language. Those rows outrank the base at read time, so a later edit to the base would stop reaching deployed databases, and the editor would show hundreds of keys as overrides the admin never wrote. The generic row is the comparison because it matches the base module key for key. A value blanked as JSON `null` moves as an empty string.

### Base content

The base module is the object that was in the seed factory, which had not kept up with the migrations that changed email copy. It has been filled in from those migrations and now matches what a deployed database holds for every key, across all ten languages. Ten values differed between the two. Each took the value the database holds, so no email copy changes as a result of this PR.

### The editor

Email is a third scope in Settings > Translations. It reuses the jurisdiction read and write endpoints, which already take a site, so only the base column differs. The public and Partners bases are bundled into the site, while the email base ships with the api and is served by a new `GET /translations/base/email/:language`. `useEmailBaseTranslations` reads it for English, and for the active language when that is not English; the scopes that bundle their base make no extra request.

That base arrives over the network, so it is absent while the request is in flight and if the request fails. The hook reports `baseReady`, and the grid waits for it before rendering rows.

Email uses the selected jurisdiction's language list, the same as the public scope, because a jurisdiction only sends email in the languages it offers.

The development seed now turns on `enableDbDrivenContent`, which gates the Translations tab, for the jurisdiction it creates. It previously created every flag through `featureFlagFactory`, which appends a random suffix to keep generated names unique, so none of its flags matched the names the code gates on and the tab never appeared.

### Staleness

A non-English override stores a hash of the English it was translated from, so the editor can flag it when that English later changes. The resolver read stored rows only, and email's English base ships with the api, so every email translation saved with no source and could never be flagged. The base now sits beneath the stored rows for this scope. A stored English override still wins as the source. Rows created by migration 71 keep a null hash: the English those translations were made from is not known, and recording today's value would assert they are current.

### The public reads

`GET /translations/jurisdictions/:id` and `/translations/byName/:name` take `site` as a query parameter, have no jurisdiction access check, and carry a public cache header, so adding `email` to the enum would have exposed a jurisdiction's email overrides to any caller with the public site's passkey. They now accept only the public and partners scopes. Admin access is unchanged, through the raw routes, which take `site` as a path parameter and check jurisdiction access.

### Removals

Four script runner endpoints and the `translations:sql` generator wrote the legacy table, and all of them are removed. Three of the four had never written anything: the helper they shared tested `translation.translation`, and the field on the model is `translations`. `upsertTranslation` is removed too, since nothing calls it once the seeds write to the email scope.

The seeds now write a jurisdiction's strings onto the email scope instead of the legacy table.

### Notes

This PR targets `refactor/translation-scope-hook` rather than `main`, because the editor builds on the `useTranslationScope` hook that #6593 introduces. That PR needs to merge first, and GitHub will retarget this one to `main` when it does. The diff here excludes its changes.

`translationFactory` is still in the seed helper. Twelve integration specs use it to build fixtures against the legacy table. Dropping that table is #6650, after this has run in a deployed environment.

`docs/feature-flags/enableDuplicatesDetailsInEmail.md` and the matching row in `docs/feature-flags.md` come from `yarn generate:client`, which this PR had to run for the new endpoint. The enum on `main` already used that flag name while the generated docs did not, so the generator corrected them. `docs/feature-flags/enableDuplicatesDetails.md` is now unreferenced, since the generator writes files but does not remove them.

1. Run a fresh setup with `yarn db:setup`.
2. Confirm the jurisdiction's footer landed on the new scope:

   ```sql
   SELECT language, key, value FROM translation_strings WHERE site = 'email';
   ```

   Expect `footer.line1` and `footer.footer` set to the jurisdiction name, and no other keys.

3. In Settings > Translations, pick the Email site. The base column should show the email strings, and none of the public site's keys.
4. Edit a string, save, then trigger the email that uses it and confirm the new text. Repeat in a non-English language.
5. Edit the English of a key that the language translates, and confirm the translation is unchanged for a recipient in that language.
6. With a non-English override saved, change the English it came from and confirm the editor marks that row stale.
7. Confirm a jurisdiction with no override still receives the base text, and that the footer shows the jurisdiction name.
8. Confirm the public and Partners scopes are unchanged, and that `GET /translations/jurisdictions/:id?site=email` returns 400.

### Verification

The api unit suite passes at 63 suites and 1291 tests, the api e2e suite at 36 suites and 1357 tests, and the Partners suite at 90 suites and 935 tests.

New coverage: migration 71 is exercised against a legacy blob that mixes keys identical to the generic row, a key that differs, and a key the generic row lacks, and it asserts only the last two move; the merge precedence cases in `translation.service.spec.ts`; the base endpoint across all ten languages in `translation.e2e-spec.ts`; the staleness round trip on the email scope; the base loading, failed and language-missing states in `useTranslationScope.test.tsx`; and saving and reverting an email override in `settings/translations.test.tsx`.

## Author Checklist:


api/prisma/seed-helpers/translation-factory.ts
```diff
@@ -1,1745 +1,41 @@
...
+        key,
+        value,
+      },
+      update: { value },
+    });
   }
-};
+}
 
 export async function upsertTranslation(
```
ludtkemorgan: nit: Looks like this isn't used anymore. It can be deleted

ludtkemorgan (changes requested):
Still making my way through the code, but here are my initial findings:
* Issue: The functionality is working properly for public site email, however any partner site email there is no way to override the values. We send more emails from the partner site than the public site so this is important to be able to override.
* suggestion: We typically don't use the seed-dev script as the seed-staging script has more realistic data to what our production sites have (the correct feature flags turned on and listings that match those setups). The dev script is most commonly used for performance testing. No code changes are required, everything you added makes sense. Just wanted to give you a heads up to let you know that you might have more accurate testing using the seed-staging
* thought: All of the production jurisdictions we have setup have used the translation table slightly different. Some use the jurisdiction overrides to add new entries and some just override the base one. I'll copy those tables locally and test with it to make sure the migrations and functionality works as expected
* question: How do you envision the process for adding a new set of keys for the email translations? When we add a new email to the system how would all jurisdictions get those new values?

seanmalbert:
> * Issue: The functionality is working properly for public site email, however any partner site email there is no way to override the values. We send more emails from the partner site than the public site so this is important to be able to override.

Users can now select "Default (no jurisdiction)" to override the values for partner site emails. These values also become the base/default for the jurisdiction email strings.

api/prisma/seed-helpers/translation-factory.ts
seanmalbert: I've got #6650 to track removing this and the table and model.

ludtkemorgan:
> > * Issue: The functionality is working properly for public site email, however any partner site email there is no way to override the values. We send more emails from the partner site than the public site so this is important to be able to override.
> 
> Users can now select "Default (no jurisdiction)" to override the values for partner site emails. These values also become the base/default for the jurisdiction email strings.

@seanmalbert I'm not seeing that drop down option. Did you push up the change?

seanmalbert:
I'm sorry, @ludtkemorgan, I did forget to push. You should see the changes now.

I also didn't answer your question:
>question: How do you envision the process for adding a new set of keys for the email translations? When we add a new email to the system how would all jurisdictions get those new values?

The base layer for email translations now lives in `api/src/locales/email-translations.ts`. That would be updated, similar to the base JSON files for public and partners.

ludtkemorgan (approved):
Looks good. And thanks for explaining how to add new translations

PR #6654, opened 2026-09-01
fix(api): check status and time out when fetching translation files

seanmalbert:
Prerequisite for reworking #6625's translation backfill into a script-runner endpoint. That endpoint fetches the same override files from GitHub.

## Description

`getTranslationFile` fetches locale files from `raw.githubusercontent.com` for `migrateDetroitToMultiselectQuestions`. It used its own `https` client, with four defects:

- The response status was ignored, so a 403 with a body of `{"message":"rate limit exceeded"}` resolved as a translation file.
- A response cut short mid-body left the promise pending forever. Node emits that error on the response object, and the handler was on the request. The command hung with its `scriptRuns` row already marked as started, so it could never be re-run.
- `body += chunk` decoded each Buffer separately, so a multi-byte character split across a chunk boundary became two replacement characters. Two of the six files are around 92 KB and 125 KB and contain non-ASCII.
- Rejections were strings rather than `Error` objects.

It now goes through `HttpService`, as `external-listing.service.ts` does.

1. `cd api && yarn test`. The suite no longer depends on a network connection; disconnect and run it if you want to confirm.
2. Read the `getTranslationFile` describe: the parsed body, the url and deadline passed to the client, a missing file, a cut-short response, the deadline being named rather than reported as a cancellation, a non-json body, and an empty body.
3. Removing the json check fails two tests. Removing the deadline fails one.
4. `migrateDetroitToMultiselectQuestions` is otherwise unchanged, so its existing assertions on script runs, query counts and the listing update still stand.

### Verification

The api unit suite passes at 64 suites and 1317 tests, and the api lint gate passes at its 2-warning threshold.

## Author Checklist:

### DIFF
--- a/api/src/modules/script-runner.module.ts
+++ b/api/src/modules/script-runner.module.ts
@@ -1,4 +1,5 @@
 import { Logger, Module } from '@nestjs/common';
+import { HttpModule } from '@nestjs/axios';
 import { ScriptRunnerController } from '../controllers/script-runner.controller';
 import { ScriptRunnerService } from '../services/script-runner.service';
 import { AmiChartModule } from './ami-chart.module';
@@ -11,6 +12,7 @@ import { PrismaModule } from './prisma.module';
 @Module({
   imports: [
     AmiChartModule,
+    HttpModule,
     EmailModule,
     FeatureFlagModule,
     MultiselectQuestionModule,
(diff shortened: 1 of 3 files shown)


ludtkemorgan (approved):
This looks good! I'll also merge this so that reviewing your other PR is easier

PR #6655, opened 2026-09-01
feature: translation data migration

seanmalbert:
This PR addresses #6519

Supersedes #6625, which should be closed. That PR shipped this as a local `yarn translations:migrate` script.

Stacked on #6651 and targets `6632/email-translation-scope` rather than `main`, because removing the blob half below depends on that PR's migration 71. The diff also has #6654's three files until that merges to `main` and reaches this branch.

## Description

`PUT /scriptRunner/migrateTranslationOverridesToKeyRows` loads the bundled public and partners override files into `translation_strings`, which the editor and the sites read. It fetches them from GitHub, with the repository, ref, and the directory under each site as request parameters.

Only a full admin can run it. The route has the permission guard against the `translation` resource, which only `admin` has a policy grant on. The controller's class-level guard admits a jurisdictional admin, so without this the endpoint would have written a jurisdiction's rows for them.

The command resolves the jurisdiction name before anything else, so an unknown name fails without consuming the script name. It then fetches and validates every file, builds rows for both sites, applies source hashes across them together, diffs each scope against what is stored, and logs the report. Only when `commit` is true does it record the run, write, and mark it complete.

Public rows are written for the named jurisdiction. Partners rows are written with no jurisdiction.

The writes go through one `$transaction` covering both sites, so a failure partway through no longer leaves partners written and public not. `createMany` uses `skipDuplicates`, so a row an admin added between the diff and the write is left as they set it rather than failing the insert.

### What the request may ask for

`repositoryUrl` must be an https url on `raw.githubusercontent.com`, with no port, credentials, query, or fragment. `@IsUrl({ protocols: ['https'] })` accepted `https://10.0.4.12:9200` and `https://169.254.169.254`, so the endpoint could be pointed at anything reachable from the api.

`gitRef`, `publicPath`, and `partnersPath` share one pattern that refuses `..` and a leading non-word character. The previous `gitRef` rule accepted `main/../../../attacker/bloom-fork/main`, which resolves to a different repository.

`languages` requires a non-empty array. A bare string passed validation and was iterated by character, requesting `e.json` and `s.json`.

`publicPath` and `partnersPath` are new. The two directories were fixed constants matching core's layout, so a fork laid out with `sites/*/src/page_content` could not run the migration against its own repository at all.

### Values the editor would refuse

Each file is checked against the editor's own key length, value length, and markup rules before it is used, and a failure stops the run naming the file and the key. A value that fails them would otherwise land in a row no admin could then edit. All 283 pairs across the eleven bundled files pass, so this rejects nothing the migration is meant to write.

### Missing files

A 404 on a non-english file is tolerated and named in the report. Partners ships english only, and a fork need not translate every language. A 404 on an english file stops the run, because english is the source every translation is hashed against.

Any other failure stops the run before anything is written.

### Dry runs

`commit: false` writes nothing and records nothing in `script_runs`, so a dry run can be repeated after correcting a ref. A committing run is recorded per jurisdiction, so a second one is refused until that row is deleted.

`commit` and `skipExisting` are both required. The validation pipe drops unknown properties, so a misspelled `skipExisting` would otherwise fall through to a default that overwrites every admin edit.

A committing run that tolerated a missing file says so in a closing report line, since the run records itself as complete either way.

### What #6625 had that this does not

The half that exploded the legacy `translations` blob into key rows with no site is gone. The email read path no longer reads those rows: it reads `api/src/locales/email-translations.ts` plus rows on the `email` scope, and migration 71, added in #6651, moves a jurisdiction's rows itself. Keeping it would also write an english source that `englishSourceHashes` reads for every site, not just email.

`withSourceHashes` loses only its fallback to that scope. No row loses a hash: every key in each of the nine non-english public override files is present in `general.json`, so the first fallback resolves for all of them, and partners has no non-english rows.

### Review comments

The seeded overridden key is now `region.name`, populated from the jurisdiction name, so a database seeded under another name shows that name on the homepage hero. Note that `region.name` is in all ten public override files as "Bloomington", so a run without `skipExisting` overwrites the seeded value. That only shows in a database seeded under a different name.

The README section opens with the disclaimer that this exists only for the sites predating database-managed content.

1. `yarn db:setup`, then start the api.
2. Sign in as an admin and take the session cookie.
3. Dry run, using the `curl` in `api/README.md` with `"commit": false`. The report is in the api log. Confirm it lists rows to create, that `translation_strings` is unchanged, and that no `script_runs` row appeared.
4. Run it again with `"commit": true`. Confirm the public rows have the jurisdiction id, the partners rows have none, and a `script_runs` row now exists.
5. Call it a third time. It should return 400.
6. Point `gitRef` at a value that does not exist. The error should name the url and the status.
7. Send a body without `commit`. It should return 400 rather than defaulting.
8. Sign in as a jurisdictional admin and call it. It should return 403.
9. Set `"repositoryUrl": "https://169.254.169.254"`, then `"gitRef": "main/../../../attacker/bloom-fork/main"`. Both should return 400 naming the field.

### Verification

The api unit suite passes at 65 suites and 1364 tests, and the api e2e suite at 36 suites and 1387 tests on a fresh database. Both lint gates pass.

The integration suite runs the endpoint with `commit: true` against the real database, with `HttpService` overridden to serve fixture files. Before this the create and the null-scoped `updateMany` had only ever run against a mocked Prisma, so neither had met the `NULLS NOT DISTINCT` index. Overriding the provider also stops the already-run case from fetching from GitHub, which it was doing because the run is refused only after the files are read.

Each fix was checked by removing it and confirming a named test fails. That pass found that the validation cases asserted only a 400 status, which a body naming an unknown jurisdiction produces anyway, so none of them were failing when the rule under test was removed. They now assert the field the validation pipe named.

## Author Checklist:


ludtkemorgan (approved):
This works great! I tested it with pointing to a couple of our forked repos and it successfully pulled in all appropriate keys

PR #6658, opened 2026-09-02
6254/professional partners page

mcgarrye:
This PR addresses [#(6254)](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/6254)

## Description

Creates new `professional-partners` content page for Doorway use. Has generic content for core

## Author Checklist:

## Review Process:


sites/public/src/pages/professional-partners.tsx
```diff
@@ -0,0 +1,83 @@
...
+      metaDescription={t("pageDescription.housingBasics")}
+    >
+      <PageHeaderLayout
+        heading={t("pageTitle.professionalPartners")}
+        subheading={t("pageDescription.housingBasics")}
+        inverse
+        className={pageStyles["site-layout"]}
+      >
+        <article className={pageStyles["site-content"]}>
+          <div className={pageStyles["item-wrapper"]}>
```
ludtkemorgan, line 48: issue: the questions get wider when opening and the aside gets squished. After some digging it looks like you just have a typo here. It should be `items-wrapper`

docs/feature-flags/enableProfessionalPartnersPage.md
```diff
@@ -0,0 +1,13 @@
...
+## Name
+
+`enableProfessionalPartnersPage`
+
+## Description
+
+When true, the professional partners page will appear in the nav header of the public site
+
+## Additional Information
+
```
ludtkemorgan, line 12: suggestion: add a screenshot here of the page

sites/public/src/static_content/generic_professional_partners_content.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return {
+    categories: [faqContentSection, faqContentSection],
+  }
+}
+
+export const getGenericProfessionalPartnersContactContent = (): ContactCardProps => {
+  const contactDescription = t("resources.contactDescription")
+  const contactInfo = t("resources.contactInfo")
+  const email = t("resources.contactEmail")
+  const heading = t("footer.contact")
```
ludtkemorgan, line 52: nit: This isn't a problem for this PR per se, but this is slightly different pattern than what we are using for the additional-resources page. The "contact" card can be updated with just changing the translation strings whereas this requires a change here. It won't be an issue for using in Doorway, and we can accommodate around this when/if we add it to Sean's changes. So I'm not worried, but if our long term plan was to keep forking repos we would want new strings for each of these.

ludtkemorgan (approved):
Looks good. Just one typo fix and the tests added it is good to merge

sites/public/src/static_content/generic_professional_partners_content.tsx
mcgarrye: This is just the generic content, the actual professional partners page will allow the use of Sean's changes like all the other content pages

ludtkemorgan: Yeah, I was more referencing how there is no specific translations keys for these fields. So without Sean's changes you would need to add new keys to the overrides file and update them here. With the resources page you can just update the strings and it will appear without any additional steps. Like I said, I'm not worried about it and can proceed with this!

PR #6659, opened 2026-09-03
feature: disclaimers and assistance render stored content

seanmalbert:
This PR addresses #6630

Targets `6594/overrides-on-remaining-pages` rather than `main`, for the `getStaticProps` #6635 adds to `/privacy` and `/disclaimer`. The diff also has #6635's `housing-basics` commit until that merges.

## Description

`/privacy`, `/disclaimer` and `/get-assisstance` now render the database stored content.

### How a stored value layers over the bundled one

Both documents follow the same three states per field, whichever page reads them:

| Stored document | What renders |
|---|---|
| no such key | the bundled value |
| the key, with text | the stored value |
| the key, empty or null | nothing |

`stored_content.tsx` expresses that two ways, because the two documents are shaped differently. `storedPageBody` takes the bundled node and returns what to render, since a disclaimers page has one field and no partial state. `getStoredContactContent` returns only the fields the document contains, and the contact card spreads that over its bundled values, following `getStoredResourcesContent`. There the third state works by assigning the key with a value of `undefined`, so it survives the spread and replaces the bundled value.

### Disclaimers, on /privacy and /disclaimer

Each of the four page components passes the bundled `static_content/generic_content.md` to `storedPageBody` and renders what comes back.

`DisclaimerSeeds`, `DisclaimerDeprecated`, `PrivacySeeds` and `PrivacyDeprecated` all change, not just the Seeds pair. `showNewSeedsDesigns` is empty unless `SHOW_NEW_SEEDS_DESIGNS` is `TRUE`, so the deprecated components are what render by default and a Seeds-only change would have no effect on most deployments.

### Contact, on /get-assistance

The card keeps its bundled heading, description and info line. A stored email replaces the bundled one. Phone, address and hours each render only when the document sets them, so a jurisdiction that stores nothing sees the card it has today. Only `email` overlapped with what the card already rendered.

`addressHtml` renders through `StoredHtml`. The phone is a `tel:` link, matching how the listing view renders a leasing agent.

`yarn db:setup` already seeds what this needs: Bloomington has `enableDbDrivenContent` on, the English content row sets both documents, and the Spanish row sets `privacyHtml` with `disclaimerHtml` explicitly null.

With the api and public site running:

1. `/privacy` and `/disclaimer` show the stored text rather than the bundled markdown.
2. `/get-assistance` shows the stored email, phone, address and hours.
3. `/es/privacy` shows the Spanish text.
4. `/es/disclaimer` shows nothing for the body. `disclaimerHtml` is null there, and it must not fall back to the bundled markdown.
5. `update jurisdiction_content set disclaimers = null where language = 'en'`, then reload `/disclaimer` and expect the bundled markdown back.

Content is cached for `cacheRevalidate` seconds, 30 by default, so a database change takes that long to appear.

Editing through Partners needs #6628, which is not in this stack.

### Verification

The public suite passes at 114 suites and 1062 tests, `tsc --noEmit` is clean, and the lint gate reports warnings only. Each behaviour is covered by a test that fails when it is reverted: falling back to the bundled page, hiding an emptied or null field, the contact card reading the stored fields, dropping a document with no row, and passing each document through `RENDERED_DOCUMENTS`.

No test added here reads a string the repository ships. The disclaimer cases count markdown headings rather than matching the bundled wording, and where a bundled string is unavoidable it is looked up through `t(key)`.

## Author Checklist:

- [ ] Made corresponding changes to the documentation (not applicable)
- [ ] Ran `yarn generate:client` and/or created a migration when required (not applicable)


ludtkemorgan (approved):
The new pages look good. After a rebase and merge conflict resolution this is good to merge

Issue #6663, opened 2026-09-03
Brand API: validation, shade-ramp derivation, and asset URLs


seanmalbert:
- [ ] General
- [x] Backend
- [ ] Public
- [ ] Partners

## Background / Issue
The jurisdiction read must return a complete, validated brand: every ramp filled in (derived from
its base where not explicit), and the logo and favicon as usable URLs. The jurisdiction update must
reject a malformed brand. Both ride the existing jurisdiction endpoints; no new endpoint.

## Technical Context
- TDD: §4.1, §4.2, D2, OQ1.
- `BrandRampDTO` (`base` required `@IsHexColor()`; `dark`/`darker`/`light`/`lighter` optional) and
  `BrandDTO` (`primary` required, `secondary` optional, `fontFamily`/`fontUrl` optional), nested on
  `JurisdictionUpdate` via `@ValidateNested()` so the existing `ValidationPipe` rejects malformed
  payloads. `brandLogoAssetId`/`brandFaviconAssetId` are `@IsUUID()` and must reference existing
  assets. Hex values normalize to uppercase on write.
- Shade-ramp derivation runs server-side at read time, never stored: absent ramp values derive from
  that ramp's `base` via the OQ1 algorithm (proportional HSL deltas per §11.1: darker = L×0.64,
  dark = L×0.88, light = L+(100−L)×0.88, lighter = L+(100−L)×0.95). Explicit values win. Runs for
  primary and for secondary when present. Settle the exact deltas before this starts if §11.1 is
  not final.
- `logoUrl`/`faviconUrl` built at serialization per storage backend (§4.1, §4.3): extend
  `cloudinaryUrlFromId` (or add a sibling) for PNG output at branding quality, since the existing
  helper hardcodes the listing-image `q_65` `.jpg`; the S3 path uses the direct
  `s3Service.urlForPublic(fileId)` URL unchanged.
- Caching: `Cache-Control: public, s-maxage=3600, stale-while-revalidate=86400` on the public
  jurisdiction read.
- Depends on: #6662 (data model).

## Acceptance Criteria
- A read of a jurisdiction with only `brand.primary.base` set returns a complete primary ramp with
  the documented derived values; explicit ramp values are returned as stored.
- `secondary` is absent from the response when the jurisdiction sets none.
- A malformed brand (bad hex, unknown shape) is rejected on write; a nonexistent asset id is
  rejected.
- `logoUrl`/`faviconUrl` are correct for both storage backends.

## QA Notes
[...]

PR #6676, opened 2026-09-04
feature: brand API validation, shade-ramp derivation, and asset URLs

seanmalbert:
This PR addresses #6663

Stacked on #6674 and targets `6662/branding-data-model`.

## Description

Brand fields are added to the existing jurisdiction endpoints. The read endpoints return them and `PUT /jurisdictions/:jurisdictionId` accepts them. No new endpoint is added.

### Write

`BrandDTO` is validated by the existing `ValidationPipe`. Ramp values must be 3- or 6-digit hex with a leading hash and are normalized to uppercase. The generic hex-color validator is not used because it also passes 4- and 8-digit and unprefixed values, which the ramp derivation and CSS variables cannot use. The primary ramp is required; the secondary ramp, font family, and font URL are optional. `brandLogoAssetId` and `brandFaviconAssetId` must be UUIDs of existing assets. An unknown id returns a 400 that includes the id, instead of a 500 from the foreign key violation.

Sending `brand: null` clears the stored brand, and sending a null asset id disconnects that asset.

The stored JSON contains only the fields the admin set. `logoUrl` and `faviconUrl` are removed before storage because the read path builds them from the asset foreign keys, and stored copies could become inconsistent with those keys. The write path connects the assets as Prisma relations, which the checked create input requires.

### Read

All jurisdiction reads build the brand response before serialization. Ramp values the admin did not set are derived from that ramp's base color with the TDD §11.1 HSL deltas (darker = L×0.64, dark = L×0.88, light = L+(100−L)×0.88, lighter = L+(100−L)×0.95). Values the admin set are returned unchanged. The expected values in the derivation test table were computed with a separate implementation (python colorsys), not with the code under test.

A stored ramp without a base color, which only a hand-edited row can produce, is returned as stored instead of failing the jurisdiction read.

`logoUrl` and `faviconUrl` are built from the referenced assets' file ids: a Cloudinary URL with `f_png,q_90` and a per-kind width, or the S3 public-bucket URL when `USE_S3_FILE_STORAGE` is `TRUE`. That env var is new and is documented in `.env.template`; the api previously had no storage-backend setting. When S3 is selected but the bucket or region is unset, no URL is built and an error is logged. A jurisdiction with an asset but no colors returns the urls. A jurisdiction with no brand data returns `brand: null`.

The two public GET routes set the shared `PUBLIC_CACHE_CONTROL` header. Its defaults are `s-maxage=300, stale-while-revalidate=600`, and both values can be changed by env var. The TDD suggests 3600 and 86400; those can be set through the env vars, and reusing the shared constant keeps a single cache policy.

### Also in the diff

The `SET NULL` test from #6674 assigned raw FK scalars alongside checked create input, which runs but fails `tsc --noEmit`. It now uses the `connect` form. The generated client gains the brand types, regenerated from an isolated instance so no other operations changed.

1. PUT a jurisdiction with `"brand": { "primary": { "base": "#77aa33" } }` and GET it back by name: the base is uppercased, the four shade values are filled in, and the database row stores only the base.
2. PUT explicit ramp values and confirm they come back unchanged.
3. PUT `"primary": { "base": "rebeccapurple" }`, `"base": "773E98"` (no hash), a brand with no primary, and a `brandLogoAssetId` that does not exist: each returns 400.
4. Create an asset row, reference it, and confirm `brand.logoUrl` for the configured storage backend.
5. PUT `"brand": null` on a branded jurisdiction and confirm the stored brand is cleared.
6. GET a jurisdiction with no brand: `brand` is null and the response has the `Cache-Control` header.

### Verification

The api unit suite passes at 67 suites and 1351 tests, and the e2e suite at 36 suites and 1345 tests on a fresh database, including six new cases through the real endpoints. Mutations each fail named tests: skipping derivation at read, storing the url fields, removing the asset-existence check, breaking a derivation delta, loosening the hex format back to the generic validator, removing the malformed-row guard, and removing the null clear. Both frontends typecheck against the regenerated client.

## Author Checklist:


api/.env.template
```diff
@@ -103,6 +103,8 @@ RECAPTCHA_THRESHOLD=0.7
 ENABLE_RECAPTCHA=TRUE
 # controls the repetition of the afs (duplicates) cron job
 DUPLICATES_PROCESSING_CRON_STRING=15 * * * *
+# When TRUE, asset urls (branding logo/favicon) are built for the S3 public bucket rather than Cloudinary
+USE_S3_FILE_STORAGE=
```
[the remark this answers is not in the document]
seanmalbert: Fixed. `S3_PUBLIC_BUCKET` is now in the template next to `S3_BUCKET`. Worth noting the gap was not specific to branding: `S3Service` reads the same variable at `s3.service.ts:30` and logs "S3_PUBLIC_BUCKET is not set" without it.

api/src/utilities/brand-asset-url.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (process.env.USE_S3_FILE_STORAGE === 'TRUE') {
+    const bucket = process.env.S3_PUBLIC_BUCKET;
+    const region = process.env.S3_REGION;
+    if (!bucket || !region) {
+      console.error(
+        'USE_S3_FILE_STORAGE is TRUE but S3_PUBLIC_BUCKET or S3_REGION is not set',
+      );
+      return undefined;
+    }
+    return `https://${bucket}.s3.${region}.amazonaws.com/${fileId}`;
```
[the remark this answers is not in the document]
seanmalbert: Fixed, thank you, this was a good one to flag. The flag is now set in `ecs_api_service.tf` alongside the `S3_PUBLIC_BUCKET` it depends on.

api/src/controllers/jurisdiction.controller.ts
```diff
@@ -46,6 +48,7 @@ export class JurisdictionController {
   }
 
   @Get(`:jurisdictionId`)
+  @Header('Cache-Control', PUBLIC_CACHE_CONTROL)
```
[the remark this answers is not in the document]
seanmalbert: Intentionally not changed. `PUBLIC_CACHE_CONTROL` is shared with the Phase 1 translation and jurisdiction-content endpoints, so raising the default to 3600/86400 changes their caching too, beyond this PR's scope. The durations are tunable per environment through `PUBLIC_CACHE_S_MAXAGE` and `PUBLIC_CACHE_STALE_WHILE_REVALIDATE`.

api/src/dtos/jurisdictions/jurisdiction.dto.ts
```diff
@@ -279,4 +282,23 @@ export class Jurisdiction extends AbstractDTO {
   @Type(() => IdDTO)
   @ApiPropertyOptional({ isArray: true, type: IdDTO })
   subJurisdictions?: IdDTO[];
+
+  @Expose()
+  @IsOptional({ groups: [ValidationsGroupsEnum.default] })
+  @ValidateNested({ groups: [ValidationsGroupsEnum.default] })
+  @Type(() => BrandDTO)
+  @ApiPropertyOptional({ type: BrandDTO })
+  brand?: BrandDTO;
```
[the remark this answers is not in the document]
seanmalbert: Fixed at the schema level. `brand`, `brandLogoAssetId`, and `brandFaviconAssetId` are now declared `nullable: true` with matching TypeScript types, so the served OpenAPI document describes what the endpoint accepts and returns.

seanmalbert:
@ludtkemorgan , Copilot's findings have been addressed.

api/test/integration/jurisdiction.e2e-spec.ts
```diff
@@ -86,6 +86,145 @@ describe('Jurisdiction Controller Tests', () => {
...
+      const res = await request(app.getHttpServer())
+        .get(`/jurisdictions/byName/${updateBody(jurisdiction.id).name}`)
+        .set({ passkey: process.env.API_PASS_KEY || '' })
+        .expect(200);
+
+      expect(res.body.brand.primary.base).toEqual('#77AA33');
+      expect(res.body.brand.primary.dark).toMatch(/^#[0-9A-F]{6}$/);
+      expect(res.body.brand.primary.darker).toMatch(/^#[0-9A-F]{6}$/);
+      expect(res.body.brand.primary.light).toMatch(/^#[0-9A-F]{6}$/);
+      expect(res.body.brand.primary.lighter).toMatch(/^#[0-9A-F]{6}$/);
```
ludtkemorgan, line 141: question: Is there a reason we can't do an exact match of the non-base values? Don't we want to test that the ramping worked properly

api/src/utilities/brand-ramp.ts
```diff
@@ -0,0 +1,79 @@
...
+export const completeRamp = (ramp: BrandRamp): Required<BrandRamp> => {
+  const base = hexToHsl(ramp.base);
+  const { l } = base;
+
+  return {
+    base: ramp.base.toUpperCase(),
+    darker: ramp.darker?.toUpperCase() ?? shifted(base, l * 0.64),
+    dark: ramp.dark?.toUpperCase() ?? shifted(base, l * 0.88),
+    light: ramp.light?.toUpperCase() ?? shifted(base, l + (100 - l) * 0.88),
+    lighter: ramp.lighter?.toUpperCase() ?? shifted(base, l + (100 - l) * 0.95),
```
ludtkemorgan, line 77: suggestion: There are a decent amount of magic numbers in this file, especially in this return statement. Maybe set them as named constants to describe what they are.

ludtkemorgan: Oh, I see at the top of the file you call out you got those values from UI-seeds. I think that is sufficient enough unless you want to move the comment to right above this return statement

api/.env.template
```diff
@@ -103,10 +103,14 @@ RECAPTCHA_THRESHOLD=0.7
...
 # Region the s3 bucket used to manage large file transfers (like csv/spreadsheet zips) is set up under
 S3_REGION=
 # The s3 bucket used to manage large file transfers (like csv/spreadsheet zips)
 S3_BUCKET=
+# The public s3 bucket that serves uploaded assets, read when USE_S3_FILE_STORAGE is TRUE
```
ludtkemorgan, line 112: suggestion: Our s3 env variables are getting a bit confusing with all of the different use cases. Also, it's been our practice to not have a "USE_X" env variable around these, rather if there is a value for the bucket it would choose that over cloudinary. I'm curious what your thought would be to remove the USE_S3_FILE_STORAGE and instead rely on the S3_PUBLIC_BUCKET existing. Don't know if that makes it more confusing

ludtkemorgan (commented):
The code looks good and the ramping seems to work. I was unable to get the PUT endpoint to work however. It kept complaining about fields not being the correct value. Like this:
`"listingFeaturesConfiguration.an unknown value was passed to the validate function"`
any suggestions on how to test it properly?

seanmalbert:
> I was unable to get the PUT endpoint to work however. It kept complaining about fields not being the correct value. Like this:
"listingFeaturesConfiguration.an unknown value was passed to the validate function"
any suggestions on how to test it properly?

@ludtkemorgan , the cause of that bug is that `defaultValidationPipeOptions` validates with `groups: [ValidationsGroupsEnum.default]`, and the validators on `ListingFeaturesConfiguration`, `ListingFeatureField`, and `ListingFeatureCategory` declare no group, so none of them run. With no applicable constraints and `forbidUnknownValues: true` set, class-validator reports the object as an unknown value, which is the message you saw. It looks like this was introduced in #5777 .

To test this PR before that one merges, omit `listingFeaturesConfiguration` from the PUT body.

api/test/integration/jurisdiction.e2e-spec.ts
seanmalbert: Good question, and there was no good reason not to test the actual values. Updated, so the test now asserts the whole primary ramp rather than the shape of each value:

```ts
expect(res.body.brand.primary).toEqual({
  base: '#77AA33',
  dark: '#69962D',
  darker: '#4C6D21',
  light: '#EFF7E4',
  lighter: '#F8FCF4',
});
```

api/src/utilities/brand-ramp.ts
seanmalbert: Moved, as you suggested in the follow-up. The explanation now sits directly above the return rather than at the top of the file.

api/.env.template
seanmalbert: Updated per your suggestion and removed `USE_S3_FILE_STORAGE`. Following the existing convention of relying on the existence of the value to use the value makes sense to me. Keeping the flag would probably create more confusion.

ludtkemorgan (approved):

Issue #6669, opened 2026-09-03
Partners Branding settings page


seanmalbert:
- [ ] General
- [ ] Backend
- [ ] Public
- [x] Partners

## Background / Issue
Admins need to edit a jurisdiction's brand without a deploy: colors, ramp overrides, logo, favicon,
and font, from a settings page in Partners.

## Technical Context
- TDD: §6.1, §6.3, §4.3.
- New `/settings/branding` page alongside `/settings/properties` and `/settings/preferences`, on
  the established primitives: react-hook-form, the shared `Field` components, SWR invalidation,
  toasts.
- Fields per §6.1: primary color picker with hex input (required once any branding field is set);
  four optional primary ramp pickers showing the derived value when empty and staying absent on
  save so derivation continues to apply; optional secondary and its four ramp pickers; logo and
  favicon uploads through the existing `fileUploader`/`CloudinaryUpload` path (S3 or Cloudinary per
  `useS3FileStorage`, §4.3), creating `Assets` rows labeled `brandLogo`/`brandFavicon`; font family
  text and font URL.
- Saves PUT the existing jurisdiction update with the `brand` object and asset ids; no new
  endpoint.
- Visibility: `superAdmin` always; `jurisdictionAdmin` when `enableDbDrivenBranding` is on for
  their jurisdiction (§6.3).
- Depends on: #6663 (API), #6664 (flag). #6670 adds the preview to this page.

## Acceptance Criteria
- An admin can set every §6.1 field, upload both assets, and save; the public site reflects the
  change after revalidation.
- An empty ramp field shows the derived value and saves as absent.
- Validation errors from the API surface on the offending fields.
- Page visibility follows the access rules.

## QA Notes
Partners. Edit each field and verify against the public site. Confirm a `jurisdictionAdmin` without
the flag cannot see the page and `superAdmin` always can. Upload on both storage backends.

Issue #6666, opened 2026-09-03
Brand logo and favicon rendering


seanmalbert:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
With the colors runtime-driven, the logo and favicon are the remaining static assets. The site
header should render the jurisdiction's uploaded logo and the document head its favicon, each
falling back to the in-repo defaults.

## Technical Context
- TDD: §5.3, §4.1 (URL building lands in #6663), OQ3.
- Pass a logo node built from `brand.logoUrl` to the ui-seeds `SiteHeader`'s `logo` prop
  (`layouts/application.tsx` behind `showNewSeedsDesigns`), falling back to
  `/images/logo_glyph.svg`. The legacy `UICSiteHeader` path is intentionally not wired (§5.3).
- Emit `<link rel="icon" href={faviconUrl}>` from #6665's `_document.tsx` when set; when unset,
  emit nothing so the browser falls back to `/favicon.ico`.
- OQ3 (favicon formats) is decided here: default to a single square PNG upload (≥256×256),
  transformed to standard sizes where the backend supports it. Multiple-format uploads
  (apple-touch-icon, `.ico`) only if review shows they are needed.
- Public site only (§5.5).
- Depends on: #6663 (URLs), #6665 (`_document.tsx` exists).

## Acceptance Criteria
- With a logo asset set, the header renders it; with none, the in-repo default.
- With a favicon asset set, the head links it; with none, no link is emitted and `/favicon.ico`
  applies.
- Both behave per storage backend.

## QA Notes
Public site. Verify header and tab icon with and without assets set, on a Cloudinary-backed and an
S3-backed environment.

Issue #6665, opened 2026-09-03
Public site brand injection via _document.tsx


seanmalbert:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
The brand colors must be on the page at first paint, including in portals and modals that render
outside the jurisdiction class wrapper. This ticket adds the public site's `_document.tsx` and the
per-request `:root` style block.

## Technical Context
- TDD: §5.1, §5.2, D3, D4, D6.
- New `sites/public/src/pages/_document.tsx` (picked up by filename convention; no build-config
  change). `getInitialProps` fetches the jurisdiction via the existing `fetchJurisdictionByName`
  (in-memory TTL cache; forward `ctx.req` for address attribution) and emits the style block from
  its `brand`.
- Emit the `--seeds-color-primary*` ramp, the `--seeds-color-secondary*` ramp when secondary is
  present, and the `--bloom-color-primary*` ramp for the legacy ui-components surface (D4). The
  seedification audit (#5903) decides whether the `--bloom-*` mapping is dropped at cleanup or
  extended; it does not block this issue (§10).
- Gate on the jurisdiction's `enableDbDrivenBranding` flag: flag off or `brand` null emits no block
  and the page renders exactly as today.
- Re-assert every interpolated value as hex at render time, so a stored row can never inject CSS
  regardless of write-path history.
- Public site only (§5.5). Partners gets no `_document.tsx` and no injection.
- Depends on: #6663 (complete ramps in the read), #6664 (flag).

## Acceptance Criteria
- With a primary set and the flag on, the SSR HTML contains the `:root` block with all five primary
  values in both namespaces; with a secondary set, the secondary variables too; with no secondary,
  they are omitted.
- With the flag off or no brand, the SSR HTML contains no brand block.
- A non-hex value in a stored row is dropped at render, not emitted.
- A modal or drawer rendered in a portal shows the brand colors.

## QA Notes
Public site. Curl the SSR HTML for a branded and an unbranded jurisdiction and diff the head.
Toggle the flag and confirm the block appears and disappears. Open a modal and confirm portal
coverage visually.

PR #6683, opened 2026-09-11
feat(public): render the brand logo and favicon

seanmalbert:
This PR addresses #6666

Stacked on #6681 and targets `6665/public-brand-injection`.

## Description

The logo and favicon are the two static assets left after #6665 made the colors runtime-driven. The site header now renders the jurisdiction's stored logo and the document head links its favicon, each falling back to what ships in the repo.

### Favicon

`_document.tsx` already resolves the brand behind `enableDbDrivenBranding`, so the favicon rides on that same resolution and needs no separate gate. It emits `<link rel="icon" href={faviconUrl}>` when one is stored and nothing at all when not, which leaves `/favicon.ico` to apply as it does today.

Two limits of that, both outside what this PR can fix. A stored favicon whose asset no longer resolves shows the browser's blank icon rather than falling back to `/favicon.ico`, because any declared `rel="icon"` stops the browser requesting the default, and a second link does not reliably serve as a fallback. And the 64px width in the delivery URL applies only on the Cloudinary path: with `USE_S3_FILE_STORAGE=TRUE` the stored object is served at its original size on every page load, so a large upload is a per-visit cost. Both are arguments for constraining the upload in #6669 rather than the render.

### Logo

The header is in the React tree rather than the document head, and `layouts/application.tsx` has no jurisdiction of its own. Phase 1 solved the same problem with `JurisdictionContentContext`: `_app.tsx` reads the value out of `pageProps` and provides it, and components consume a hook. This follows that route. `fetchSharedPageProps` already returns the whole jurisdiction, so `_app` applies `isFeatureFlagOn` and provides the brand through a new `BrandContext`, and the layout reads it with `useBrand`. The alternative, passing the brand into `Layout` from every page, is what the Phase 1 work rejected.

`headerLogo` is exported from the layout so the choice between the stored logo and the fallback can be tested without rendering the layout, the same reason `brandStyleBlock` is exported from the document. A second test renders `Layout` itself, so removing the `logo` prop or the context read fails rather than passing silently.

The logo slot in `SiteHeader.module.scss` capped height only. It now caps width as well, at 12rem, with `object-fit: contain`, so a wide wordmark cannot push the header row out of shape at mobile widths. That is the one change here no test covers, since Jest does not evaluate the stylesheet.

### Two to review

The image renders with `alt=""`. The site title sits beside it inside the same link, so a populated alt would announce the same thing twice, and the brand model has no alt-text field to draw from. `CustomSiteFooter` takes the other approach with a stored `logoAltText`, but that field belongs to Phase 1 content rather than to the brand.

The fallback stays the `HomeIcon` the header renders today. The issue and TDD §5.3 both describe the fallback as `/images/logo_glyph.svg`, which is not what the seeds header currently uses; switching to it would change how every unbranded site looks, which is outside what this issue asks for.

The legacy `UICSiteHeader` is deliberately not wired, per §5.3, and Partners is untouched.

Set up from #6681 step 1 applies: the jurisdiction needs a brand and `enableDbDrivenBranding` on. These steps add the two assets. `psql` connects to the database in the api's `DATABASE_URL`, and each change needs `CACHE_REVALIDATE` seconds (30 by default) before a render picks it up.

**1.** Point the jurisdiction at a logo and a favicon. The `file_id` values have to exist in the environment's image store, since the API builds the URL from them rather than storing it.

```sql
INSERT INTO assets (file_id, label, updated_at)
VALUES ('dev/bloom_logo', 'brandLogo', now()), ('dev/bloom_favicon', 'brandFavicon', now());

UPDATE jurisdictions SET
  brand_logo_asset_id = (SELECT id FROM assets WHERE label = 'brandLogo' LIMIT 1),
  brand_favicon_asset_id = (SELECT id FROM assets WHERE label = 'brandFavicon' LIMIT 1)
WHERE name = 'Bloomington';
```

**2.** Confirm the API composes both URLs. On a Cloudinary-backed environment they carry the per-kind width, 400 for the logo and 64 for the favicon; with `USE_S3_FILE_STORAGE=TRUE` they are public bucket URLs instead.

```bash
KEY=$(grep '^API_PASS_KEY' api/.env | cut -d= -f2- | tr -d '"')
curl -s -H "passkey: $KEY" \
  "http://localhost:3100/jurisdictions/byName/Bloomington" \
  | python3 -c "import json,sys; b=json.load(sys.stdin)['brand']; print(b['logoUrl']); print(b['faviconUrl'])"
```

The passkey is required: without it the endpoint returns a 401 body with no `brand` key. The two file ids have to exist in the environment's image store, or the URLs resolve to nothing and the header shows a broken image.

**3.** Confirm the head links the favicon, and that the tab icon changes in the browser.

```bash
curl -s http://localhost:3000/ | grep -o '<link rel="icon"[^>]*>'
```

**4.** Load the site and confirm the header shows the stored logo rather than the house icon. Worth doing with a wide wordmark and at a narrow viewport, since the width cap is not covered by any test.

**5.** Clear both assets and confirm the page returns to the in-repo defaults: the house icon in the header, no `rel="icon"` in the head, and the browser falling back to `/favicon.ico`.

```sql
UPDATE jurisdictions SET brand_logo_asset_id = NULL, brand_favicon_asset_id = NULL
WHERE name = 'Bloomington';
```

**6.** Turn `enableDbDrivenBranding` off with both assets set, and confirm both revert to the defaults.

### Verification

The public suite passes at 117 suites and 1094 tests, including 11 new cases, and the site typechecks and lints clean. Run it with `--runInBand`: in parallel, `create-account` and `finder` fail intermittently on 5s timeouts, with a different set of test names each run and each suite passing on its own.

The new cases cover the favicon link with an asset, without one, and with the flag off; `headerLogo` with a stored logo, with no brand, and with a brand that has no logo; the `_app` gating with the flag on, with it off, and with no jurisdiction in props; and `Layout` itself rendering the stored logo or the fallback icon. The two `Layout` cases were checked against mutations: deleting the `logo` prop fails one, and severing the context read fails the other.

## Author Checklist:

- [ ] Made corresponding changes to the documentation (none required)
- [ ] Ran `yarn generate:client` and/or created a migration when required (not required)

### DIFF
--- a/sites/public/__tests__/layouts/application.test.tsx
+++ b/sites/public/__tests__/layouts/application.test.tsx
@@ -1,0 +1,41 @@
+import React from "react"
+import { BrandDTO } from "@bloom-housing/shared-helpers/src/types/backend-swagger"
+import { render, mockNextRouter } from "../testUtils"
+import Layout from "../../src/layouts/application"
+import { BrandContext } from "../../src/lib/BrandContext"
+
+const renderLayout = (brand: BrandDTO | null) =>
+  render(
+    <BrandContext.Provider value={brand}>
+      <Layout>
+        <div>page</div>
+      </Layout>
+    </BrandContext.Provider>
+  )
+
+describe("<Layout>", () => {
+  beforeEach(() => {
+    mockNextRouter()
+    process.env.showNewSeedsDesigns = "TRUE"
+  })
+
+  afterAll(() => delete process.env.showNewSeedsDesigns)
+
+  it("puts the stored logo in the header", () => {
+    const { container } = renderLayout({
+      logoUrl: "https://example.test/logo.png",
+    } as unknown as BrandDTO)
+
+    expect(container.querySelector("header img")).toHaveAttribute(
+      "src",
+      "https://example.test/logo.png"
+    )
+  })
+
+  it("keeps the in-repo icon when the jurisdiction has no brand", () => {
+    const { container } = renderLayout(null)
+
+    expect(container.querySelector("header img")).toBeNull()
+    expect(container.querySelector("header svg")).toBeInTheDocument()
+  })
+})
--- a/sites/public/__tests__/layouts/headerLogo.test.tsx
+++ b/sites/public/__tests__/layouts/headerLogo.test.tsx
@@ -1,0 +1,28 @@
+import { render } from "@testing-library/react"
+import { BrandDTO } from "@bloom-housing/shared-helpers/src/types/backend-swagger"
+import { headerLogo } from "../../src/layouts/application"
+
+const brandWith = (logoUrl?: string) => ({ logoUrl } as unknown as BrandDTO)
+
+describe("headerLogo", () => {
+  it("renders the stored logo when the jurisdiction has one", () => {
+    const { container } = render(headerLogo(brandWith("https://example.test/logo.png")))
+
+    const image = container.querySelector("img")
+    expect(image).toHaveAttribute("src", "https://example.test/logo.png")
+    expect(image).toHaveAttribute("alt", "")
+  })
+
+  it("falls back to the in-repo icon when there is no brand", () => {
+    const { container } = render(headerLogo(null))
+
+    expect(container.querySelector("img")).toBeNull()
+    expect(container.querySelector("svg")).toBeInTheDocument()
+  })
+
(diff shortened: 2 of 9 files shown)


ludtkemorgan (approved):
Looks great!

Image

Issue #6573, opened 2026-08-11
Header "eyebrow" - core changes


ludtkemorgan:
- [ ] General
- [ ] Backend
- [x] Public
- [ ] Partners

## Background / Issue
For the Doorway implementation of Bloom we need to support an "eyebrow" above the main nav bar. This is to allow an additional logo as well as text on the right side.

<img width="688" height="101" alt="Image" src="https://github.com/user-attachments/assets/bbf2fca2-5870-42e3-ad2a-2a4c27c0011e" />

## Technical Context
* Designs: https://www.figma.com/design/7txDwTSN3ITUWI14598NyE/Corification-updates?node-id=5116-4680&p=f&m=dev 
* The "eyebrow" should be optional. Only display if the content exists
* In core a logo should not exist. By default nothing is shown in the core repo.
* It should be built the same way that the content pages (e.g. `faq.tsx` and `Resources.tsx`)  are built. It fetches from a separate file a `getJurisdictionContent` to determine what content to display. This separate file will be updated in each forked repo that wants it.

## Acceptance Criteria
* The ability to add a customized "eyebrow" for a jurisdiction

PR #6689, opened 2026-09-15
fix: only render eyebrow text if mounted

ludtkemorgan:
This PR addresses #6573

## Description

When turned on, the eyebrow causes issues in the CI cypress tests. This is because the tIfExists returns false on the first render before the translations have been added. This issue doesn't appear outside of CI, but it could theoretically happen so this is a fix to only render the text after initial mount.

Note: when eyebrow is enabled there might be a small size increase of the eyebrow when the text appears

Update `jurisdiction_eyebrow_image.tsx` to have an image and add nav.eyebrow.text in the locale_overrides make sure the eyebrow still appears.

## Author Checklist:

## Review Process:


mcgarrye (approved):
lgtm

PR #6691, opened 2026-09-15
feat: expand home resource cards

mcgarrye:
This PR addresses [#(6497)](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/6497) [#(6260)](https://app.zenhub.com/workspaces/bloom-5dc32d7144bd400001315dac/issues/gh/bloom-housing/bloom/6260)

## Description

For the new Doorway site, two new resource cards need to be added to the home page. This add the card to link to the '/get-assistance' page and a card to link to an external site to display data (getAssistanceCard and seeOurDataCard respectively).
With more card options, the system needed to be more dynamic. The grid will now adjust columns in the following way on desktop/tablet:
1 card -> 2 * 1
even # cards -> 2 * n
odd # cards -> 3 * n

Additionally, the resourcesCard was tied to `enableResources` but needed to be decoupled from this flag and is now tied to a new flag `enableResourcesCard`

Note: the unit test folder for home components was at the wrong level, folder was moved

Run `yarn setup:region` and have public pointed to `Bridge Bay`
Navigate to the public site
Verify getAssistanceCard, notificationCard and seeOurDataCard all appear in that order
Play around with the feature flags to test the dynamic grid layout: [`enableAdditionalResources`, `enableGetAssistanceCard`, `enableResourcesCard`, `enableSeeOurData`]. notificationCard will always appear as jurisdiction has `notificationsSignUpUrl` set. 

## Author Checklist:

## Review Process:


shared-helpers/src/locales/ar.json
```diff
@@ -299,7 +299,7 @@
   "application.household.addMembers.done": "تمت إضافة الأشخاص",
   "application.household.addMembers.doubleCheck": "يرجى التحقق مرة أخرى من المعلومات الخاصة بكل فرد من أفراد الأسرة.",
   "application.household.addMembers.title": "أخبرنا عن منزلك",
-  "application.household.assistanceUrl": "هتبص://إكسيج.كوم/",
```
ludtkemorgan, line 302: 😆

sites/public/src/components/home/HomeResources.tsx
```diff
@@ -18,115 +18,234 @@ export const HomeResources = (props: HomeResourcesProps) => {
...
+      </Grid>
+    )
+  } else if (enabled.length > 0) {
+    gridLayout = (
+      <Grid spacing="lg">
+        <Grid.Row columns={3}>{cards}</Grid.Row>
+      </Grid>
+    )
+  }
+  return gridLayout
```
ludtkemorgan: issue: when there is only 1 card it should not have just 1 column but instead be 2
```suggestion
return (
    <Grid spacing="lg">
      <Grid.Row columns={enabled.length % 2 === 0 || enabled.length === 1 ? 2 : 3}>
        {cards}
      </Grid.Row>
    </Grid>
  )
```

ludtkemorgan: For reference this is what the 1 column looks like 

Image

sites/public/src/components/home/HomeResources.tsx
```diff
@@ -18,115 +18,234 @@ export const HomeResources = (props: HomeResourcesProps) => {
...
+  }
+  if (enableResourcesCard) {
+    enabled.push(enableResourcesCard)
+  }
+  if (enableSeeOurData) {
+    enabled.push(enableSeeOurData)
+  }
+  if (showNotificationsCard) {
+    enabled.push(showNotificationsCard)
+  }
```
ludtkemorgan: nit: instead of adding these all as booleans, you could move this enabled to the bottom and add the cards. 
```
  const enabled = []

  if (enableAdditionalResources) {
    enabled.push(additionalResourcesCard)
  }
  if (enableGetAssistanceCard) {
    enabled.push(getAssistanceCard)
  }
  if (enableResourcesCard) {
    enabled.push(resourcesCard)
  }
  if (enableSeeOurData) {
    enabled.push(seeOurDataCard)
  }
  if (showNotificationsCard) {
    enabled.push(notificationsCard)
  }
```
and then update the return to be 
```
 <Grid spacing="lg">
        <Grid.Row columns={2}>{enabled}</Grid.Row>
      </Grid>
```

Note: you'll need to adjust the order to the correct order

ludtkemorgan (changes requested):
Just one issue and one suggestion

sites/public/src/components/home/HomeResources.tsx
mcgarrye: I wasn't 100% sure how we wanted to handle a single card, but can do

mcgarrye:
Screenshot 2026-09-15 at 3 31 46 PM
Just a call out that if no cards are turned on we still have the header appearing. Should be a followup imo

ludtkemorgan (approved):

PR #6692, opened 2026-09-15
fix: can remove rent info from unit

YazeedLoonat:

## Description
When a unit was saved with rent info you were allowed to progress out of the unit form but when you went to save the listing you would get a nondescript error 

This change allows a unit that had rent info to have that rent info removed without locking the listing form from saving

Edit or use a listing that has a unit that has rent info for the unit
remove the rent info from the unit
the listing should save properly 

## Author Checklist:

## Review Process:


ludtkemorgan (approved):
question: I thought some of these fields are required fields in some jurisdictions. Was that just being enforced by this weird error?

Anyway since this solves LA immediate problem I'm good with us merging it and we can follow-up to make sure this works for other jurisdictions

PR #6696, opened 2026-09-15
fix(api): apply the default validation group to jurisdiction feature …

seanmalbert:
Fixes a validation bug that makes `PUT /jurisdictions/:id` fail whenever the body carries `listingFeaturesConfiguration`, including the exact shape a GET returns.

Targets `main`. Found while reviewing #6676, where it blocked a reviewer from exercising the jurisdiction update endpoint at all.

## Description

`defaultValidationPipeOptions` validates with `groups: [ValidationsGroupsEnum.default]`. A decorator that declares no group never runs under that configuration, so the validators on `ListingFeaturesConfiguration`, `ListingFeatureField`, and `ListingFeatureCategory` were all inert. With no applicable constraints on the nested object and `forbidUnknownValues: true` set, class-validator reports it as an unknown value:

```
listingFeaturesConfiguration.an unknown value was passed to the validate function
```

Every validator in those three DTOs now declares the default group. The one decorator that already had it, `required` on `ListingFeatureCategory`, is what suggests the omission was an oversight rather than a decision.

Two consequences worth naming. The endpoint rejected a payload it should accept, so any caller round-tripping a jurisdiction had to strip the field. And the constraints that were supposed to guard the payload were never enforced, so a malformed configuration would have passed had the object not been rejected wholesale.

Behavior elsewhere is unchanged: the field is optional, and a jurisdiction sent without it validated before and still does.

**1.** Read a jurisdiction, then send it straight back:

```bash
KEY=$(grep '^API_PASS_KEY' api/.env | cut -d= -f2- | tr -d '"')
curl -s -H "passkey: $KEY" "http://localhost:3100/jurisdictions/byName/Bloomington" > j.json
```

Send that body to `PUT /jurisdictions/<id>` with an admin session. On `main` it returns 400 with the message above. On this branch it returns 200.

**2.** Confirm the validation now actually applies, rather than being skipped in the other direction. A configuration with a non-string id should return 400:

```json
{ "listingFeaturesConfiguration": { "fields": [{ "id": 5 }] } }
```

**3.** Confirm a jurisdiction with no `listingFeaturesConfiguration` still saves.

### Verification

The api suite passes at 67 suites and 1398 tests, and the api builds and lints clean.

The new spec at `api/test/unit/dtos/listing-features-config.dto.spec.ts` drives the real `ValidationPipe` rather than calling `validate` directly, so it fails the way an API caller does. Its four cases are the flat field list a read returns, categorized features, a non-string id still rejected, and no configuration at all. The first two fail without the DTO change, which I confirmed by stashing it.

## Author Checklist:

- [ ] Made corresponding changes to the documentation (none required)
- [ ] Ran `yarn generate:client` and/or created a migration when required (not required)


ludtkemorgan (approved):
Thanks for addressing this!

Issue #6668, opened 2026-09-03
Component-token overrides


seanmalbert:
- [ ] General
- [x] Backend
- [x] Public
- [ ] Partners

## Background / Issue
Forks override component tokens beyond color and font in their SCSS today (`.lakeview` sets
`--button-border-radius-sm`/`-md`; other blocks adjust card padding). Without a runtime path for
these, per-fork SCSS survives deforking. This implements OQ7 Option A: an allowlisted `tokens` map
in the brand JSON, emitted through the same injection.

## Technical Context
- TDD: §11.7 (Option A).
- Audit the forks' `overrides.scss` component blocks to produce the allowlist of token names a
  jurisdiction may override.
- Add optional `tokens: Record<string, string>` to `BrandDTO`. Write-time validation rejects any
  token name outside the allowlist and any value outside a constrained grammar per token: a number
  with unit, a keyword, or a `var(--seeds-*)` reference. Free-form strings would be a CSS-injection
  vector, unlike the hex-validated colors, so the grammar is the security boundary.
- `_document.tsx` emits the validated tokens on `:root` alongside the ramps, re-asserting the
  grammar at render the way the colors re-assert hex.
- The seed migration (#6671) moves each fork's allowlisted overrides into `tokens`; anything
  outside the allowlist is reconciled to a single core value or stays a documented code-level
  exception.
- Depends on: #6663 (DTO), #6665 (injection).

## Acceptance Criteria
- An allowlisted token with a valid value is emitted in the style block.
- A token name outside the allowlist, or a value outside its grammar, is rejected on write; a
  stored bad value is dropped at render.
- The allowlist is documented with each token's permitted grammar.

## QA Notes
Backend and public. Write and render a radius token end to end; attempt an unlisted name and a
free-form value and confirm both are rejected on write and at render.

Issue #6667, opened 2026-09-03
Brand font via Google Fonts


seanmalbert:
- [ ] General
- [x] Backend
- [x] Public
- [ ] Partners

## Background / Issue
A jurisdiction can set its font as a Google Fonts stylesheet URL plus a family name. Per OQ2's
recommended approach, the Google path is the only one built in Phase 2; self-hosted fonts are a
recorded follow-on.

## Technical Context
- TDD: §5.4, OQ2.
- Write-time validation restricts `brand.fontUrl` to the Google Fonts hosts
  (`fonts.googleapis.com` / `fonts.gstatic.com`), tightening #6663's `@IsUrl()`.
- `_document.tsx` emits `<link rel="preload" as="style">` then `<link rel="stylesheet">` for the
  URL, and sets `--seeds-font-sans` to the family with the fallback stack
  (`system-ui, sans-serif`) so text stays readable during load.
- When `fontUrl` is absent, no links are emitted and the default ui-seeds stack applies.
- The follow-on self-hosted design (structured `font` object, `faces[]` uploads, inline
  `@font-face`, license metadata outside `brand`) stays in TDD OQ2; nothing here forecloses it.
- Depends on: #6663 (validation home), #6665 (`_document.tsx`).

## Acceptance Criteria
- A Google Fonts URL and family render the font at first paint with preload ahead of stylesheet.
- A non-Google `fontUrl` is rejected on write.
- No font set emits no links and renders the default stack.

## QA Notes
Public site plus write validation. Set a Google font in staging and confirm the links, the CSS
variable, and readable text during a throttled load.

PR #6705, opened 2026-09-16
feat: allow a jurisdiction to override allowlisted component tokens

seanmalbert:
This PR addresses #6668

Stacked on #6693 and targets `6667/brand-font`.

## Description

Implements an allowlisted `tokens` map in the brand JSON, emitted through the same `:root:root` injection as the colors and fonts.

### What the audit found

The issue asks for the allowlist to come from auditing the forks' `overrides.scss`. Diffing each fork's public overrides against core's, ignoring color declarations:

| Fork | Fork-specific non-color declarations |
| --- | --- |
| bloom-hba | none |
| bloom-detroit | 3, all font stacks |
| bloom-la | 3, all font stacks |
| doorway | 69 |

Three of the four forks need nothing beyond what the colors and #6667's fonts already cover, except the serif font slot, which Detroit and LA both set. Doorway is the whole case for this feature.

Doorway's 69 do not fit the grammar the issue specifies. 17 are a `var(--seeds-*)` reference and 16 a plain length, which the grammar allows. The other 36 do not: 18 reference `--bloom-*` tokens from the legacy design system, 14 are keywords or other values, and 4 use `calc()` or multi-value shorthand such as `var(--bloom-s4) var(--bloom-s3) var(--bloom-s3) var(--bloom-s3)`.

The allowlist is the part of that set the grammar can safely express, and Doorway keeps the rest in SCSS. Covering Doorway fully would mean admitting `calc()`, multi-value shorthand, and `--bloom-*` references.

### The allowlist

| Token | Grammar | Evidence |
| --- | --- | --- |
| `--button-border-radius-sm` | a length (`0`, px, rem, em, %) or `var(--seeds-*)` | core's `.lakeview`, doorway |
| `--button-border-radius-md` | same | core's `.lakeview`, doorway |
| `--button-border-radius-lg` | same | doorway |
| `--seeds-font-serif` | a family name | bloom-detroit, bloom-la |

The grammar is the security boundary, so it is enforced twice: by an `AllowlistedTokens` validator on write, and again in `_document.tsx` at render, because a row written by a seed, a migration, or direct SQL doesn't hit the DTO. A token outside the list, or a value outside its grammar, is refused on write and dropped at render.

`--seeds-font-serif` takes a family name rather than a full stack, matching how #6667 handles the sans and heading slots. That keeps one family rule across all three font slots and avoids parsing user-supplied quotes and commas inside a style block.

### Font fallbacks move to CSS

The document previously repeated a literal fallback stack for each font slot, and they did not match the design system: the sans and heading slots both fell back to `system-ui, sans-serif` while ui-seeds uses Open Sans and Lato respectively.

`overrides.scss` now defines `--brand-font-fallback-sans`, `-alt-sans`, and `-serif`, each mirroring its ui-seeds stack, and the document references them. A brand font that fails to load now degrades to the design system's own appearance, the stacks live in CSS beside the rest of the styling, and a ui-seeds change is a one-file edit. This corrects #6667's behavior as well as covering the serif slot added here.

### One thing to know for review

The token map is stored correctly but came back as `{}` through the endpoint. `mapTo` passes `enableImplicitConversion: true`, and under that class-transformer empties a map of dynamic keys. The field therefore reads from the source object, `@Transform(({ obj }) => obj?.tokens)`. Without implicit conversion it serializes fine, which is why it passes in isolation and failed end to end.

The generated client types the field as `tokens?: object`, so the render side checks each entry rather than trusting the type.

Set up from #6681 step 1 applies: the jurisdiction needs a brand and `enableDbDrivenBranding` on.

**1.** Store an allowlisted token through the API, with an admin session:

```json
{ "brand": { "primary": { "base": "#773E98" },
             "tokens": { "--button-border-radius-md": "var(--seeds-rounded-3xl)",
                         "--seeds-font-serif": "Noto Serif" } } }
```

**2.** Confirm both reach the page, and that the serif slot carries its fallback:

```bash
curl -s http://localhost:3000/ | grep -o -- '--\(button-border-radius-md\|seeds-font-serif\):[^;]*;'
# --seeds-font-serif: "Noto Serif", var(--brand-font-fallback-serif);
# --button-border-radius-md: var(--seeds-rounded-3xl);
```

**3.** Confirm buttons actually round in the browser, which is the part no test covers.

**4.** Try the rejections. Each should return 400: a name outside the allowlist such as `--link-text-color`, a value with `calc()`, a `var(--bloom-*)` reference, multi-value shorthand, and a value that closes the declaration such as `0; } body { display: none } .x {`.

**5.** Write one of those bad values straight into the `brand` column with SQL, wait out `CACHE_REVALIDATE`, and confirm the page drops that token while still emitting the colors.

### Verification

The api suite passes at 70 suites and 1463 tests, and the public suite at 117 suites and 1121 tests. The api builds and both halves lint and typecheck clean. Run the public suite with `--runInBand`: in parallel, `create-account` and `finder` fail intermittently on 5s timeouts, with each suite passing on its own.

The jurisdiction e2e spec was run against a live database and passes at 23 tests, including a token stored and returned, a name outside the allowlist rejected, and a `calc()` value rejected. Note that spec assumes the fresh database `yarn test:e2e` creates: it inserts fixed names such as `new jurisdiction`, so a second run against the same database fails on the unique-name constraint.

I also ran this against the running site rather than only in tests: with a font and two tokens stored, the page emits the four expected declarations, and the `--brand-font-fallback-*` definitions they reference are present in the CSS the app serves.

## Author Checklist:


sites/public/styles/overrides.scss
```diff
@@ -1,6 +1,18 @@
 :root {
+  // Verbatim copies of the ui-seeds stacks, referenced by _document.tsx.
+  // TODO(ui-seeds#155): replace with var(--seeds-font-default-*) once those tokens ship.
```
ludtkemorgan, line 3: question: are you planning on addressing this issue? Or is this ok to ship like this for now?

ludtkemorgan (approved):
Just one question

sites/public/styles/overrides.scss
seanmalbert: This is safe to ship like this. Updating seeds directly would just allow us to remove this from overrides.scss. I didn't want to introduce the seeds repo and deployment as a blocker to this.

